PR #2041, opened 2013-03-26
Add Stylus port to foundation framework (in addition to Sass)

blai:
I think porting Foundation to Stylus would give Node.js (particularly Express.js) application the same capability of modularizing the theme and extend it.

When use the Sass version (default) with Ruby applications, `@import 'some/Component'` would resolve to `some/Component.sass`

When use the Stylus port with Express.js applications, `@import 'some/Component'` would resolve to `some/Component.styl`

I also generated the diff between the output of Sass and Stylus port [here](https://github.com/blai/foundation-sass-stylus-diff/blob/master/css.diff), most css are identical, with minor differences for colors generated by 'darken' and 'lighten' functions.

I will soon checkin the enhancement I did for https://github.com/blai/generator-foundation to allow user to choose either Sass or Stylus.


soulcreate:
+1

GCheung55:
+1

blai:
I am not sure if anyone would have time to dedicate on this. But if anyone wants to take this up, I think I still have ownership on the npm package name `foundation`.

atilkan:
I am willing to contribute stylus port. 
I think switching to stylus would be better if you are to choose best tool for the job.

IamManchanda:
cc @kball

atilkan:
@IamManchanda 
I would understand if you knew stylus. I wonder how you don't like something you don't know. No offence but this is the pure definition of ignorance.

IamManchanda:
I am not against it ... I am definately in favour of the port but do we have the resources to do that is a big question? 

Not ignoring it!

robmuh:
With all the new CSS grid stuff I do not use Foundation or Bootstrap or any other layout grid any longer. I'm unsubscribing but thanks for the responses.

atilkan:
stylus port is not rocket science. Variable defination and include statements are the differences. The rest would work in stylus because of flexible syntax.

IamManchanda:
Hmmn... I see 
Awaiting your reply @kball

PR #2173, opened 2013-04-12
Fix for visibility display bug

lukebussey:
The visibility classes use the property of "inherit" which causes the element they are set on to inherit the display property of the parent element.

This causes an issue where an inline element is used within a block element as demonstrated on the following page where a label is used within a paragraph of text:

http://lukebussey.com/foundation-visibility-classes-bug/

Screen Shot 2013-04-12 at 3 35 54 PM

The property should be set to "initial" rather than "inherit" which gives the element the original display property rather than inheriting it from its parent which may be incorrect.

The fix is demonstrated here:

Screen Shot 2013-04-12 at 3 36 14 PM

http://lukebussey.com/foundation-visibility-classes-fix/

BTW: I didn't modify any of the table styles in the _visibility.scss file, the same fix could be applied to them but I haven't tested it.

Thanks!
Luke

### DIFF
--- a/scss/foundation/components/_visibility.scss
+++ b/scss/foundation/components/_visibility.scss
@@ -6,7 +6,7 @@
   /* Foundation Visibility HTML Classes */
   .show-for-small,
   .show-for-medium-down,
-  .show-for-large-down { display: inherit !important; }
+  .show-for-large-down { display: initial !important; }
   
   .show-for-medium,
   .show-for-medium-up,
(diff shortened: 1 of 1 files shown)


DivinoAG:
Could I request further consideration to be given to this issue? I haven't been using Foundation for that long, but on my very first project I already ran into this problem 3 times, because this makes visibility classes almost useless when in conjunction with inline elements, and hiding/displaying different elements according to screen size is a pretty common situation. There are a few suggestions here that solve this problem without breaking basic HTML components.

PR #2671, opened 2013-06-29
In engine installation

ShadowKoBolt:
Related to this closed issue: https://github.com/zurb/foundation/issues/1225

When generating a mountable engine with rails plugin new name_of_plugin --mountable (rails version 4.0.0) found I was unable to run the foundation:install generator in the newly generated plugin (after adding zurb-foundation to either the plugin gemfile or gemspec).  

I have moved the generators into lib/generators/foundation which means it is now detected in the plugin as an available generator.  I've also added the --engine=ENGINE option to the install generator to allow you to specify the namespace which the files are nested within.  This adjusts the layout template and the lines appended into application.js and application.css

E.g. in a plugin generated via:
  rails plugin new some_name --mountable
After adding zurb-foundation to the gemfile or gemspec you should run:
  bundle exec rails g foundation:install --engine=some_name

This should carry out the normal actions, adding namespace where needed and also requiring vendor/jquery in the application.js file as jquery is not standard in a new plugin.

I have not attempted to use this change against a rails --full type, only --mountable plugin.


jimboslicethat:
Has anyone come up with a solution to this yet? I have been racking my head for quite some time trying to find someone out there who was finally able to get the foundation generator working inside an engine.

r3wt:
@jimboslicethat no, I haven't been able to get it working at all. :-1:

jimboslicethat:
What about manual installation? I have been trying that and it seems to install foundation, but overriding defaults in foundation is a completely different story.

PR #3157, opened 2013-08-31
add checkboxes required validation

huchenme:
Originally Abide does not support checkboxes group required validation (or it might be a bug). I have added the validation for checkboxes and updated the documentation as well


rossgzc:
+1! This would be incredibly useful :)

Elyasin:
+1 Would be top.

Akarienta:
Check this SO question, it is possible if you write your own abide validator: http://stackoverflow.com/questions/26890023/display-error-if-no-checkbox-is-checked-in-checkbox-group

mikejoyceio:
+1

PR #3829, opened 2014-01-06
Orbit height not set on accordion

seantimm:
Orbit slider container height is not being calculated, when Orbit slider is not visible (when inside closed accordion).

Orbit's container height normally is being calculated from the largest image/element in it. But, if the container is not being displayed, it can not calculate the height, because jQuery can't calculate image height, if parent element is not displayed `display: none`... It is how it its.

In the code ( https://github.com/zurb/foundation/blob/master/js/foundation/foundation.orbit.js#L266 ) there I see those lines:

``` javascript
$(window).on('resize', self.compute_dimensions);
$(window).on('load', self.compute_dimensions);
```

If somehow you could trigger `compute_dimensions` after opening an accordion's tab... For example, you could trigger `accordionOpened` event or something and bind Orbit to listen the event... At least give me a direct access to `compute_dimensions` !

Currently I fixed this problem by calling

``` javascript
$(window).resize();
```

on accordion's tab click.

### DIFF
--- a/js/foundation/foundation.orbit.js
+++ b/js/foundation/foundation.orbit.js
@@ -412,21 +412,38 @@
     init: function (scope, method, options) {
       var self = this;
 
-      if (typeof method === 'object') {
-        $.extend(true, self.settings, method);
-      }
+      this.bindings(method, options);
+    },
 
-      if ($(scope).is('[data-orbit]')) {
-        var $el = $(scope);
-        var opts = self.data_options($el);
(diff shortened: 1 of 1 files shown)


smerth:
Hi there I am having this exact issue but I am unclear how to implement the javascript solutions talked about above...  This is a ver old thread and I am using the latest Foundation 6 but the issue is the same..  Can anyone help me understand how to implement a fix.

PR #4754, opened 2014-03-21
Added grunt-autoprefixer

tmugford:
As per #4123, I've added **grunt-autoprefixer** and **grunt-contrib-cssmin** (to create a minified version instead of running the sass and autoprefix tasks twice). All vendor prefixed properties and associated mixins have been removed, except those that have been included: 
- to work around specific browser issues (several places in `_offcanvas.scss` for example)
- as a hack to target particular browsers  (`@-moz-document url-prefix()` in `_forms.scss` for example)
- or because they're for non standard properties (`-webkit-font-smoothing` and `-webkit-appearance` for example)

As discussed in the [associated issue](https://github.com/zurb/foundation/issues/4123), I've also added a mixin to control the inclusion of the aforementioned explicitly vendor prefixed properties (repurposing the  `$experimental` flag). Though I've not used the mixin in this PR, the intended use would be something like the following:

``` scss
@include experimental {
  -webkit-appearance: none !important;
}
```

I've tested it as much as I can with the browsers/devices that I have access to. Let me know if you'd like the changes squashed in to a single commit.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -57,14 +57,22 @@ module.exports = function(grunt) {
           'dist/assets/css/normalize.css': 'scss/normalize.scss',
           'dist/docs/assets/css/docs.css': 'doc/assets/scss/docs.scss'
         }
+      }
+    },
+
+    autoprefixer: {
+      options: {
+        browsers: ['Explorer >= 9', 'Android >= 2', 'iOS >= 5', '> 1%']
       },
+      dist: {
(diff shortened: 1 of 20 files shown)


quakenul:
I am actually still very interested in that. The other day I was having a problem with transform in a current safari, and was pretty surprised to learn the autoprefixer, which I kind of took for granted, was gone. 

As far as I can see, no substitute mixins were added. Why would you remove autoprefixer? What is the proposed solution?

PR #5237, opened 2014-05-29
Add comma support for Abide number validation...

prestancedesign:
...because some countries use comma for decimal separator instead point.


antoniopiha:
Redundant escape character before comma sign:
In patterns number instead of 
[\.\,]
it should be:
[\.,]

because comma is not special character.

PR #5308, opened 2014-06-11
Orbit is alive!!!! Full revert back to when Orbit worked

joeworkman:
In Foundation v5.2, Orbit was overhauled to include CSS animations. Everyone has been plagued with issues ever since. This release rolls back Orbit to its former glory.

I also went through all of the commits to Orbit since 5.2 and tried to add all of the bug fixes and features back in. The only known feature that was not added back in was the ability to completely turn off the Orbit timer. The code base was too different to easily merge that feature back in.


brody182:
orbit does not work anymore!!!

brody182:
Orbit is dead!!!

PR #5388, opened 2014-06-27
Created option to skip element in the carousel and updated documentation...

asommer70:
Good day,

Submitting these few small changes to add the option to Clearing to be able to configure an element that will be skipped when navigating the carousel.  Please let me know if I need to change anything, or you need additional information.

Thanks all,
Adam


gorelog:
Hello,

This does not seem to skip if there are multiple skippable elements in a row. If one is skipped, it just skips to the next one regardless if that too, has a skip_selector designated to be skipped. See: http://codepen.io/anon/pen/waNMWm

Thanks.

PR #5583, opened 2014-08-07
Trigger all known resize events on window load

joeworkman:
I won't cry if you don't accept this one. I am sure there are better ways to accomplish this. This was the quick and dirty way. 

I thought that it was an interesting idea that could solve a lot of sporadic edge case issues with things perceiving not to load properly on page load.

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -116,7 +116,7 @@
         }
       });
     }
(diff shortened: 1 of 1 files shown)


MarcGuay:
This commit causes issues with range sliders.  https://github.com/zurb/foundation/issues/6606

PR #5640, opened 2014-08-20
When deep_linking:true, allow tab a[href] to have full url, only matching the anchor in the end.

lkol:
Setting `data-options="deep_linking:true"` for tabs would not work unless `href` attribute had only the anchor part. This is not sufficient when the user wants to deep link tabs across pages. Simple change to CSS selector fixes this problem (only match the end of `href` attribute).


rafibomb:
Thanks for this! Looks like there have been a lot of updates to tabs since the PR that have resolved the underlying issue. https://github.com/zurb/foundation/commits/master/js/foundation/foundation.tab.js 

If you still experience it, please let us know. Thanks!

PR #5747, opened 2014-09-06
Added text input types function

ghost:
As per the request https://github.com/zurb/foundation/issues/5712

Accepts a custom selector, which defaults to 'input'. Can also accept a list of types if having all types is overkill.

Example:

``` scss
#{text-inputs( email url textarea, '.form-control')} {
    border-color: $oil;
    &:focus {
      border-color: $primary-color;
    }
}
```

### DIFF
--- a/scss/foundation/_functions.scss
+++ b/scss/foundation/_functions.scss
@@ -6,7 +6,7 @@
 $rem-base: 16px !default;
 
 // IMPORT ONCE
-// We use this to prevent styles from being loaded multiple times for compenents that rely on other components. 
+// We use this to prevent styles from being loaded multiple times for compenents that rely on other components.
 $modules: () !default;
 @mixin exports($name) {
   @if(not index($modules, $name)) {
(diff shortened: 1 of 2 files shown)


Grawl:
When `$all-text-input-types` is defined inside this function, I cannot get it for my own use or redefining it. I suggest to make it global and `!default`. PR maybe?

gakimball:
@Grawl Why do you need to redefine it? Is something missing?

Grawl:
@gakimball for a reason. I may want to add my custom-typed controls here. Or a class.

Now I just want to use these classes in my own code to do something different that Foundation do not have to do. But I have to copy them to my code.

gakimball:
There's a finite number of valid values for `<input type>`, so `$all-text-input-types` is only going to contain those values.

If you're interested in targeting a class instead of `input[type=""]` to apply form styles, you can use the form mixins, like so:

```scss
.text-input {
  @include form-element;
}
```

Either way, Foundation 5 is at end-of-life, so we're no longer adding new features, just fixing bugs.

PR #5836, opened 2014-09-19
Fix incorrect import path

Weeblewonder:
_functions.scss is on the same level as file not in a foundation folder.


kamalgill:
@rafibomb: Why was this reverted?  It seems to be an important fix, and Travis build https://travis-ci.org/zurb/foundation/builds/35710065 seemed to pass.

PR #5966, opened 2014-10-21
[foundation.reveal.js] replace bind(this) with local context variable

ridiculous:
Fix compatibility issue with the use of `.bind(this)` and the Capybara (2.1.0) testing framework. Exceptions were encountered when using the Capybara WebKit driver or PhantomJS  Poltergeist driver.

### DIFF
--- a/js/foundation/foundation.reveal.js
+++ b/js/foundation/foundation.reveal.js
@@ -257,7 +257,8 @@
       // is modal
       if (css) {
         var settings = el.data(this.attr_name(true) + '-init') || this.settings,
-            root_element = settings.root_element;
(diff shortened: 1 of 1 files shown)


ridiculous:
@gakimball any update on this?

gakimball:
So the fix is just to make it work with Capybara, which ordinarily wouldn't concern is, but as long as it doesn't mess with any existing functionality it should be fine. Have you tested the change in multiple browsers?

ridiculous:
Yup :D

gakimball:
That works for me, then! Thanks for the PR, and sorry we didn't get to it sooner. :)

PR #6038, opened 2014-11-14
Reveal Modal Fixes

Kolodij:
These additions allow a user to choose both the background root element and the modal root element independently of each other. it also changes the background element so that it covers the entire element, not just the currently visible portion in the window. Fixes many issues seen throughout the forums in regards to Reveal Modals.

### DIFF
--- a/js/foundation/foundation.reveal.js
+++ b/js/foundation/foundation.reveal.js
@@ -15,6 +15,7 @@
       close_on_esc: true,
       dismiss_modal_class: 'close-reveal-modal',
       bg_class: 'reveal-modal-bg',
+      bg_root_element: 'body',
       root_element: 'body',
       open: function(){},
       opened: function(){},
@@ -238,10 +239,13 @@
       return base;
     },
 
-    toggle_bg : function (modal, state) {
+    toggle_bg : function (el, modal, state) {
+      var settings = el.data(this.attr_name(true) + '-init') || this.settings,
+            bg_root_element = settings.bg_root_element; // Adding option to specify the background root element fixes scrolling issue
+      
       if (this.S('.' + this.settings.bg_class).length === 0) {
         this.settings.bg = $('<div />', {'class': this.settings.bg_class})
-          .appendTo('body').hide();
+          .appendTo(bg_root_element).hide();
       }
 
       var visible = this.settings.bg.filter(':visible').length > 0;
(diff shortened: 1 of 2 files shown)


Kolodij:
@AbuJunayd i see the reveal background element is covering the entire body element, although in firefox the sticky footer and form are pushed outside of the body.

the changes were proposed so that developers could choose what elements the reveal and reveal background were attached to independently. they both default to the body currently. You can change this behaviour however in your .js file with these:

bg_root_element:
root_element:

Hope that helps!

AbuJunayd:
@Kolodij thanks. But it seems the results depend on the browser as you can read here: https://github.com/zurb/foundation/issues/6154

Kolodij:
From what I can gather reveal was attaching itself directly after the root element, and the background was always attaching itself to the window. the problems with that are that the modal was always added at the top of the page, and the background only covered the window area. If a user scrolled down the page without closing the modal the background remained only covering where the window was at the time it was opened. Also, if there was a link in a footer that opened a modal when clicked the user was scrolled to the top of the screen automatically this is bad UX it’s not an expected action unless of course when close they were auto scrolled back to the original position… but that just seems weird. 

On your example page you have content being pushed outside the body, but the modal_bg is being attached to the body. it won’t cover that extra content

the update to reveal was checked by me in firefox / i.e. / chrome / and safari not opera though, as well as on mobile. so I don’t believe anything browser specific is coming from reveal itself but more than likely some other element(s) css issues

in your example page when resizing the browser your content is flowing outside of the body class, actually it’s flowing outside of the html element as well. for some reason though, your #wrapper-content seems to contain all your elements. so this will attach the modal bg and the modal to that.

I’m not sure if I’m supposed to post code here… But try adding this to your foundation init.

   reveal:
   {
       animation: 'fadeAndPop',
       animation_speed: 250,
       close_on_background_click: true,
       dismiss_modal_class: 'close-reveal-modal',
       bg_class: 'reveal-modal-bg',
       bg_root_element: ‘#wrapper-content',
       root_element: '#wrapper-content',
       bg : $('.reveal-modal-bg'),
       css : {
           open : {
               'opacity': 0,
               'visibility': 'visible',
               'display' : 'block'
           },
           close : {
               'opacity': 1,
               'visibility': 'hidden',
               'display': 'none'
           }
       }
   }

AbuJunayd:
I have added this code but it doesn't change anything. The background doesn't cover all the content and the modal stays at the top, regardless of the scroll.

I've decided to use bPopup instead of reveal in my project as it has been working in all browsers I have tested so far. 

But thanks for your help.

PR #6040, opened 2014-11-16
refactor grunt to load tasks only when required (JIT)

ctalkington:
The main Grunt file sets the key paths as variables, which are then available for use in the tasks. This helps with a few things including:
- reusable configs across different projects
- git history is cleaner with modular approach
- easier to avoid merge conflicts


rafibomb:
We'll be working on this one Friday. Thanks!

PR #6050, opened 2014-11-20
No !important for .hide

wolfgang-demeter:
`.hide` should not be that !important, as it breaks a heck lot of jquery/JS-Code using .show(), .hide() or .toggle()!
https://github.com/zurb/foundation/issues/6024


andycochran:
I think `.hide { display: none !important; … }` is a valid method by which the framework hides stuff. In an app-specific situation, if I need to use other means to show/hide elements (e.g. jQuery), I simply avoid using the `.hide` class. 

You can always `$( ".target" ).removeClass('hide');`.

PR #6121, opened 2014-12-17
Clean up javascript code style

walbo:
Be more consequent in JS files. It's differences in all files.

Maybe use node-jscs (grunt-jscs) to make it easier to write consistent markup?


gakimball:
Merged in around here: f8f3ce3234ea54e07c340bbd56dcae5fd448ebb7

For the next version of the framework, the plan is to add Sass and JS linters from the start.

PR #6150, opened 2014-12-24
Reveal modal: Multiple opened

markbaas:
In my app (marksexpress), I have implemented a wysiwyg editor (summernote-foundation). In admin mode the edit form has been implemented through a modal box, however the editor also uses the modal boxes. So when inserting an image, the editor modal box would be closed. This patch adds an option that allows previous modals to remain open.
12-24-2014 8-51-49 am
12-24-2014 8-53-00 am


gjastrab:
:+1: we could use this in our app as well, hopefully this makes it in the next release!

PR #6159, opened 2014-12-29
Semantic Tabs

GUESShimself:
I swapped out the dl and dd's for a ul and li's for the tabs to make them more semantic. Per the [spec](http://www.w3.org/html/wg/drafts/html/master/grouping-content.html#the-dl-element), the dl element must contain one or more dt element, each followed by one or more dd elements.

Using list items (li's) allows the tabs to be used with unordered and ordered lists, as well as in a menu element if necessary.


rafibomb:
I could have sworn I did this already :) Thanks!

rafibomb:
The ```<li>```'s need a tab-title class I've also added.

PR #6161, opened 2014-12-29
fixed positioning of icon pip for multi line split buttons

DrSammyD:
Using Ems to get consistent icon centering based on font sizes. It now works like the existing pip for vertical centering.


tdhartwick:
Thanks DrSammyD.

PR #6163, opened 2014-12-30
Reverts #5551e12, commit has breaking changes for non-WebKit browsers.

### LINKED ISSUE #6154: Reveal Modal out of visible area on Firefox
Hello,

I just wanted to report this bug about reveal modal.

While opening a modal from the documentation page on firefox: http://foundation.zurb.com/docs/components/reveal.html

I noticed that the modal was not opening on the visible portion of the page. Rather it's opening at the top, so we have to scroll up to the top of the page to see it.

I have only been able to test it on chrome 39.0.2171.95 (it is working on chrome) and on firefox 34.0 (not working)

jonathanmelville:
5551e12853bb29de7ba685bea7dcdc8628bc186b introduced a bug to Reveal where page scroll was not taken into account on non-WebKit browsers. This had the effect of launching a modal at the top of a page, even if the user opened it while scrolled to the bottom.

This commit simply reverts the to the last working version and will fix #6154.

### DIFF
--- a/js/foundation/foundation.reveal.js
+++ b/js/foundation/foundation.reveal.js
@@ -15,7 +15,6 @@
       close_on_esc: true,
       dismiss_modal_class: 'close-reveal-modal',
       bg_class: 'reveal-modal-bg',
-      bg_root_element: 'body',
       root_element: 'body',
       open: function(){},
       opened: function(){},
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks Jon!

jonathanmelville:
Thank you sir!

PR #6177, opened 2015-01-07
[Orbit] use outerHeight() instead of height()

alfredbez:
I had some slides with an 1px-border and there were 2px missing on the container.

see http://codepen.io/alfredbez/pen/KwaKBp


rafibomb:
Sounds good, thanks!

PR #6178, opened 2015-01-07
Add column setting for every support

stephane-r:
Choosing the number column on all support (small, medium and large)


rafibomb:
We're not looking to make this kind of change. It kind of would work the way our older grids worked. Other people might want to grab the changes from this PR. We'll close it here. Thanks!

PR #6188, opened 2015-01-12
Removed unnecessary variable '$pip-color' in dropdown button component

hohertz:
The variable '$pip-color' is always applied as a parameter of the mixing 'dropdown-button' with the default value '$white'. For this reason the pip stays white, no matter which value the variable '$dropdown-button-pip-color' has. In my opinion the '$pip-color' variable is unnecessary and should be removed.


gakimball:
Shouldn't the `$pip-color` default value just be `$dropdown-button-pip-color`, then?

rafibomb:
We can just use the default color in the argument

rafibomb:
https://github.com/zurb/foundation/commit/0f2d9d0ee9650436a3d5099e9660426f7b9f8770

PR #6189, opened 2015-01-12
Update foundation.js

MikeNGarrett:
Any reason you would want to append a random number query string to an image url? Happy to be wrong, just very confused by these lines.

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -133,14 +133,6 @@
 
     function bindLoad () {
       this.one('load', loaded);
-
-      if (/MSIE (\d+\.\d+);/.test(navigator.userAgent)) {
-        var src = this.attr( 'src' ),
-            param = src.match( /\?/ ) ? '&' : '?';
-
-        param += 'random=' + (new Date()).getTime();
-        this.attr('src', src + param);
-      }
     }
 
     if (!image.attr('src')) {


MikeNGarrett:
I reject this merge as an answer. 

@dougaxe1 found the basis of these lines. This hack appears to be referenced on CSS Tricks: http://css-tricks.com/snippets/jquery/fixing-load-in-ie-for-cached-images/

An alternative was proposed by @catearcher to fix this issue: 
`$("img").attr("src", $("img").attr("src"));`

I haven't tested this as of yet, but I think it's a better solution to fix the issue being "solved" in the deleted code.

oller:
I'm having problems with this block of code on IE10 triggering multiple requests for the same image.  

As you can see the cache busting means the image is being downloaded twice, with another attempt aborted.

IE10 Network Pane

Shall investigate further, is anyone else encountering this issue? It seems specific to IE10

MikeNGarrett:
Hey Oller, 

I can tell you that Foundation has an IE-targeted script that appends a query string with a random number to the end of an image src. 
This code seems to be related to this snippet discussion: https://css-tricks.com/snippets/jquery/fixing-load-in-ie-for-cached-images/
At the end of the discussion there's a better method: https://github.com/desandro/imagesloaded

oller:
Hi @MikeNGarrett 

Thanks for the quick response.  I guess my concern is why it's running multiple times, which event is firing multiple times to trigger this?  I'll update this thread once I've some progress on this matter.

MikeNGarrett:
Not sure. To be clear, I'm not affiliated with Zurb or Foundation at all. I just filed the ticket. ;-)

PR #6209, opened 2015-01-21
top-bar auto-width only for h1 in title-area

karland:
Search engine optimization (SEO) suggests to only use one `h1` per page. Therefore one should not use up the `h1` element already in the title. If `h2` is used instead the width is not `auto` and there is a line break, if the title name is too long.


gakimball:
Could you update it to have all six header elements? Might as well for consistency's sake, since usually when we broadly target headers we target all of them.

karland:
OK. I have produced an update [#6223](https://github.com/zurb/foundation/pull/6223) and closing this PR.

PR #6215, opened 2015-01-22
Removed duplicate and unnecessary properties

tomsky:
Updated range-slider, reveal, sub-nav and tabs. This is only maintenance and cause no visual change.


rafibomb:
Good stuff!

PR #6223, opened 2015-01-25
 top-bar auto-width only for h1 in title-area (Update)

karland:
SEO suggests to only use one h1 per page. Therefore one should not use
up the h1 element already in the title. If h2 is used instead the width
is not auto and there is a line break, if title name is too long.


gakimball:
Awesome, thanks!

PR #6230, opened 2015-01-27
Wrap _visibility.scss exported html in the export function

domjtalbot:
I found when multiple components referenced the _visibility.scss it was pulled into the css multiple times.
Wrapping '@if $include-html-visibility-classes{}'  with the exports mixin, stopped this problem.


tdhartwick:
Thanks @domjtalbot!

PR #6231, opened 2015-01-27
Create variable for callout panel background

adamj88:
I wanted to be able to set the callout panel background color to one of my choosing, it seemed a bit odd that this was tied to a specific shade of the `$primary-color` variable. I couldn't find any reason for it to be so specific so I've added a variable in.

### DIFF
--- a/scss/foundation/_settings.scss
+++ b/scss/foundation/_settings.scss
@@ -807,6 +807,7 @@
 // $panel-bg: scale-color(#fff, $lightness: -5%);
 // $panel-border-style: solid;
 // $panel-border-size: 1px;
+// $callout-panel-bg: scale-color($primary-color, $lightness: 94%);
 
 // We use this % to control how much we darken things on hover
 // $panel-function-factor: -11%;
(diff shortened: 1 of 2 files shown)


rafibomb:
Thanks! Just realized this needs to go in docs too. I got it.

OndrejVasicek:
Hi,
I just wanted to ask, how docs corresponds with the actual “last” version, cause I’ve just tried to use $callout-panel-bg, which I had founded here [http://foundation.zurb.com/docs/components/panels.html] and was confused, why it didn’t work. After a while I found this pull request, which I manually put into my foundation, so it works now.

But it’s clear, that the docs doesn’t fit with the last downloadable version from the web/bower. Is that a “normal” state or is it just a bug? Or is there a new release just behind the corner?

Thanks for the answer.

OndrejVasicek:
And just one another thing. There isn’t an option to modify the color of text in callout. It would be nice to have an option to set something like $panel-callout-font-color and also make the $panel-font-color-alt (which is used only when the $bg is dark enough) work with the $callout-panel-bg as well. It works only with the $panel-bg now.

PR #6239, opened 2015-01-30
Update _settings.scss

shoaibik:
Re added the column gutter variable.


rafibomb:
Thanks, it was moved to global from grid but never made it into settings.

PR #6241, opened 2015-01-30
Ensure that the validate function returns the full list of validations

nickhammond:
The validate method is expected to return an array of validation checks(true/false), the line of code that I've removed set the array to the entire set of validations status. The original selector is `S(this).find('input, textarea, select')` which will indeed select hidden elements regardless of if you've specified that they should be validated. When we go to call `focus()` if you have `focus_on_invalid`(the default) and you have a hidden element it's going to grab that first field if any field has failed validation. I've demonstrated this in a spec as well, let me know if you have any questions.

Related commit and comment: https://github.com/zurb/foundation/commit/0436102f90741556db2a69c33129316eed00842e

Thanks to @laantorchaweb for bringing it up in that commit, much appreciated.

### DIFF
--- a/js/foundation/foundation.abide.js
+++ b/js/foundation/foundation.abide.js
@@ -242,7 +242,6 @@
         }
         validations.push(el_validations[0]);
       }
-      validations = [validations.every(function (valid) {return valid;})];
       return validations;
     },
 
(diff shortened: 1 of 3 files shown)


emptyflask:
:+1: This bug breaks validation on the majority of forms.

PR #6257, opened 2015-02-06
commas in list should follow a space

marcobiedermann:


zurbrandon:
Nice! We'll be adding a linter in the next version, but thanks so much for the fixes!

marcobiedermann:
@zurbrandon Yeah, adding a linter would be great. I've seen some inconsistancy with the coding style. If you like I can tackle some 'mistakes'

PR #6268, opened 2015-02-07
remove leading zero

marcobiedermann:
LeadingZero should be written without a leading zero

SassGuidelines
http://sass-guidelin.es/#zeros


rafibomb:
We believe that a leading zero is easier to read and the guidelines say that single decimal place numbers can have it. Having said that, It's ok for now to make it consistent so we'll merge it in. For the next version, we'll come up with a styleguide for things like this.

PR #6270, opened 2015-02-08
Fixing a vendor-prefixes and outlines

tomsky:
Adding compatible vendor-prefixes where possible. 
Adjusting usage of "outline: none". This should only be used with :focus and only if other visual changes are made.


rafibomb:
For accessibility you need the dropdown content to be tab accessible. What's the reason you removed the tab index on the content?

tomsky:
Not sure how adding tabindex to the "ul" solves anything, it just gives focus to something that the user can not interact with. Maybe I'm missing something. 

Regardless, setting "outline: none" (and not giving any other visual hint that it is in focus), to an element that receives focus is not a good option.

To provide better accessibility the example could make better use of WAI-ARIA. Running this example through VoiceOver seems to work produce a good results.

```html
<a id="dropdown-button" href="#" data-dropdown="drop1" aria-controls="drop1" aria-haspopup="true" aria-expanded="false" class="button dropdown">Dropdown Button</a>
<ul id="drop1" data-dropdown-content class="f-dropdown" role="menu" aria-hidden="true" aria-labelledby="dropdown-button">
  <li><a href="#" role="menuitem">This is a link</a></li>
  <li><a href="#" role="menuitem">This is another</a></li>
  <li><a href="#" role="menuitem">Yet another</a></li>
</ul>
```

PR #6287, opened 2015-02-12
Fixes input element animation

jkole:
When $input-include-glowing-effect == false, adding the .error class to an input element resulted in unwanted animation of the input bottom margin down to zero.  The error message div would appear (without animation) with a gap between it and the input element, then quickly animate up to its desired position along the bottom of the input....pretty jarring.  Proposed fix is to replace the single-transition mixin with specific animation of the border-color and background properties.  Not sure whether Foundation relies on animating any other input properties when $input-include-glowing-effect is false, but these are the only two I could find.


rafibomb:
Nice! Thanks!

PR #6290, opened 2015-02-12
Equalizer fix pictures with predefined height

Ziemien:
After this equaliser doesn't wait until pictures will be loaded if all the
pictures in element to be equalised are provided with ‘height’
attribute.

This fixes the error when using AngularJS and pictures come with 404
error. Before this equaliser didn't work at all.

Also: https://github.com/zurb/foundation/issues/6276

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -599,7 +599,19 @@
         var self = this,
             unloaded = images.length;
 
-        if (unloaded === 0) {
+        function pictures_has_height(images) {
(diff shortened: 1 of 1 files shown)


Ziemien:
Also, the issue might be that, if the response for loading picture is 404 the JS may trigger 'error' event instead of 'load' so callback in foundation.js, line 137 - https://github.com/Ziemien/foundation/blob/equalizer-picture-pre-height/js/foundation/foundation.js#L137

I'll look into that later on.

PR #6295, opened 2015-02-14
@else should be placed on same line as previous curly brace

marcobiedermann:
http://sass-guidelin.es/#conditional-statements


rafibomb:
Thanks! Standards are good - can't auto-merge this but we'll get it in soon.

marcobiedermann:
I just got one question. What if there is a comment above the `@else` condition?
See https://github.com/zurb/foundation/blob/master/scss%2Ffoundation%2Fcomponents%2F_grid.scss#L54

gakimball:
You could move the comment inside the block, and then you'd be able to pair all the `@else` blocks together.

marcobiedermann:
Ok, I'll fix it soon

gakimball:
Hey, really sorry for letting this get super stale! Now that Foundation 6 is out, we won't be making any more codebase improvements to Foundation 5. We're always looking for contributions to the new codebase, though :)

PR #6296, opened 2015-02-14
UnnecessaryParentReference: Unnecessary parent selector (&)

marcobiedermann:
http://sass-guidelin.es/#selector-nesting


gakimball:
Thanks for doing all that work! :)

marcobiedermann:
@gakimball You are very welcome. I think it is important to have clean code and than extend it. :)

PR #6305, opened 2015-02-18
Focus ability for reveal

Owlbertz:
Allow modals created with reveal to be accessible for tabbing by setting tabindex-attribute.


rafibomb:
Nice!

PR #6307, opened 2015-02-20
Fixed the close tag </strong>

imabdulsal:


abdullahsalem:
I will separate my commits into different branches to be more organised

Issue #6297, opened 2015-02-14
Reveal modal bug with no animation & scrolling


mitja-p:
If you do not use animation and have a page where you scroll the position of the modal dialog is from the top of page instead of from the current scrollTop.

No animation code:
reveal: {
  animation: 'none',
  animation_speed: 250,
  close_on_background_click: true,
  dismiss_modal_class: 'close-reveal-modal',
  bg_class: 'reveal-modal-bg',
  root_element: 'body',
  bg : $('.reveal-modal-bg'),
  css : {
    open : {
      'opacity': 0,
      'visibility': 'visible',
      'display' : 'block'
    },
    close : {
      'opacity': 1,
      'visibility': 'hidden',
      'display': 'none'
    }
  }
}

For now I have fixed for myself using this code:
        css.top = $(window).scrollTop() + el.data('css-top') + 'px'; // ADDED LINE
        var animData = getAnimationData(settings.animation);
        if (!animData.animate) {
          this.locked = false;
        }
        if (animData.pop) {
          css.top = $(window).scrollTop() - el.data('offset') + 'px';
          var end_css = {
            top: $(window).scrollTop() + el.data('css-top') + 'px',
            opacity: 1
          };

          return setTimeout(function () {

rafibomb:
More info http://foundation.zurb.com/forum/posts/22812-reveal-modal-bug-with-no-animation--scrolling-version-511

jeaniec:
Fixed with #6309.

PR #6309, opened 2015-02-20
Fix reveal height at different vertical scroll positions

cpg1111:
Hi there, this is my fix for issue #6297 so I figured I'd submit a pull request for it.


ClemensSchneider:
Wondering if  `modal.scrollY` was the intended property as it results in NaN for me, too..

PR #6313, opened 2015-02-21
Remove non-namespaced JS events

### LINKED ISSUE #5330: Deprecate Non-Namespaced JS Event Triggers in 5.4

javiercejudo:
Fully remove non-namespaced JS events. I didn't really know what to say in the docs' deprecation notices. Let me know if you'd like to change the wording.

I noticed only abide has tests regarding events. I didn't want to add tests for that in this PR, but might add tests separately.

And an open question: is this necessary? https://github.com/zurb/foundation/blob/624c2e35facc1d5b98bc3cacf8d4ef936f427678/js/foundation/foundation.reveal.js#L179 Maybe the check for namespace in fefc24c03b64438835f2859d18c248b9a723484e can be removed as well.

closes #5330

### DIFF
--- a/doc/pages/components/abide.html
+++ b/doc/pages/components/abide.html
@@ -83,7 +83,7 @@ If a submit event is fired, a `valid.fndtn.abide` event is triggered when the fo
 
 <div class="panel">
   <p><strong>Deprecation Notice</strong></p>
-  <p>Previous versions of the abide plugin emitted un-namespaced `valid` and `invalid` events, however, these have been replaced by the namespaced `valid.fndtn.abide` and `invalid.fndtn.abide` events. The un-namespaced events will be fully deprecated when Foundation 5.4 is released.</p>
+  <p>Previous versions of the abide plugin emitted un-namespaced `valid` and `invalid` events, however, these have been replaced by the namespaced `valid.fndtn.abide` and `invalid.fndtn.abide` events. The un-namespaced events have been fully deprecated.</p>
 </div>
 
 You can bind to these events and fire your own callback:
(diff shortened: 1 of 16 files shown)


oller:
Is anyone else encountering issues with the namespaced events not firing?

I had no trouble prior to updating to 5.5.2.

Specifically with the abide `invalid.fndtn.abide` and `valid.fndtn.abide` events.  I had no troubles prior to the namespace refactoring, when abide simply triggered `invalid` and `valid` events.

javiercejudo:
Just to be sure, are you listening on the namespaced events on the form element? Just checked and those seem to work. Notice that un-namespaced (and undocumented?) `valid` and `invalid` events are triggered on the individual inputs/selects... to imitate the native events that would happen without novalidate (https://developer.mozilla.org/en-US/docs/Web/Events/invalid)

oller:
Hi @javiercejudo, that's right I am listening on the form element. 

I've had to roll back to 5.5.1 version of abide, there's quite a few changes to that file in the latest bump and I'm not sure what's causing teh exact problem.  Going to try and debug a little more now.  There's certainly been some regression in abide with 5.5.2 - it may not be related to the event namespacing.

In my instance, after an invalid event is fired, a subsequent valid event is not fired when the form is valid.

javiercejudo:
At http://foundation.zurb.com/docs/components/abide.html, adding 
```js
$('#testForm').on('invalid.fndtn.abide', function() { alert(0); });
$('#testForm').on('valid.fndtn.abide', function() { alert(1); });
```
in the console and manipulating the first form seems to do what it should, and it is using foundation 5.5.2. It'd be great to see a JSFiddle/CodePen or similar showing the issue.

oller:
I agree, I can't recreate on the docs page.  I'll investigate further and update this thread with any findings, or create an issue if unrelated to the namespacing.

Thanks for the feedback @javiercejudo!

Issue #6308, opened 2015-02-20
Abide `valid.fndtn.abide` not working


trutty:
Hi there,

I am using Foundation 5.5.0 from [this CDN](https://cdnjs.cloudflare.com/ajax/libs/foundation/5.5.0/js/foundation.min.js)

I noticed that the abide example from the [docs](http://foundation.zurb.com/docs/components/abide.html#events) does not work. The HTML is:

    <form data-abide="ajax" id="myform">
    <div class="name-field">
        <label>Your name <small>required</small>
            <input type="text" name="users_name" required pattern="[a-zA-Z]+">
        </label>
        <small class="error">Name is required and must be a string.</small>
    </div>
    <button type="submit">Submit</button>
    </form>

JS:

    $('#myform').on('valid.fndtn.abide', function() {
      // Handle the submission of the form
    });

Instead of `valid.fndtn.abide` I had to use `valid`, which is deprecated since version 5.4.

regdoug:
I can't reproduce your issue, even using the same CDN link you supplied. http://codepen.io/regdoug/pen/ZYoJPX

rafibomb:
Looks like a duplicate of https://github.com/zurb/foundation/issues/5392#issuecomment-71432576

We'll close it here in favor of that one.

PR #6317, opened 2015-02-23
Added tests for namespaced abide events

regdoug:
While looking at #6308 I noticed that the test cases for abide.js didn't test for the new namespaced events. This PR fixes that by adding new expectations to the tests.


regdoug:
Whoops, didn't see #6313

PR #6322, opened 2015-02-25
Add $info-color to possible background colors for topbar buttons

MarcGuay:
There isn't an obious reason for this class to be excluded so I'm assuming it's desired.


MarcGuay:
Build seems to fail for an unrelated reason "abide: basic validation should trigger correct events for all required fields FAILED"

PR #6324, opened 2015-02-26
Move aria label into $include-html-nav-classes conditional

jaiew:
If $include-html-nav-classes is set to false the [aria-label="breadcrumbs"] css was still be output


rafibomb:
Thanks!

PR #6325, opened 2015-02-26
(chore): cleanup structure of _offcanvas.scss

jaiew:
Move the mixins above the css classes and also move the .right-submen and text-direction inside the exports


tdhartwick:
@jaiew Thanks!

PR #6335, opened 2015-02-27
square-grid possibility added

EasterPeanut:
Square columns are possible with this little added code. A div with class "inner-square" can be wrapped with for example "large-square-6". Maybe it's useable for others. The content goes in the "inner-square" div.


tdhartwick:
@EasterPeanut Thanks for pointing this out. It's an interesting mechanic, but we think the use-cases for this is on the rarer side. We'll be able to point people to this for future reference however. Thanks!

PR #6339, opened 2015-02-28
SingleQuotess: Prefer single quoted strings

marcobiedermann:
SingleQuotess: Prefer single quoted strings

See: http://sass-guidelin.es/#strings


rafibomb:
Thanks! This reminds us to delete the toolbar file.

PR #6347, opened 2015-03-02
adding decimal place to breakpoints/ranges in both _globals.scss and _se...

sir-sharkey:
...ttings.scss as '.063' ends up leaving a tailing .008 in the calculation, and Chrome rounds up (meaning the 641 med range won't kick in until 642, large won't kick in until 1026, etc; this creates some really messed up displays right at the breakpoints as of right now

### DIFF
--- a/scss/foundation/_settings.scss
+++ b/scss/foundation/_settings.scss
@@ -152,10 +152,10 @@ $include-html-global-classes: $include-html-classes;
 // - - - - - - - - - - - - - - - - - - - - - - - - -
 
 // $small-range: (0em, 40em);
-// $medium-range: (40.063em, 64em);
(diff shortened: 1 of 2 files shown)


rafibomb:
I think this will help, thanks!

jsharkey90:
No problem, I'm glad to help! This was killing me for a while as I wasn't sure if I was somehow applying my own styling and messing it up, but then I decided to check out settings/globals and noticed that the breakpoints had an extra tailing .008 at the end when calculated. I didn't realize it til now but I guess all browsers will round that .008 up to the next pixel vs round down, as this issue appeared in all browsers for me. Glad it's fixed!

PR #6351, opened 2015-03-03
Add some love to breakpoints

karland:
1. Add breakpoints for media queries, which are a) easier to change and can b) be used as variables elsewhere.
2. Make them `rem`-compatible and more consistent with foundation. This also makes them more meaningful, i.e. `rem-calc(1024)` is easier than `64em`

### DIFF
--- a/doc/pages/media-queries.html
+++ b/doc/pages/media-queries.html
@@ -8,7 +8,7 @@ title: Media Queries
 
 <h3>Basic</h3>
 
-Media queries are built using ems in Foundation.
+Media queries are built using rems in Foundation.
 
 <h4>CSS</h4>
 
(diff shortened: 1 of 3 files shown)


sdhull:
I like it, personally. It's small but definitely more readable :+1:

khanh:
I like this change too :)

rafibomb:
We like the idea because this is closer to what we're doing for the next version of Foundation. 

Concerning the switch to rem's, has anything changed in the way these work in media queries? According to http://blog.cloudfour.com/the-ems-have-it-proportional-media-queries-ftw/ ems are safer. Also has this issue really been resolved in webkit? https://bugs.webkit.org/show_bug.cgi?id=78295

Doesn't appear to be...

karland:
Good point. I did not have any problems with rem's in media-queries, but now I double check. However, the idea is not really dependent on rem's. If foundation still offered `em-calc`, which was deprecated, I would have probably taken `em-calc`. My idea was not so much about rems or ems but to have a more intuitiv approach to setting breakpoints. 

If you want to "rescue" the idea I see two solutions to the problem:
+ re-vive the deprecated `em-calc` only in this context until the bug is fixed
+ write a special purpose function, e.g. `bp-calc` that translates px in ems (which is just reusing the old em-calc under a different name)

If you do not want to do anything either, I am happy to close the issue.

rafibomb:
I discussed it with the team. It's a step in the right direction. Let's get this changed to EM's and we'll merge it in.

karland:
Fine. So, I am going to issue a new PR.  I am going to write a new function `bp-em-calc` to make clear that it is for breakpoints only and to not re-introduce a deprecated function. When the rem-bug in media-queries is fixed, this function can then also be deprecated. Would you be happy with this?

karland:
New [PR #6439](https://github.com/zurb/foundation/pull/6439)

PR #6356, opened 2015-03-03
Only apply $topbar-dropdown-bg to dropdown li's

MarcGuay:
Otherwise the property is applied to all topbar list items, rather than specifically to dropdown list items as the variable name describes.  I'm not familiar enough to know if this is a clean solution but I hope that it illustrates the problem.


rafibomb:
Thanks!
This variable is only applied on the mobile styles. Moving it to the `<li>`'s removes the mobile dropdown background color. Are you trying to address a naming confusion, or something else?

We'll close this PR.

MarcGuay:
I can't reproduce the original problem I had.

PR #6365, opened 2015-03-05
Accessibility for accordions

Owlbertz:
Added a bunch of ARIA attributes to the accordion JavaScript and added an Accessibility chapter to the docs, including an example with enhanced ARIA attributes and roles.


rafibomb:
Thanks!

PR #6370, opened 2015-03-06
Adding default split button border color and making it so the value passed actually affects the color of the border.

grgcnnr:
Currently, $span-border in the split button mixin is documented as accepting a color to use as the divider however any truthy value sets the border color to rgba(255,255,255,.5)

This PR adds a default $split-button-span-border-color and sets the split button border color to whatever is passed into the mixin.


grgcnnr:
Sorry if my PR is flaky,first one here...

rafibomb:
Got the merge conflicts fixed, thanks for the PR!

grgcnnr:
:D

PR #6373, opened 2015-03-06
Make Equalizer Responsive

tyrantkhan:
#4882 Made Reference to this

The issue comes up if you have set equalize_on_stack to true.  I opted for a solution where you set the media query on the parent, and not on a per element  basis. 

Maybe there is value in having it there for both objects, but I'm not sure that's feasible given the current implementation.  I was not looking to do a whole rewrite.

Finally, I would have liked to put the value in just the data-equalizer attribute, but Nested elements are actually using that space, so i opted for a new attribute data-equalizer-mq. If you set it to some non-existant query like `data-equalizer-mq="only-megalarge"`, the plugin will just ignore your media query request and continue as if you passed no media query.

### DIFF
--- a/doc/pages/components/equalizer.html
+++ b/doc/pages/components/equalizer.html
@@ -64,7 +64,7 @@ You can create an equal height container using a few data attributes. Apply the 
 
 ## Responsive Equalizer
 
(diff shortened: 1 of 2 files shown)


khanh:
i messed up my fork a bit by doing the PR off master, so i purged it, i'm going to resubmit this again in a bit.

PR #6377, opened 2015-03-07
#6352 input-transition was hardcoded

### LINKED ISSUE #6352: Hard Coded single-transition in forms
When $input-include-glowing-effect is false, we've got a hard-code single-transition. I know i could just override the class, butIt would be nice if we could get some variables to customize it in settings.

@include single-transition(all, 0.15s, linear);

I could probably submit a PR for it, if there is interest.

tyrantkhan:
This resolves #6352


rafibomb:
Thanks! I removed the !default from the settings file and added the new variables to the docs. https://github.com/zurb/foundation/commit/4eafb38351a31ed111bf348d943ae6a9575c856f

PR #6381, opened 2015-03-09
Hover colour fix for topbar

Pleochism:
Changed the hover colours to be different from the default background colour, and as close to the original colour as possible (ie. #272727)


rafibomb:
That's good! Can you also update the settings file and docs? Thanks!

Pleochism:
I've updated the settings. Not sure what changes to make in the docs, since there wasn't a mention of hover colouring even before it was accidentally removed. Perhaps just a line stating that there are colour changes on hover and they can be changed in the settings?

rafibomb:
Thanks, I'll update the docs. It's the variables at the bottom of each page.

rafibomb:
https://github.com/zurb/foundation/commit/d38d038eabd3a6f2f5a7eca1efce846aa2c44ffc

PR #6382, opened 2015-03-09
Issue #6371 - fix accordion mixin when using custom class names

### LINKED ISSUE #6371: Accordion fails to expand when clicked when using the mixin
This commit https://github.com/zurb/foundation/pull/5884 broke the accordion mixin because it hard codes the class `.accordion-navigation` into the click selector in the JS. 

If you follow the Foundation documentation for creating an accordion mixin by making your own class names, they fail to expand when clicked on.  
See http://foundation.zurb.com/docs/components/accordion.html#semantic-markup-with-sass-mixins

davegross:
This fixes #6371. Since the docs show that you can create an accordion mixin with custom class names, the JavaScript shouldn't hard code `.accordion-navigation` in the selectors.

### DIFF
--- a/js/foundation/foundation.accordion.js
+++ b/js/foundation/foundation.accordion.js
@@ -23,24 +23,24 @@
       var S = this.S;
       S(this.scope)
       .off('.fndtn.accordion')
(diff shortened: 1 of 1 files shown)


rafibomb:
Sweet, thanks!

PR #6389, opened 2015-03-11
Fix markdown syntax-highlighting at javascript-utilities doc page.

guilhermepontes:
Today we have a markdown syntax-hightlighting problem at [javascript-utilities](http://foundation.zurb.com/docs/javascript-utilities.html). <br />
captura de tela 2015-03-11 as 4 17 56 pm

I've just closed the tag.


rafibomb:
Thanks!

PR #6407, opened 2015-03-20
Exclude hidden input fields from validation.

briemens:
Exclude hidden input fields from validation. 

Validation broke for me when using an wysiwyg editor that injects a hidden field. The error message element (<small class="error" />) is shared between an textarea and hidden input field where the hidden field makes the actual error message disappear. 

Can't think of a real world case for this behaviour and if there is a use for it create an attribute or custom validator to enable hidden field validation.


rafibomb:
Thanks!

PR #6412, opened 2015-03-23
Fix top-bar style for select

stenlyk:


rafibomb:
Makes sense, thanks!

stenlyk:
You are welcome :)

PR #6421, opened 2015-03-25
Replace display with a mixin

### LINKED ISSUE #6358: `.hide` class no longer works on `.error` elements
I've just updated from Foundation 5.4.7 to 5.5.1 and I had lots of `small.error` elements in my forms that I always had rendered but added and remove `.hide` class via JS when necessary (it's a Single Page App).

I was looking for a fix at first, but I'm not this familiar with foundation code base to go straight to a Pull Request. With some pointers from you guy's I'll be happy to be the one fixing that.

In `_forms.scss:306` there's the reason:
```sass
@mixin form-error-message($bg:$input-error-message-bg-color) {
  display: block;
  ...
}
```
That's included in `small.error` in `_forms.scss:534`
```sass
span.error, small.error {
  @include form-error-message;
}
``` 
Therefore `display: block` in `small.error` takes precedence over `display: none` of `.hide`

For now I can monkey-path this with such code in my project-specific stylesheet:
```sass
small.error.hide {
  display: none;
}
```
But definitely it's not the final fix for this problem. Maybe a mixin like this would be a solution:
```sass
@mixin display-block {
  display: block;
  &.hide {
    display: none;
  }
}
```

kruszczynski:
Fixes #6358


gakimball:
#6358 was closed as being not an issue (just an unfortunate mix-up), so we won't need this fix.

PR #6436, opened 2015-04-02
Fixing tooltip placement for screen sizes larger than medium

yinzara:
When the screen size was larger than medium, the "small" function in foundation.tooltips was returning true causing all tooltips to be placed at 12.5em left margin from the page.

### DIFF
--- a/js/foundation/foundation.tooltip.js
+++ b/js/foundation/foundation.tooltip.js
@@ -233,8 +233,11 @@
     },
 
     small : function () {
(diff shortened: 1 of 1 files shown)


zurbchris:
I think that could all be replaced with `Foundation.utils.is_small_only()` as it will return a boolean based on the same parameters you're checking. If you could check that out and resubmit, I'll have a look, but based on just glancing at your PR, I think that would be the ticket.

kball:
This PR has long gone stale.  If this is still an issue please reopen against the V5 branch.

PR #6439, opened 2015-04-04
Add some love to breakpoints

karland:


rafibomb:
Thanks! Ok, we're ready to pull this in. As long as it's backwards compatible. Can you rename the bp-em-calc to just em-calc and we'll pull it in?

karland:
I do not understand. Currently `em-calc` is a deprecated function and the code in `_functions.scss` reads like this:

```scss
    // OLD EM CALC
    // Deprecated: We'll drop support for this in 5.1.0, use rem-calc()
    @function emCalc($values) {
      @return rem-calc($values);
    }
    
    // OLD EM CALC
    // Deprecated: We'll drop support for this in 5.1.0, use rem-calc()
    @function em-calc($values) {
      @return rem-calc($values);
    }
```

If I use the current `em-calc` for breakpoints I re-introduce REMs, but we did not want REMs for breakpoints. So do you want me to replace the current `em-calc` function? In case YES, do you want

1. the code for the old `em-calc` function? In this case, can you point me to the old code before it was deprecated?
2. the code from `bp-em-calc`?
3. What about the statement "// Deprecated: We'll drop support for this in 5.1.0, use rem-calc()"?

Thanks.

rafibomb:
You can keep your code and just rename the function to em-calc. We'll remove the deprecation warning. Cool?

karland:
Can't detect any errors. Started from scratch. Result is in [PR 6502](https://github.com/zurb/foundation/pull/6502).

PR #6454, opened 2015-04-09
Add retina media query to scss.

razor-x:
I've created a `$retina` media query variable based on the one used by Interchange.

Not sure how much more to add to the docs for this one, or where else this variable might need to be referenced in the code or docs.


gakimball:
That should be good, thanks for adding it!

Issue #4710, opened 2014-03-17
F6 Wishlist


rvinay88:
- [x] Incorporate Sass 3.3 features such as maps and provide a settings API for all Sass components
 - [x] .editorconfig #4720
 - [ ] Unit tests for Sass https://github.com/ericam/true
 - [ ] Support both em's and rem's using a generic calc function #4633 
 - [ ] Experimental advanced grid system - Singularity, Susy, Zen Grids rely more on mixins, make less assumptions and do not constraint design #4134 
 - [ ] Animation library for use with all js plugins - http://h5bp.github.io/Effeckt.css/
 - [ ] Megamenu - http://geedmo.github.io/yamm3/#
 - [ ] Labels with close buttons (like alerts) - Useful in search contexts
 - [x] Accessibility 

Will add more 
Edit: Added accessibility

rnaud:
Less overwriting of base html elements. I'd rather have a table.table {} definition than a table {} one that makes it hard to use table elements with the default behavior.

Also +1 on no reflow. 

My last one would be making more elements able to load content remotely. Modals can do it but tooltips cannot.

gakimball:
We've gotten a number of requests for date and time pickers. #6176

cautionbug:
@gakimball YESYESYESYES! There is a calendar picker out there that *mostly* works (i've had to do some monkeywrenching with how it handles empty values), but it might be a decent starting point:

https://github.com/jockmac22/foundation_calendar_date_time_picker

rafibomb:
@cautionbug Thanks for the tip! We'll check it out!

cautionbug:
@rafibomb i may set up a fork of it, since it doesn't seem there have been any significant changes in quite a while. i'll commit what i did there & follow up here.

So far, the list is looking really good. Ambitious, well-considered... looking forward to what actually makes it into the road map.

anish274:
is foundation planning for material design? I love zurb but my website needs material design. Any Solution?

ryanmargheriti:
Your website doesnt *need* material design @anish274. Nor do developers *need* another material design theme. 

Use one of these instead: 
- https://fezvrasta.github.io/bootstrap-material-design/
- https://github.com/angular/material

anish274:
@ryanmargheriti but my site is already implemented in zurb and the 1st link is the material version of bootstrap. So if my site is bootstrap I can convert it to material design using above link ( I think not sure).
I am totally unaware of angular.

[...]

PR #6485, opened 2015-04-20
Include package.json in dist/assets

mattdsteele:
Module distribution was already discussed a little in #4710, but I'd like to kick-start one portion of it: make Foundation installable via npm. (My use case: I'm attempting to wean myself off Bower and install all dependencies via npm)

[npm allows Git repositories to be used as endpoints](https://docs.npmjs.com/files/package.json#github-urls), but the repo requires a valid `package.json`. This commit simply copies the existing `package.json` file into `dist/assets`.

You could publish specifically to npm but it's not strictly necessary (perhaps wait until F6?). For now, one could install Foundation as an npm dependency with this configuration:

```
"dependencies": {
  "foundation": "zurb/bower-foundation"
}
```

### DIFF
--- a/grunt/config/copy.js
+++ b/grunt/config/copy.js
@@ -23,7 +23,7 @@ module.exports = {
       dest: '<%= paths.dist %>assets/scss/',
       filter: 'isFile'
     },{
-      src: 'bower.json',
+      src: ['bower.json', 'package.json'],
       dest: '<%= paths.dist %>assets/'
     }]
   }


gakimball:
I think we're going to get this in for our next deploy, which should be this week. We can start publishing to npm, but the directory structure will change with version 6.

gakimball:
Also went ahead and changed the package name to `foundation-sites`, since somebody is using the `foundation` name on npm, but we need to push the new naming conventions for the frameworks. We'll do the same for Bower with version 6. 2a71b1c2f403d2cfb3a0d62168a79a5733b7a7d9

mattdsteele:
Thanks! :+1:

gakimball:
It's live now with version 5.2.2.

```bash
npm install foundation-sites
```

PR #6498, opened 2015-04-24
Removing duplicated android top-bar fix

oller:
Somehow this block was duplicated with 2 separate commits, removing one...


rafibomb:
good catch!

PR #6502, opened 2015-04-25
Add some love to breakpoints (v3)

karland:
Changes within [PR 6439](https://github.com/zurb/foundation/pull/6439) led to a Travis error.

### DIFF
--- a/doc/pages/media-queries.html
+++ b/doc/pages/media-queries.html
@@ -8,7 +8,7 @@ title: Media Queries
 
 <h3>Basic</h3>
 
-Media queries are built using ems in Foundation.
+Media queries are built using ems in Foundation. At some point in time we will move to rem's, however, at the moment not all browsers like rem's in media queries.
 
 <h4>CSS</h4>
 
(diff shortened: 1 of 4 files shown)


karland:
From looking at the details the error seems to stem from `foundation.equalizer.js`? Foundation was failing from before this PR.

gakimball:
This should fix it: b2a8fd483c977b0ba8948b3a8035fe0974adc542

Try rebasing or merging.

gakimball:
@karland This may have introduced a bug related to how breakpoints are calculated when the base font size is changed (see [here](http://foundation.zurb.com/forum/posts/25180-update-v-550-to-552-calculates-wrong-breakpoints)). Any ideas?

karland:
@gakimball Thanks for pointing me to this is issue.

Stefan4UK:
@karland: Thanks for your effort, but I think in line 139 you made a little typo. I would suggest:
-  $remValues: rem-calc($values, $base-value: $rem-base);
+  $remValues: rem-calc($values, $base-value);

Unfortunately the breakpoint issue isn't completely solved as I point out here: http://foundation.zurb.com/forum/posts/25180-update-v-550-to-552-calculates-wrong-breakpoints
Your fix works within classes but strangely doesn't have any effect on the media-queries. Any ideas why?

karland:
@Stefan4UK I have tested what you said. Here are my conclusions:

1. The "typo" that you identified is not really a typo. The code should work fine as it stands, however, I agree, going by your suggestion is less redundant. Here is the logic:
If you call `em-calc()` without a `$base-value` which normally is the case, then `$base-value` is set to `$rem-base`. In this case the argument `$base-value` to the function call of `rem-calc($values, $base-value: $rem-base)` is defined and set to a value and therefore will never use the default. So the default value in `rem-calc()` is somewhat redundant, but should not get in the way. I would assume, that your test case from [here](http://foundation.zurb.com/forum/posts/25180-update-v-550-to-552-calculates-wrong-breakpoints) works the same with or without the "typo". Let me know, if this is not the case. 

2. I tested the code again and it works fine with me. So my hunch is that you might have the wrong order in your `_settings.scss`. You must specifiy `$rem-base` BEFORE the `@import 'foundation/functions'` statement. Otherwise `em-calc()` and `rem-calc()` use the default `$rem-base: 16px !default`. The other thing I can think of is that you did not properly specifiy the breakpoints:

```scss
$small-breakpoint:  em-calc(640);
$medium-breakpoint: em-calc(1024);
$large-breakpoint:  em-calc(1440);
$xlarge-breakpoint: em-calc(1920);
```

For a `$rem-base: 14px;` I get the following values for media-queries min-width as expected:

+ medium (641px): 45.78571em
+ large (1025px): 73.21429em

Let me know how it goes...

Stefan4UK:
@karland: Thanks a lot! You finally found the reason for why it was still not working for me :). I had my rem-base inlcuded after the _function.scss include in the _settings.scss file. What a stupid mistake...

Therefore without even knowing it my breakpoints where 560px,896px,1260px and 1680px. These are all default values multiplied by *(14/16).

Because of my wrong order the rem-base was always 16px...

By using the values I calculated above my design finally looks like it should be ;).
Do you think using these different breakpoints is a problem? Or is it not important which breakpoints are set as long as the design looks great on every screen width?
Many thanks again for your great commit and effort to help me.

karland:
@Stefan4UK You're welcome! I am glad that you got your problem resolved. It would be nice if you could comment on the issue solution in the Foundation forum to complete the thread.

Other than that I think the default breakpoints from Foundation are good default values. However, I never use them because I want to have more differentiation on smaller smartphone/tablet screens. Here are my defaults:

```scss
$include-xl-html-grid-classes: true;

$small-breakpoint:  rem-calc(320);
$medium-breakpoint: rem-calc(600);
$large-breakpoint:  rem-calc(800);
$xlarge-breakpoint: rem-calc(1024);
```

PR #6512, opened 2015-04-29
Reveal: Fix top positioning when no animation is used.

choladay:
As reported here http://foundation.zurb.com/forum/posts/22812-reveal-modal-bug-with-no-animation--scrolling-version-511

When using `animation: 'none'` opening the Foundation reveal modal, the modal does not adjust to the current position of the window and always appears `$reveal-position-top` from the top of the page.  

Inserting the the inline `top` CSS property is done along with animation effects.  This will add the property should no animation effect be used.

### DIFF
--- a/js/foundation/foundation.reveal.js
+++ b/js/foundation/foundation.reveal.js
@@ -352,6 +352,11 @@
           }, settings.animation_speed / 2);
         }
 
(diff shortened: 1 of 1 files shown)


zurbchris:
This issue should be resolved, thanks for contributing!

PR #6514, opened 2015-04-30
Added the ability to stack progress bars.

cwperry:


rafibomb:
Thanks for the PR! We're not planing to have this feature in Foundation Sites 6 but it will help people on 5. If you're cool with it, we can add a Foundation 6 version of it to the building blocks - http://zurb.com/building-blocks

cwperry:
Sounds good to me.

PR #6517, opened 2015-04-30
Update foundation.reveal.js to ensure target is not undefined in open event

chrisjsherm:
Ensure open event has a target before accessing the 'data-reveal-id' attribute. I ran into target === undefined after calling $('#myModal').foundation('reveal', 'open');


rafibomb:
Thanks Chris! We'll test this soon and merge it in.

PR #6523, opened 2015-05-04
Add active_class to initially opened accordion-navigation

fergaldoyle:
An active_class is added to .accordion-navigation items when the user clicks on an accordion, but is not added to the initially active accordion.

The active_class should be added both initially and after a user opens an accordion to allow for consistent styling of active accordions.


rafibomb:
That's not what I'm seeing here. If I load the docs page and inspect, the active class is present.
screen shot 2015-06-19 at 3 54 56 pm

Can you explain what your seeing differently and how it causes an issue?

rafibomb:
Thanks!

PR #6537, opened 2015-05-08
Corrected kitchen sink layout and joyride functions

jennz0r:
Just some small changes since I noticed the `Start Joyride` button wasn't working on the `kitchen-sink.html` page.


rafibomb:
Thanks!

Issue #6248, opened 2015-02-03
Top Bar mobile_show_parent_link only appears at Small breakpoint


ricardo-lewis:
The JS generated markup:
```
<li class="parent-link show-for-small-only">
<a class="parent-link js-generated" href="/products-services">Products &amp; Services</a>
</li>
```
It would be nice if the Parent Link would show at the ```$topbar-breakpoint``` instead of at Small only.

rafibomb:
Can you submit a PR because this would help others?

rafibomb:
The parent links visibility needs to match the top-bar breakpoint, not the grid breakpoint.

jeaniec:
@gakimball Do we have a class that can signify a 'topbar breakpoint show/hide' behavior?

ricardo-lewis:
@jeaniec that's a good idea – I can already think of places outside the Top Bar where I would use this. Quite a few elements change appearance at the TB breakpoint...

vita10gy:
I've recently added this as a breakpoint, and came here to add this issue of the parent link as a request .

I'd be willing to make a pull request if those in the know think there's a good chance they'd accept it. 

What would be a good name for this breakpoint? Is there "official" foundation lingo for the full desktop menu vs the mobile menu?

jeanieshark:
@vita10gy If you want to submit a pull request and all looks good, we'd be happy to pull it in. There's no official lingo, but exporting a class something along the lines of "hide-for-topbar-mobile" or "hide-for-topbar-desktop" should be fine. After you submit the pull request, the team will look over it and see if there are any changes that we might want to commit on our end as well.

vita10gy:
I also added this breakpoint in to the grid so I could do something the equiv of topbar-desktop-5 That way I could make the header above the menu play ball with the menu breakpoint and visibility classes we're talking about. 

Would that interest people, or it that getting too specific? 

Also, one "catch" to this change we're discussing is that, as far as I can tell at least, though truth be told I didn't spend a ton of time on trying alternatives: the breakpoint needs to be defined with the others in _globals and not _topbar. 

Is that too "breaking" to justify for this?

rafibomb:
Thanks for pointing this out. Ideally the visibility would match the top-bar breakpoints. We're not adding any features to Foundation 5 as we're concentrating on getting Foundation 6 out.

PR #6538, opened 2015-05-08
Top Bar Breakpoint Visibility

vita10gy:
Discussed in #6248 

Add visibility classes based on top-bar's breakpoint.

Make top-bar show and hide parent-links based on these new visibility
classes instead of medium.

Add grid classes based on top-bar breakpoint. (Optional  - off by
default)

### DIFF
--- a/js/foundation/foundation.topbar.js
+++ b/js/foundation/foundation.topbar.js
@@ -378,7 +378,7 @@
         if (!$dropdown.find('.title.back').length) {
 
           if (settings.mobile_show_parent_link == true && url) {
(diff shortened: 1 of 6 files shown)


vita10gy:
If this gets pulled as is I'd also be willing to update any documentation, but no point in doing that if things might change.

rafibomb:
Hey, thanks for the PR. Sorry this has sat a while. We're working on getting Foundation 6 out for public launch therefor we're only merging bug fixes. New features are not on the roadmap because we're making sure the final release of Foundation 5 is very stable.  We'll close this here. Focusing on bugs and Foundation 6. Thanks!

vita10gy:
I think one could argue the fact that topbar doesn't use the topbar breakpoint in a decision point and instead has an entirely different one hardcoded into it is indeed a bug.

PR #6544, opened 2015-05-11
fix float if last-column is true

lancedikson:
If you write something like:

```
@include grid-column($columns: 4, $last-column: true)
```

you will have this CSS:

```
.main__header-btns {
  padding-left: 0.9375rem;
  padding-right: 0.9375rem;
  width: 33.33333%;
  float: right;
  float: left; }
```

where last float is wrong. Solution for fix this is one of both to set `$float` as `false` by default or don't add some `float` if `$last-column` was defined as `true`.

Now I have correct CSS:

```
.main__header-btns {
  padding-left: 0.9375rem;
  padding-right: 0.9375rem;
  width: 33.33333%;
  float: right; }
```


rafibomb:
Thanks!

PR #6549, opened 2015-05-12
Updating postfix mixin call since additional parameter added

oller:
Fixes the problem noted in https://github.com/zurb/foundation/issues/6546

Suspect problem brought in with merge of https://github.com/xtianus79/foundation/commit/ed777bbbac2d0aa7262b217e93e9eef35c28ddd6

Postfix mixin had additional parameter added, but calls to the mixin not updated.


rafibomb:
Thanks!

PR #6554, opened 2015-05-13
Add ability to adjust slash position in bread crumb 

blacklizard:
Add ability to adjust slash position in bread crumb

### DIFF
--- a/scss/foundation/_settings.scss
+++ b/scss/foundation/_settings.scss
@@ -6,7 +6,7 @@
 
 // Table of Contents
 // Foundation Settings
-//
+//brea
 // a. Base
 // b. Grid
 // c. Global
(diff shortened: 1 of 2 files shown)


zurbrandon:
I'm wondering the use case here. Does it have to do with changing text size? if so i'm wondering if we can try and do this better through sass math as opposed to a variable.

blacklizard:
@zurbrandon hi there.. The use case here is when you use different font, for example google font, for my experience, the slash is not aligned with the text, its either up or down, that's the reason i create this PR. Also this helps when you use different separator other than slash. Its not for changing text sizes :)

SalahAdDin:
Thanks man :D, i have the same problem, but, when i change the slash with &raquo.

Issue #6547, opened 2015-05-12
custom validation executing 3 times


caycek:
A custom validation I wrote is executing 3 times consistently. When the content changes. I'm thinking this is a keyup validation check error, but I haven't found out yet. I'm using the Abide Foundation 5. I'm not sure right now what other information to provide, but if anyone has had this problem or resolved it I would be grateful to hear about it.

caycek:
[6406](https://github.com/zurb/foundation/pull/6406)

Using this did fix the issue, but caused a MAJOR lag in connection for every validation. The current validation is speedy. I just don't need it executing more than once per check.

caycek:
Upgraded to 5.5.2 (only with abide) and it has major lag. I will attempt to diagnose where...

rafibomb:
We added this comment in our planning board for the next version. It would good to make this more performant, however still balancing all the features that people are asking for. 

A PR would be valuable to the community, otherwise It will be addressed then. You can see and vote on the proposed updates here: https://trello.com/b/QBPKlQqf/foundation-sites

PR #6558, opened 2015-05-14
known issue "Abide - custom validation slow" #6547

ghost:
I fixed some 4 spaces to 2 spaces and some spelling errors in the code.
nothing major there. But I also fixed the abide to be faster by lowering
the check time from 1second to 1/10th of a second. This resolves the
feeling of lag and does not seem to hinder any other function that I am
seeing. This does not resolve the 3 times execution as of yet. but as of
now does resolve the speed issue.

### DIFF
--- a/js/foundation/foundation.abide.js
+++ b/js/foundation/foundation.abide.js
@@ -94,10 +94,16 @@
         .find('input, textarea, select').not(":hidden, [data-abide-ignore]")
           .off('.abide')
           .on('blur.fndtn.abide change.fndtn.abide', function (e) {
+              var id = this.getAttribute('id'),
+                  eqTo = form.find('[data-equalto="'+ id +'"]').get(0);
             // old settings fallback
             // will be deprecated with F6 release
             if (settings.validate_on_blur && settings.validate_on_blur === true) {
               validate(this, e);
+            }
+            // checks if there is an equalTo equivilant related by id
+            if(typeof eqTo !== "undefined"){
+              validate(eqTo,e);
             }
             // new settings combining validate options into one setting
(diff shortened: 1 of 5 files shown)


caycek:
Trying to fix two problems here with abide... Speed was my first priority. Hopefully you feel this is fixed. I do not see any major drawbacks in the way that I'm using it,  but cannot think of an issue which would cause it to be a poor push. Also, the fix with the equalTo check.... I just realized as I was typing this an error that could happen so that is the Hot fix. The check now goes back and finds any elements in the same form that have the id value set to X if so then it checks their value and so on so forth. 

Questions or concerns please let me know.

rafibomb:
Love the PR! Thanks! Only thing we think should be changed is the timeout. We added a little documentation for it just now, we think the default should be left at 1s. It's a useful feature and not that it's pointed out in the docs (when we deploy) people can set it themselves.

caycek:
That's understandable. I thought that might be the case! Do you want me to change it back to 1s or would you prefer to do that yourself?

Also, I could look into building an option into this build if you wanted me to. Not sure how that would work off the top of my head, but if not it can be left for 6 later. Let me know!

caycek:
@rafibomb Hey hope you're well. It has been some time since this was posted. I wanted to double check if / when you guys do accepting of pull requests. I did my branching incorrectly and only now finally understand how to do branching more correctly. If this has limited your ability to merge this request I can go through and reapply the changes or do what is needed on a separate branch.

Let me know when you can!

Thanks!

rafibomb:
We just need the conflicts cleared and we're good to go!

zurbchris:
Thanks for the contribution!

PR #6564, opened 2015-05-15
Include missing data_options on main Javascript documentation table

jacobjpope:
This includes all the missing data_options that are available on the individual Javascript components documentation pages.  I did not include any that are functions.  Note: some of the existing data options from this documentation are not on the individual components documentation pages.  Dropdowns, Reveal Modal, Tooltips are some examples of that.


gakimball:
Awesome, thanks!

PR #6589, opened 2015-05-29
Adding convenience variables important when topbar breakpoint != medium-breakpoint

Larzans:
This new variables allow for a more convenient media queries for the topbar:

```
@media #{$topbar-only} { ...
```

and

```
@media #{$topbar-up} { ...
```

This is very useful in cases when the tobbar does not break at the same time as medium which is often the case because of the width of the navigation, depending on the number of entries.

At least on my pages the navigation breaks usually independently of the ranges because of its width.


vita10gy:
Something like this is covered in #6538

Larzans:
Indeed, looks like you covered that also.
But as your PR has not been merged yet and includes a lot of other changes, it would be cleaner this way as this PR is only about the media query for the topbar.

rafibomb:
Thanks for the PR! Sorry we didn't get to it sooner. We're working on getting Foundation 6 out for public launch therefor we're only merging bug fixes. New features are not on the roadmap because we're making sure the final release of Foundation 5 is very stable. 

Focusing on bugs and Foundation 6 so we'll close this here. Thanks!

PR #6597, opened 2015-06-02
Fix inline-list margin issue caused by re-ordering of css properties

oller:
The commit https://github.com/zurb/foundation/commit/c65a8e9e1f422b48ce7be9665292e848b188c293 alphabetically sorting a lot of .scss properties has caused some issues.

This is fine when all properties are unique but this has broken the margins on `.inline-list.scss` as the margin properties are deliberately declared twice, with one overwriting the other, in this case the order matters and changing it has broken the margins.

This is especially apparent if you're using a css minification tool.

### DIFF
--- a/scss/foundation/components/_inline-lists.scss
+++ b/scss/foundation/components/_inline-lists.scss
@@ -33,7 +33,8 @@ $inline-list-children-display: block !default;
 // We use this mixin to create inline lists
 @mixin inline-list {
   list-style: none;
-  margin: $inline-list-top-margin auto $inline-list-bottom-margin auto;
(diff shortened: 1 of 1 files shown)


gakimball:
Do you know why margins are declared twice in the first place?

oller:
I do not, this PR is just to revert the order to get it working again.

It looks like you could break the shorthand `margin` declaration to just the individual `margin-top` and `margin-bottom` as the `auto` values will always be overwritten by the dynamically generated side margin values

gakimball:
Right, yeah. Well, for now I'll merge this in as-is. Thanks!

PR #6607, opened 2015-06-04
Deep-Linking Tabs prevent back button in IE from working correctly.

sammarks:
I'm using the tabs plugin in a site and was testing it with Internet Explorer when I noticed that the deep-linking tabs were "breaking" the back button in the browser (creating an infinite loop as described in the pull request I mention below).

I did some Googling and found #6118. I saw that it had been accepted and updated to the latest version of Foundation to test the fix. It worked in Chrome (and I'm assuming other non-IE browsers but did not do tests to verify), but not Internet Explorer.

I re-visited the code and simplified it (getting rid of recording the entry point because it seems that the tabs work fine without having to record the entry point). I did some final testing in IE (as well as Chrome to represent the cooler browsers) and found that it worked fine.

(I also fixed some formatting at the top of the tabs plugin)


rafibomb:
Nice one!

PR #6622, opened 2015-06-11
Default input type

ghost:
Fix to https://github.com/zurb/foundation/issues/6416
"The default type is text, if this attribute is not specified." (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-type)


rafibomb:
Thanks!

Issue #6619, opened 2015-06-10
tfoot is missing in Visibility Class 


neonoxGER:
Hi,

when using .hide-for-small-only or .show-for-small-only or any other visibility class, it doesn't work with tfoot. At the moment, only thead and tbody are supported.
Please add support for tfoot.

vita10gy:
If you need this asap, and you're using the scss version, you call grab _visibility.scss from my pull request there.

neonoxGER:
thx for your mail, I updated my scss yesterday, but thanks :)

best regards
Benjamin

rafibomb:
Closing in favor of PR - thanks!

PR #6623, opened 2015-06-11
add tfoot to visibility classes

vita10gy:
Addresses #6619 

I made tfoot do whatever thead did (with the exception of using "table-footer-group", obviously.)

One big problem I found, however, is that the "only" classes don't work like the standard ones do, but this issue applies to the existing thead and tbody, as well as the new tfoot classes.

https://jsfiddle.net/0fpr9vvw/

show-for-small-only is behaving like show-for-small-up, and show-for-medium-only is behaving like show-for-medium-up

Still, this PR will make tfoot do whatever it does now for the other table classes, which is better than just  forgetting about them for now.

### DIFF
--- a/scss/foundation/components/_visibility.scss
+++ b/scss/foundation/components/_visibility.scss
@@ -41,6 +41,7 @@ $visibility-breakpoint-queries:
     $visibility-table-list: ();
     $visibility-table-header-group-list: ();
     $visibility-table-row-group-list: ();
+    $visibility-table-footer-group-list: ();
     $visibility-table-row-list: ();
     $visibility-table-cell-list: ();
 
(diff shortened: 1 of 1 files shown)


vita10gy:
I'm thinking it might be dueling !importants

vita10gy:
I believe we can get rid of the dance of important! tags by matching the specificity (and moving the none to last). 

thead.show-for-small-only{ show} isn't overridden by .show-for-small-only{none} even though it's above it, because it's more specific. 

I don't have time to flesh this out right now, but I'll see what I can do next week if this is still an open issue. Might be a 10 second "oh duh" for someone who understands what's going on in there better than I.

neonoxGER:
This issue is still open -.- please fix that.

gakimball:
Sorry for the late response, would it be possible for you to open a new PR that targets the `V5` branch instead? Thanks :)

kball:
Closing based on last comment.  Please reopen targeting V5 branch.

Issue #6631, opened 2015-06-17
Sticky Menu Unscrollable


vita10gy:
If a sticky menu is taller than the window you can't scroll to all the items.

zurbrandon:
We're adding this feature to a brandnew plugin for Foundation 6, but since Foundation doesn;t currently support a sticky sidebar we won;t be able to port that over. For long scrollable sections we've had great luck with Sticky Kit in the past.

vita10gy:
This was in reference to topbar.

If the menu is sticky, and taller than the phone when open, you can't get to the bottom items.

PR #6632, opened 2015-06-17
Sticky menu height protection and topbar breakpoint

vita10gy:
Add protection so a menu won't be sticky if it's taller than the screen (generally, once opened).

Otherwise you're never able to get to the end items.

I've also added the topbar breakpoint option (which seems like the most
likely one a person really means) and cleaned the function up to use the
existing small() medium(), and large() functions it was already half
using.

Addresses issue #6631

### DIFF
--- a/js/foundation/foundation.topbar.js
+++ b/js/foundation/foundation.topbar.js
@@ -63,25 +63,32 @@
     },
 
     is_sticky : function (topbar, topbarContainer, settings) {
-      var sticky     = topbarContainer.hasClass(settings.sticky_class);
-      var smallMatch = matchMedia(Foundation.media_queries.small).matches;
-      var medMatch   = matchMedia(Foundation.media_queries.medium).matches;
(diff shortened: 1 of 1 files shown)


vita10gy:
It would be a non backwards compatible change, but I think it aught to be considered that this function take in arbitrary queries (small-up, medium-only, large), and work like the standard queries already do. Great pains are taken to ensure small means "small only" medium means "medium only", etc, even though this is generally the opposite of what is meant.

zurbchris:
At this time, we aren't adding any large features or changes that may affect current usability. We are, however, working hard on Foundation 6, and will keep this in mind. Thanks for contributing!

vita10gy:
I'd file this as more of a bug fix than a "feature". Currently a menu is made sticky even if it's taller than the current window, meaning you can't ever get to the bottom of the menu.

This check makes it so it only applies the sticky if a sticky would fit, otherwise it would scroll as a normal in place menu so you can get to the "missing" options.

PR #6636, opened 2015-06-18
Fix select focus state

serg0x:
I gave the dropdown select boxes the same behaviour, like the normal input fields for the good overall consistency.


gakimball:
Why not select both `:hover` and `:focus`? Adding the focus state makes sense, but we'd also like to keep the hover state most likely.

serglo:
Easy, because standard form elements also have no hover effect see http://foundation.zurb.com/docs/components/forms.html

rafibomb:
I get the visual consistency aspect of it, but isn't a better interaction to have the visual queue on hover?

serglo:
Than we need to add hover states to all other form elements, because right now only dropdown element has hover state.

tdhartwick:
Thanks!

PR #6657, opened 2015-06-26
Removed Deprecated JS Component

ghost:
Orbit is deprecated, so i removed it from humans.txt file. Nothing special

### DIFF
--- a/humans.txt
+++ b/humans.txt
@@ -4,5 +4,5 @@
 
 /* SITE */
   Standards: HTML5, CSS3
-  Components: jQuery, Orbit, Reveal
+  Components: jQuery, Reveal
   Software: Sublime, Git, Sass


SalahAdDin:
Why? Why you deprecated Orbit if Orbit is very useful? I try use other sliders but are very hard, i only need the all slider with the same height, but they put me the slider in different heights :'(

atilkan:
I removed it from humans.txt file. It was just an example. You can use older foundation version if you really need it badly.

gakimball:
Hm, I suppose for that line to make sense, it needs to include *every* JavaScript component, of which there are about a dozen.

I'm not really sure why we have a humans.txt there anyway, though. That file doesn't get copied to the Foundation marketing site or docs, and it's not meant to be used in Git repos. It's supposed to be at the root of a website where a robots.txt would be.

gakimball:
Closing this in favor of just deleting the `humans.txt`, because this repository isn't a website, it's a codebase.

PR #6664, opened 2015-06-30
Make media_queries work when Foundation JS is loaded asynchronously

carey:
This change defers resolution of CSS media queries until they are first used, rather than when the `foundation.js` file is loaded. Apart from a performance improvement from not evaluating unused CSS rules, which is probably too small to measure, this allows reordering and asynchronous loading of Foundation CSS and JS, as described in #5683. This is the kind of change which is recommended by Google PageSpeed's [Optimize CSS Delivery](https://developers.google.com/speed/docs/insights/OptimizeCSSDelivery) and [Remove Render-Blocking JavaScript](https://developers.google.com/speed/docs/insights/BlockingJS) rules.

Using a JavaScript object's `toString()` method to return a media query will always work with `matchMedia`. In the case of the polyfill, the query is concatenated with other strings, so its `toString()` method will be called. For the real function, it takes a `DOMString` as a parameter, which the relevant specs say will be obtained by calling `toString()` on any non-primitive object.

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -10,14 +10,12 @@
   'use strict';
 
   var header_helpers = function (class_array) {
-    var i = class_array.length;
     var head = $('head');
-
-    while (i--) {
-      if (head.has('.' + class_array[i]).length === 0) {
-        head.append('<meta class="' + class_array[i] + '" />');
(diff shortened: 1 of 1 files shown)


gakimball:
Looks solid, and thanks for cleaning up the code while you were at it :)

PR #6675, opened 2015-07-02
New mixin for css transitions

somewebmedia:
This mixin can accept single or multiple transitions. It also accepts 4th parameter for animation delay.

Sample usage:

```
@include transition(background-color, .15s, ease-in, .2s);

// multiple transition properties with same duration and ease
@include transition((color, background-color), .15s, ease-in);

// multiple transitions
@include transition((
  (opacity, .2s, ease-in-out),
  (visibility, 0s, ease, .2s)
));
```

Old mixin single-transition() is left in the code for backward compatibility.

### DIFF
--- a/doc/includes/examples_global_rendered.html
+++ b/doc/includes/examples_global_rendered.html
@@ -72,9 +72,76 @@ $base-line-height: 150% !default;
 // We use this to add transitions to elements
 // $property - Default: all, Options: http://www.w3.org/TR/css3-transitions/#animatable-properties
 // $speed - Default: 300ms
-// $ease - Default:ease-out, Options: http://css-tricks.com/almanac/properties/t/transition-timing-function/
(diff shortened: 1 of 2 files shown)


zurbrandon:
Sounds great! Thanks so much for this!

s-whaley:
There is a conflict when using this with compass/css mixins. Compass/css @include transition() does not use a comma delimited list of params. Using the compass/css3 mixin works in Foundation 5.5.2

error scss/app.scss (Line 124 of bower_components/foundation/scss/foundation/components/_global.scss: Undefined variable: "$_property".)

somewebmedia:
@s-whaley Well they have their own mixin called **transition** so they are clashing https://github.com/Compass/compass/blob/stable/core/stylesheets/compass/css3/_transition.scss#L141

You shouldn't use those two together, instead choose and import which one you want to use.

somewebmedia:
@zurbrandon You're welcome! :)

SalahAdDin:
:+1:

PR #6678, opened 2015-07-04
Fixed some positioning issues of dropdowns and tooltips

msdm:
- Fixed missRight and missLeft calculation bug
- Fixed positioning problems for RTL

### DIFF
--- a/js/foundation/foundation.tooltip.js
+++ b/js/foundation/foundation.tooltip.js
@@ -234,8 +234,6 @@
         }).end();
       };
 
-      objPos(tip, (target.offset().top + target.outerHeight() + 10), 'auto', 'auto', target.offset().left);
-
       if (this.small()) {
         objPos(tip, (target.offset().top + target.outerHeight() + 10), 'auto', 'auto', 12.5, $(this.scope).width());
         tip.addClass('tip-override');
--- a/scss/foundation/components/_tooltips.scss
+++ b/scss/foundation/components/_tooltips.scss
@@ -74,11 +74,11 @@ $tooltip-max-width: 300px !default;
         position: absolute;
         top: -($tooltip-pip-size * 2);
         width: 0;
-        #{$default-float}: $tooltip-pip-size;
+        left: $tooltip-pip-size;
 
         &.rtl {
           left: auto;
-          #{$opposite-direction}: $tooltip-pip-size;
+          right: $tooltip-pip-size;
         }
       }
 
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -293,15 +293,17 @@
           }
 
           //miss right
-          if (t.offset().left + this.outerWidth() > t.offset().left + actualMarginWidth && t.offset().left - actualMarginWidth > this.outerWidth()) {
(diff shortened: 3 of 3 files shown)


gakimball:
Could you explain in more detail the changes you made to fix these issues?

scss/foundation/components/_tooltips.scss
```diff
@@ -74,11 +74,11 @@ $tooltip-max-width: 300px !default;
         position: absolute;
         top: -($tooltip-pip-size * 2);
         width: 0;
-        #{$default-float}: $tooltip-pip-size;
```
msdm: Why do we set the tooltip position based on $default-float value? the positioning is handled in the js file.

js/foundation/foundation.tooltip.js
```diff
@@ -234,8 +234,6 @@
         }).end();
       };
 
-      objPos(tip, (target.offset().top + target.outerHeight() + 10), 'auto', 'auto', target.offset().left);
```
msdm: This line was redundant and could generate some problems in the calculation of "left" value

msdm:
Sure,
**Tooltips**
The scss assignment of left/right positioning based on $default-float was wrong because ositioning is handled in js. Also the removed line from the associated js file was redundant and could generate some problems in the calculation of "left" value.

**Dropdowns**
The calculation of missRight and missLeft in _base() (lines 296 to 305 of foundation.dropdowns.js) was wrong. Also they were coupled to each other with no reason. Added if-condition in this function is to correctly calculate initial "left" value for RTL.
The changed codes in top() and bottom() were for preventing out of screen dropdowns when either missRight or missLeft is true. Notice that if both missRight and missLeft is true, the default position is selected based on LTR/RTL.
For left() and right(), when missLeft or missRight is true respectively, we just need to call bottom() to fallback to the default position.

Behzad63:
Committed changes have fixed positioning issues in right to left mode.

zurbchris:
We pulled in a change to speed up the repositioning for dropdowns, if you pull that in and resubmit this with those changes, I would like to take a look at it.

zurbchris:
I pulled this down, and while it passes testing, it does some weird things it shouldn't. I've take two screenshots, one from the current Foundation Docs page, one from my local dev environment with your PR. Notice the pip positioning is wrong in the first image. 
screen shot 2015-09-04 at 9 54 18 am
screen shot 2015-09-04 at 9 54 28 am

msdm:
Strange. I tested it again right now and evertything seems work just fine on my local dev env. The wrongly positioned pip in the first screenshot can only happen for RTL pages on small screens (when this.small() returns true) because the current implementation of the dropdowns does not locate the pip in small screens according to the target element position. So, please let me know if this screenshot belongs ot an LTR or RTL page.

zurbchris:
LTR, medium window size.

msdm:
I could not reproduce the case. I even downloaded the corresponding Foundation doc page and added the modified foundation.dropdown.js as the last script file to it but the pip was correctly located in the left side.
I wonder how the pip is in the right side in your screenshot because its location direction is always set according to webpage direction (left for LTR and right for RTL).

kball:
This PR has gone stale and is no longer relevant to master.  Please resubmit against the V5 branch.

PR #6680, opened 2015-07-06
Allow easier customization of alternate button bg

joelworsham:
Right now, in order to override alternate (secondary, alert, warning, etc) buttons, you are required to modify the sass file or override the CSS. I don't like this. I think it's sloppy for me to do that.

A prime example is using ghost buttons in your theme. The primary button is easy enough, just set `$button-bg-color: transparent`; but you can't do this to any other button.

I've added variables for the bg color of _every_ button. This would make it SO much easier to modify the alternate button bg's.


rafibomb:
Thanks Joel! In the next version we've added transparent buttons. In the meantime this works for people using F4S 5.

joelworsham:
Great!

PR #6686, opened 2015-07-07
Resolve Order Issues - Fieldset Legend Margin

JamesJCole:
Having margin:0 after negates the earlier margin-#{direction etc} call before on render.


rafibomb:
Yup, thanks!

PR #6695, opened 2015-07-10
slider: add support for changing by clicking on the slider without having started to drag from the handle

Burtchen:
The idea is that in many other implementation of a slider, it's possible to directly click/tap on a target value to change the slider, rather than to grab the handle first, which can be a bit annoying. There's a bit of almost-duplicate code compared to the mousemove event handler, but it seemed overkill to create another function before get_cursor_position for this.


rafibomb:
Thanks! We love this!

Issue #6524, opened 2015-05-05
Publicly accessible methods for Accordion


winghouchan:
Howdy all!

A user on Foundation Forum had an issue where they wanted to toggle accordion panels according to the viewport size.

Going through the Accordion's Docs there is no documented method on how to toggle accordion panels and looking through the code it does not seem possible.

It would be greatly useful to have a publicly accessible method to toggle Accordion panels, similar to how Off Canvas has publicly available methods to programmatically show, hide and toggle.

Looking forward to such functionality being introduced. Feel free to discuss below and open a PR.

Original thread on Foundation Forum: http://foundation.zurb.com/forum/posts/24648-open-all-accordions-on-medium-up

zurbrandon:
Looks like this was merged in I'll close this for now unless i'm mistaken.

PR #6700, opened 2015-07-14
Accordion: Added publicly accessible toggle, open and close methods. #6524

pauleds:
Issue #6524

### DIFF
--- a/js/foundation/foundation.accordion.js
+++ b/js/foundation/foundation.accordion.js
@@ -81,41 +81,42 @@
       }
     },
 	
-	toggle : function(options) {
-		var options = typeof options !== 'undefined' ? options : {};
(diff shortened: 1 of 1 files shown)


zurbchris:
I'm pretty sure you can use `.triggerHandler()` here instead of `.trigger()`, which would prevent the event from bubbling up to the `window`. Also, if you're creating events, please namespace them, i.e. `'click.fndtn.accordion'`. If you make a couple of changes, I'll take another look at it. Thanks for contributing!

pauleds:
I tried ```.triggerHandler()``` initially but that didn't work I'm afraid. I've now namespaced the click event as recommended and also used Big S for $item selection. I'm happy to make any other changes / improvements that might be useful.

zurbchris:
Thanks for the contribution!

PR #6703, opened 2015-07-15
Fix to include query arguments to magellan nav links.

quinncomendant:
The animation bound to `click.fndtn.magellan` sends the user to a URL lacking the query string arguments. The URL generation was missing the `anchor.search` component.

Before, when viewing a page, e.g.:

```
http://example.com/report.php?survey_id=6
```

Clicking a magellan nav link would scroll and the URL location would change to:

```
http://example.com/report.php#location-3
```

Adding the `anchor.search` component to the URL correctly shows the full URL including query args:

```
http://example.com/report.php?survey_id=6#location-3
```

(I also fixed some JS formatting.)

### DIFF
--- a/js/foundation/foundation.magellan.js
+++ b/js/foundation/foundation.magellan.js
@@ -59,11 +59,10 @@
               'scrollTop' : scroll_top
             }, settings.duration, settings.easing, function () {
               if (history.pushState) {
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks!

PR #6705, opened 2015-07-15
Dropdown hover fixes

brianlenz:
Fixing dropdown hover issues with multiple dropdown targets and the open event being triggered unnecessarily after the dropdown is already open.

https://github.com/zurb/foundation/issues/6704

### DIFF
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -63,7 +63,7 @@
           }
 
           if (settings.is_hover) {
-            if(!dropdown.is('.open')) {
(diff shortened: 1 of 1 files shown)


zurbchris:
jQuery's `.hasClass()` method is faster than `.is()`, if you change that I'll check it again. Thanks!

brianlenz:
Sure thing! All done :)

zurbchris:
Looks like I'm still getting multiple event firings with this. We'll make sure this isn't an issue in Foundation 6. Thanks for looking out, and please continue to help improve Foundation!

PR #6713, opened 2015-07-19
Only fire change event if set_ui is coming from user interaction

MarcGuay:
Fixes some of the problems discussed here: https://github.com/zurb/foundation/issues/6606


MarcGuay:
Unfortunately this breaks the display_selector input because it calls set_value, needs some tweaking.

MarcGuay:
Replaced by https://github.com/zurb/foundation/pull/6720

PR #6714, opened 2015-07-19
Remove reflow from set_value

MarcGuay:
This causes set_ui to be called twice.  Issue discussed here https://github.com/zurb/foundation/issues/6606.


MarcGuay:
Does not seem to be the case when combined with my other 2 pull requests on this file.  Closing.

PR #6715, opened 2015-07-19
Remove reflow from init

MarcGuay:
This causes set_ui to be called twice on page load.  The other comes from the resize event fired in foundation.js.  Related https://github.com/zurb/foundation/issues/6606

### DIFF
--- a/js/foundation/foundation.slider.js
+++ b/js/foundation/foundation.slider.js
@@ -23,7 +23,6 @@
     init : function (scope, method, options) {
       Foundation.inherit(this, 'throttle');
       this.bindings(method, options);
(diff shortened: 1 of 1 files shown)


MarcGuay:
Not sure about this now, it makes it so that the sliders aren't drawn until load(), rather than ready().  Maybe removing the trigger from foundation.js is smarter...

MarcGuay:
But if we leave the reflow in the init, every time you set_value it reflows all of the sliders... hergmmm.

MarcGuay:
Replaced by https://github.com/zurb/foundation/pull/6720

PR #6720, opened 2015-07-20
Fix slider issues with over-reflowing

MarcGuay:
- Add a way to distinguish between user interaction and reflowing (firechange argument).  
- Do not reflow on init (init is called every time set_value done and we don't need to reflow all sliders to change a single one)
- Sliders are reflow'd on page load in foundation.js.

### DIFF
--- a/js/foundation/foundation.slider.js
+++ b/js/foundation/foundation.slider.js
@@ -23,7 +23,6 @@
     init : function (scope, method, options) {
       Foundation.inherit(this, 'throttle');
       this.bindings(method, options);
-      this.reflow();
     },
 
     events : function () {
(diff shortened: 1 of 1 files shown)


MarcGuay:
I just remembered why only have the sliders init'ed on page load is a problem. I'm trying to get the value from one of the sliders to apply some filters to a catalog and it's not available to me until when?

zurbchris:
That value should be available to you as soon as `$(document).foundation()` has completed. You can add your callback function to the `foundation` object, or register an event listener within your own script to watch for changes. See: http://foundation.zurb.com/docs/components/range_slider.html
I'm closing this issue, but thanks for contributing!

Issue #5810, opened 2014-09-16
Reveal modal not in the page when set options animation: null


chuangbo:
If I set `data-options="animation:null"` to the modal div to disable the animation, the modal will not have a `top` in it. So that the modal might not be shown on the correct position.

Like this Pen http://codepen.io/anon/pen/holkL

```html
<a href="#" data-reveal-id="myModal">This modal not good</a>

<div id="myModal" class="reveal-modal" data-reveal data-options="animation:null">
    <h2>This modal not that good</h2>
    <p>The only different is this modal with options animation:null</p>
    <a class="close-reveal-modal">&#215;</a>
</div>
```

rafibomb:
We added this to our "known issues" section in our planning board for the next version. A PR would be valuable to the community, otherwise It will be addressed then. You can see and vote on the proposed updates here: https://trello.com/b/QBPKlQqf/foundation-sites

mitja-p:
I cannot vote on the issue on trello. But I would really like to have this bug fixed. It is really unexpected that animation works, but you turn off animation and it is broken.

PR #6721, opened 2015-07-20
Fix for #5810, reveal modal without animation bug

mitja-p:
Top position was not set, when animation was disabled.


zurbchris:
This issue should be resolved, thank you for contributing!

PR #6729, opened 2015-07-26
Adding full height responsiveness to Orbit

chrissmejia:
Orbit has some problems with responsiveness, this push solve some of that problems.
1. Allow you to set your slider to height 100%.
2. Automatically center your's sliders
3. Adjust to fit perfectly in every screen (mobile, tablet, pc)
4. Don't trim the content if it takes more than the screen size.
5. Allow you to define a maximum height to looks good even in huge screens.
6. Provide several customisation over how this height works (you can add different extra padding for top or bottom, add or remove size e.g if you need to remove the top menu height).

You can see it working at:
http://52.6.143.233

### DIFF
--- a/doc/pages/components/orbit.html
+++ b/doc/pages/components/orbit.html
@@ -219,6 +219,11 @@ $(document).foundation({
       timer: true, // Does the slider have a timer active? Setting to false disables the timer.
       variable_height: false, // Does the slider have variable height content?
       swipe: true,
+      full_height : false, // Take all window height
+      full_height_top_offset : 0, // Add padding at top
+      full_height_bottom_offset : 0, // Add padding at bottom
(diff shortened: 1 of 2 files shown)


SalahAdDin:
:+1: THis will be great, because, i trying use slick-slider but don't works fine, the thinks that i cauld make with orbit i can't make now with slider, change the caption background, margin, etc., center the slick block, it's crazy, and sad :'(

zurbchris:
Orbit is deprecated in Foundation 5, but we're bringing it back for Foundation 6. Keep checking in for updates and thanks for your contribution.

SalahAdDin:
@zurbchris It's really? You will come back Orbit?
:+1: 
But, now i'm using Owl Slider :'(

chrissmejia:
That's really good news :)

PR #6735, opened 2015-08-02
Add class to block-grid <li> to specify block size

jxw125:
In a block-grid, add class="block-X" to a <li> to specify how many blocks the element should take up.

For example:
This is a block-grid:
[X] [X] [X] 
[X] [X] [X]

Changing the first <li> to <li class="block-2"> will give this:
[--X--] [X]
[X] [X] [X] 
[X]

### DIFF
--- a/scss/foundation/components/_block-grid.scss
+++ b/scss/foundation/components/_block-grid.scss
@@ -72,6 +72,11 @@ $block-grid-media-queries: true !default;
         @include block-grid-aligned($per-row, $spacing);
       }
     }
+    @for $i from 1 through $per-row {
+      .block-#{($i)} {
+        width: (100%*$i)/$per-row;
+      }
+    }
   }
 }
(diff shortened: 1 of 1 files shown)


gakimball:
Did you encounter any problems with clearing of floats?

In your example, `*-block-grid-3` puts three items on each row, *and* it adds a float clear to every 3rd item, so things wrap properly. But if the first item is double the width, then the third item has wrapped around to the next row, and it's going to get a float clear, which would wrap the following items to the next line.

So you'd get a layout like this:

```
[    ][ ]
[ ]
[ ][ ]
```

rafibomb:
Thanks for the PR! Piggybacking on Geoff's explanation, it would seem the grid handle the use case you provided well. The Block Grid is best for same width blocks. It's not the direction we're going with this. We'll close it.

PR #6741, opened 2015-08-07
Fix tooltip nub on small screen

gray-one:
On narrow screen toolip nub wasn't triangle but white square. One css rule override another.


rafibomb:
Thanks!

PR #6742, opened 2015-08-07
Added additional settings for abide scrolling on focus

Siqu:
Added three additional options for abide scroll on input focus:
- scroll_disable: disable scroll completely
- scroll_offset: remove this from scroll top e.g. site has a fixed navigation and scroll would scroll element behind the navigation without the offset
- scroll_media: overwrite regex for user agent detection to enable/disable
  scroll for additional user agents


zurbchris:
Thanks, but we're gonna close this one.

PR #6743, opened 2015-08-07
Interchange Fix - Add additional check on el

vita10gy:
Fixes a "Cannot read property 'nodeName' of undefined" error that people
seem to run into now and again.

I don't know what I did to get in a situation to get this error,
interchange works in other places on the same site fine, but adding this
little extra check fixed one usage that threw it.

Even if it was "my fault" somehow probably never a bad idea to check
what really matters here.

### DIFF
--- a/js/foundation/foundation.interchange.js
+++ b/js/foundation/foundation.interchange.js
@@ -48,7 +48,7 @@
           //   console.log($(this).html(), a, b, c);
           // });
 
-          if (el !== null && /IMG/.test(el[0].nodeName)) {
+          if (el !== null && el.length > 0 && /IMG/.test(el[0].nodeName)) {
             var orig_path = $.each(el, function(){this.src = path;});
             // var orig_path = el[0].src;
 


vita10gy:
Although this fix makes the error go away, and thus JS not crap itself and have everything stop working, I think I've tracked down what actually causes this error to happen. 

It seems to be from having an image and non image interchange on the same page. 

Can anyone else confirm this? I've even confirmed using the examples straight from the example page, and have pulling interchange.js straight from master.

zurbchris:
I'm gonna close this until we have this issue reported from other users.

vita10gy:
Well the actual change in question is kind of a no downside little check.

rphelan:
@zurbchris @vita10gy I can confirm this is still an issue.  I have some elements that get initialized with interchange and later removed.  Their UUIDs stay in the cache and the lookup returns an empty set, so accessing el[0] throws an error.  You could improve on this PR by removing the offending UUIDs from the cache, but this fix would be a great start and produces no negative side effects.  Looks like someone tried to fix this in #6743 too but missed the mark slightly.

PR #6830, opened 2015-08-09
Update _top-bar.scss

Braden1996:
Fixed a minor bug which resulted in '$topbar-dropdown-bg' being applied as the background to every top-bar-section list-item Also, the variable wasn't even being used to set the background of the dropdown menu; so I simply moved it down.


rafibomb:
Thanks!

PR #6840, opened 2015-08-14
Remove Tabindex section from Accessibility page

joe-watkins:
Hello all :)

I reached out via Twitter to note that on the Accessibility page there is a section that suggests the usage of tabindex="1+" to establish a tab order to a document. This is actually not such a good idea unless you really know what you are doing. @rafibomb reached back and asked if I could maybe help with this page.

I believe that entire chunk could be removed. The keyboard accessibility section leads well into how to test for keyboard accessibility as well. That is great content btw :) It is important that when exposing markup examples to the world that they are good examples. I worry young developers seeing the declaration of 1+ values on the attribute rendering their UI unusable.

Here is some further reading about this topic:
http://www.paciellogroup.com/blog/2014/08/using-the-tabindex-attribute/

Browsers and Assistive Technology work naturally with the markup order. The user can flow through the document without the assistance of tabindex if the document is marked up correctly. Skip links are the best way to allow jumps/hops for users.

I'll make another PR for a section covering CSS `outline: none;` which is a great high level topic that would fit in well with this page.

Thanks for taking the time to include a special page covering accessibility! I'd love to contribute more accessibility enhancements to the framework if needed :)

### DIFF
--- a/doc/pages/accessibility.html
+++ b/doc/pages/accessibility.html
@@ -23,28 +23,9 @@ Aside from accessibility features that has been built into Foundation's componen
 
 <p>The most common impairments for web users are those with problems seeing, hearing or a physical inability to use a mouse. For that reason, the site must be navigatable by keyboard. Most commonly the tab key is used to tab through the content. For a vision impaired person will have a <a href="http://en.wikipedia.org/wiki/Screen_reader">screenreader</a> installed that reads the content out loud. We used <a href="http://www.chromevox.com/">Chromevox</a> to test with. You can find a list of popular screen readers below in the resource list.</p>
 
-### Tab-index
-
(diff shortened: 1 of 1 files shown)


rafibomb:
@joe-watkins We agree with removing the tab-index details for now. I'd like to see a better explanation of the use cases for these things soon.

A PR for a section covering CSS outline: none; would be a great topic!

rafibomb:
There are merge conflicts that need to be cleared before we can merge this. Can you look at those?

joe-watkins:
@rafibomb Hmm.. I rebuilt and attempted a merge into `Master` locally and things look good.. try again? Ideas?

joe-watkins:
Gonna create a fresh PR for this.

PR #6841, opened 2015-08-14
Fix Typo, Add aria-labelledby, add resources

joe-watkins:
Hello :)

Fixed a typo in Tools, used `aria-labelledby` for lists and added a couple cool #a11y resources.

a11y-project: An awesome project making accessibility a bit easier to understand for developers and designers with great code examples.

Tenon.io - A wonderful online accessibility testing app.

### DIFF
--- a/doc/pages/accessibility.html
+++ b/doc/pages/accessibility.html
@@ -25,7 +25,7 @@ Aside from accessibility features that has been built into Foundation's componen
 
 ### Tab-index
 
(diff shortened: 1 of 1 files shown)


joe-watkins:
Also removed `role="menuitem"` from `<li>` as that is generally used for menu items that are part of a menu bar.. like in an application and not to be used like that in an unordered list. 

I also removed `role="navigation"` from each `<ul>` as well since you want to reserve that for major site navigation components and not just a list of links that mostly to external sites.

rafibomb:
Good stuff - thanks!

PR #6847, opened 2015-08-17
Create balancer.js

coreysyms:
jQuery "balancing" for block grid odd numbered blocks. This plugin will manipulate the orphans to either "center" up based off the siblings width, or span the full with in a balanced manor. This can be accomplished via setting respectSiblingWidth = false in initialization.

Example of respecting siblings. http://www.tangerineindustries.com/download/foundation-balancer/respect.html

Example of disrespecting siblings. http://www.tangerineindustries.com/download/foundation-balancer/disrespect.html

### DIFF
--- a/js/foundation/balancer.js
+++ b/js/foundation/balancer.js
@@ -1,0 +1,58 @@
+;(function ($, window, document, undefined) {
+  'use strict';
+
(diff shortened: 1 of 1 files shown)


SalahAdDin:
:+1:

PR #6855, opened 2015-08-23
Added title text for orbit for better accessibility

Owlbertz:
The Orbit initializing now takes two additional parameters allowing to set the text for the arrows that control the slider.
(I am not quite sure if this is considered bad practice to have English fallback strings?)


zurbchris:
Orbit is deprecated in Foundation 5, we are updating it for Foundation 6 and it will include accessibility features. Stay tuned for more updates, and thanks!

PR #6861, opened 2015-08-26
Changed skip_selector if statement to a while loop.

asommer70:
New pull request based on feedback in previous [request](https://github.com/zurb/foundation/pull/5388#issuecomment-127771141) from @gorelog.

The change works fine in my testing, but if there's anything I need to change/update please let me know.

Thanks and Party On!

### DIFF
--- a/js/foundation/foundation.clearing.js
+++ b/js/foundation/foundation.clearing.js
@@ -471,7 +471,7 @@
           target = current[direction]();
 
       // Check for skip selector.
-      if (this.settings.skip_selector && target.find(this.settings.skip_selector).length != 0) {
+      while (this.settings.skip_selector && target.find(this.settings.skip_selector).length != 0) {
         target = target[direction]();
       }
 


rafibomb:
Not seeing a performance gain for this. We're safer leaving it the way it is. Thanks!

gorelog:
Performance gain? It's not about performance, it's about functioning the way it was intended (from my point of view anyway). Again, if you have 2 images in the middle of 4 and you skip from the first to the second, it then goes to the third which it is also supposed to skip. It should go from image 1 to image 4.

I've created a version with this functionality I hope to share soon. It not only lets you skip multiple images in a row, but the skippable ones are also are not visible to the end user in the carousel.

So say you have 6 images and click a toggle to filter those 6 down to 3. When you bring up the clearing lightbox, the ones that are hidden remain hidden in the carousel and skip as they should.

asommer70:
I agree, @rafibomb not sure where you think it's about a "performance gain"?  (maybe you were thinking about another issue?).  I appreciate the feedback though.

@gorelog I'd love to see your version.  Changing the **if** statement does allow skipping multiple images, but they do show up in the bottom preview area.  That would be great if you could address that part.

gorelog:
@asommer70 I'll spend some time on it this weekend. It's pretty hacked together and I haven't look at it for about a month. One thing that is different from your implementation is that I target the list item instead of the image.

gorelog:
@asommer70 Here is my [little demo](http://clearing.gorelog.com) with some brief explanation. Not too much needed to be adjusted. I'm sure there is some performance impact since it uses ':visible' to filter out the visible list items. However, I have this working on a gallery where you are able to show/hide photos by tag and it works pretty well.

asommer70:
Good day @gorelog finally had a chance to look at your demo.  Here's some notes I have after checking out your code:

* Not sure that the visibility aspect should matter.
* If they are marked as **skip** the images should just be skipped whether they are visible or not.
* Integrating visibility into the mix feels like a whole different feature than just having the option to skip certain images in the carousel.

I've committed a couple of additional [changes](https://github.com/asommer70/foundation/commit/3ab204fbdcd6c09ec1102e7a596d90f9df47784d) to my fork to toggle the **hide** class to the **skip_selector** elements right after the **[open/close].fndtn.clearing** events are triggered.  This will hide the skip_selector elements in the bottom carousel as well as skip them when paging through the images.

@gorelog I'd love to get your feedback on these changes.

PR #6869, opened 2015-08-31
dropdown performance boost

dragthor:
timing went from ~65 milliseconds down to ~45 milliseconds for a long
list of items

An example: http://www.barenecessities.com/Bras_catalog_nxs,31.htm

Mobile view... the Size dropdown is now 20 milliseconds faster on desktop (even better on mobile).

### DIFF
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -275,23 +275,25 @@
         //lets see if the panel will be off the screen
         //get the actual width of the page and store it
         var actualBodyWidth;
+        var windowWidth = window.innerWidth;
+        
         if (document.getElementsByClassName('row')[0]) {
           actualBodyWidth = document.getElementsByClassName('row')[0].clientWidth;
         } else {
-          actualBodyWidth = window.innerWidth;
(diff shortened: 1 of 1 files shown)


dragthor:
Utilizing the Chrome profiler it cuts down on recalculate style time.

zurbchris:
This is a good fix, but, could you fix the indentation on lines 288 and 289? Also maybe set `window.innerWidth` to a variable, it's being read a few times, no reason to not save it. `windowWidth` perhaps?

dragthor:
Okay working on it now.  I should just make another pull request?

zurbchris:
If you work on your local fork and push to the branch you're submitting the PR from, it should auto-update the original PR.

dragthor:
I also have similar changes to top, bottom, left, and right too that I will run through the profile, and then commit.

dragthor:
On a related note... I am still trying to determine some calls to get height/get width that might cause invalidation/recalculation.  Not going to check in anything at this time until I am 100% sure, tested, and profiled.

zurbchris:
Thanks for the contribution! Do send us another PR when you've tested anything new.

PR #6872, opened 2015-09-01
Close dropdown when a reveal modal link is clicked from inside

tjasko:
I noticed that Foundation does not close a dropdown when a reveal modal link is a child of the unordered list. I figured this is a behavior most people would want, so here is some code to add it to Foundation's core.

If there's a better way to accomplish this, let me know.


rafibomb:
Seems to already work like that in 5.5.2 - http://codepen.io/rafibomb/pen/QjWVVW?editors=110

tjasko:
Hmm, strange. I'll have to dig into this why it's not like this for 5.5.1; possibly I'm doing something strange.

PR #6879, opened 2015-09-10
Include $input-placeholder-font-color variable

serg0x:
The $input-placeholder-font-color: #cccccc !default; is defined in the forms.scss on line 26 and was forgotten in the settings.scss


rafibomb:
Thanks! We've removed all the hard coded colors to be variables. This one is $iron but I think we have too many anyways. We'll leave it as is.

PR #6882, opened 2015-09-11
Abide skip validation if no fields validatable

### LINKED ISSUE #6881: Abide throws an error if form has no validatable fields
Although we originally attempted to distinguish 100% between forms that were validated and forms that were not, there is an exception case.

We have a form that dynamically hides fields after they have been loaded, when custom widgets are used to select values which must then be hidden from view (e.g. token replacement with proper names)

The issue is simply to ensure that execution continues without error in this case. Currently we see:

    Uncaught TypeError: Cannot read property '0' of undefined    foundation.abide.js:207

charneykaye:
Although we originally attempted to distinguish 100% between forms that were validated and forms that were not, there is an exception case.

We have a form that dynamically hides fields after they have been loaded, when custom widgets are used to select values which must then be hidden from view (e.g. token replacement with proper names)

The issue is simply to ensure that execution continues without error in this case. Currently we see:

```
Uncaught TypeError: Cannot read property '0' of undefined    foundation.abide.js:207
```

### DIFF
--- a/js/foundation/foundation.abide.js
+++ b/js/foundation/foundation.abide.js
@@ -217,8 +217,11 @@
     // TODO: Break this up into smaller methods, getting hard to read.
     check_validation_and_apply_styles : function (el_patterns) {
       var i = el_patterns.length,
-          validations = [],
-          form = this.S(el_patterns[0][0]).closest('[data-' + this.attr_name(true) + ']'),
+          validations = [];
+      if (i == 0) {
+        return validations;
+      }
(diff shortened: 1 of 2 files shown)


charneykaye:
Closes #6881

rafibomb:
Thanks! We'll be testing this soon for a merge!

zurbchris:
Thanks for the contribution!

charneykaye:
Absolutely! Thank **you!**

PR #6892, opened 2015-09-19
Move functions and settings imports to make them available globally

a-dg:
This makes the import paths clearer and more consistent. Since functions and settings do not generate any markup themselves, the resulting output is not affected by this change.


rafibomb:
I'm not sure how this will do what you want.
_settings.scss is always imported before foundation.scss, which means it has to import that functions file
So changing the order within foundation.scss doesn't affect _settings.scss

We'll close this here for now. Let us know if we've missed something.

PR #6893, opened 2015-09-19
Consolidate block grid clearing styles

a-dg:
Using `&:nth-of-type(1n)` selects every element of that type, and appears to be unnecessary. Simply moving `clear: none;` to the main selector produces the same result with fewer lines.


rafibomb:
This may be like that for specificity. We'll test it out and see.

rafibomb:

PR #6897, opened 2015-09-23
Fix 'dd a active' for tabs

ysalmin:
Why in tag 'a' (line 42) you are using '>' but in line 55 you are not?


rafibomb:
Looks like an oversight. Thanks!

NeXTs:
@ysalmin лол. приятного

PR #6902, opened 2015-09-24
Form emits `valid.fndtn.abide` when there are no validatable fields 

### LINKED ISSUE #6901: Form ought to emit `valid.fndtn.abide` when there are no validatable fields
This issue is follow-up to the original issue #6881 Abide throws an error if form has no validatable fields, and will be accompanied by a pull request within a couple hours.

The original issue #6881 is still an improvement, because prior to that Abide was simply throwing a JavaScript error in this case: **a form that hides its validatable elements before submission**

However, we identified a paradox that made it very difficult to actually implement such a form without hacking our way around Abide's blocking of the original form `submit` event.

So, the solution to the original issue #6881 was fine in that it prevented a JavaScript error. However, the failure to emit `valid.fndtn.abide` was caused by a totally separate oversight in how Abide identifies the form element to which the validation belongs.

The expected enhanced behavior would still *skip validation,* but then it would also still emit the `valid.fndtn.abide` event.

charneykaye:
closes #6901

### DIFF
--- a/js/foundation/foundation.abide.js
+++ b/js/foundation/foundation.abide.js
@@ -154,7 +154,7 @@
     validate : function (els, e, is_ajax) {
       var validations = this.parse_patterns(els),
           validation_count = validations.length,
-          form = this.S(els[0]).closest('form'),
+          form = this.S(els.length ? els[0] : e.target).closest('form'),
           submit_event = /submit/.test(e.type);
 
       // Has to count up to make sure the focus gets applied to the top error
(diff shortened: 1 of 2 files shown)


rafibomb:
Thanks for this! We'll be testing for merge soon.

charneykaye:
Hiya @rafibomb please let me know if there's anything I can do here to help.

rafibomb:
Hi - thanks for the PR! It looks like this go caught in out Foundation 6 release cycle. 

We still want this to go into Foundation 5 and we'll be pushing a patch next week. Are you able to re submit on the V5 branch so we can get it in? Thanks!

charneykaye:
Thanks @rafibomb - created #7335 and closing this one.

PR #6903, opened 2015-09-25
dropdown performance minus the pip

dragthor:
Hey guys.  We were having a major dropdown performance issue, a show stopper.  In order to receive business approval to “go live” we resorted to a gorilla tactic (described below).  
Our dropdowns (small, medium, large) do not visually utilize the PIP.  We hide it with the following CSS:

`#ourDrp.f-dropdown:before, #ourDrp.f-dropdown:after {
content: "";
display: none;
width: 0;
height: 0;
border: none;
border-color: transparent;
border-bottom-style: none;
top: 0px;
left: 0px; } }`

The profilers (Chrome, IE11, and Visual Studio) all pointed to dropdown::adjust_pip().  Lots and lots of layout thrashing.  Thus we decided to “kill” adjust_pip() altogether –

`Foundation.libs.dropdown.adjust_pip = function () { };`

Success (kind of).  Since we are not using the PIP why calculate its position?  We are now “live” in production.

Anyway I want to get your thoughts on a more elegant solution (below).

Add an additional setting to dropdown.js (defaulting to existing Zurb behavior) -

`no_pip : false`

Then in the adjust_pip() function just add –

`if (settings.no_pip === true) return;`

Utilize the feature like this (javascript):

```
$(document).foundation({
    dropdown: {
      no_pip: true
    }
});
```

Or declarative syntax (html):

`<a data-dropdown="drop1" aria-controls="drop1" aria-expanded="false" data-options="no_pip:true">Has Dropdown</a>`

`<ul id="drop1" class="f-dropdown" data-dropdown-content aria-hidden="true" tabindex="-1">
  <li><a href="#">This is a link</a></li>
  <li><a href="#">This is another</a></li>
  <li><a href="#">Yet another</a></li>
</ul>`


zurbchris:
Looks good, but bump up that `if` statement to top level in the `adjust_pip` function. No need to declare variables and such if we aren't running the rest of the function. 

An even better way might be to put a conditional statement on the function calls to `adjust_pip`, since they are all calling on jQuery functions to determine if the pip needs to be moved, which is just overhead for no purpose if there's no pip.

For instance, change line 391 to:

```javascript
        if (!self.settings.no_pip && (t.outerWidth() < this.outerWidth() || self.small() || this.hasClass(s.mega_menu))) {
          self.adjust_pip(this, t, s, p);
        }
```

dragthor:
@zurbchris good call.

PR #6904, opened 2015-09-25
Equalizer nested tests added

dragthor:
Changed:

I added the "assert" prior to the foreach because it would still pass if for some strange reason it was an empty list.

Addition:

Tests for nested equalizer.


zurbchris:
I like it!

PR #6912, opened 2015-10-01
Tooltip unused var; Nub outerWidth

dragthor:
Removed unused variable. Fixed nubWidth using outerHeight instead of outerWidth.  Fortunately most of the time they are equal.


rafibomb:
Another good one! Thanks!

Issue #6888, opened 2015-09-16
Accessibility with Tabs


JoshuaVB:
## Specs
Browser: Chrome v. 45, Firefox v. 40, IE 11 (essentially every browser we've tried)
Foundation: 5.5.2

Related Ticket: https://github.com/zurb/foundation/issues/6663 

## Issue
Hi All,

We've been using Foundation for the bulk of our new web project and have come across a snag.
When keyboard navigating the Tabs element, users are stuck within the Tabs and cannot navigate out.

This behavior is present on the Tabs examples from the Docs as well. Trying to use the keyboard to navigate further down the page, or into one of the tabs, doesn't seem to be possible.

This is very concerning for us, as our site is supposed to fully "Accessible", and screen-reader users must be able to navigate easily.

This behavior prevents any screen-reader user a chance to use the Tabs item properly. Even using the full "Accessibility" suggestions for the Tab object, it is completely useless for blind users.

Has there been any development on this item since the 6663 ticket?

If the Tabs object cannot be used for screen-reader/keyboard navigators then we will have to completely remove it from our project and code something that is compliant, which is, of course, undesirable.

dragthor:
@JoshuaVB Josh I will check out #6663

JoshuaVB:
@dragthor Thanks for looking into this item.

After further digging, we believe part of the offending piece of code is the following from the base foundation JS file, starting line 5503:

        if ($target.length) {
              $original.attr({
                'tabindex' : '-1',
                'aria-selected' : null
              });
              $target.attr({
                'tabindex' : '0',
                'aria-selected' : true
              }).focus();
            }

This causes the other non-focused tabs to be removed from the keyboard navigation schema.

JoshuaVB:
@dragthor A follow-up to the previous item, we've discovered a way to modify the Foundation.js code to partially accommodate our needs. Here's our changes:

#### Original
Line 5521: $('#' + $(document.activeElement).attr('href').substring(1))
              .attr('aria-hidden', null);

#### Updated
Line 5521: $('#' + $(document.activeElement).attr('href').substring(1))
              .attr('aria-hidden', null).find('a').eq(0).focus();

Since we have links in the tab content, this modification works for us. However, it is still not possible to navigate between tabs without first hitting "Tab" or "Enter" on the tab first. It is also not possible for this change to work for users without links in their tab content.

rafibomb:
[...]

PR #6914, opened 2015-10-04
Accessibility for Tabs

Owlbertz:
Did a bunch of stuff to make tabs usable with keyboard as discussed in #6888.
I had to outcomment the tabindex adjustments, which do not make sense to me just now...
However, tabs should now me usable with tab-key, which means they can be selected with tabbing and then opened with enter or space.


rafibomb:
Sweet! We tested it and it seems to work great so far. We're pushing a release out this week and will test it some more. Great work!

PR #6930, opened 2015-10-10
Use correct active link color in side_nav mixin

exciton:
The side_nav mixin takes link-color-active as a parameter, but it is not used properly (the global side-nav-link-color-active is used instead, when it is supposed to just be a default).

This patch fixes it.


rafibomb:
👍Thanks!

PR #6931, opened 2015-10-10
Disable text-decoration on reveal-close

mik-laj:


rafibomb:
Thanks! Can you describe the scenario where this change is helpful?

mik-laj:
It is important for me, because I have set $anchor-text-decoration: underline. It looks ugly. This icon is sufficient information that the object requires action. Everyone knows that the button in the corner is used to close. Underlining the icon is very misleading.

rafibomb:
👍Agreed - this should not be applied in a modal close.

mik-laj:
pic related: http://imgur.com/tXodw78

PR #6934, opened 2015-10-12
fix label.inline, for label font size is not 14

wy-z:
whether this is more appropriate, when label font size is not 14.


rafibomb:
Thanks for the PR! Can you explain how this improves usability of the forms?

weiyoung-zhou:
JSbin
http://jsbin.com/leqabu/8/edit?html,output
have label vertically center against an input with `line-height`, but not `padding`

rafibomb:
Ah, nice - thanks!

PR #6935, opened 2015-10-13
Fix event binding

### LINKED ISSUE #6839: Clearing Lightbox does not work with 'data-clearing-interchange' Attribute
Foundation 5.5.2.

[Example including images here](http://foundation.zurb.com/forum/posts/25598-images-not-advancing-in-clearing-lightbox-using-interchange)

Codepen with example here: http://codepen.io/anon/pen/NqZpyb

Clearing Lightbox does not work with 'data-clearing-interchange' Attribute

If the 'data-clearing-interchange' Attribute is used, it is not possible to switch between images in the lightbox.
The first thumbnail that was clicked is then always shown in the lightbox no matter which other thumnail clicked.

szinya:
Fixes #6839

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -106,7 +106,7 @@
     var self = this,
         bind = function(){
           var $this = S(this),
-              should_bind_events = !$this.data(self.attr_name(true) + '-init');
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks for this!

Issue #6939, opened 2015-10-14
Tooltip nub issues for small devices


nategraves:
The documentation for tooltips states
>On a small device, the tooltips are full-width and bottom aligned

This works as expected, however, the nubs for the tooltips don't get updated. This means that for `tip-top`, `tip-left`, and `tip-right` the nubs look and are positioned the same as on larger sizes. Here's an example of a `tip-left`:
![screen shot 2015-10-14 at 3 36 09 pm](https://cloud.githubusercontent.com/assets/788530/10495410/b61bc2a6-7289-11e5-91e4-bec142369f9e.png)

The nubs for small size tooltips should probably use the same styling as the `tip-bottom` class as that's how they are being treated.

rafibomb:
Thanks for the PR! We'll close this in favor the PR and look to merge soon.

PR #6940, opened 2015-10-14
Fixed appearance and positioning of tooltip nubs for small devices

nategraves:
This is a fix for issue #6939. 

On small devices, all tooltips are treated as `tip-bottom` but the direction and positioning of the nubs wasn't being adjusted. 

I figured there were a couple of ways to go about fixing this. I opted for adjusting `border` and--in the case of `tip-left` and `tip-right`--the `margin-top` based on the presence of the tip-override class. 

This could also be done with a media query or potentially in the tooltip JS, so let me know if one of those options would be preferable.

Before:
screen shot 2015-10-14 at 3 36 09 pm

After:
screen shot 2015-10-14 at 3 49 24 pm

### DIFF
--- a/scss/foundation/components/_tooltips.scss
+++ b/scss/foundation/components/_tooltips.scss
@@ -134,9 +134,18 @@ $tooltip-max-width: 300px !default;
           right: auto;
           top: 50%;
         }
+      }
 
+      .tooltip.tip-override {
+        > .nub {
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks for this! Email foundation at zurb.com for some sweet Yeti stickers!

PR #6943, opened 2015-10-16
fix form select focus background

skinnp91:
When focusing a <select> box in a form, the dropdown arrow is wiped out because its set by background-url.  Not sure if this is intended, but it makes it very difficult to _not_ wipe out the arrow if you do decide you want it.  Changing background to background-color fixes this.


rafibomb:
Thanks!

PR #6945, opened 2015-10-17
FIXED hover_delay in tooltip

ynonp:
hover_delay was read from global settings even if a different value
was set via data-options.
1. Add hover_delay() function that gets/sets the value using $.data
2. Implemented reflow() function that sets hover_delay value for
   all tooltips from their data-option value or global settings.
3. Changed code in showTip to take current hover_delay value
   into account
4. Add 2 specs:
   - hover_delay value is read from data-options if specified
   - hover_delay value is read from global settings if no data-options
     value.

### DIFF
--- a/js/foundation/foundation.tooltip.js
+++ b/js/foundation/foundation.tooltip.js
@@ -27,6 +27,7 @@
 
     init : function (scope, method, options) {
       Foundation.inherit(this, 'random_str');
+      this.reflow();
       this.bindings(method, options);
     },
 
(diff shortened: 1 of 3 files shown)


rafibomb:
Thanks for this! We'll be testing it for merge soon!

rafibomb:

rafibomb:
We'll close it since GitHub doesn't allow us to re-target it. We'd still like to merge it in if it's pointed at the v5 branch. Thanks!

PR #6956, opened 2015-10-21
Prevent Tooltip from evaluating scripts, which may lead to XSS!

abods:


TrevorBramble:
@abods This looks like it may have been fixed in v6. Could you re-target your patch against the `V5` branch?

@gakimball, @arthurzurb, @rafibomb is that all it'll take to get this patch in?

gakimball:
@TrevorBramble @abods Yes, older pull requests need to compare against `V5` now.

rafibomb:

gakimball:
Closing in favor of #7359

Issue #6942, opened 2015-10-16
Sass precompiler does not work with abide form error class


jontro:
Currently when generating the css abide will not generate correct css

`"[data-abide]" .error small.error, "[data-abide]" .error span.error, "[data-abide]" span.error, "[data-abide]" small.error {`

the correct css should look like
`[data-abide] .error small.error, [data-abide] .error span.error, [data-abide] span.error, [data-abide] small.error {`

gakimball:
We tested this in Libsass 3.3.0 and Ruby Sass 3.4.20 yesterday, and both compilers rendered `[data-abide]` without quotes.

I believe the issue was related to a bug in Libsass that's been fixed in more recent versions. Since Foundation 5.5, our recommendation has been to always use the latest versions of Libsass or Ruby Sass.

PR #6958, opened 2015-10-21
Fix for Sass precompiler does not work with abide form error class #6942

ollicle:
In libsass the attribute selector (e.g. [data-abide]) is rendered to CSS with quoted (e.g. "[data-abide]"). Wrapping the isolated interpolated value in parenthesis resolves the problem.


rafibomb:

gakimball:
No longer an issue in Libsass 3.3 and up, just tested!

PR #6960, opened 2015-10-23
Abide hidden validation settings

hughkelsey:
Move the Abide validation filtering in to a setting which would allow users to specifically ignore (or not ignore) fields for validation.

Before this change ALL hidden fields are filtered out and in projects there are SOME hidden fields (those controlled by a range slider) that may require validation.

I've added specs which currently do not pass but I think that's because I'm not building the project correctly locally. I would appreciate an extra set of eyes on this and am happy to make any changes.


rafibomb:
Wnat to test it for merge, can you look into the errors first?

hughkelsey:
@rafibomb It looks like it's failing because the code is somehow incorrect. Can I ask for a more experienced set of eyes to look at the code?

zurbchris:
This is for Foundation 5. If you were still interested in submitting this pr, please reopen and target the Foundation 5 branch.

liyaodong:
think this is a very good solution, I think maybe we should reopen this issue.

Issue #6434, opened 2015-04-02
100% height on body and html adds extra page when printing the page


abderrahmane-tj:
Like the title says, having the following in the css

    html, body {
        height: 100%; }

always adds an extra page to the printed document.

adding this : 

    @media print {
        html, body {
        height: auto; }
    }

solves it for me

rafibomb:
A PR would really help here. We'll add it in!

caycek:
Hey I can get this fixed I'm pretty sure... I don't want to submit a secondary issue, but instead if we do `min-height: 100%` on both the `html` and `body` this should resolve the issue. If you want me to dive into this I will asap!

caycek:
Would it be worth changing up the body and html 100%?

if we changed the css

    html,body{
        min-height: 100%;
        float: left;
        width: 100%;
    }

and just got rid of the body having the css of `position: relative` this would leave the `html` to be the full body of the page and it is always filled with the content. No matter if there is floated content or not everything is covered. This would allow large screen applications to use position absolute without too much tweaking of the body css. 

I'm thinking this might be a good move, but I would like feed back before I do anything else.

thiagotomasi:
@abderrahmane-tj I just made a PR if your suggestion!
You can see it here: #6967

gakimball:
Fixed in the `V5` branch: 1f46b0d85dee94c6c176ad42a433972a4273f869

PR #6967, opened 2015-10-29
[RFR] Fix body height in print pages

thiagotomasi:
This PR is a change proposed in issue #6434.


gakimball:
This PR broke when we shifted the branches around from v5 to v6 (our bad!), so I made the change myself: 1f46b0d85dee94c6c176ad42a433972a4273f869

PR #6972, opened 2015-10-31
Adding Right Click support to dropdowns

al1b:
As you can see I have added an option **_is_right_click**_. If the user enable this option the dropdown menu will show up when the user right-clicks on the item ( or touch and press for about 1 second ).

Of course on Safari web browser the user should set the CSS rule `-webkit-touch-callout: none;` to avoid Safari open its own menu.

I tested this code on Google Chrome 46 (and its mobile emulator) and Internet Explorer 11.

### DIFF
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -35,19 +35,15 @@
                
           var settings = S(this).data(self.attr_name(true) + '-init') || self.settings;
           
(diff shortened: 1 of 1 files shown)


zurbchris:
I'm going to close this as we will no longer be adding features to Foundation 5.x 
If you would like to submit a similar pr to Foundation 6, I'd be interested to look at it. Thanks!

PR #6975, opened 2015-11-02
Update foundation.topbar.js

reecube:
Fixed an issue created by the intial-scroll of Google Chrome.

**Description of the issue and the solution:**
If you scroll down your foundation-website and use the sticky-tobbar and then you reload the page on Google Chrome, Chrome will directly jump to your last scroll position.
Because `start_offset` has the default value of `0`, the script won't toggle the `sticky` and `fixed` classes. So I've set the default value of `start_offset` to `false` (still `0` in JS arithmetics) and on the initialization of the script, it searches for the current offset of the topbar and sets it to the `start_offset` variable, if the `start_offset` equals to false.

### DIFF
--- a/js/foundation/foundation.topbar.js
+++ b/js/foundation/foundation.topbar.js
@@ -8,7 +8,7 @@
 
     settings : {
       index : 0,
-      start_offset : 0,
+      start_offset : false,
       sticky_class : 'sticky',
       custom_back_text : true,
       back_text : 'Back',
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks for the PR! Since this is for Foundation 5, it needs to be pointed to the V5 branch in this repo. Are you able to resubmit there so we can test and merge?

kball:
Closing based on last comment.  Please resubmit against V5.

PR #6979, opened 2015-11-04
Add $off-canvas-height variable to settings

MarcGuay:
The _settings.php file was missing the $off-canvas-height variable.


rafibomb:
Thanks!

PR #6983, opened 2015-11-05
Enable dynamic media queries and classes

a-dg:
Adds a unified method of defining and modifying breakpoint names and media query ranges. Creating new breakpoints is simply a matter of modifying one map:

``` scss
$regular-breakpoints: (
  small: 0,
  little: 421, // New breakpoint
  medium: 641,
  average: 801, // New breakpoint
  large: 1025,
);
```

This defines new media query ranges and all of the attendant grid, block grid, text alignment, and icon bar classes. Extra large breakpoints are defined in a separate map to preserve the functionality of settings like `$include-xl-html-grid-classes`.
### Breakpoint values

Breakpoints must be defined in order, with the smallest one first and the largest one last. This is important for establishing the relationship between adjacent breakpoints, and for outputting proper media queries.

The breakpoint values are defined using the bottom of each media query range. That's why `small` is defined as `0`, because it's the first breakpoint and therefore goes all the way down to the smallest viewport possible.

I would normally define the other breakpoints by using nice, round numbers (`640`, `1024`, `1440`, `1920`), but I've added `1` to each of them in order to preserve the resulting media query ranges in the previous version. It's not very likely that most people would notice a `1px` difference in the media queries, but I wanted to avoid any unnecessary changes.
### Differences in generated CSS

Generating the Foundation CSS from these new components results in a few differences, though they should have no effect on practical usage:
1. There is no `max-width` for the largest breakpoint, so `xxlarge-only` is identical to `xxlarge-up`.
2. There is a new `meta.foundation-mq-xxlarge-only` rule, which was missing from the previous version. Since all breakpoint-related classes are generated automatically, this one gets added. Again, there is no `max-width` for the largest breakpoint.
3. Legacy `push-*` and `pull-*` classes are extracted from the `medium` and `large` media queries.
4. `small` classes for icon bars are defined separately from the default icon bar styles. This introduces a negligible amount of redundancy in the CSS, and is again due to all breakpoint-related classes being generated automatically.
5. Redundant icon bar classes are removed.
### Hard-coded breakpoints

Please note that there are still places throughout the components that specify a particular breakpoint by name. For example, in `_reveal.scss`:

``` scss
@media #{map-get($queries-up, medium)} { // Was: @media #{$medium-up}
  top: $top-offset;
}
```

Some of these could possibly be extracted into new variables (like `$reveal-top-offset-size` for the above example), but others are a bit more deeply ingrained or are otherwise more difficult to extract in a meaningful way. I've left all of these cases as they are. If a user is savvy enough to define their own breakpoint names that _do not_ include the standard ones, they can certainly troubleshoot any errors resulting from a missing map key.


a-dg:
Uh, yeah, so I just saw your [announcement detailing dynamic breakpoints](http://zurb.com/article/1412/a-look-under-the-hood-of-foundation-6) in Foundation 6. So, never mind!

gakimball:
@a-dg You read our mind! :) Sorry we can't merge this in, since it's clear you put a lot of work into it. But with 6 out, 5 is now at a feature lock.

That being said, there's a whole new codebase to make contributions to, so I look forward to seeing more from you in the future. :)

a-dg:
No problem at all, I'm thrilled that you went in a similar direction for Foundation 6. Looking forward to trying it out!

PR #6985, opened 2015-11-05
Bring settings in line with change made in #6731

petecoop:
Change made in #6731 should be reflected in the settings file.


rafibomb:
Thanks!

PR #6997, opened 2015-11-12
Fix mixup between show-for-print and print-only visibility classes

cesarizu:
In foundation 5.5.X the semantics of the `show-for-print` visibility class changed. Before attaching that to an element wouldn't hide it when displaying the document on the screen but would force the element to be visible when printing. In 5.5.X the element is forcibly hidden for screens. The `print-only` class serves that purpose. This pull request restores the previous functionality.

This is related to https://github.com/cz848/foundation/pull/1 (523a434), https://github.com/zurb/foundation/issues/6445 and https://github.com/zurb/foundation/issues/3837 where, IMHO, those two classes, `show-for-print` and `print-only` are confused.

### DIFF
--- a/doc/includes/sidebar.html
+++ b/doc/includes/sidebar.html
@@ -5,7 +5,7 @@
     <input tabindex="1" id="autocomplete" type="search" placeholder="Search Docs: e.g. forms">
   </form>
 
-  <a href="/develop/download.html" class="download button expand">Download Foundation</a>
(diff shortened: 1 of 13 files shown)


acactown:
:+1:

gakimball:
Sorry for the super late reply, would it be possible for you to open a new pull request that targets the `V5` branch instead? Thanks :)

cesarizu:
Done in #7795

PR #7010, opened 2015-11-17
Delte accessability promise

evertalbers:
Since Orbit is deprecated, I don't expect this to happen anymore, so I suggest deleting these lines.


zurbchris:
But it's back, and accessible!

evertalbers:
That's good news.

PR #7013, opened 2015-11-18
Remove deprecated licenses section from package.json

manwithacat:
Replace with license key referencing SPDX, specifying MIT License, and remove LICENSE file


gakimball:
Closing, see my comment on #7014

Issue #7018, opened 2015-11-19
Meteor package - duplications - corrections needed


juliancwirko:
Hi, in the package.js there are files added from dist foder. There is foundation.js and foundation.min.js added and this is bad because it will be bundled in the app's js code. So you will have 2x foundation js code. We need only foundation.js (it will be minified by meteor in the production) 

The same with css files I don't think that we need it at all when we have .scss files. Meteor will compile it into css using fourseven:scss package. For now we will have 3x foundation css code in the bundle. All scss files should have '_' prefix even 'foundation.scss' and there could be a separate readme for Meteor on Atmosphere. Which will tell how to import the .scss files from Meteor Foundation package. So we could import all files or only some of them. 
Example: [zf5 package](https://github.com/juliancwirko/meteor-zf5) 

If someone don't want to use Scss he could just coppy foundation.css and foundation.js into Meteor app.

You can configure the path to the readme in: 
```
Package.describe({
  name: 'zurb:foundation-sites',
  summary: 'The most advanced responsive front-end framework in the world.',
  version: '6.0.0',
  git: 'https://github.com/zurb/foundation-sites.git',
  documentation: 'Meteor-README.md'
});
```
I would create PR for this docs etc but I am not english speaker.

(ps. you can provide version format like : 6.0.0_1 (for Meteor) to keep current Foundation version in it)

gakimball:
We plan on doing 6.0.1 today or tomorrow anyway, so if you want to submit this fix as a PR, it'll be in by the end of the week.

juliancwirko:
Ok I'll try to prepare PR but we also need Readme for Meteor usage I could try to add it but there might be some grammar errors. Anyway it is a short readme to show how to import and use include mixins. It will be the readme visible at https://atmospherejs.com/zurb/foundation-sites

juliancwirko:
merged

PR #7027, opened 2015-11-19
Meteor package adjustments and docs for atmospherejs.com

juliancwirko:
It is related with: https://github.com/zurb/foundation-sites/issues/7018
I tested it with Meteor. It should work now. Before there was css and js code duplicated, and some other blockers. Also there is no _joyride.scss file in the repo.


zurbchris:
Thanks!

Issue #7040, opened 2015-11-20
_banner.scss does not exist


colin-marshall:
Following a `git clone` of this repo and `npm install`, I get the following when I run `npm start`:

```
Error in plugin 'sass'
Message:
    docs/assets/scss/docs.scss
Error: File to import not found or unreadable: banner
       Parent style sheet: stdin
        on line 26 of stdin
>> @import 'banner';
   ^
```

I don't see a `_banner.scss` file in that folder.

anik786:
This issue has been fixed in the following pull request: https://github.com/zurb/foundation-sites/pull/7051
Hopefully will be merged soon!

gakimball:
When we merged the Foundation 6 branch into 5, somehow the `docs` folder ended up in `.gitignore`. By removing that, I was able to commit three files that were only on my computer as a result.

PR #7051, opened 2015-11-20
Fix issue - import non-existent file

jojoee:
Hi :smile: ,
I've remove the code that import non-existent file, please review.
import non-existent file


gakimball:
There was a different issue causing this error, see my closing comment in #7040.

PR #7082, opened 2015-11-20
Fixes bug where the ".end" class is overridden

yahya-uddin:
The `.end` class is currently being overridden by `.columns:last-child:not(:first-child)` as shown by the following:
https://jsfiddle.net/anik786/qy45s94e/embedded/result/

The fixed version is as so:
https://jsfiddle.net/anik786/gc58hhcg/1/embedded/result/


anik786:
This seems to be already fixed by a previous pull request

Issue #7101, opened 2015-11-21
[Global Styles] Variable '$global-margin' described twice in docs


YuriGor:
variable '$global-margin' [described](http://foundation.zurb.com/sites/docs/global.html#sass-variables) twice with different description:

Name | Type | Default Value | Description
-------- | ------- | ------------------ | ---------------
$global-margin | Number | 1rem	 | Global value used for margin on components.
$global-padding | Number | 1rem | Global value used for padding on components.
$global-margin | Number | 1rem	 | Global value used for margin between components.

Same issue at _settings.scss itself:
```scss
$global-margin: 1rem;
$global-padding: 1rem;
$global-margin: 1rem;
```

gakimball:
Merged in #7107

PR #7107, opened 2015-11-21
Remove duplicate global margin

aibrean:
global margin variable was being declared twice - removed duplicate


aibrean:
Note: this needs the docs to be adjusted as well (https://github.com/zurb/foundation-sites/issues/7101), but I wasn't sure which comment was the right one.

gakimball:
The docs are automatically generated from this list, so it updates automatically.

PR #7113, opened 2015-11-21
[Meteor] Package config fixes and documentation update

juliancwirko:
- Meteor has in its core jQuery 1.11.4 so you can't use jQuery 2 in the package, if you are using Meteor.versionsFrom you don't have to provide version for jQuery because this is a core library like for example 'ecmascript'
- There should be Foundation javascript file, it was removed in Foundation 6.0.3 (probably wrong merge)
- _settings.scss path was wrong
- added motion-ui because some of the Foundation's plugins don't work without it (ex. Reveal, Orbit) :/ This is weird, I tought that this is a separated library.
- what-input will not work that way it will be not included in the bundle. Npm.depends will only download it from npm and we should add the file to the repo or just do something like: api.addFiles('.npm/package/node_modules/what-input/what-input.js', 'client'); Anyway there is a problem with 'what-input' and Meteor. The document.body is not accesible because there isn't a body when the script is loaded. I don't know exactly what 'what-input' is supposed to do, but with Meteor it should be handled in some other way. I have removed 'what-input' for now because it does nothing in the package that way.
- **there will be needed 'meteor publish' after merge**

### DIFF
--- a/meteor-README.md
+++ b/meteor-README.md
@@ -1,8 +1,14 @@
-# [Foundation for Sites](http://foundation.zurb.com) (v6.0)
+# [Foundation for Sites](http://foundation.zurb.com) (v6.0.3)
 
 Foundation is the most advanced responsive front-end framework in the world. Quickly go from prototype to production, building sites or apps that work on any kind of device with Foundation. Includes layout constructs, like a fully customizable, responsive grid, commonly used JavaScript plugins, and full A11Y support.
 
 ## Usage in Meteor
+
+- [Scss guide](#scss-guide)
+- [JavaScript guide](#javascript-guide)
+
+
+## Scss Guide
 
 ### 1. Add the package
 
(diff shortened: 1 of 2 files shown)


gakimball:
what-input is more or less an optional dependency. It removes CSS outlines if the user is using a mouse, but keeps them enabled if they're using a keyboard. This allows us to keep outlines, which are necessary for accessibility, but only if the user actually needs them (i.e., they're using a keyboard).

juliancwirko:
it is hard to integrate it with Meteor because it assumes that the document.body is ready and with Meteor it is not when the package is initialized/loaded.

gakimball:
Understandable, in that case maybe the Meteor readme can include a remark about this. It doesn't appear to be possible to re-initialize the plugin programmatically.

juliancwirko:
Yeah, I was trying to integrate it somehow but this should be rewrite in some way. :/ 
I think it could be added to the meteor-README.md Could you do that? You'll do it better than me. Also there are probably some grammar errors to fix. You could work on the new readme version from this PR. Thanks.

PR #7117, opened 2015-11-21
Change overflow-y to overflow-x for scrollable tables

natewiebe13:
Using `overflow-x` instead of `overflow-y` seems to be the target effect. It also doesn't force a vertical scrollbar.


zurbrandon:
overflow-x is the desired effect so that on mobile screens it scrolls left and right when it's too long. What bug are you solving with this?

natewiebe13:
Currently, foundation is using `overflow-y: scroll;`. This forces a vertical scrollbar, and adds horizontal scrolling when needed. Ideally, it should be:
```
overflow-x: auto;
overflow-y: visible;
```

gakimball:
We're unable to auto-merge this PR now, so I made the change: 3e64b6c06e71e1ae133ee29198fcca46450b34a0

I also used `auto` instead of `scroll`, so you don't get the always-visible scroll bar on Windows.

PR #7134, opened 2015-11-22
[RFR] not is an operator, not a function

KittyGiraudel:
Tiny minor change. Cheers. :)


gakimball:
I think this was a holdover from an older version of Libsass, but I'm guessing the issue was resolved.

PR #7140, opened 2015-11-23
Remove duplicate $global-margin in _settings.scss

jasonmccallister:
Noticed a duplicate $global-margin in settings on one of my recent projects.


gakimball:
Merged in #7107 which resolves this.

Issue #7037, opened 2015-11-20
[scss] Dropdown arrow appears right for RTL, and top value is not consistent with menu item padding


imranomari:
1. When direction is configured to RTL, dropdown arrows `.is-down-arrow ` should be at the end of the text (`left`):
![correct](https://cloud.githubusercontent.com/assets/2339897/11288549/f9cc15a0-8f1c-11e5-8ff6-23e0ec38bfea.png)

 but they appear before the text instead:
![uncorrect](https://cloud.githubusercontent.com/assets/2339897/11288558/0f299850-8f1d-11e5-89c8-eb1e77d06a16.png)

to fix this, the `left` word both in values and and property name, should be replaced by the global direction variables as in the previous version:

```diff
// components/_dropdown-menu.scss
a::after {
-   float: right;
+   float: $global-right;
    margin-top: 3px;
-   margin-left: 10px;
+   margin-#{$global-left}: 10px;
}

@if $dropdownmenu-arrows {
        &.is-down-arrow a {
-            padding-right: 1.5rem;
+            padding-#{$global-right}: 1.5rem;
            position: relative;
        }
        &.is-down-arrow > a::after {
            @include css-triangle(5px, $anchor-color, down);
            position: absolute;
            top: 12px;
-           right: 5px;
+           #{$global-right}: 5px;
        }
       // ...
}
```

```scss
// components/_dropdown-menu.scss modified
a::after {
  float: $global-right;
  margin-top: 3px;
  margin-#{$global-left}: 10px;
}

@if $dropdownmenu-arrows {
    &.is-down-arrow a {
          padding-#{$global-right}: 1.5rem;
          position: relative;
    }
    &.is-down-arrow > a::after {
        @include css-triangle(5px, $anchor-color, down);
        position: absolute;
        top: 12px;
        #{$global-right}: 5px;
    }
  // ...
}
```

2. Also the top value of the arrow is hard coded to `12px` , which makes it inconsistent with the padding of menu items configurable by `$menu-item-padding`:
```scss
// components/_dropdown-menu.scss
&.is-down-arrow > a::after {
  @include css-triangle(5px, $anchor-color, down);
  position: absolute;
  top: 12px; // ,<-- should be calculated based on $menu-item-padding
  right: 5px;
}
```

PR #7175, opened 2015-11-24
fix arrows direction and spacing in dropdown menus, fixes: #7037

imranomari:


gakimball:
Just one comment on this. Instead of using `nth()` to find a padding value, you can use Foundation's `get-side()` function to pull out a more precise value.

```scss
// Second parameter is top, right, bottom, or left
get-side($menu-item-padding, top);
```

zurbrandon:
Hey @imranomari we want to pull this in, but wanted to see if you can fix the issue @gakimball mentioned above.

imranomari:
yes @zurbrandon, this commit 44c6a17 does exactly that, it's included in the pull request, sorry I didn't mention it when I pushed it.

gakimball:
@imranomari Nice, thank you :) Because generic padding settings like that could have 1–4 values, `get-side()` allows us to account for anything and always get the same number.

Issue #7215, opened 2015-11-25
css-triangle mixin up/down direction keywords


TvrtkoM:
Direction keywords parameters are not consistent with documentation. 'up' should be 'top' and 'down' should be 'bottom'

TvrtkoM:
seems like a lot of components use up/down keywords so probably docs should be modified

TvrtkoM:
created a pull request to fix the documentation - #7234

PR #7216, opened 2015-11-25
 fixes: #7215 change up/down direction keywords to top/bottom

TvrtkoM:


TvrtkoM:
there are a lot of occurences of up/down in the project so i'll close it. maybe it's better to change the docs

PR #7254, opened 2015-11-27
Media query utils are required

mafrosis:
A minor update to the docs at http://foundation.zurb.com/sites/docs/javascript.html, to make sure that people know that `foundation.util.mediaQuery.js` is required.

My feeling is that if it really is required, then it should be in core. Having this in the docs will save others wasting time debugging jscript errors!

Without `foundation.util.mediaQuery.js` included I get this error in Chrome:

```
Uncaught TypeError: Cannot read property '_init' of undefined
  foundation @ foundation.core.js:254
  (anonymous function) @ (index):273
```

### DIFF
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -12,10 +12,11 @@ Once you have the files, add links to jQuery and Foundation as `<script>` tags a
 ```html
 <script src="js/jquery.min.js"></script>
 <script src="js/foundation.min.js"></script>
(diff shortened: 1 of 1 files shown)


dherran:
+1 for this. Spent some time trying to understand where the error came from. Should be definitely be flagged in the docs as a dependency for foundation.core.js.

zurbchris:
I think we're going to move util.mediaQuery into .core for simplicity. Likely in 6.1.

PR #7268, opened 2015-11-27
Add drilldown sub menu background variable

brettsmason:
Added drilldown submenu background variable so that the background colour can be overwritten. I noticed I couldn't change this when working on a project!

### DIFF
--- a/scss/components/_drilldown.scss
+++ b/scss/components/_drilldown.scss
@@ -16,7 +16,7 @@ $drilldown-arrows: true !default;
 
 /// Background color for drilldown submenus.
 /// @type Color
-$drilldown-sub-background: $white !default;
+$drilldown-background: $white !default;
 
 @mixin foundation-drilldown-menu {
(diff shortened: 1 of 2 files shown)


Renkas:
I think you sholud add it to _settings.scss also

brettsmason:
I've added that to settings and renamed the variable - all should be good to include hopefully!

scss/components/_drilldown.scss
```diff
@@ -29,7 +33,7 @@ $drilldown-arrows: true !default;
     z-index: -1;
     height: 100%;
     width: 100%;
-    background: $white;
+    background: $drilldown-sub-background;
```
Renkas: You forgot to rename variable here

brettsmason:
Thanks - I did it in a hurry and forgot! That's now changed.

zurbrandon:
Sweet! Thanks!

PR #7270, opened 2015-11-27
Replaces `&` at the middle of the media-object rule

### LINKED ISSUE #7245: media-object-section incorrect markup?
The _media-object.scss file contains the line:

`&:last-child:not(+ &:first-child) {`

When compiled it translates to the rule:

`.media-object-section:last-child:not( + &:first-child) {`

Now, this doesn't seem at all valid CSS to me, but maybe I'm misinterpreting/not getting something?  Don't think it's just my compilation, though, as it's the same on the Zurb site.

filipekiss:
As stated in Sass documentation

> & must appear at the beginning of a compound selector

So the way this mixin was written was causing invalid CSS to be generated. Assigning the selector to a variable allows it to be used anywhere in the compound selector.

This fixes #7245

### DIFF
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -27,6 +27,7 @@ $mediaobject-image-width-stacked: 100% !default;
 /// Adds styles for sections within a media object.
 /// @param {Number} $padding [$mediaobject-section-padding] - Padding between sections.
 @mixin media-object-section($padding: $mediaobject-section-padding) {
+  $selector: &;
   display: table-cell;
   vertical-align: top;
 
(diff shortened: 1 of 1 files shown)


R4c00n:
+1

rafibomb:
Thanks!

R4c00n:
Shouldn't it be `&:last-child:not(+ #{$selector}:first-child) {` instead of `&:last-child:not(+ $selector:first-child) {`? 

Your version throws an Compass error `Invalid CSS after "...st-child:not(+ ": expected ")", was "$selector:first...")`

filipekiss:
@R4c00n Hm. Maybe. I don't know, I've been using `libsass` and it works fine. I'll confirm this and open a new pull-request if needed. :)

R4c00n:
I opened a new PR ;) #7346

Issue #7265, opened 2015-11-27
grid-column mixin not working with (n of n)


whereizmymind:
The grid-column mixin will not work if I write : `@include grid-column(6 of 16);`
It returns an error "Error: may only compare numbers".

The problem come from the mixin that compare the value : 
```scss
@if ($columns <= $grid-column-count) {
    @include grid-column-size($columns);
}
```

PR #7285, opened 2015-11-28
issue #7265 - enable grid-column mixin to parse list/fraction (n of m)

TvrtkoM:
PR fixes issue #7265

### DIFF
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -57,7 +57,12 @@
   $columns: $grid-column-count,
   $gutter: $grid-column-gutter
 ) {
-  @if ($columns <= $grid-column-count) {
+  @if type_of($columns) == 'list' {
+    @if nth($columns, 1) <= nth($columns, 3) {
+      @include grid-column-size($columns);
+    }
+  }
+  @else if ($columns <= $grid-column-count) {
     @include grid-column-size($columns);
   }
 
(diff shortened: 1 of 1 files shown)


gakimball:
Can you verify that the tests for this function continue to pass? Run `npm start test:sass`.

TvrtkoM:
Hi, I'm not sure now if I want this change in mixin. We already have a function grid-column (which is tested) and it can be used in some cases on width css property.

I've ran the tests with gulp test and got 3 unrelated fails

    Breakpoint converts a named breakpoint to an em value {
      ✗: Expected "(min-width: 40em)" to be "(min-width: 32em)";
      Test: Failed; }

    Breakpoint creates an only range out of a named breakpoint {
      ✗: Expected "(min-width: 40em) and (max-width: 63.9375em)" to be "(min-width: 32em) and (max-width: 63.9375em)";
      Test: Failed; }

    Text inputs creates a selector out of a list of text input types {
      ✗: Expected "[type='text'], [type='password']" to be ["[type=", "text", "], [type=", "password", "]"];
      Test: Failed; }

Note docs need to be fixed for grid-column mixin if this feature is not needed

gakimball:
Yeah, I think there's another way to fix this. I'll write another comment in #7265.

gakimball:
The fix ended up being a lot simpler—just remove the `$columns <= $grid-column-count` check. It doesn't affect the CSS output at all.

PR #7295, opened 2015-11-29
ADD Map for css in build doc

IllusionPerdu:
It's add module in npm gulp-sourcemaps
When genrate the doc it add sourcemap for the css


gakimball:
Are the source maps not supposed to be written in the same folder as the original file? (e.g., `dist/assets/css/app.css.map`)

IllusionPerdu:
I have tested and apparently Sourcemap directory is the root of the destination directory ('_build/assets/css'). 
With ('.') it is written in the same directory as the file css.
(it's translate by google translate sorry for my bad english)

PR #7326, opened 2015-12-01
Simplify expanded menus

### LINKED ISSUE #7192: Simplify expanded menus
Expanded menus are using quantity queries to define the item widths. This generates a lot of unnecessary CSS output. There is no need to define a width if table-layout: fixed is used.

Here’s a little demo: http://codepen.io/anon/pen/zvQXQq

slavanga:
Fixes #7192

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -46,16 +46,20 @@ $buttongroup-expand-max: 6 !default;
 /// @param {Keyword|Number} $count [auto] - Number of buttons inside the button group. Set to `auto` to generate CSS that will account for a variable number of buttons.
 /// @param {String} $selector [$buttongroup-child-selector] - Selector for the buttons inside a button group.
 @mixin button-group-expand(
-  $count: auto,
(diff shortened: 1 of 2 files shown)


andycochran:
Why `table-layout: fixed` instead of `auto`?

slavanga:
Because the auto-width mixin will output a large amount of unneeded CSS.
See: https://github.com/zurb/foundation-sites/blob/develop/dist/foundation.css#L2279-L2293

Setting ```table-layout: fixed``` will equalize the items just as well.

gakimball:
@slavanga Thanks for putting this together! What browsers were you able to test in?

gakimball:
Also, we use the auto layout mixin on button groups as well, if you're interested in implementing it there as well.

andycochran:
Sorry, I was confused (was thinking about inline menus). :+1:

slavanga:
@gakimball This method is basically supported in every browser that can handle table display ([caniuse] (http://caniuse.com/#feat=css-table))

Issue #7328, opened 2015-12-01
all variables are initally uncommented in _settings.scss?


yucelm:
The foundation doc says in order to make a change, variables must first be uncommented (http://foundation.zurb.com/apps/docs/#!/sass - "The Settings File" section Paragraph 3) Yet when I create a project, all of the variables inside the _settings.scss file are already uncommented? 

If I were to comment any of them, that component becomes unusable. No default styling is available. Any ideas why?

abdullahsalem:
Hi @yucelm,
I've commented the whole variables in scss/settings/settings.scss to be matched what is written in the documentation.
However, you should not face any problem either it's commented or uncommented, because the default variables should work instead!

yucelm:
Thanks. I am just checking it out. 
I just wanna ask you another quick question if you don't mind. 

In the app.scss file, the main foundation file is already imported at the top "@import 'foundation';"
When I comment out one of the includes however (i.e @include foundation-grid;) it breaks that component. 

Aren't all components already included in the main foundation file anyways? Should it not work regardless even though I comment out the individual components?

Thanks a lot.

abdullahsalem:
I understood that you are trying to @import the Foundation's component, but NOT @include its mixin, if you did so, that component will NOT work, because you only @import its FILE but didn't call its mixin, and this is a major difference between F5 and F6. In F5, you only need to @import the component's file but in F6 you need to @include its mixin too.

yucelm:
Hello,

Here is my app.scss and settings.scss files.

![settings scss](https://cloud.githubusercontent.com/assets/10710822/11499276/9642f886-982d-11e5-9277-4f06aebcdd03.png)
![app scss](https://cloud.githubusercontent.com/assets/10710822/11499277/9644a01e-982d-11e5-92d9-c58bb06c6db2.png)

yucelm:
I see that is something I wasn't aware of. It is clear now.

abdullahsalem:
I hope everything does well..
Good luck,

Knight-Yoshi:
@yucelm So that when using a component the classes are not automatically added into the CSS and to remove the need for a variable to check if it should include the CSS at a global level and on a per component level (I think to clean up fragmentation).

If you want to include all the Foundation classes use @include foundation-everything();

yucelm:
@Knight-Yoshi It is super clear now. I haven't had enough time fully review all the changes in F6. Many thanks.

gakimball:
[...]

PR #7329, opened 2015-12-01
Commented the whole settings variables #7328

imabdulsal:
#7328


gakimball:
See [my comment](https://github.com/zurb/foundation-sites/issues/7328#issuecomment-161037445) on #7328 to see why the uncommenting is intentional.

PR #7335, opened 2015-12-01
Form emits `valid.fndtn.abide` when there are no validatable fields

### LINKED ISSUE #6901: Form ought to emit `valid.fndtn.abide` when there are no validatable fields

charneykaye:
closes #6901

### DIFF
--- a/js/foundation/foundation.abide.js
+++ b/js/foundation/foundation.abide.js
@@ -154,7 +154,7 @@
     validate : function (els, e, is_ajax) {
       var validations = this.parse_patterns(els),
           validation_count = validations.length,
-          form = this.S(els[0]).closest('form'),
+          form = this.S(els.length ? els[0] : e.target).closest('form'),
           submit_event = /submit/.test(e.type);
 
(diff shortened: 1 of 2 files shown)


kball:
@coreysyms can you take a look at this?  Thanks

coreysyms:
this looks good to me

charneykaye:
Thanks!

PR #7346, opened 2015-12-02
Fix "media-object-section" mixin

R4c00n:
Fix `media-object-section` mixin in order to work when compiling with
Compass.

### DIFF
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -35,7 +35,7 @@ $mediaobject-image-width-stacked: 100% !default;
     padding-#{$global-right}: $padding;
   }
 
-  &:last-child:not(+ $selector:first-child) {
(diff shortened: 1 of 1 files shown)


tjhiggins:
+1

gakimball:
Closing in favor of #7349 which is a slightly cleaner solution.

IllusionPerdu:
@gakimball the  https://github.com/zurb/foundation-sites/pull/7349 generate this css
```css
  .media-object-section:last-child:not( + &:first-child) {
```
this genrate the good css

R4c00n:
@gakimball PR #7349 is causing a bug that was fixed a few days ago. Therefore the $selector variable was introduced in #7270, but it was used a little wrong. This PR #7346 is fixing the issue by wrapping the Variable in #{}, like it it is done multiple times in the foundation core.

PR #7349, opened 2015-12-02
Fixes for django-compressor compressing foundation.

tjhiggins:
Running the latest version of django compressor/sass ruby.

scss/components/_media-object.scss
    Breaks when trying to compress the file becauses of $selector.
scss/util/_mixins.scss
    Compiles but corrupts css output making screen unusable.

Not sure why this is happening since it supports ruby sass 3.4+. Any ideas?


tjhiggins:
Oh missed this pr right before mine...
https://github.com/zurb/foundation-sites/pull/7346

gakimball:
Merging this one because it's a slightly cleaner solution.

PR #7353, opened 2015-12-02
[Dropdown] Add back closeOnClick functionality for dropdowns.

tjhiggins:
Use this functionality heavily on our site.
Unsure why it was excluded in 6 so I disabled by default.

Shamelessly copied this commented out code from DropdownMenu. Then retooled to fit the new flow.

Do I also have to rebuild foundation.js in this pr?

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -69,7 +69,13 @@
      * @option
      * @example true
      */
-    autoFocus: false
+    autoFocus: false,
+    /**
+     * Allows a click on the body to close the dropdown.
+     * @option
+     * @example true
+     */
+    closeOnClick: false
   };
   /**
    * Initializes the plugin by setting/checking options and attributes, adding helper variables, and saving the anchor.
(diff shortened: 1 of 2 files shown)


tjhiggins:
Looks like this breaks anything with data-close in dropdown. Going to fix and repush.

tjhiggins:
Updated. Solution is a little heavy handed, but powerful. Suggestions?

zurbchris:
seems like it would be much faster and less code to say: 
```js
if(_this.$element.find(e.target).length){
  return;
}
_this.close();
$body.off...
```
for the check if the click was in the dd or not. Otherwise, I think this is a fine feature.

tjhiggins:
@zurbchris Updated

tjhiggins:
Found another bug with my addBodyHandler, and a unused settings var in dropdown.scss

zurbchris:
This is a pretty good feature. I've added it to the 6.1 Milestone list, and there will be a new 6.1 specific branch shortly if you could change this pr to that branch. Don't worry about compiling a new copy of the js, we'll take care of it.

tjhiggins:
@zurbchris https://github.com/zurb/foundation-sites/pull/7478
Should I delete this pr?

zurbchris:
I'll take it in case the new dropdown isn't complete in time. I do like the feature :)

rafibomb:
@tjhiggins Thanks!

PR #7354, opened 2015-12-02
Fix rule for block grid.

stenlyk:
Fix rule for block grid.
In this example you can see http://codepen.io/anon/pen/obvdaN that grid inside block grid doesn't work correctly. It's apply size from block grid instead from grid definition.


zurbrandon:
This has actually been fixed for 6.0.5, that was an old CDN link

This pen has a new one. 
http://codepen.io/zurbrandon/pen/YwzXJx

Thanks though we always appreciate the help.

PR #7360, opened 2015-12-03
Make custom drilldown backButton markup working

pavelleonidov:
The drilldown back event currently works only with the default markup. The reason is it's currently triggered only when a custom markup doesn't exist. Fix by detaching _this._back($menu) from the existing condition.

See http://foundation.zurb.com/forum/posts/36661-bug-in-custom-drilldown-backbutton-markup


zurbchris:
Nice! Thanks for that little fix.

PR #7374, opened 2015-12-03
Fixed plugin 'double dip' in foundation.reflow.fn

flipnode:
Checking for “double dip” on plugins should use the jQuery ‘data’
instead of ‘attr’ since ‘zf-plugin’ is set as a jQuery data key when
the instance of the plugin on the element is created.


zurbchris:
haha! simple things... thanks.

PR #7389, opened 2015-12-04
auto width on button inputs in topbar

gpspake:
Inputs in the topbar use the fixed $topbar-input-width and are 200px wide by default.
That applies to button inputs as well which is probably not desirable for a search submit button, for instance.
This just excludes inputs with the button class from that rule.


gpspake:
just a note, it's probably ideal to use <button type="submit" class="button"> in this situation to avoid this problem, but sometimes search forms use <input type="submit" class="button">

PR #7412, opened 2015-12-05
Fix button padding & dropdown menu line height

muhammetm:
Fix button padding & dropdown menu line height

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -68,7 +68,7 @@ $button-opacity-disabled: 0.25 !default;
   vertical-align: middle;
   border: 1px solid transparent;
   border-radius: $button-radius;
-  padding: $button-padding;
+  padding: $button-padding !important;
   margin: $button-margin;
   font-size: map-get($button-sizes, default);
 }
@@ -114,7 +114,7 @@ $button-opacity-disabled: 0.25 !default;
   }
 
   background: $background;
-  color: $color;
+  color: $color !important;
 
   &:hover, &:focus {
     background: $background-hover;
@@ -134,7 +134,7 @@ $button-opacity-disabled: 0.25 !default;
   $color-hover: scale-color($color, $lightness: -50%);
 
   border: 1px solid $color;
-  color: $color;
+  color: $color !important;
 
   &:hover, &:focus {
     border-color: $color-hover;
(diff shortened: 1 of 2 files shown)


zurbrandon:
Is this a fix for buttons or the center class?

muhammetm:
@zurbrandon  I updated the comment, it's for button padding and dropdown menu line height.

andycochran:
Surely we can avoid `!important` here (which seems to address a separate issue).

zurbrandon:
@muhammetm is there a way to do this same thing without an !important

muhammetm:
@zurbrandon  I couldn't find a way, top bar's very specific selectors override button styles.

andycochran:
Maybe add :not(.button) to the topbar styles?

muhammetm:
What if it's a button with a custom class name created by mixin?

andycochran:
Ack! I see the problem. Still, I'd still rather not have !important anywhere. Maybe using `.button` inside `.topbar .menu` is an app-specific edge case that requires custom style rules.

gakimball:
Currently reviewing this. Can you highlight which top bar selectors are creating issues? I'd like to tackle this from the other direction, making top bar CSS more generic if possible.

muhammetm:
It's getting overwritten by the selector below:

`.menu > li:not(.menu-text) > a {
    display: block;
    line-height: 1;
    padding: 0.7rem 1rem;
}`

capture

gakimball:
@muhammetm You might consider using a `<button>` or `<input>` element instead. "Search" is an action, so either of those elements will be more appropriate than `<a>`. From an accessibility perspective, `<a>` is for links to other pages, while `<button>` and `<input>` perform actions on the current page.

muhammetm:
I understand. It was just an example. There are cases when button is linked to another page, but it's not much of a big issue after all.

gakimball:
@muhammetm It's definitely something we should try to fix, but adding `!important` to buttons, a component used in many places, is a little spooky.

We did cut back the specificity of that menu selector to `.menu > li > a`, but that's not enough for a single `.button` to win.

muhammetm:
Yes, you are right about `!important`.

gakimball:
Going to close this, as this isn't the way we want to solve this problem. We're aware it's an issue, however.

Issue #7281, opened 2015-11-28
Reveal modal pop z-index incorrect on Safari on iOS 8 using Foundation 6.0.4


rickcurran:
Reveal popups are appearing incorrectly in Safari on iOS8, there appears to be a z-index issue and the modal is appearing behind the overlay. This is Foundation 6.0.4

aibrean:
I noticed that while using reveals in tabs.

zurbrandon:
All fixed up in 6.0.6

Thanks for all the Issues on this and being so thorough, man!

Issue #7280, opened 2015-11-28
Reveal modals are not opening correctly in Safari on iOS 9 using Foundation 6.0.4


rickcurran:
Reveal modals are not opening correctly in Safari on iOS 9 using Foundation 6.0.4, the background overlay appears but the actual popup modal itself doesn't appear. Changing orientation of the device sometimes makes them appear. Please see issue #7279 which is for the same issue on OSX: https://github.com/zurb/foundation-sites/issues/7279

This is happening in my own usage but you can see the same behaviour on the Reveal Docs page:

http://foundation.zurb.com/sites/docs/reveal.html

aibrean:
![reveal-zindex](https://cloud.githubusercontent.com/assets/6740252/11573831/567daf5c-99d6-11e5-9c5a-6e57e0c52497.png)

rickcurran:
I was getting this issue without off-canvas in my page. One other thing to note is that on iOS all browsers use the same browser framework underneath (part of the deal with being on iOS) so this issue should happen in all iOS browsers.

aibrean:
There is a PR right now to resolve this although it looks like offcanvas specific.

zurbrandon:
This looks to be all fixed up now in 6.0.6.

PR #7415, opened 2015-12-05
Fixed issue with Reveal modal being rendered below the backdrop on iOS

### LINKED ISSUE #7380: [Reveal][Off-Canvas] Z-index issue on iOS (v9 iPhone 6 tested on Chrome)
F6.0.5
Use the actual device (I know it definitely is a bug with iOS 9 on an iPhone 6). The Chrome emulator doesn't replicate it. I was running Chrome on there. Seems to be iOS specific rather than browser in iOS.

I'm thinking that: 
+ it's directly related to having `off-canvas` on the site 
+ it's related to where `reveal-overlay` is inserted - specifically outside of the off-canvas wrappers.

Proper z-indexing so the reveal is above the overlay. Seems to be isolated on sites using off-canvas.

**What happened instead?**

Everything is darkened.

**Test case**

reveal-zindex

Owlbertz:
Fixes #7380, as well as #7280, #7281.
See example here http://codepen.io/Owlbertz/pen/eJOQVv.

### DIFF
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -53,7 +53,7 @@ $maincontent-shadow: 0 0 10px rgba($black, 0.5) !default;
     overflow-x: hidden;
     position: relative;
     backface-visibility: hidden;
-    -webkit-overflow-scrolling: touch;
+    -webkit-overflow-scrolling: auto;
   }
 
   .off-canvas-wrapper-inner {


aibrean:
Nice :)

gakimball:
image

Owlbertz:
I really wish all PRs would be merge this way :laughing:

PR #7427, opened 2015-12-07
fixes #7423 [Equalizer] reflow , init is buggy | please take a look

### LINKED ISSUE #7423: [Equalizer] reflow , init is buggy
in the reflow (init) of Equalizer is
```js
$('[' + this.attr + ']').each(function() {
      var $eqParent       = $(this),
          adjustedHeights = [],
          $images = $eqParent.find('img');

      if ($images.length) {
        Foundation.onImagesLoaded($images, function() {
          adjustedHeights = self.getHeights($eqParent);
          self.applyHeight($eqParent, adjustedHeights);
        });
      }
      else {
        adjustedHeights = self.getHeights($eqParent);
        self.applyHeight($eqParent, adjustedHeights);
      }
    });
```
this causes the functions to execute as many data-equalizer elements you have in the markup with the wrong settings from the actual equalizer
Because there is the each function
for Example
you got 3 data-equalizer tags in the html so the plugin executes 3 tims and 3 time the each loop gets the heights and adjuststhe heights but with the settings from the wrong element

1. settings equalizeOnStack=true
2. settings equalizeOnStack=false
3. settings equalizeOnStack=true

so the 2. equalizer gets executed last in the each loop with settings equalizeOnStack=true what are not the settings of the 2. equalizer

so why the each loop is done in the reflow???

i hope its understandable ;-)

designerno1:
this fixes the reflow problem for me

### DIFF
--- a/js/foundation.equalizer.js
+++ b/js/foundation.equalizer.js
@@ -57,7 +57,9 @@
     this.$window
       .off('.equalizer')
       .on('resize.fndtn.equalizer', Foundation.util.throttle(function () {
-        self._reflow();
+        $('[' + self.attr + ']').each(function(){
+          $(this).data('zfPlugin')._reflow();
+        });
       }, self.options.throttleInterval));
   };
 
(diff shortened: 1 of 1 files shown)


zurbchris:
There's a new version of Equalizer, coming to a repo near you, for the 6.1 release. This should no longer be an issue at that time.

zurbrandon:
We've rewritten a large chunk of Equalizer to remedy this. It'll be out very soon. I'll close this since the code won;t match up, but thanks for the PR regardless.

PR #7432, opened 2015-12-07
[Dropdown] Add new dropdown option to persist hover over pane.

tjhiggins:
This reimplements old v5 default functionality.


zurbchris:
Thanks!

PR #7442, opened 2015-12-08
V5

ropic:
foundation.dropdown.js on close action only set aria-expanded=false on first element, my fix:

```
dropdown.each(function (key, value) {
var original_target = $('[' + self.attr_name() + '=' + dropdown[key].id + ']') || $('aria-controls=' + dropdown[key].id+ ']');
```

Using key and not 0


zurbchris:
I think this pr needs to target the V5 branch instead of develop.

gakimball:
This PR is comparing `develop` and `V5`, neither of which are your branch. If you can open a new PR that compares your fork with `develop`, that would be great :) Thanks!

Issue #7443, opened 2015-12-08
Top-Bar Link Color is White


JeremyEnglert:
By default, $topbar-link-color is white - which is really hard to see on the topbar which is also a light color. 

![screen shot 2015-12-07 at 9 08 39 pm](https://cloud.githubusercontent.com/assets/6110968/11647688/4860a32a-9d28-11e5-8b98-d7e7ee54aa39.png)

twd3:
Yeah, that was added in 6.0.5 for some reason. It comes from:
```.top-bar li > a, .top-bar ul li > a {
      color: #fff; }```

gakimball:
Yep, this has been fixed but hasn't landed in a release yet. We'll likely do 6.0.6 by the end of the week. 9bf0552ef5f737c1e209e9efd5c2e76b05afbdbe

PR #7444, opened 2015-12-08
Makes default topbar link color more legible

JeremyEnglert:
Fixes https://github.com/zurb/foundation-sites/issues/7443


gakimball:
Already fixed via 9bf0552ef5f737c1e209e9efd5c2e76b05afbdbe

The settings file is automatically generated, but since we only generate it right before we tag a release, it's out of date right now.

PR #7450, opened 2015-12-08
Fixes #7449 [Magellan] Last menu element active on load if body has no margin

### LINKED ISSUE #7449: [Magellan] Last menu element active on load if body has no margin

1. Add `body { margin: 0 }` to your css

Note that setting `html, body { height: 100% }` resolves this issue. 

First menu element having `active` class when page first loads.

**What happened instead?**

Last menu element has `active` class instead.

**Test case**

http://s.codepen.io/afonsoduarte/debug/LGELmz

afonsoduarte:
Uses `document.documentElement.clientHeight` to calculate viewport height instead of `document.body.clientHeight`. See http://jsfiddle.net/tzdcx/1/ for differences between the two.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -93,7 +93,7 @@
         html = document.documentElement;
 
     this.points = [];
-    this.winHeight = Math.round(Math.max(window.innerHeight, document.body.clientHeight));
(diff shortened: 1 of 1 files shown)


zurbchris:
Looks good, thanks for catching that.

PR #7454, opened 2015-12-08
[Table] Don't at-root the inner table styles. Makes it impossible to namespace

tjhiggins:
Unsure the original reason for the @at-root, but this mixin gets set on the table anyways.

### DIFF
--- a/scss/components/_table.scss
+++ b/scss/components/_table.scss
@@ -61,67 +61,65 @@ $table-head-font-color: $body-font-color !default;
 $show-header-for-stacked: false !default;
 
 /// Adds the general styles for tables.
-/// @param {Keyword} $stripe [$table-stripe] - Uses kewords even, odd, or none to darken rows of the table. The defualt value is even.
+/// @param {Keyword} $stripe [$table-stripe] - Uses keywords even, odd, or none to darken rows of the table. The default value is even.
 @mixin table($stripe: $table-stripe) {
   margin-bottom: $global-margin;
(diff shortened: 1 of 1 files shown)


gakimball:
I'd like to make this configurable instead. The reason we use @at-root is to keep the specificity lower. For example, the selector `thead` is better than `table thead`. *However*, that's not what you want if you're applying the mixin to a class.

You could add a parameter to the mixin like `$nest`. If it's `false`, it uses @at-root, and if not, it doesn't.

tjhiggins:
@gakimball What would be the correct way to conditionally @at-root. Only thing that I can think of is another helper mixin.

gakimball:
It'll look a bit messy, but yeah, that's a solution. It can be a private mixin, so we won't document it anywhere. The code structure will look something like this:

```scss
@mixin -zf-table-children-styles {
  // ...
}

@mixin table(
  $stripe: $table-stripe,
  $nest: false
) {
  @if $nest {
    @include -zf-table-children-styles;
  }
  @else {
    @at-root {
      @include -zf-table-children-styles;
    }
  }
}
```

tjhiggins:
@gakimball Updated. Is this what you had in mind or should there be a global variable (table-nest) in _settings?

gakimball:
Nah, no global variable is needed. Thanks for implementing the feedback! :)

xzyfer:
It's worth nothing this usage of `at-root` is currently incompatible with LibSass.

PR #7456, opened 2015-12-08
Add note to block grid documentation

NeonArray:
Add note to block grid documentation pointing out that Foundation defaults to a maximum of six columns.


zurbrandon:
Thanks!

PR #7459, opened 2015-12-09
New Equalizer

zurbchris:
Rewrites equalizer. Adds new option: equalizeOn:"breakpointStr", improves performance of nested plugins, and removes window resize event listener, and (now) unused debounce option. Also fixes an issue in the docs with collapsing margins on nested plugins. It's pretty dope...


docs/pages/equalizer.md
```diff
@@ -35,7 +35,7 @@ To use one Equalizer inside another, each container needs a unique ID, assigned
 In the below example, the first set of Equalizer elements have the value `foo`, while the inside elements have the value `bar`.
 
 ```html
-<div class="row" data-equalizer="foo">
+<div class="row" datadata-equalizer="foo">
```
Renkas: datadata - seems lika a typo

Issue #7452, opened 2015-12-08
*-uncollapse causes double padding


tamcy:
Please refer to this example:
http://codepen.io/anon/pen/rxaZvp

If I understand correctly, the `*-collapse` class in the grid (and its corresponding mixin) removes the left and right gutter/padding of a column, while `*-uncollapse` class restores it. So for the code above, I expect to see:
(1) No gutter in small screen size.
(2) The gutter should be restored in medium and large screen size, i.e. it should align with the second row without any `*-collapse` or `*-uncollapse` class. 

However, in reality, (2) behaves differently than what I expected. It looks like double padding is added to the `medium-uncollapse` columns, which makes it narrower than the row below it.

The same markup with foundation 5.5.0 seems to behavior more properly:
http://codepen.io/anon/pen/jWEvxV

As I am completely new to foundation I'm not sure whether this change is intended or not, but the doc leads me to think the new behavior isn't that correct. Thanks.

zurbrandon:
This has been fixed by a PR, so i'll close this out. Thanks!

PR #7463, opened 2015-12-09
fix incorrect gutter padding to *-uncollapse classes

tamcy:
Here is a proposed fix to issue #7452. In addition, it now accepts a $gutter parameter so that columns with non-default gutter value can make use of this mixin.

### DIFF
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -13,9 +13,12 @@
 }
 
 /// Un-collapse the gutters on a column by re-adding the padding.
-@mixin grid-column-uncollapse() {
-  padding-left: $grid-column-gutter;
-  padding-right: $grid-column-gutter;
+///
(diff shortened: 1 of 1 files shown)


smg6511:
In a project I'm working on in which there is a nested row with the *-uncollapse class applied, I still am seeing the unaltered (double) padding being applied (i.e., in my _settings where there is a 30px $grid-column-responsive-gutter for the medium breakpoint ... this value is being applied on my medium-uncollapse child columns instead of the rem_calc(30px)/2). Any ideas why this might happen?

thomasHolman:
I'm having the same issue.

See screenshot below. Default gutter set to 24px resulting in 0.75rem (12px) column gutters.
When using uncollapse, padding is restored using the full 24px gutter width.

screen shot 2016-02-05 at 18 07 05

PR #7464, opened 2015-12-09
Fix focus state/simplify select triangles

tjhiggins:
Tested in Chrome/IE 10


gakimball:
The `background` to `background-color` change is good. However, the color in the SVG has to be `rgb()`, because Firefox doesn't like hex values in embedded SVGs.

tjhiggins:
@gakimball Updated the pr. I had to encode the rgb parens because it was compiling and adding a single quote which broke the url.

Tested in:
Chrome, Firefox, IE10

gakimball:
Awesome, thanks!

Issue #7441, opened 2015-12-08
[Accordion] Accordion can only use <ul> and <li>


oolongcha:
**How can we reproduce this bug?**

Tried to make an accordion with dl and dd. Would not work until I changed it all to ul and li

**What did you expect to happen?**

I expected the accordion to work with dl and dd, as stated in the docs, "we use ul, but you can use any element you want"

http://foundation.zurb.com/sites/docs/accordion.html

**What happened instead?**

Accordion would only display first panel and none of the others.

Knight-Yoshi:
If you're using Foundation 6 the accordion component is apart of the general menu component which uses unordered lists.
http://foundation.zurb.com/sites/docs/accordion-menu.html

oolongcha:
Ah, right, I should clarify, this is not a bug for the accordion menu. Thanks for pointing that out.

Knight-Yoshi:
Yeah the accordion doesn't seem to work with anything other than lists

ocularrhythm:
I created a pull request for this issue. Line 58 of `foundation.accordion.js` was using a general `li` selector, instead of the `accordion-item` class. I changed it and tested it with nested divs and it worked correctly.

PR #7469, opened 2015-12-09
Issue #7441 Enable non-list elements in Accordions

NeonArray:
Fixes [#7441](https://github.com/zurb/foundation-sites/issues/7441) Instead of selecting an `li` element, I set the selector to `accordion-item`. I tested using nested div elements instead of a list and it worked correctly.

### DIFF
--- a/js/foundation.accordion.js
+++ b/js/foundation.accordion.js
@@ -55,7 +55,10 @@
    */
   Accordion.prototype._init = function() {
     this.$element.attr('role', 'tablist');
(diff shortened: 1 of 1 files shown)


zurbchris:
See #7231 
Would love to get this in before patch time.

ocularrhythm:
I've added an additional check for the data attribute in the logic. I have it first check for the list element and if zero are returned, then look for the attribute. I tested it using two separate accordion components, one using nested `div` and the other using `ul` and `li`

Issue #7406, opened 2015-12-05
Can't use Type in Abide Validation


calmass:
Validation with Abide only works with the "pattern" attribute set, not based on the type. In fact, when the "type" attribute is anything other than "text", the validation always fails.  I know you guys are going for code reduction, but the behavior should be to attempt to read the pattern  first, then the type to check for special validation.  I'll write my own javascript shim and use that for now, but out-of-the-box would be prettier and make updates easier.  I'll repost when complete and tested.

zurbchris:
Should be fixed with #7590, please comment and submit pr's to the `abide` branch until 6.1 release.

Issue #7357, opened 2015-12-03
Form validation with data-abide


anothergituser:
Same as #7127
Impossible to submit for after validation is ok.
Please check..

anothergituser:
Well, to answer my own problem, edit foundation.abide.js
These are the changes

```javascript
Abide.prototype._events = function() {
    var self = this;
    this.$element
        .off('.abide')
        .on('reset.fndtn.abide', function(e) {
            self.resetForm($(this));
        })
        .on('submit.fndtn.abide', function(e) {
            // e.preventDefault();
            // comment e.preventDefault and returnt the result of self.validateForm
            // just like abide from foundation 5 does
            return self.validateForm(self.$element);
        })
        .find('input, textarea, select')
        .off('.abide')
        .on('blur.fndtn.abide change.fndtn.abide', function(e) {
            if (self.options.validateOn === 'fieldChange') {
                self.validateInput($(e.target), self.$element);
            }
            // self.validateForm(self.$element);
        })
        .on('keydown.fndtn.abide', function(e) {
            // if (settings.live_validate === true && e.which != 9) {
            //   clearTimeout(self.timer);
            //   self.timer = setTimeout(function () {
            //     self.validate([this], e);
            //   }.bind(this), settings.timeout);
            // }
            // self.validateForm(self.$element);
        });

},
```

and the validation function
```javascript
Abide.prototype.validateForm = function($form) {
    var self = this,
        inputs = $form.find('input'),
        inputCount = $form.find('input').length,
        counter = 0;

    while (counter < inputCount) {
        self.validateInput($(inputs[counter]), $form);
        counter++;
    }

    // what are all the things that can go wrong with a form?
    if ($form.find('.form-error.is-visible').length || $form.find('.is-invalid-label').length) {
        $form.find('[data-abide-error]').css('display', 'block');
        return false; // don't submit when there are errors
    } else {
        $form.find('[data-abide-error]').css('display', 'none');
        return true; // ok to submit
    }
};
```

Also this is my gulp file in case anyone needs it to compile this
```javascript
var gulp = require('gulp');
var $    = require('gulp-load-plugins')();

var uglify = require('gulp-uglify');
var rename = require('gulp-rename');
var concat = require('gulp-concat');

var sassPaths = [
  'bower_components/foundation-sites/scss',
  'bower_components/motion-ui/src'
];

gulp.task('sass', function() {
  return gulp.src('scss/app.scss')
    .pipe($.sass({
[...]

Issue #7345, opened 2015-12-02
Abide Required Email & Password Always Invalid


Knight-Yoshi:
Similar to this issue,  https://github.com/zurb/foundation-sites/issues/7304, required email and password fields are always invalid.

There's also a small syntax error with the Abide Error State example on the site. The is-invalid-input is in the type attribute, not the class
 <input type="text is-invalid-input" required aria-invalid aria-describedby="uuid">

johanneslamers:
Have the same problem. Always invalid on email.

````
<input autocomplete="on" id="fields-emailAddress" name="fields[emailAddress]" placeholder="Email" type="email" value="" required>

jamilservicos:
fix for password input => [#7300](https://github.com/zurb/foundation-sites/pull/7300/files)

example working of email input: `<input type="text" pattern="email" placeholder="Email" id="Email" aria-describedby="registerEmail" required="">`

zurbchris:
Should be fixed with #7590, please comment and submit pr's to the `abide` branch until 6.1 release.

Issue #7304, opened 2015-11-30
Abide required textarea always invalid


ariordan:
If a textarea is created empty, then abide will always report the textarea as invalid, even when it has a value. However, if there area is created with a value, validation will subsequently operate as expected, even when empty.

This appears be caused by the call "$el.is(':empty')" at line 154 of foundation.abide.js, returning true.

foundation - 6.0.4
jquery - 2.1.4

zurbchris:
Should be fixed with #7590, please comment and submit pr's to the `abide` branch until 6.1 release.

PR #7470, opened 2015-12-09
adds missing abide features without splitting inputs into types (code…

### LINKED ISSUE #7424: [Abide] Password type fields couldn't passed the required rule

1. Use a type "password" into input fields with required attribute <input type="password" name="password"  required />
2. Put a value to the password input
3. Submit the form. It's always throws an error and form can't pass

It should pass the validation since It's already has a value

**What happened instead?**
Did not pass the form validation

Give us a link to a [CodePen](http://codepen.io) or [JSFiddle](http://jsfiddle.net) that recreates the issue.

designerno1:
… reduction)
This should fix #7424 #7406 #7357 #7345 #7304 #7230 #7228 #7226 #7209 #7196 #7127
adds support for html5 input types, selects (multi), textareas
adds support for html5 elements through patternLib
adds support for pattern and data-pattern through patternLib and own patterns

everything is tested in a testcase, for me it works without splitting the types.
For those input elements (email..) which return a el.type for example email and a pattern exists in the patternLib it will be validated with this.

what have i done?
1. fix the default case of requiredCheck because is(':emtpy') checks for innerhtml elements in inputs and there aren't
2. remove not used

``` js
textInput = $form.find('input[type="text"]'),
checkInput = $form.find('input[type="checkbox"]'),
```
1. inputcount improvement
2. submit form if valid
3. pattern support
   in html5 type overrides pattern
   in my abide i handle it like this

```
Data-pattern overrides
Pattern overrides
type
```

so you can flexible choose whatyou would like to use
even a input and for example type email, or pattern, without required, can be validated if there is something typed in.

i tested the following elements
input types and elements
    hidden
    text
    password
    color
    date
    datetime
    datetime-local
    email
    month
    number
    range
    search
    tel
    time
    url
    week
    select
    select multiple
    textarea
    radio
    checkbox
    reset
    submit

if you want to i cann paste my test html

please have a look maybe this fixes the issues without splitting the input types and bloat the code

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -62,9 +62,9 @@
       color : /^#?([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$/
     },
     validators: {
-      equalTo: function (el, required, parent) {
-        var from  = document.getElementById(el.getAttribute(this.add_namespace('data-equalto'))).value,
-            to    = el.value,
+      equalTo: function ($el, required, parent) {
+        var from = $('#'+$el.attr('data-equalto')).val(),
+            to    = $el.val(),
             valid = (from === to);
 
         return valid;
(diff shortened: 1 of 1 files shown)


designerno1:
validator can also be added later

zurbchris:
See #7407 
We want to push out Abide with the 6.1 release, I could use some community support in making sure all the pieces are there beforehand.

designerno1:
maybe for 6.1 see #7481

PR #7478, opened 2015-12-10
[Dropdown] Add back closeOnClick functionality for dropdowns.

tjhiggins:
Use this functionality heavily on our site.
Unsure why it was excluded in 6 so I disabled by default.


zurbchris:
added to develop

PR #7479, opened 2015-12-10
[Dropdown] Fix positioning of dropdown when overflowing top/bottom.

tjhiggins:
Reset position class when resizing window back to no-overflow.

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -166,7 +166,7 @@
         param = (direction === 'top') ? 'height' : 'width',
         offset = (param === 'height') ? this.options.vOffset : this.options.hOffset;
 
-    if(($eleDims.width >= $eleDims.windowDims.width) || (!this.counter && !Foundation.Box.ImNotTouchingYou(this.$element, null, true, false))){
+    if($eleDims.width >= $eleDims.windowDims.width){
       this.$element.offset(Foundation.Box.GetOffsets(this.$element, this.$anchor, 'center bottom', this.options.vOffset, this.options.hOffset, true)).css({
         'width': $eleDims.windowDims.width - (this.options.hOffset * 2),
         'height': 'auto'
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -142,9 +142,12 @@
     }else if(position === 'right' && (this.usedPositions.indexOf('left') > -1) && (this.usedPositions.indexOf('bottom') < 0)){
       this.$element.removeClass(position);
     }
-    //if nothing cleared, set to bottom
+    //if nothing cleared, set to positionClass otherwise bottom
     else{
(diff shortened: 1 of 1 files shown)


js/foundation.dropdown.js
```diff
@@ -157,6 +157,17 @@
    */
   Dropdown.prototype._setPosition = function(){
     if(this.$anchor.attr('aria-expanded') === 'false'){ return false; }
+
+    if(this.classChanged){
+      var curPositionClass = this.getPositionClass();
```
tjhiggins: Copied reset class code from close function. Was on the fence about creating a helper method for both. Zurbs call.

zurbchris:
Do you have an issue to point to for this pr?

tjhiggins:
I didn't create one. 

The first issue is that the dropdown's width gets set to the window on overflow of top/bottom, but you're just scrolling the dropdown off the window. This change already exists in the reveal setPosition. 

Here is a gif of the second issue. The dropdown's width gets set to full when window gets small, but doesn't reset width when window resizes again.
http://i.imgur.com/PfCFXYo.gifv

tjhiggins:
Moved the resetting outside of every _setPosition, and it fixed this bug too:
http://i.imgur.com/IsnzZin.gifv

js/foundation.dropdown.js
```diff
@@ -165,7 +166,7 @@
         param = (direction === 'top') ? 'height' : 'width',
         offset = (param === 'height') ? this.options.vOffset : this.options.hOffset;
 
-    if(($eleDims.width >= $eleDims.windowDims.width) || (!this.counter && !Foundation.Box.ImNotTouchingYou(this.$element))){
+    if(($eleDims.width >= $eleDims.windowDims.width) || (!this.counter && !Foundation.Box.ImNotTouchingYou(this.$element, null, true, false))){
```
tjhiggins: This is still a problem because of the repositioning. If the dropdown overflows with the top then it repositions itself and could overlap with the left/right. Tempted to remove.
```(!this.counter && !Foundation.Box.ImNotTouchingYou(this.$element, null, true, false)```
Not understanding when you would want to dynamically set the width except when:
```$eleDims.width >= $eleDims.windowDims.width```

zurbchris: I would be fine with that change. If you want to adjust this pr for that condition, that would be great.

tjhiggins:
@zurbchris Updated. Hope you all had a good break!

tjhiggins:
Also would like to question why the dropdown defaults to bottom if all repositioning fails. 
For example:
I specify I want my dropdown to open to the left, and it collides with the bottom in all positions.
Then it will open on the bottom which it is colliding with... 
Seems like a default of bottom is fine, but why not open with the position specified (left).

PR #7481, opened 2015-12-10
adds missing abide features see #7470 PR for 6.1

designerno1:
#7470

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -62,9 +62,9 @@
       color : /^#?([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$/
     },
     validators: {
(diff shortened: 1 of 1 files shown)


designerno1:
ok i trigger the submit with a formValid handler so it is possible to
```js
$('[data-abide]').off('formValid').on('formValid.fndtn.abide', function() {
  console.log('do something');
});
```
in an own integration and do a ajax submit

zurbchris:
@designerno1 a couple things, I'm working on Abide right now, and I've set up a new branch for it. Also, could you jump on gitter for a quick chat or email me?

zurbchris:
Shouldn't be necessary with #7590. Submit pr's directly on the `abide` branch.

PR #7483, opened 2015-12-10
adds support for equalizeByRow - Responsive Equalizer

designerno1:
Responsive Equalizer
a very helpful feature for example for Block Grid or Elements that are stack on mobile but should be equalized on medium or large, but medium ist different rows to large

i make a commit with doc examples to show the features

### DIFF
--- a/docs/pages/equalizer.md
+++ b/docs/pages/equalizer.md
@@ -79,3 +79,146 @@ In the below example, the first set of Equalizer elements have the value `foo`, 
     </div>
   </div>
 </div>
+
+## Responsive Equalizer - equalize row by row
+
+Equalizer works the Elements row by row, so every data-equalizer-watch on one row has the highest height from the row.
+Usefull for Rows that should be the Same height on every breakpoint because it works responsive
+
+<div class="callout success">
+  <strong>Works great with Block-Grid</strong>
+</div>
+
+<div class="callout primary">
+  <strong>Under the Hood</strong><br>
+  Equalizer splits the data-equalizer-watch elements into groups by checking the offset().top position and then sets the max height to each element in a row<br>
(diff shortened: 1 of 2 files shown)


designerno1:
2. commit are docs examples but the have to be worked over.
Only to show the feature

zurbchris:
OOOoooooohhhh WEEEeeeeeeee, that's a doozie. I'll have a look.

designerno1:
maybe some parts can be refined, i take a deeper look tomorrow.
i wrote the feature for the old version and adapted it to the new-eq.
i have to check some parts in the new version but that's only code reduction and better code.

zurbchris:
:open_mouth: This is hoooooooooOOOOOOTTT! Visually speaking. I haven't yet checkout out the code.

designerno1:
i think in code:
array handling could be improved (it's little tricky in JS)
offset().top performance on each element could be faster ;-) i checked native offsetTop but this isn't really faster

zurbchris:
Would love to merge this. I played a bit with the methods used, and while offsetTop gives some unexpected results, `el.getBoundingClientRect().top` works great.

zurbchris:
```js
      var elOffsetTop = ~~(this.$watched[i].getBoundingClientRect().top);
```
this is pretty hot.

zurbchris:
Switching to getBoundingClientRect and using native js Array.map shaved off a few ms in chrome console profiler, not much, but something.
I'm going to apply those changes, merge this in, and merge it to 6.1

rafibomb:
@designerno1 Thanks for your help on the Framework! We want to mail you a Yeti shirt! Email foundation at zurb .com and we'll set you up!

rafibomb:
@designerno1 I found your address! Sending you a shirt!

Issue #7160, opened 2015-11-23
Disabled <button> elements doesn't get styled 


JoelESvensson:
I've included the button mixin but my `<button>` element doesn't get the `button[disabled]` style

ocularrhythm:
I tried reproducing this bug using Foundation 6.0.5 but the button imports the correct styling.

JoelESvensson:
Are you really sure that you get the styling from `[disabled]` and not `.disabled`?

ocularrhythm:
I created a test class `.testbutton` and inside the rule I included the mixin `@include button-disabled;`

I then created a button element `<button class="button testbutton">Test</button>` which had the CSS properties of `opacity: 0.25` `cursor: not-allowed` `pointer-events: none`.

If you are expecting the mixin to add a `disabled` attribute to the button, that is not what it does. This only adds the three CSS properties to the rule. From the docs `Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.`

JoelESvensson:
Okay. My main point is that I expected `@include foundation-button;` to include that styling for the disabled attribute as well. In my opinion it should because it seems like a very standard use case to add that

JoelESvensson:
(I accidentally clicked close)

ocularrhythm:
My mistake, I totally misunderstood what you meant. Looking in the `_buttons.scss` it is clear that there is no styling being given to the `disabled` attribute. I'm not sure how Zurb wants to approach this, but I have made a [fix](https://github.com/zurb/foundation-sites/pull/7485).

JoelESvensson:
:) Thanks!

zurbrandon:
Merged this in @ocularrhythm . We'll close out this issue. Thanks!

Issue #7387, opened 2015-12-04
Meet AA contrast requirements


gakimball:
I'm combining all of our open issues around contrast into one. Our goal is AA, not AAA.

**Buttons:**
- [ ] [Primary](http://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=2199e8) #7103
- [ ] [Secondary](http://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=777777) #7104
- [ ] [Success](http://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=3adb76) #7376
- [ ] [Alert](http://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=ec5840) #7377
- [ ] [Warning](http://webaim.org/resources/contrastchecker/?fcolor=ffffff&bcolor=ffae00) #7378
- [x] Disabled #7379

Disabled is an open question, as according to the W3C's guidelines, contrast requirements are not necessary for disabled elements.

**If any other components don't meet AA requirements, let's add them here also.**

Owlbertz:
@gakimball the difference between `disabled` and `aria-disabled` should be, that not all elements support `disabled` according to HTML standard and that `disabled` actually disables the element, while `aria-disabled` just indicates a disabled UI component to AT. This means, `buttons` would go with the `disabled` attribute, and `a` would need to use `aria-disabled` or `.disabled`. I feel like we should include all three ways in the CSS selector, and then recommend `disabled` for `button` and `input` in the docs.

Owlbertz:
Another place with too low ratio is the title of Off-canvas list titles: http://webaim.org/resources/contrastchecker/?fcolor=8a8a8a&bcolor=2c2c2c

andycochran:
Also of too low ratio: 
* regular `$primary-color` links
* links in callouts
* `header small`
* `.tabs-title` (especially `.is-active`)
* blockquotes
* `.subheader`

I'm not sure about the requirements for progress bars. It's not text on a background, but the ratio is too low.

davefuentes:
@Owlbertz is right. Disabled elements don't need to meet WCAG criteria. We can skip that one.

davefuentes:
I think if we check colors with backgrounds and foregrounds, we can fix all possible issues:

```
$primary-color: #2199e8;
$secondary-color: #777;
$success-color: #3adb76;
$warning-color: #ffae00;
$alert-color: #ec5840;
[...]

Issue #7379, opened 2015-12-03
[Global Styles] Disabled buttons' colors are not accessible


andycochran:
_This is related to #7378, #7377, #7376, #7104, and #7103._

Accessibility Question: 
Does it matter that an element is accessible if it cannot be interacted with? 
(Perhaps @Owlbertz might weigh in.)

Owlbertz:
I think there are two things to consider: people that are (partly) not able to see colors and screen readers (for people that cannot see at all). 
For the first case, the disabled style is not accessible for sure because of the pretty low (its 3.01 to 1, needs to be at least 4.5 to 1). I am not sure if there are any hard requirements for the style of disabled states, maybe the cursor could change to the disabled state when hovering? 
Regarding screen readers everything should be fine as long as there is the disabled attribute. This is currently not set, but I think this - rather than the class - should indicate the disabled state.

zurbrandon:
Based on this conversation i'm inclined to say we could use a disabled cursor on them but that its ok they didn't have a correct contrast ratio. 

http://ux.stackexchange.com/questions/80881/do-disabled-buttons-still-need-to-be-contrast-compliant-for-accessibility

andycochran:
I agree. 

Since W3C says… 

> 1.4.3 Contrast (Minimum): The visual presentation of text and images of text has a contrast ratio of at least 4.5:1, except for the following:
>
>*  (Level AA) Large Text: Large-scale text and images of large-scale text have a contrast ratio of at least 3:1;
* Incidental: Text or images of text that are part of an **inactive user interface component**, that are pure decoration, that are not visible to anyone, or that are part of a picture that contains significant other visual content, have no contrast requirement.
* Logotypes: Text that is part of a logo or brand name has no minimum contrast requirement.

I'm closing this.

andycochran:
Of course, I might make a PR with 
```
.button[aria-disabled="true"] { 
  cursor: not-allowed; 
}
``` 
Is this what you're suggesting, @Owlbertz?

gakimball:
@andycochran There's also `.button.disabled` and `button[disabled]`.

*Although*, I'm inclined to say we should recommend just one way to indicate a disabled button, and then only have CSS for that.

@Owlbertz @andycochran What's the difference between using `disabled` and `aria-disabled`?

gakimball:
I'm collecting all of the contrast issues into one: #7387

PR #7485, opened 2015-12-11
Issue #7160 Add styling for disabled attribute button

NeonArray:
Add disabled attribute selector to include the disabled styling for buttons.


Owlbertz:
I guess this partly addresses #7387 and #7379:
Maybe the `&[aria-disabled]` should also be in the selector. @andycochran @davefuentes

Issue #7493, opened 2015-12-11
[Global Styles] Changing global font size


therealbenhogan:
Using 6.0.5

The docs state that you can change the base font size by setting $rem-base to $global-font-size and making sure that $global-font-size is a pixel value.

In the _settings scss file I did that eg

    // $global-font-size: 100%;
    $global-font-size: 14px;

    // $rem-base: 16px;
    $rem-base: $global-font-size

However, when you do that, the max width of the site from 

    $global-width: rem-calc(1200);

Is no longer 1200px ... how can we change the global font sizes and still retain the correct global width (and not have it affect the grid breakpoints)

ocularrhythm:
After digging around, it seems that `$rem-base` is being defaulted in the `util/_unit.scss` file. Since the `utils/utils` are being imported/compiled before the `global`, the variable is never overridden with the custom value. Declaring the variable before the `@import 'util/util` in the `_settings` file seems to do the trick.

therealbenhogan:
Hi, that's great - thanks for helping out!

PR #7496, opened 2015-12-11
Issue #7493 Move $rem-base before utils import

NeonArray:
The $rem-base variable would never change to a custom value, since the default was being initialized prior to parsing the global overrides. Moving the variable directly before the utils import corrects this.

Now changing $global-font-size to a pixel based value such as 14px, as well as changing $rem-base: $global-font-size outputs a max-width: 85.71429rem

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -39,13 +39,14 @@
 //  34. Tooltip
 //  35. Top Bar
 
+$global-font-size: 100%;
+$rem-base: 16px;
+
 @import 'util/util';
 
 // 1. Global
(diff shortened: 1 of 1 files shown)


zurbrandon:
Hey @gakimball, we'll need to fix this on our end because the settings file is generated dynamically through octophant, correct?

gakimball:
@zurbrandon Yes, or maybe there's another way around it. I'll look into it.

zurbrandon:
Hey @ocularrhythm  We'll need to fix on our end since that file is generated on deploy. Thanks for the PR though.

PR #7525, opened 2015-12-14
[Abide] add .is-invalid-label to .input-group

viniciusban:
Abide currently applies the `.is-invalid-label` class only to a `<label>` element, but sometimes we have another DOM structure, like this one:

```
<div id="id-name-input-group" class="input-group">
  <label id="id-name-label" for="id-name">
    Your Name <small>(Required)</small>
  </label>
  <input required type="text" name="name" id="id-name">
  <p id="id-name-help-text" class="help-text">As it is in your documents.</p>
  <p id="id-name-error" class="form-error">Your name cannot be empty.</p>
</div>
```

In this case, the `.is-invalid-label` should be applied to `#id-name-input-group` div, instead of the `<label>` element.

It's very useful for backend generated fields, like Django, and helps us to better customize error rendering.

For compatibility, we use the `.input-group` class, already used by Foundation.

BTW, this is the recommended way to mark a field with error by both [U.S. Web Design Standards](https://playbook.cio.gov/designstandards/form-controls/#text-inputs) and [UK's Government Service Design Manual](http://govuk-elements.herokuapp.com/errors/#highlight-errors).

What do you think? Could it be accepted into Abide?

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -159,7 +159,11 @@
     }
   };
   /**
-   * Checks whether or not a form element has the required attribute and if it's checked or not
+   * Get the first element in this order:
+   * 1. The closest parent ".input-group" class.
+   * 2. The <label> next to the element.
+   * 3. The closest parent <label>
+   *
    * @param {Object} element - jQuery object to check for required attribute
    * @returns {Boolean} Boolean value depends on whether or not attribute is checked or empty
    */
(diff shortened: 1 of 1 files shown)


zurbchris:
We shouldn't be targeting classes directly like that if they can be user-configurable. I think this is a useful feature, and if you want to rewrite it to be more inline with how we use class selectors, I'd be happy to take a look at a pr submitted to the `abide` branch before the 6.1 release.

viniciusban:
Nice. I see you made a big refactor on Abide code. :-)

I'll work on this pr tomorrow morning.

Thank you for your attention.

PR #7526, opened 2015-12-14
[Abide] Closes #7304 - fix validation of required textarea

### LINKED ISSUE #7304: Abide required textarea always invalid

viniciusban:
Closes #7304, where required <textarea> was wrongly reported as empty.

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -151,7 +151,7 @@
         }
         break;
       default:
-        if ($el.attr('required') && (!$el.val() || !$el.val().length || $el.is(':empty'))) {
+        if ($el.attr('required') && (!$el.val() || !$el.val().length)) {
           return false;
         } else {
           return true;


zurbchris:
Would you resubmit this to the `abide` branch

viniciusban:
I've just tried, but Github says:

> zurb:abide is up to date with all commits from viniciusban:abide-fix-validation-of-textarea.

viniciusban:
BTW I saw you merged into `abide` branch from my repo.

zurbchris:
Shouldn't be necessary with #7590. Submit pr's directly on the `abide` branch. And yes, I pulled yours into abide, so you should get credit for the pr there.

viniciusban:
Relax.

I'm not worried about credit. ;-)

PR #7529, opened 2015-12-14
Adding JS label to those corresponding components

jamigibbs:
In the previous docs (v5), there was a `JS` label next to each component in the navigation. This was a handy reference that I'd like to see in the v6 docs too.


zurbrandon:
Thanks!

Issue #7502, opened 2015-12-11
Top Bar stack breakpoint


andycochran:
If your inline menu items list is too wide to fit on medium screens, you may want to stack it on both medium and small screens. Currently… 

> A top bar (.top-bar) can have two sections: a left-hand section (.top-bar-left) and a right-hand section (.top-bar-right). On small screens, these sections stack on top of each other.

I don't see a way to set the breakpoint at which the Top Bar stacks. Could we create a variable for this?

andycochran:
I'm trying to do this: 

```
<div class="top-bar">
	<button id="mainMenuToggler" class="hide-for-large float-right" type="button" data-toggle="mainMenu"><span class="show-for-sr">Menu</span></button>
	<div class="top-bar-left">
		<ul class="menu">
			<li><a href="#">Site Name</a></li>
		</ul>
	</div>
	<div id="mainMenu" class="top-bar-right show-for-large" data-toggler=".show-for-large">

		...

	</div>
</div>
```

Maybe I'm going about it wrong? Is there a better F6 way?

TvrtkoM:
maybe this could help. added classes 'stacked', 'stacked-for-small' and 'stacked-for-medium'.

top bar doesn't stack sections default with this and 'stacked' class should be applied in order to work

andycochran:
Relatedly, #7566 will allow you to stack your menus at the same breakpoint at which your top bar stacks.

gakimball:
Added a `$topbar-unstack-breakpoint` variable to make the shift from mobile to desktop mode configurable.

PR #7541, opened 2015-12-15
#7502 - topbar - enable stacking for medium screens

TvrtkoM:
Issue #7502 
- added stack breakpoints for medium and small screens
- 'stacked' class must be applied to 'top-bar' for stacking to work.

### DIFF
--- a/scss/components/_top-bar.scss
+++ b/scss/components/_top-bar.scss
@@ -48,6 +48,7 @@ $topbar-input-width: 200px !default;
 
 /// makes sections stacked
 @mixin top-bar-stacked() {
+  // Sub-sections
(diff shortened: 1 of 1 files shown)


andycochran:
Shouldn't this include every breakpoint by do something like...
```
  .stacked-for-#{$size} {
        ...
      }
    }
}
```

zurbrandon:
Hey @TvrtkoM, thanks for the PR @andycochran is correct that we'd need to use that Sass snippet to make sure dynamically created breakpoints are accounted for. Are you able to make that chnage, so we can test it?

TvrtkoM:
Yes sure, working on it...

PR #7546, opened 2015-12-15
Remove Border in Full Screen

Ne-Ne:
Not needed in full screen. CSS targets fullscreen only.

Test case - black background, border is visible.


zurbrandon:
Makes sense. Thanks!

PR #7566, opened 2015-12-15
Revise Button Group stacking to target particular breakpoints

andycochran:
I've revised Button Group so that you can stack it on particular screen sizes. The `.stacked-for-medium` class will stack a button group on medium screens or smaller. Using the `.stacked` class in combination with `.unstacked-for-large`, the menu is stacked for small and medium screen, but unstacked for large and larger.

### DIFF
--- a/docs/pages/button-group.md
+++ b/docs/pages/button-group.md
@@ -76,10 +76,12 @@ Add the `.expanded` class to the container to make a full-width button group. Ea
 
 ## Stacking
 
-A button group can be made vertical with the `.stacked` class. Or, you can use `.stacked-for-small` to only stack a button group on small screens.
+A button group can be made vertical with the `.stacked` class.
+
+You can also stack a button group on particular screen sizes. The `.stacked-for-medium` class will stack a button group on medium screens or smaller. Or use the `.stacked` class in combination with `.unstacked-for-large` to unstack the button group on particular screen sizes. In the example below, the menu is stacked for small and medium screen, but unstacked for large and larger.
 
 ```html_example
-<div class="stacked-for-small button-group">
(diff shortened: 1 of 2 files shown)


zurbrandon:
So is this going against the mobile first classes of the framework?

andycochran:
Hmm, good point. 

Using `.stacked` and `.unstacked-for-[breakpoint]` is totally a mobile first approach. 

But I see your point that `.stacked-for-[breakpoint]` seems less mobile first. Could just be semantics tho—you could think of it as "stacked for all screens up to [breakpoint]" and rename it appropriately. But it really serves the same function as `.stacked` and `.unstacked-for-[breakpoint]`. 

I try reworking this so that `.stacked-for-[breakpoint]` only targets that particular breakpoint (not also those which are smaller). I'll try: 

* `.stacked` – stacks at all screen sizes
* `.unstacked-for-[breakpoint]` – used w/ `.stacked`, unstacks at a particular breakpoint
* `.stacked-for-[breakpoint]` – stacks at a particular breakpoint

Would this be better, @zurbrandon?

gakimball:
After discussing this internally, we decided we aren't interested in having responsive versions of these button group features. However, we were able to fix the issue where `.expanded` and `.stacked-for-small` conflict with each other.

We will at least add a `.stacked-for-medium` class to give button groups a little more flexibility, however.

marcusconnor:
> However, we were able to fix the issue where `.expanded` and `.stacked-for-small` conflict with each other.

Was this fixed, and then broke again? These 2 classes don't work together. Would someone please share the fix for this?

rafibomb:
PR open #10703

Issue #7555, opened 2015-12-15
Dropdown menu - alignment issue when not enough room


tamcy:
Hello,

Consider this dropdown menu example (items copied from the foundation website):
http://codepen.io/anon/pen/jWWBvO

Normally, subitems are left aligned with the parent item. However, when there is not enough room for the subitems to appear (try shrink the window), the box is placed so that its right border touches to left border of the parent item, which gives a somewhat weird result:

![dropdown-menu-issue](https://cloud.githubusercontent.com/assets/959433/11818005/cc5e4c20-a392-11e5-9fe5-6e2bc170e5f5.png)

I suppose the expected behavior should be right aligned, as indicated in the screenshot?

designerno1:
#7251

gakimball:
These issues were resolved in Foundation 6.2. If I update your CodePen to use the 6.2.0 CSS and JavaScript, the flipped menus align properly.

Issue #7251, opened 2015-11-27
Dropdown Menu: Last Item Dropdown Positioning Bug


JeremyEnglert:
If the last item in a menu is a dropdown, the dropdown isn't positioned quite right. I'm assuming its shifting left to avoid the edge of the browser, but it's shifting pretty far left.

The dropdown below is the the "Page With Posts" menu item.

![screen shot 2015-11-26 at 8 52 41 pm](https://cloud.githubusercontent.com/assets/6110968/11433797/de7bbc48-947f-11e5-8696-b003c97fd408.png)

zurbchris:
@designerno1 @JeremyEnglert 
damn... caused by the border, I suspect. Edge detection just tells the JS if something is clear, not by how much or in what direction. Possibly solvable, albeit temporarily, with an equivalent margin-right on the last link in the dropdown.
That aside, lemme think a for a couple minutes on this one.

apcrane:
I am also seeing this issue on the dropdown menu and I am using the latest version (6.0.5). I was just wondering if there was a solution for this yet? 

![dropdown-issue](https://cloud.githubusercontent.com/assets/5689025/11714112/ae5ca01c-9f2f-11e5-81ee-7a568b12c316.PNG)

apcrane:
Found a fix that seems to work for me, although haven't fully tested this yet. But I just changed right:100% to right: 0 via an overrides.scss file:

```
.dropdown.menu .is-dropdown-submenu-parent.opens-left .submenu {
  right: 0;
}
```

designerno1:
+1
looks good for me
_dropdown-menu.scss
change
```scss
.is-dropdown-submenu-parent.opens-left .submenu {
      left: auto;
      right: 100%;
    }
```
to
```scss
.is-dropdown-submenu-parent.opens-left .submenu {
      left: auto;
      right: 0;
    }
```
solve the problem for me too
don't checked if this effects for example align-right menus

nska02:
@apcrane @designerno1 fyi, your fix makes nested dropdown menus overlap each other

<img width="231" alt="screen shot 2015-12-16 at 14 54 19" src="https://cloud.githubusercontent.com/assets/11373782/11842614/f20c187c-a404-11e5-82a0-e66763c85593.png">

designerno1:
you are right thinking about it

designerno1:
quick and dirty fix
```css
.is-dropdown-submenu-parent.opens-left .submenu {
    left: auto;
[...]

PR #7567, opened 2015-12-15
Issue #7555 - align submenu of dd menu below parent if no room to right

TvrtkoM:
Issue #7555 Changing the right css property to 0 will position submenu below parent menu item.


designerno1:
#7251

PR #7568, opened 2015-12-16
Updated jsDelivr links to 6.0.6

intrepion:


gakimball:
Good catch, thanks! Our deploy process bumps those version numbers manually, but that file was missing from the list. So I added it: 22e48f8d73ad40fd38f6261d9437e510d2bf8994

Issue #7557, opened 2015-12-15
$button-fill behavior


gakimball:
When `$button-fill` is set to `hollow`, we get these visual bugs:

![image](https://cloud.githubusercontent.com/assets/1142001/11818140/93a8cb92-a30d-11e5-8df0-0fa3db13c20d.png)

![image](https://cloud.githubusercontent.com/assets/1142001/11818146/9bdbd908-a30d-11e5-8d6d-d9b012254190.png)

gakimball:
Fixed via #7571

PR #7571, opened 2015-12-16
Issue #7557 Apply correct styling to hollow button group

NeonArray:
Added logic to test for a hollow property within the button-fill variable. If it is present, applies the correct hollow styling to the buttons, even if there is a wrapping color class such as secondary or primary. Also fixes the transparent border being applied to non-stacked and stacked button groups.


andycochran:
:+1:

gakimball:
Awesome, thanks so much!

Issue #7468, opened 2015-12-09
[Top Bar] Parent of nested sub-menu instantly navigates to its link when clicked and hover is disabled.


ZachWatkins:
I am currently experiencing this issue in Foundation 5.5.3, importing the bower package. I believe this started in 5.2.0, as I have switched between many versions of 5.x trying to find one that works. This behavior does work as I expect it to in 5.0.2. This issue seems to be occurring because the "focus" event only triggers on "&lt;a&gt;" elements which have an "href" attribute, and "focus" resolves at the end of the "mousedown" event and therefore before the "click" event. The preventDefault behavior is in a click event on line 216 of foundation.topbar.js, and the focus event which limits preventDefault is on line 296.

How can we reproduce this bug?
1. Add an “href” attribute to the “&lt;a&gt;” element arranged before a dropdown “&lt;ul&gt;” element.
2. Click the “&lt;a&gt;” element.

What did you expect to happen?
After step two, the “hover” class would be applied to the parent of the “&lt;a&gt;” element. Clicking the “&lt;a&gt;” element a second time consecutively, and the page would navigate to the URL in its “href” attribute. 
In other words, how the first level of Top Bar's menu items behave when they have sub-menus.

What happened instead?
After step two, the page navigated to the URL in the “&lt;a&gt;” element’s “href” attribute.

Demonstration: http://codepen.io/zw/pen/pjmbEd?editors=100

Documentation from other users experiencing the same issue: http://foundation.zurb.com/forum/posts/26150-top-bar---problem-with-dropdown-when-clickable

ZachWatkins:
I have a solution, but I do not know how to create this repository at version 5.5.4 locally. I cloned the repository but it is at the latest subversion of 6.

ZachWatkins:
Please ignore the first two times I references this issue. I was having difficulty with revising the commit and decided to refork the repository instead.

ZachWatkins:
I have submitted a pull request to address this issue: https://github.com/zurb/foundation-sites/pull/7572

PR #7572, opened 2015-12-16
Issue #7468 - Allow submenu parent links

ZachWatkins:
This change is intended to allow top bar elements which have the "data-options" attribute "is_hover: false" to have linked submenu parents by adding the "last-clicked" class to submenu parents the first time they are clicked and removing it from any others within that top bar element.

### DIFF
--- a/js/foundation/foundation.topbar.js
+++ b/js/foundation/foundation.topbar.js
@@ -255,7 +255,9 @@
       });
 
       S('body').off('.topbar').on('click.fndtn.topbar', function (e) {
-        var parent = S(e.target).closest('li').closest('li.hover');
+        var parent = S(e.target).closest('li').closest('li.hover'),
+            topbar = S(e.target).closest('[' + self.attr_name() + ']'),
+            settings = topbar.data(self.attr_name(true) + '-init');
 
         if (parent.length > 0) {
           return;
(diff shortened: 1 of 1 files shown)


kball:
@coreysyms I know you've been doing work in Foundation 5, can you take a look at this, test it, and merge if it seems reasonable?

coreysyms:
I'll take a look tomorrow

js/foundation/foundation.topbar.js
```diff
@@ -258,6 +262,10 @@
         }
 
         S('[' + self.attr_name() + '] li.hover').removeClass('hover');
+        
+        if(!settings.is_hover){
```
coreysyms: settings.is_hover returns undefined, please add in settings to the scope, IE line 258 should now be 
```
var parent = S(e.target).closest('li').closest('li.hover'),
topbar = S(e.target).closest('[' + self.attr_name() + ']'),
settings = topbar.data(self.attr_name(true) + '-init');
```

coreysyms:
This looks like it will work once @ZachWatkins modifies his code to my note, settings.is_hover was undefined in the scope of the click function. One a more broad note, this was a browser specific issue, Safari retuned behavior as it should be, however chrome did not, the user's code did work across all browsers though and should be merged after his revision.

ZachWatkins:
Sorry, I overlooked that originally and fixed it without pushing back to the branch. Thanks for taking this up!

coreysyms:
looks good, merged, thanks!

PR #7573, opened 2015-12-16
Fixed util import in settings/_settings.scss.

ruction:


gakimball:
The path on that `@import` statement is correct if you're adding Foundation as an import path in your Sass compiler. So if you add, for example, `bower_components/foundation-sites/scss` as an import path, the settings file will be able to find `util/util` no problem.

PR #7575, opened 2015-12-16
Add button expanded class for each breakpoint

### LINKED ISSUE #7570: [Button Proposal] add support for small-expanded...
it would be nice to have classes for expanded on small medium large etc

andycochran:
Closes #7570.

### DIFF
--- a/docs/pages/button.md
+++ b/docs/pages/button.md
@@ -39,7 +39,7 @@ Additional classes can be added to your button to change its size and shape.
 <a class="button" href="#">So Basic</a>
 <a class="large button" href="#">So Large</a>
 <a class="expanded button" href="#">Such Expand</a>
-<a class="small expanded button" href="#">Wow, Small Expand</a>
+<a class="small expanded-for-large button" href="#">Wow, Small + Expand for Large</a>
 ```
 
 ---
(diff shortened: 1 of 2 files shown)


gakimball:
Similar to [my comment here](https://github.com/zurb/foundation-sites/pull/7566#issuecomment-173040940), we aren't interested in having responsive versions of these button group features.

DaSchTour:
Came across this while fighting with responsive buttons. Now I'll override `button-expand` mixin to prevent expanded buttons from my mobile template to create gigantic buttons on desktop. Why exactly this feature was rejected? I've seen this as a common pattern to have full width buttons for mobile and "normal" buttons for desktop.

andycochran:
If it's to prevent bloat, perhaps the way forward (v7?) with features like this is settings vars to enable them? 

```scss
$responsive-expanded-button: true;
```
Maybe this could be resolved in the "Full menu rework" ticket #9421.

DaSchTour:
I would like to just have the buttons there. I guess a lot of other features were added without such a configuration that add a lot of more "bloat". So maybe we could reconsider just this PR?

IamManchanda:
I think this really makes sense specially as andy mentioned with 

```scss
$responsive-expanded-button: true; // maybe `false` by default
```

Reopening this and cc / @kball

IamManchanda:
Also, 
Code Brevity => `expanded-large` is enough!

DaSchTour:
Closing in favor of #10653

PR #7579, opened 2015-12-16
Keyboard

### LINKED ISSUE #7536: [Keyboard access] uglify version "Component not defined!"
if i uglify the output with gulp-uglify
Keyboard access not works error in Console:
Component not defined!
```js
var uglify = require('gulp-uglify');
gulp.task('javascript:foundation', function() {
  return gulp.src(FOUNDATION)
    .pipe(concat('foundation.js')).pipe(uglify())
    .pipe(gulp.dest('_build/assets/js'));
});
```
http://localhost:3000/dropdown-menu.html
press tab several times and then the info in the console appears

in normal version it seams to work fine but not in uglifyed version

Owlbertz:
Replaces the parameter for the Keyboard Util from Component Object to String (Component name).
Also adjusts the docs accordingly.
Fixes #7536.


zurbchris:
Thanks, stupid mangling. Almost need something like ng-annotate for our build process...

Issue #7500, opened 2015-12-11
Media Queries and base font size


n-dawson:
Changing the base font size affects the media queries in an unexpected way.

The medium breakpoint for example is defined as 640px which at the default base is 40em. When the base is changed to 14px the media query will update to 45.714...em. 

ems are relative to the font size defined in the browser settings when used in media queries and not the HTML element which means the medium breakpoint comes into effect at 732px instead of 640px.

I believe the solution would be to calculate the queries in em without taking into account the setting used for global font size.

gakimball:
Yes, we've been talking about this internally. We need to change how breakpoints are printed so that the "pixel value" of the breakpoint is always the same.

So the em value for medium should always equal 640px, regardless of the page's base font size. This will involve making some adjustments to the `-zf-bp-to-em` function, which does the em calculation for breakpoints.

jensengelmann:
I described the problem here: https://github.com/zurb/foundation-sites/issues/7021#issuecomment-163913431

In my eyes a simple fix would be changing https://github.com/zurb/foundation-sites/blob/develop/scss/util/_unit.scss#L54 to ```$value: rem-calc($value,16px);```.

Or you assign a new variable instead of hardcoding the ```16px```...

gakimball:
Going to close this in favor in #7021, which has a more detailed discussion of the issue.

Issue #7021, opened 2015-11-19
Changing base font size DOES affect grid breakpoints


designerdean:
I added this issue to the foundation-sites-6 repository but then the official version 6 was released. I tested this bug with the latest release and it is still a problem.

The documentation's indicates that, "if you want your base font size to be different and not have it affect the grid breakpoints, set $rem-base to $global-font-size and make sure $global-font-size is a pixel value" (http://foundation.zurb.com/sites-6-docs/global-styles.html#font-sizing). I've done that but it actually does affect the grid sizing.

As you can see from the attached screenshots, I have set $global-font-size to 22px; and $rem-base to $global-font-size. In the attached browser screenshot, you can see that the "small" breakpoint is being applied before hitting the 550px point (the point I defined in my breakpoint variables, also attached).

![1](https://cloud.githubusercontent.com/assets/729893/11281373/8e325922-8eb7-11e5-8d15-e170840193ae.png)
![2](https://cloud.githubusercontent.com/assets/729893/11281374/8e424d8c-8eb7-11e5-8c8e-b75a90227387.jpg)
![3](https://cloud.githubusercontent.com/assets/729893/11281375/8e49193c-8eb7-11e5-9eb8-de4cafa3afc9.png)

erredeco:
If I remenber, In version 5 I solved this kind of problems in this way:
I defined a function that converts px values to %:
```
@function convert-to-percent($value, $base-value: $rem-base){
  $value: strip-unit($value) / strip-unit($base-value) * 100%;
    @if ($value == 0%) { $value: 0; }
    @return $value;
}
// $rem-base: 16px
```
then on _global.scss:

```
html{
  font-size:100%
}
body{
  font-size: convert-to-percent($base-font-size);
}
```
I will try if the same approach works for FF6

designerdean:
Foundation team: Any update on this?

@erredeco's suggestion made me realize that changing $global-font-size to a bigger or smaller **percentage** kind of works without messing up the media query breakpoints. However, since both `<html>` and `<body>' are given the same value, it takes a percentage of the percentage, making it bigger or smaller than intended.

As I mentioned when opening this issue, the Foundation documentation gives different instructions on setting the global font size.

What is the correct Foundation way to change the base font size?

kesy85:
Same problem here, any news? It's a basic function, please fix it

jensengelmann:
For my understanding it's just a bug in the breakpoint calculation.

[...]

PR #7580, opened 2015-12-16
Global font size and rem/em calculation

gakimball:
This pull request addresses #7021, #7500, and others. With these changes, the base font size can be shifted up or down without affecting any `rem` or `em` based lengths, such as the `.row` width, or the widths of breakpoints.

Here are the changes:
- Removed `$rem-base`.
- `rem-calc()` references `$base-font-size` for rem calculation.
- `-zf-bp-to-em()` always uses a 16px base when calculating breakpoint widths.

Still need to figure out what migration process would be necessary.

### DIFF
--- a/docs/assets/scss/docs.scss
+++ b/docs/assets/scss/docs.scss
@@ -7,6 +7,7 @@
 //
 // Feel free to submit pull requests to foundation-docs like you would Foundation itself!
 
+@import 'settings/settings';
 @import 'foundation';
 @import 'motion-ui';
 
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -118,8 +118,7 @@ $foundation-colors: (
     font-family: '#{-zf-bp-serialize($breakpoints)}';
   }
 
-  html,
-  body {
+  html {
     font-size: $global-font-size;
     box-sizing: border-box;
   }
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -6,13 +6,13 @@
 /// @group global
 ////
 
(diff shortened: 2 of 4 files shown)


tjhiggins:
+1 Not sure about fix, but having the same problem with breakpoints being incorrect.

jensengelmann:
I don't understand the following lines (https://github.com/zurb/foundation-sites/commit/3391c30fd72ffd69756f182f21904353c6d0b81e#diff-53036f9871421f19686387c09cc9e13cR37):

```scss
$base: $global-font-size;

@if unit($base) == '%' {
  $base: 16px;
}
```

Maybe I'm wrong, but you're checking if ```$global-font-size``` ist a percentage value and then set the base to ```16px```. What if I'm setting ```$global-font-size: 75%``` - the base for them rem-calculation should then be ```12px```.

In my opinion you have to convert the percentage value to the corresponding pixel values... someting like:

```scss
// base is 75%
@if unit($base) == '%' {
  $base:($base/100%) * 16px;
  //now base is 12px (.75 * 16)
}
````

Not tested - don't know if the syntax is right.

gakimball:
@jensengelmann Good catch, I implemented that change: 4b718ad8da5552fb5abb1c051087d9748771b185

Last thing to do is figure out how backwards-compatible this change is.

gakimball:
Alright, this change is totally backwards-compatible. All you'll need to do when you upgrade is move the `$global-font-size` variable above any references to `rem-calc()`, so the correct value is used.

Issue #7196, opened 2015-11-25
Abide Validation Radio Field required


designerno1:
i got a form
```html
<form id="create_account" action="/create_account.php" method="post" data-abide novalidate>
  <div class="row" data-equalizer>
    <fieldset data-equalizer-watch>
      <legend>Some heading</legend>
      <input required type="radio" name="gender" value="m" id="93b5777073645a556be5027b4f5a9b08" />
      <label for="93b5777073645a556be5027b4f5a9b08">yes</label>
      <input required type="radio" name="gender" value="f" id="48b31e1ebc91b3271f4f267ffb83ab31" />
      <label for="48b31e1ebc91b3271f4f267ffb83ab31">no</label>
      *
    </fieldset>
  </div>
  <button class="button" type="submit" value="Submit">Submit</button>
</form>
```
the gender gets marked as error even if i check one radio

if i remove the "required" it doesn't get the error but doesn't even get's submitted

zurbchris:
This one is tricky. I think for this use case the `required` attribute should go on the field set, and one of the elements within that fieldset should pass validation. Thoughts? Note, the feature I just described is not included in #7590 at the moment.

designerno1:
i thought there was no problem with radio in my solution of abide #7470

designerno1:
pattern on textarea does not work?

FlorianBoe:
To open up a new discussion on this issue I looked into the specification.
W3C and WHATWG are quite specific on the use of required attributes on radio inputs:
> To avoid confusion as to whether a radio button group is required or not, authors are encouraged to specify the attribute on all the radio buttons in a group. Indeed, in general, authors are encouraged to avoid having radio button groups that do not have any initially checked controls in the first place, as this is a state that the user cannot return to, and is therefore generally considered a poor user interface.
> - https://www.w3.org/TR/html/forms.html#the-required-attribute

Handling this use case should be part of the solution.

gakimball:
Fixed via cc2f2c668bb4f795cfad130915baf1b5f45fd79b. I wrote a test case [here](https://github.com/zurb/foundation-sites/blob/cc2f2c668bb4f795cfad130915baf1b5f45fd79b/test/visual/abide/abide-radio.html) with two Abide forms: one with a required radio group, and one with an optional radio group, showing both working as expected.

Regarding the W3C spec, to be nice to developers, a radio group is considered "required" if *any* of the `<input>`s in the group have the `required` attribute.

[...]

PR #7590, opened 2015-12-17
Abide that works. I hope.

zurbchris:
@viniciusban @designerno1 @briantan888 @milose @jamilservicos 
Would you folks take a look, see what you like/don't like. Submit any proposed changes directly to the `abide` branch. This will be merged in before the 6.1 release.


designerno1:
i take a look later that day, and run it through my testcase

viniciusban:
I'll test later, today.

semikolon:
I'm going to try this today or tomorrow.

designerno1:
the type of an input like email isn't checked against the options.patterns

designerno1:
radio buttons does not work for me see
```html
<fieldset class="large-6 columns">
      <legend>Choose Your Favorite</legend>
      <input type="radio" name="pockets" value="Red" id="pocketsRed"><label for="pocketsRed">Red</label>
      <input type="radio" name="pockets" value="Blue" id="pocketsBlue"><label for="pocketsBlue">Blue</label>
      <input type="radio" name="pockets" value="Yellow" id="pocketsYellow"><label for="pocketsYellow">Yellow</label>
    </fieldset>
```

zurbchris:
@designerno1 I commented elsewhere (#7196) on the radio buttons. 
For the type checking, oversight, simple fix.

designerno1:
ok and question before?

jamilservicos:
input email:
<input type="text" pattern="email" placeholder="Email" id="Email"
aria-describedby="Email" required>

Att.

Gervásio Júnior, (Jamil).

2015-12-17 18:25 GMT-03:00 designerno1 <notifications@github.com>:

semikolon:
I ported this branch to a fork of the foundation-rails gem. Available here for testing: https://github.com/semikolon/foundation-rails

designerno1:
pattern on textarea does not work?

designerno1:
for me radio does not work or whats the markup?

zurbchris:
The same as it was?

designerno1:
ok it doesn't work ;-)

zurbchris:
A little more specific please?

briantan888:
Created pr #7628 

Required attribute for input radio does not make much sense. Proposed .radio-group wrapper container for radio require check. Function validateRadio is no longer needed.

    <label for="pokemonRadio">Pick one pokemon</label>
    <div id="pokemonRadio" class="radio-group" required>
      <input type="radio" name="pokemon" value="red">Red
      <input type="radio" name="pokemon" value="blue">Blue
      <input type="radio" name="pokemon" value="yellow">Yellow
    </div>
    <div class="form-error">Please pick one pokemon</div>

Similarly for checkbox, (though retain required check for individual checkbox input, probably not much use too ).
  
    <label for="pokemonCheckbox">Pick at least one pokemon</label>
    <div id="pokemonCheckbox" class="checkbox-group" required>
      <input type="checkbox" name="pokemon" value="red">Red
      <input type="checkbox" name="pokemon" value="blue">Blue
      <input type="checkbox" name="pokemon" value="yellow">Yellow
    </div>
    <div class="form-error">Please select at lease one pokemon</div>

briantan888:
unable to bind 'valid.zf.abide' and 'invalid.zf.abide' for form. Is this the right way? only submit is triggered

    $("#myform").on('submit', function() {
      console.log("submit triggered");
      return false;
    });
    $("#myform").on('invalid.zf.abide', function() {
      console.log("invalid.zf.abide triggered");
    });
    $("#myform").on('valid.zf.abide', function() {
      console.log("valid.zf.abide triggered");
      // ajax post.
    });

zurbchris:
Add form in there for full form events `on('formvalid.zf.abide', …`

designerno1:
But we can check radio group without .radio-group with the given name attribute?

briantan888:
But where do you specify the required attribute?

briantan888:
on('formvalid.zf.abide',...) does not work either. I see from the js code that 'valid.zf.abide' is emitted (line 321)

designerno1:
The required attribute could be specified in one radio button. I think there is no way in radios with same name to be 2 checked at same time. So  we can specify it in one radio of the group or in all

briantan888:
Tried it. Didn't work. If I put "required" on one of the option, then "that" option must be checked. If I put required on all options, then validation will not get through. That is why I said "required" on radio input does not make sense.

          <label for="which_one_should_I_link_to">Monster Radio</label>
          <input id="monster1" name="monster" type="radio" value="red" required>
                <label for="monster1">Red</label>
          <div class="form-error">Must choose an option</div>
          <input id="monster2" name="monster" type="radio" value="blue" >
              <label for="monster2">Blue</label>
          <div class="form-error">Must choose an option</div>
          <input id="monster3" name="monster" type="radio" value="yellow" >
              <label for="monster3">Yellow</label>
          <div class="form-error">Must choose an option</div>

designerno1:
i give it a try tomorrow using a different checkRequired and Validation function for Radio groups.
My plan is to check if a radio enters the function the whole group for a required attribute, and if required for a check.
maybe caching the groups on init i think because if we got a radio group with 10 input radios and the submit button is clicked the form validates all inputs and so the script has to run 10x10 times through the radio group required and checked function.
because of that i think of caching the groups on init in a array
thoughts?

milose:
Using require on a single check-box makes sense i.e.: I agree.
But for individual radios - not at all.

mediabeastnz:
Hi, i'm using 6.0.5 for a project and obviously there is this Abide issue. Is there a quick patch or download somewhere?

briantan888:
@designerno1 validateInput for radio/checkbox are still needed for field-level validation, but validateForm can skip radio and checkbox elements, but loop radio-group and checkbox-group classes, but unnecessarily complicate the codes. IMO, why use 10 radio inputs when select-single is probably more effective and takes up less real-estate. :)

zurbchris:
Ok, sorry for leaving you guys to the discussion without me. I've reverted the work on radio groups to what it was last week, because the pr's I had were not doing what we wanted them to do. However, the pattern matching is good and we'll keep that in there. As for the direction of the radio/checkbox issue, here's what a group should do:

 Check the overall group for a required attribute. If the group is named (and it should be), the parent of the named group would contain the required attribute. This should be a fieldset, but we shouldn't limit it to that. If the group is required, obviously only one of them *must* be checked, so on validation, make sure at least 1 input is :checked (could be more for a checkbox group). 
However, if a developer wants to require a *specific* radio/checkbox/whatever, placing the required attribute on that element would force it to ensure that it is in fact, :checked. 

I realize that for radios, more than one required statement would break the form, and that's ok, we're not quality control for other peoples markup.

I won't be able to get to this before ZURB goes on winter break, but it will be on my list of things for the new year. If someone can implement this feature over the break, I would love to see it and get it pulled in the next patch version. That being the case, I'll file an issue about this shortly. Thanks everyone for the input, feedback, and pull requests. Please, keep them coming.

designerno1:
ok regrettable that it was only nearly what you are looking for

briantan888:
The required attribute for checkbox grouping may be a problem when the checkboxes are embedded in switches. The proposed checkbox-group class eliminates that ambiguity. It also make group error highlighting straight forward.

simar88:
Hi everyone, I'm using 6.1 and i'm unable to get  "formvalid.zf.abide" and "forminvalid.zf.abide" event.
Any fix at the moment?
Thanks in advance..

briantan888:
I submitted a pr #7674 . You can apply this patch on validateForm to your foundation.js

simar88:
Hi @briantan888 i've found that these two event are triggered only if is present the code:
<div data-abide-error class="alert callout" style="display: none;">
    <p><i class="fi-alert"></i> There are some errors in your form.</p>
  </div>

briantan888:
Bug. Fixed in next release.

eliotc1986:
What's the status of this radio button/radio button group abide issue? 
Using 6.1.1 and still can only validate a single radio button (multiple radio buttons with 'required' fail validation/ applying "required" to fieldset doesn't validate)

Issue #7596, opened 2015-12-17
Nested unordered/ordered lists


brettsmason:
I'm getting a problem with nested unordered/ordered lists, see screenshot:
![list-issue](https://cloud.githubusercontent.com/assets/73229/11873404/4eb16780-a4d2-11e5-9105-95cbc3c4dfb3.jpg)

This is taken from my WordPress testing install with a theme I'm making. I've found the issue:

```
// Nested unordered/ordered lists
  ul, ol {
    & & {
      margin-#{$global-left}: $list-nested-side-margin;
      margin-bottom: 0;
      list-style-type: inherit; <---- removing this sorts out the issue
    }
  }
```

I'll create a PR for it now.

PR #7597, opened 2015-12-17
Nested unordered/ordered lists

brettsmason:
Fixes nested unordered/ordered lists - see [7596](https://github.com/zurb/foundation-sites/issues/7596)


zurbrandon:
Thanks Brett!

PR #7603, opened 2015-12-17
Make default colors accessible

zurbrandon:
The newest attempt at making the default colors accessible in Foundation. Still a little drab, so if anyone would like to take a stab it would be appreciated.


kball:
@andycochran this PR is stale but related to #7387, which I just saw you take on.  Going to close it but feel free to use this as context for the work on that.

andycochran:
@kball I've picked back up on #7387. (See my latest note there.)

PR #7604, opened 2015-12-17
Update foundation.drilldown.js

ericjung:
If I use one of the visibility classes (http://foundation.zurb.com/sites/docs/visibility.html) to conditionally hide the drilldown, then _getMaxDims() throws an exception during initialization on line 295 when dereferencing the empy this.$menuItems array. That prevent other javascript on the page from executing. This change ensures the array is not empty in such cases, providing of course there are actually li elements in the DOM.


zurbchris:
Good catch!

PR #7605, opened 2015-12-18
Added missing variables to settings file.

russelporosky:
The variables `$rem-base` from `util/_unit.scss` and `$flexvideo-padding-top` from `components/_flex-video.scss` were missing from the `settings/_settings.scss` file.


gakimball:
Those variables are being removed in 6.1.

PR #7618, opened 2015-12-18
.column - Min-Height 1px

Ne-Ne:
Stops the elements from crashing due to float value. Helpful when setting up multiple columns during dev.

Also a good fall back while data-equalizer kicks in to match height.

### DIFF
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -61,6 +61,7 @@
 
   @include grid-column-size($columns);
   float: $global-left;
+  min-height: 1px;
   padding-left: $gutter;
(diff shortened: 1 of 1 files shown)


gakimball:
Is the use-case for this when you're setting up your HTML and all the columns are empty?

Ne-Ne:
Hi @gakimball, for me yes.

I use this a lot for prototyping at work.

gakimball:
Since it's just for prototyping, it's hard to justify.

We have considered creating a prototyping mode for the grid, which would add color overlays to rows/columns, and this property would fit well in a concept like that.

PR #7624, opened 2015-12-18
adds support for own RegEx Patterns through pattern attribute to be h…

designerno1:
…tml5 conform. F5 had this feature also

http://www.w3schools.com/tags/att_input_pattern.asp
#7590


zurbchris:
I dig it.

SalahAdDin:
:+1:

designerno1:
there is a little bug in my PR now it validates normal type="text" and non required with the regex /text/ or /submit/ or /reset/
have to think about this..........

Issue #7587, opened 2015-12-16
[ruby-sass] breakpoint(small down) doesn't render media query


TvrtkoM:
In compiled foundation.css there's a media query for *@ include breakpoint(small down)* mixin. This can be seen at the end of the foundation.css: 

    @media screen and (max-width: 39.9375em) {
      .top-bar-right {
        width: 100%; }
      .top-bar-left {
        width: 100%; } }

But if I use *@ include foundation-everything*  same breakpoint renders following css (no media query):

    .top-bar-right {
      width: 100%; }

    .top-bar-left {
      width: 100%; }

This makes top bar stack on all screen sizes if scss is used.

TvrtkoM:
I see Foundation uses node-sass for scss translation to css. Translation is correct with node-sass but not with ruby-sass.

Seems like ruby-sass doesn't make difference between '0' and '0em' and scss under 

    @if $bp > 0 or $dir == 'only' {

... block never runs

TvrtkoM:
here's the difference between minimized dist/foundation.css and css generated by ruby-sass with change from ab1b25a

https://www.diffchecker.com/4tscduxg

Looks like no difference is related to media queries, but I'm not really sure will it break anything.
running gulp test i get this:

    Breakpoint returns an empty string for the values zero down or zero up {
      ✗: Expected "(max-width: 39.9375em)" to be "";
      ✗: Expected "(max-width: 0em)" to be "";
      Test: Failed; }

gakimball:
It looks like you're working on the fix, do you want to submit a pull request for it?

I changed the comparisons in the `breakpoint()` function to always compare `em` and `em`. It's technically okay, but the Sass developers have said eventually it won't work that way. 857a8204028536fe1a6b72f42b0ebdc95b6a0386

TvrtkoM:
Hey, I've been testing things out and doing small research.

https://github.com/CSSLint/csslint/wiki/Disallow-units-for-zero-values

> It's recommended to remove units for all zero length values because these units aren't being used by the browser and therefore can be safely removed to save bytes.

So, I wouldn't personally put units by zeroes but let breakpoint mixin parse 'small down' value - ab1b25a. which should create media query by ruby-sass as noted in my first post.
The other solution is to make top-bar not stack for small screens by default but to require 'stacked' class to be applied (26d93b6) which requires more css.

I can prepare a PR but think this should be handled by someone from Zurb cause it could require change in spec.

gakimball:
[...]

PR #7632, opened 2015-12-19
don't skip 'small down' value for breakpoint

TvrtkoM:
- sass tests
- revert to comparison with unitless 0

releated to: #7587 

With this, breakpoint(small down) will transpile to media query css, but breakpoint(0 down) will be skipped. I also took liberty to modify sass tests.


pwnall:
+1

PR #7633, opened 2015-12-19
fixes radio requiredCheck

designerno1:
as required attribute is on one radio element, group [name] is matched and error applied/removed to group [name]
#7590

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -113,6 +113,7 @@
   Abide.prototype._init = function(){
     this.$inputs = this.$element.find('input, textarea, select').not('[data-abide-ignore]');
 
+    this.radioRequiredGroups();
     this._events();
   };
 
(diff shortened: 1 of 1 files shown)


js/foundation.abide.js
```diff
@@ -261,6 +262,7 @@
 
       case 'radio':
         validated = this.validateRadio($el.attr('name'));
+        clearRequire = validated;
```
designerno1: //because clearRequire checks only el.required and el.checked and not group

js/foundation.abide.js
```diff
@@ -343,18 +345,23 @@
...
+      labels.push(self.findLabel($(this)));
+      if(this.checked){
+        counter = true;
+      }
     });
-
-    return counter.indexOf(false) === -1;
+    if (counter) {
+      for(var i = 0, len = labels.length; i < len; i++){
+        self.removeErrorClasses(labels[i]);
```
designerno1: //the closest in findLabel is my friend, error added because of new group check

Issue #7219, opened 2015-11-26
!default missing from $foundation-colors prevents overridding.


umutahmet:
Can !default be added to the $foundation-colors list so it can be overridden? 
Thanks

dherran:
This is an odd one. The colours you define in _settings.scss wont be taken into account when generating the $foundation-colors map.

It seems like the map is being rendered in SASS with the default values, even though the settings file is being imported *before* the foundation-global-styles include.

The only solution I found after many tries is to have the !default flag in $foundation-colors and redeclare the map inside the settings file. I wonder if there is a way for the $foundation-colors map to grab the colour values already preset in the settings file.

umutahmet:
Yeah, that's exactly how I've got it working. The only other thing I can think of is to create a mixin which can help extend / merge the two maps together, but this seems overkill. "$button-sizes" mapping is another example, except this time it's set up with the !default flag allowing overriding..

gakimball:
`$foundation-colors` is considered to be an internal variable—we use it in `@each` loops to generate coloring classes for buttons, callouts, and so on. It has no `!default` flag because it's intentionally not in the settings file. If you want to change the default colors, use the variables themselves, like `$primary-color` and `$alert-color` and so on. Or are you saying you did this and colors still aren't changing?

dherran:
@gakimball exactly, I changed the colours, like $alert-color, $primary-color, etc and nothing happens. SASS is still taking the default values for these vars whenever $foundation-colors is used.

You can try for example the callouts. If you change the $alert-color to a different shade of red, it will be ignored and the default value in _global.scss will be rendered. You will see the callout rendered with the default shade of red instead of the new one.

andycochran:
This give me an idea. Could we instead define our own color palette quantities? Not every project will use all the buttons/badges/labels/callouts/etc generated by the various colors. Of course, `$primary-color` is required, as it's used outside of `@each` loops. But I can imagine projects that would benefit by removing the cruft other colors make, or by defining their own color palette in the settings.

For instance, if I don't need alerts, but I do need additional colors to match brand guidelines:  
```
$foundation-colors: (
  secondary: #ff19f0,
  tertiary: #9c45ff,
  success: #3adb76,
//  alert: #ec5840,
  warning: #ffae00,
);
```
Then all the `@each` loops could loop through both `$primary-color` and `$foundation-colors`.

gakimball:
[...]

PR #7637, opened 2015-12-20
Add '!default' to $foundation-colors.

pedzed:
This allows you to overwrite `$foundation-colors`.


andycochran:
Looks like this is part of the v6.1 milestone, according to #7219.

pedzed:
@andycochran Thanks, I didn't notice that an issue was created for it.

To me personally, it makes more sense to have it in `v6.0.7`, but I don't mind having it in `v6.1.0`. As long as the pull request gets accepted soon.

PR #7647, opened 2015-12-21
Responsive gutters

gakimball:
Responsive gutters will allow you to define different gutter sizes at different breakpoints. It's configured like this:

``` scss
$grid-column-responsive-gutter: (
  small: 20px,
  medium: 30px,
);
```

The default `.column` class gets new media queries for the higher-up breakpoints. The `.[size]-uncollapse` classes also take these values into account.

Work left to do:
- Make it backwards compatible with existing projects. For 6.1 it'll be an optional feature, and for 6.2 it will replace the static gutter setting entirely.
- Add it to the flex grid. That won't take long.

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -277,6 +277,32 @@ In order to work around browsers' different rounding behaviors, Foundation will 
 
 ---
 
+### Responsive Gutters
+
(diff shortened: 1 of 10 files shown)


andycochran:
Yas!

PR #7654, opened 2015-12-22
Fix: shrink wider than 100%

derweili:
If the objects inside a div with .shrink class are wider than the browser on iOS devices, the responsive layout got broken.
With a max-with of 100% this bug is getting fixed.


gakimball:
Thanks for the fix! We'll need to test this in other browsers first to make sure it doesn't break anything else.

PR #7677, opened 2015-12-23
FIx reveals with multiple anchors

thewunder:
Currently if you have more than one anchor opening the same reveal, they will all be set to the id of the first anchor.  This besides creating duplicate ids will break things.


zurbchris:
Not exactly the correct way to do this, due to aria attributes. I think Reveal needs a minor overhaul in managing accessibility features. I'll turn this one down for now, but if you resubmit that accounts for accessibility, I'd love to see it.

PR #6601, opened 2015-06-03
Adding a top/bottom functionality to zurb offcanvas with push,overlap and submenu

tomsah:
You can see the adding functionality in action on this demo page

http://tomsah.co.uk/offcanvas-demo/

Please let me know if any improvement is needed or more need to be done.

very best

PR #7683, opened 2015-12-24
Add top off-canvas

spacepants:
Foundation 6 is pretty great so far. Top/bottom off-canvas is the feature I missed the most from 5.

This (re)adds top and bottom functionality to off-canvas. I've tried to keep it as minimal as possible so that it's more in line with the new svelter codebase.

Please let me know if this needs any additional work. :heart:

### DIFF
--- a/docs/pages/off-canvas.md
+++ b/docs/pages/off-canvas.md
@@ -114,3 +114,40 @@ The main content area (`.off-canvas-content`) will be padded to the left or righ
 ```
 
 <button type="button" class="button" data-docs-example-ofc>Toggle Reveal Class</button>
+
+---
+
+## Off-canvas from Any Direction
+
+In addition to left and right off-canvas menus, you can also have top or bottom off-canvases as well.
+
+```html
+<body>
+  <div class="off-canvas-wrapper">
+    <div class="off-canvas-wrapper-inner" data-off-canvas-wrapper>
+      <div class="off-canvas position-top" id="offCanvasTop" data-off-canvas data-position="top"></div>
+      <div class="off-canvas position-bottom" id="offCanvasBottom" data-off-canvas data-position="bottom"></div>
+      <div class="off-canvas-content" data-off-canvas-content></div>
+    </div>
+  </div>
+</body>
+```
+
+<button class="button" type="button" data-toggle="offCanvasTop">Open Top Menu</button>
+<button class="button" type="button" data-toggle="offCanvasBottom">Open Bottom Menu</button>
+
+<div class="primary callout">
+  <p>When using the `title-bar`, the title-bar icons are still either `title-bar-left` or `title-bar-right`, but the off-canvas can be set to top or bottom.</p>
+</div>
+
+```html_example
+<div class="title-bar">
+  <div class="title-bar-left">
+    <button class="menu-icon" type="button" data-open="offCanvasTop"></button>
+    <span class="title-bar-title">Foundation title bar with vertical off-canvas</span>
+  </div>
+  <div class="title-bar-right">
+    <button class="menu-icon" type="button" data-open="offCanvasBottom"></button>
+  </div>
+</div>
+```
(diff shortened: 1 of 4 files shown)


zurbrandon:
This just seemed slightly odd to us, since the bottom off canvas will just scroll you to the bottom to open it up. Do you have thoughts on this?

spacepants:
Yeah, the bottom off-canvas example is kinda janky. I thought about displaying it as an iframe, but that seemed a bit much. I've only ever seen bottom off-canvas work well with the toggle at the bottom of the page in a footer or a floating element. I figured I'd throw it in while I was playing around with translateY().

Top off-canvas is the real champ, though. I'm currently building a new enterprise UI library with Foundation 6, and I really wanted to be able to use top off-canvas to help folks navigate between different applications. Here's a sneak peek: https://gfycat.com/SharpDimAmbushbug

It looks pretty nice and took so little additional work that I figured I'd submit it in case anyone else could find it useful. If you'd like, I can drop the bottom example and just mention the functionality in the docs.

rafibomb:
The top example looks great and we can see a use case for. The bottom menu seems a lot less useful and the interaction is odd. If you remove the CSS for the bottom menu we can pull this in!

spacepants:
Okay, should be good to go.

zurbrandon:
HAHA love that commit! I'll pull this down this afternoon and take a look! Thanks a ton for the help here.

andycochran:
@spacepants :laughing: Nice -m

zurbchris:
@spacepants Side note: Do you live in Austin?? I see St. Ed's in your giphy. :+1: for the message.

zurbchris:
The only thing I see that's a bit weird is the little gap between the top off-canvas and the menu bar, was that intentional?

spacepants:
@zurbchris Yep, I work at St. Ed's. Reppin' that 512.

That gap comes from the fact that I'd opted for the `.mobile-ofc` class from `foundation-docs` for styling the example. In practice you'd want to adjust the `$offcanvas-vertical-size` to match the element it houses. I'd just left it at a nice round default as it was close enough, and I thought that the gap might help provide a bit of element distinction.

For a cleaner example, it might be easier to add a separate class to the `foundation-docs` for this in the same way that `foundation-sites` has the `.display` class in the styleguide, but that's just what I can see from digging through the projects for a little bit — you guys may very well have a better idea.

zurbchris:
Awesome, I lived in the '04 for a while during my 7 year tenure in Austin. 

As for the pr at hand, I'm going to pass this off to @gakimball since he did the sass for off-canvas to begin with. Solid work from what I can tell though!

jacobalvarez:
Whatever happened to #6601?  Didn't that provide the same functionality?

I think the bottom overlap function from that demo would be useful, since it doesn't scroll you down the page and provides another option for hiding menus/other stuff.

adamj88:
I just wanted to question the decision to remove the CSS for the bottom off canvas menu in 4bfed53.

I have a use case for it where I have a fixed icon bar at the bottom of the screen (which, to me, seems to be a common pattern for mobile) and then use the bottom off canvas to display content depending on the triggered icon. Admittedly this would be better/more useful if there were overlay classes like those in v5.

I know we're trying to keep the CSS to a minimum but I think it would be a bit unusual to add an off canvas for all  edges of the screen except one at the cost of ~10 lines of CSS.

Any further thoughts?

CC: @spacepants @zurbrandon @rafibomb

kball:
This looks like it got dropped.  @andycochran @zurbrandon from your perspective is this ready to go if we redirect it to v6.3.0?

andycochran:
Honestly, I haven't looked closely. I just loved the commit message. Looking now…

andycochran:
Works for me. Although, I don't dig the way it scrolls the page to the top. But that's how left/right off-canvas works too. So… yeah.

kball:
Ok, sounds like this is a go for 6.3.  However, this PR is against develop, which is only for bugfixes, so I'm going to close it. @spacepants can you resubmit against the v6.3 branch please?

PR #7689, opened 2015-12-25
Abide: Change target parm on element level trigger to jQuery 

web2client:
Abide plugin: validateInput

Change param to jquery selector to standardize across plugins. DOM element already accessible via event.target


zurbchris:
Cool!

PR #7712, opened 2015-12-29
Requirements

afonsoduarte:
Declare plugin dependency for those who wish to load js files individually.

### DIFF
--- a/docs/pages/magellan.md
+++ b/docs/pages/magellan.md
@@ -5,6 +5,13 @@ js: js/foundation.magellan.js
 tags:
   - navigation
 ---
+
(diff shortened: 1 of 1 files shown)


zurbchris:
But, Magellan does not require Motion-UI. I think this is a great idea across the board to declare dependencies in a callout at the top of the JavaScript section for each plugin though.

afonsoduarte:
I'm really sorry, I'm an idiot. I meant foundation.util.motion.js, not Motion-UI. And actually it depends on foundation.util.triggers.js as well. This information should probably be under Javascript Reference. I'll submit a new pull request.

PR #7717, opened 2015-12-29
Update installation.md with manual installation instructions

joeyred:
Added instructions for manual set up of templates, taken from the readme files included in each template.

I feel that the manual set up should be included in the installation section of the documentation, due to my own experience in needing the information and having a bit of a tough time finding it in a timely manner. This way it is easily accessible without adding confusion.

### DIFF
--- a/docs/pages/installation.md
+++ b/docs/pages/installation.md
@@ -31,6 +31,52 @@ foundation new
 
 ---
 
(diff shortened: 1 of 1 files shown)


gakimball:
Thanks for adding this! Although you didn't quite spell **ZURB** right ;)

Issue #7433, opened 2015-12-07
[Dropdown] Allow for multiple anchor toggles.


tjhiggins:
v5 allowed you to have multiple anchors for the same dropdown. Simplifies the code to only allow one, but is much more efficient to reuse a dropdown with knockoutjs etc.

zurbchris:
How would you see that implemented in 6? A single dropdown pane that will attach to whichever anchor point was clicked? Then which one does it anchor to for a programmatic open? It also makes `aria` attributes much more difficult to manage.

tjhiggins:
@zurbchris 
<b>A single dropdown pane that will attach to whichever anchor point was clicked?</b>
Yes

<b>Then which one does it anchor to for a programmatic open?</b>
Default to first anchor, but open could optionally take an anchor?

<b>It also makes aria attributes much more difficult to manage.</b>
True, but our use case cannot afford to recreate essentially the same dropdown multiple times. The cleanup logic could be kept generic for all anchors. You would probably need to have a this.$anchors and this.$active_anchor. I was working on a potential pr, but decided to wait until https://github.com/zurb/foundation-sites/pull/7353 was resolved. 

It might be too heavy handed, but we currently utilize that functionality with v5.

andycochran:
@tjhiggins Now that #7353 is closed, does that help with your PR? I'm going to close this for now. Please reopen if you'd like to submit that PR.

PR #7732, opened 2015-12-31
[Dropdown] Allow for multiple anchors for the same dropdown

tjhiggins:
Here is my first pass at allowing for multiple anchor points. While defaulting to the first anchor. Unsure what to do for the keyboard open.
http://codepen.io/TJHiggins/pen/JGEPLK

Original issue: https://github.com/zurb/foundation-sites/issues/7433

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -91,8 +91,8 @@
   Dropdown.prototype._init = function(){
     var $id = this.$element.attr('id');
 
-    this.$anchor = $('[data-toggle="' + $id + '"]') || $('[data-open="' + $id + '"]');
-    this.$anchor.attr({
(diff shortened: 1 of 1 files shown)


zurbchris:
After talking about this with the team, we're going to pass on this pr. The reason being, we never want to remove aria-attributes, if anything, we want to add to the current support. It seems it would be prohibitively complicated for accessibility devices to discern if panes had multiple anchors.

fvonellerts:
This is also possible without overwriting Foundation:
```html
<div data-multitoggle></div>
```
```javascript
const dropdown = new Foundation.Dropdown($('#dropdown'))

$(document).on('mouseenter', '[data-multitoggle]', e => {
  dropdown.$currentAnchor = $(e.currentTarget)
  dropdown.open()
}).on('mouseleave', '[data-multitoggle]', () => dropdown.close())
```

PR #7734, opened 2016-01-01
Update off-canvas.md

web2client:
In ##Setup, data-off-canvas -> data-off-canvas-content
Add example for responsive off-canvas menus

### DIFF
--- a/docs/pages/off-canvas.md
+++ b/docs/pages/off-canvas.md
@@ -114,3 +114,88 @@ The main content area (`.off-canvas-content`) will be padded to the left or righ
 ```
 
 <button type="button" class="button" data-docs-example-ofc>Toggle Reveal Class</button>
+
+## Responsive Off-Canvas (Putting it all together)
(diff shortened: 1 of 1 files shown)


rafibomb:
Brian, thanks for the PR and we'll be reviewing the update and merging it soon! Always great to have more thorough docs!

briantan888:
So after 6.1.1 is 6.2?

rafibomb:
@briantan888 I like the idea of having a complete copy/paste solution. We don't think the docs is the best place to have it. What we can do is make a Building Block, then link it in the docs.

rafibomb:
I made the building block here http://zurb.com/building-blocks/top-bar-with-off-canvas

rafibomb:
Thanks Brian! Added your suggestions here: https://github.com/zurb/foundation-sites/commit/21026b4e99cc8fb97a6b6ed093a66d2d79977472

briantan888:
Thanks Rafi for posting the example and adding content filler image.

PR #7736, opened 2016-01-01
Adds individual bar shifting to hamburger mixin

benjaminknox:
I just needed to shift one of the bars for a design somebody gave me and I thought I might make a pull request with what I changed.

Thanks!


zurbrandon:
This feels a little more specific a use case. May make a great example in a codepen though.

PR #7750, opened 2016-01-03
Last accordion's item border fixed

sashaegorov:
Hi,
Last accordion's border disabled on the wrong element. It should have a bottom border on the title and it shouldn't have it on contents.

Here are pictures instead of thousands of words.
**Before and wrong**
before
Note, there is no border on the title, and double border on contents (it's made of last item's content and whole accordion).

**After and correct**
after
Last item's content has no border.

Happy New Year!
:tada:


zurbrandon:
Thanks!

PR #7752, opened 2016-01-04
Add "data-options" to "foundation.tabs.js" to bypass Keyboar navigation

AlexLaforge:
Sometimes, we do NOT want key navigation to be used inside the plugins, as plugins can have conflicts when used together.

In "foundation.tabs.js", it would be very useful to be able to disable (bypass) the following key mapping.
in "foundation.dropdownMenu.js" too.

That is particularely annoying having this keys mapped, when we use FORMS inside the "tabs" component ! That leads the users being unable to use the RIGHT and LEFT key of their keyboard...

```
Foundation.Keyboard.register('Tabs', {
  'ENTER': 'open',
  'SPACE': 'open',
  'ARROW_RIGHT': 'next',
  'ARROW_UP': 'previous',
  'ARROW_DOWN': 'next',
  'ARROW_LEFT': 'previous'
  // 'TAB': 'next',
  // 'SHIFT_TAB': 'previous'
});
```


zurbrandon:
This pull request isn't referencing your branch. We're also not sure this is the best way to handle that bug, there may be a way to fix this issue without disabling the accessibility.

Issue #7758, opened 2016-01-04
The Great ES2015 Shift


gakimball:
We missed our chance to ship Foundation 6 in ES2015, but we're interested in upgrading the codebase.

To compile ES6 to 5 we'd use Babel, but *only* features that don't require polyfills. A basic list includes:
- Fat arrows
- ~~Classes~~ #8086
- Multiline strings
- ~~Template strings~~ #8018
- Destructuring
- Default function parameters
- Trailing function parameters
- Spread parameters
- `let` and `const`

This will also get us easier UMD support, by using Babel's module transpiler.

We need the shift to be **entirely backwards compatible**, which should be totally doable.

If you're interested in contributing, there's a `6.2` branch you can work on.

**Edit:** I also remembered there's some weirdness with how Babel's module transpiler works regarding browser globals, so we need to work around that. It doesn't mesh with the way we assign browser globals, since all of our plugins add to `window.Foundation`, not `window`.

gakimball:
We have Babel in our build process now! Next step is to do some experimenting with hooking up the `import` and `export` syntax, since the UMD support is arguably the most important thing to square away initially.

One tricky thing is with the Babel transforms available. We need UMD, but UMD transformers tend to assign each module to a variable on the window. We have our own process for loading plugins into `window.Foundation`, and we also don't want each little utility library to get its own variable in the global namespace, i.e. `window.foundationUtilMotion`.

Instead, in each plugin, we can do a check for an absence of an AMD/CommonJS environment, and then assign things the way we want to. For example:

```js
if (typeof module !== 'object' && typeof define !== 'function') {
  // For plugins
  Foundation.plugin(PluginClass);

  // For utilities
  Foundation.util.motion = Motion;
}
```

*However*, this doesn't work unless we transform to *UMD without browser globals*, which doesn't appear to be a thing that Babel's UMD transformer does, unless it's configurable in some way.

chrisjlee:
Where can i start? What's some low hanging fruit? Can i start creating an issue for each component?

gakimball:
@chrisjlee Some high-level stuff:

- Convert plugins to classes
- Use arrow functions when possible
- Use template strings instead of string concatenation

Remember that it's just the basic ES2015 stuff, nothing wild like generators or async functions. Our `.babelrc` has a list of every feature we're using.

Don't worry about module stuff just yet, but we will be looking to use `import` and `export` so we can compile to CommonJS and AMD.

[...]

PR #7762, opened 2016-01-04
[6.2][ES2015] Add babel to gulp pipeline

colin-marshall:
To get #7758 going.


gakimball:
Thanks for being on top of it! :)

Instead of the `es2015` preset, can we load only the plugins for features we're going to use? This might help keep us reigned in as we rework the codebase. (Our starter projects will probably just use the `es2015` preset, but we need to limit our codebase to only features that don't require polyfills.)

Here's the list. Each one has a matching Babel plugin:
- Fat arrows
- Classes
- Multiline strings
- Template strings
- Destructuring
- Default function parameters
- Trailing function parameters
- Spread parameters
- Let and const

colin-marshall:
Here are the plugins I came up with that match the features list you provided. Could use some help on the ones with question marks. 

* Fat Arrows = [es2015-arrow-functions](http://babeljs.io/docs/plugins/transform-es2015-arrow-functions)
* Classes = [es2015-classes](http://babeljs.io/docs/plugins/transform-es2015-classes)
* Multiline Strings = [es2015-template-literals](http://babeljs.io/docs/plugins/transform-es2015-template-literals) ???
* Template Strings = [es2015-template-literals](http://babeljs.io/docs/plugins/transform-es2015-template-literals)
* Destructuring = [es2015-template-literals](http://babeljs.io/docs/plugins/transform-es2015-destructuring
)
* Default function parameters = [es2015-parameters](http://babeljs.io/docs/plugins/transform-es2015-parameters)
* Trailing function parameters = ???
* Spread parameters = [es2015-spread](http://babeljs.io/docs/plugins/transform-es2015-spread) ???
* Let and const = [es2015-block-scoping](http://babeljs.io/docs/plugins/transform-es2015-block-scoping/) and/or [es2015-block-scoped-functions](http://babeljs.io/docs/plugins/transform-es2015-block-scoped-functions)???

gakimball:
All of those look fine. As for `es2015-block-scoped-functions`, from the readme:

> Babel plugin to ensure function declarations at the block level are block scoped.

That sounds like it doesn't require a polyfill, so that's fine.

colin-marshall:
@gakimball PR is updated with plugins.

gakimball:
Nice, thank you :)

PR #7763, opened 2016-01-05
[Button Group] Remove font-size: inherit; from button group

ksherman:
In my sites, the font-size: inherit conflicts with the button sizes (tiny, small, large). Ends up overwriting tiny with the default button font size. Without inherit, the sizes seem correct.

Here is a reduced test-case using the JSDeliver CDN CSS: http://codepen.io/ksherman/details/ZQLgPo/

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -34,7 +34,6 @@ $buttongroup-expand-max: 6 !default;
   #{$child-selector} {
     float: #{$global-left};
     margin: 0;
-    font-size: inherit;
 
     &:not(:last-child) {
       border-#{$global-right}: $buttongroup-spacing solid $body-background;


gakimball:
Are you trying to mix button sizes within one button group?

Silvester20:
I got the same issue, please remove the font-size inherit as it conflicts with the buttons.

andycochran:
With button groups, the size class goes on the `.button-group`, not the `.button`. If you put a size class on a `.button` within a group, it'll be overridden by the group's size class.

ksherman:
@gakimball In the attached CodePen in the OP, I'm not trying to use different sizes within the button-group, just trying to get the whole group to have the same size. 

@andycochran Interesting, you're right definitely because of the size: inherit. If I add .large to the .button-group, then the whole group does change to that size. I suppose that is preferable and I just missed that in the docs for the new button-group.

PR #7784, opened 2016-01-06
CSS Triangle Fix

### LINKED ISSUE #7645: Tooltip Hovering
Small issue with the tooltips. When you hover over the letters where the tooltip arrow appears it causes the tooltip to constantly interfere with hovering the tooltip-ed element in Firefox. In Chrome it only does it while moving the mouse
http://foundation.zurb.com/sites/docs/tooltip.html
In Firefox if you hover over the B in scarabaeus you can see what I'm talking about. As I said, Chrome does the same thing, but only when the mouse is moving.

andycochran:
This PR removes the unnecessary transparent borders in `@mixin css-triangle()`, which caused tooltips to flicker on hover. 

Fixes #7645.

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -25,18 +25,22 @@
   @if ($triangle-direction == down) {
     border-color: $triangle-color transparent transparent;
     border-top-style: solid;
+    border-bottom-width: 0;
   }
   @if ($triangle-direction == up) {
     border-color: transparent transparent $triangle-color;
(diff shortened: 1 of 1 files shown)


gakimball:
Nice, thanks!

zurbchris:
Thanks!

rafibomb:
@andycochran Thanks for the help! Email foundation at zurb .com and we'll send you a Yeti t-shirt!

PR #7790, opened 2016-01-06
Added and Clarified CDN option of Motion UI

andrewbiddinger:
I think this helps clarify that Motion UI is an optional thing to add to Foundation and give another option.


gakimball:
Can you remove the reference to the JS file? It's out of date with the motion library we use in F6, and we're going to drop it from that package eventually. Just referencing the CSS is fine.

andrewbiddinger:
OK. I created another pull request: #7794

gakimball:
Thanks!

Issue #7661, opened 2015-12-22
[Kitchen Sink] Add missing components


gakimball:
These components are missing from the kitchen sink in the docs:

- Grid
- Flex grid
- Visibility classes
- Magellan
- Responsive menu
- Responsive toggler
- Off-canvas
- Reveal
- Progress bar
- Abide
- Equalizer
- Interchange
- Toggler
- Sticky

zurbrandon:
Looks like @TvrtkoM crushed this one :)

TvrtkoM:
hey people! thanks, but it's issue is complete. some items are still missing in kitchen sink - the once i didn't know how to put in (grid, flex grid, magellan, off-canvas, sticky). 
I should have probably posted some kind of comment about this earlier. oops.

gakimball:
- For grid and flex grid, borrow the examples from the grid and flex grid docs page. They add extra styling to visualize the grid.
- For off-canvas, add a static HTML example and a click trigger underneath that opens the docs's off-canvas menu.
- For magellan, add an example as-is and see what happens. Even if it sticks to the top that's probably fine?
- For sticky, there's an example in the docs you could borrow with an image floating next to text.

PR #7804, opened 2016-01-07
add rest of missing components to kichen sink

TvrtkoM:
related to #7661

### DIFF
--- a/docs/pages/kitchen-sink.md
+++ b/docs/pages/kitchen-sink.md
@@ -392,6 +392,21 @@ description: Everything but.
 
 ---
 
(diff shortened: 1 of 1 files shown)


gakimball:
For the grid examples, can you have separate examples and live code? The samples shouldn't have the `.display` class in them; that's our docs-specific class.

Basically use `html` instead of `html_example`, and copy and paste the example below.

TvrtkoM:
@gakimball i did update on the docs. will it be OK?

gakimball:
Looks great! Thanks again.

PR #7808, opened 2016-01-08
[abide.md] Change event listener

web2client:
from .bind() to .on() as bind may fail before foundation is initialized.

Source: http://api.jquery.com/bind/


zurbchris:
I dig it!

PR #7810, opened 2016-01-08
fix hash problem with full url href hash

designerno1:
Error: Syntax error, unrecognized expression: ttp://www.example.com/#panel1
http://ajax.googleapis.com/ajax/libs/jquery/2.1.4/jquery.min.js
Line 2

Please take a look at the whole hash element finding thing i don't now if it is best practics as it is right now from Foundation 6

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -88,7 +88,7 @@
       var $elem = $(this),
           $link = $elem.find('a'),
           isActive = $elem.hasClass('is-active'),
-          hash = $link.attr('href').slice(1),
+          hash = $link[0].hash.slice(1),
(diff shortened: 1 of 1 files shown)


zurbchris:
Wow, haven't seen this one. There was/is one tiny problem with your pr though, in the _init method, the `tabContent` variable wasn't running through jQuery correctly due to a lacking `#`, I've fixed that. ARIA matters!
On a side note, I prefer the vanilla JS approach, we're trying to push more that direction anyway.

designerno1:
ok sorry missed the #

zurbchris:
no worries! Thanks for the fix for a problem I didn't know existed :smiley_cat:

PR #7814, opened 2016-01-08
Update accordion.md

jackpendleton:
Documentation updated to clarify the use of role and ARIA elements/attributes following basic markup.


zurbchris:
The JS manages the state of ARIA attributes already though. The user doesn't need to include any of that stuff.

jackpendleton:
Acknowledged. Working from the official documentation (
http://foundation.zurb.com/sites/docs/accordion.html) my commit allowed
that example to work properly.
Now removing role and ARIA elements/attributes, I see they are redundant.

The unattributed *data-tab-content* may have been the issue all along.

Thanks for the guidance.

PR #7840, opened 2016-01-11
Fix off-canvas for small sites

Grafikart:
Forced height of .off-canvas-wrapper-inner to be at least 100% 
## What it fixes ?

If the content inside `.off-canvas-content` is not tall enough, and doesn't fill the window. The menu is truncated 

**before**

**after**

### DIFF
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -60,6 +60,7 @@ $maincontent-shadow: 0 0 10px rgba($black, 0.5) !default;
     @include clearfix;
     position: relative;
     width: 100%;
+    min-height: 100%; 
     transition: transform $offcanvas-transition-length $offcanvas-transition-timing;
   }
 


gakimball:
What browsers did you test in? The off-canvas CSS is highly susceptible to cross-browser differences, especially in Safari and iOS.

Grafikart:
the problem appeared on chrome linux, didn't tried if it happened somewhere else

kball:
@Grafikart I see this got lost... I'd like to push this forward, and what would be very helpful to do so is a visual regression test (see https://github.com/zurb/foundation-sites/wiki/Testing-Guide) showing the bug that would let us quickly test across all environments to make sure this 1. fixes the issue and 2. doesn't cause other regressions.

Thanks!

zgianos:
I have seen a similar issue with the `off-canvas` class rules: `max-height` is set to `100%` but `min-height` is not.

*Without* `min-height: 100%`:
screen shot 2016-07-11 at 10 23 37 am

With `min-height: 100%`:
screen shot 2016-07-11 at 10 25 04 am

rafibomb:
@zgianos Are you confirming this fix is a good solution? I'd like to get it merged in

zgianos:
I believe it is a good fix. I have not tested it extensively, but it works for my test case. I believe `min-height: 100%` was removed from a previous version, possibly due to a refactoring error.

kball:
Looks like this was removed in 88e2af5306dde1c9d7b2c493b7aa7cdef0444932 which appears to have been focused on the case where content is too long.  Having both min-height and max-height seems like a reasonable fix to me.

luisarriojas:
I have downloaded Foundation 6.2.4 and have not compiled with SASS, only used file from dist folder.

Height of .off-canvas is not working properly. If the .off-canvas-content doesn't have enough content to increase the height of the whole .off-canvas-wrapper-inner then .off-canvas doesn't show its content as this bug was reported initially.

One solution is to set:
.off-canvas-wrapper-inner {
   height: 100vh;
}

This solution works in Chrome, Safari and Firefox on Mac. Don't know on Windows.

By the way, I have cannot increase the height of .off-canvas without break the layout at the bottom of .off-canvas when it is set to:
.off-canvas{
   height:100%;
}

michalzyka:
When I use `.off-canvas-wrapper-inner {
height: 100vh;
}` it messes up the momentum scrolling in iOS (the scrolling is much less user friendly this way).

kball:
Sigh, sorry to hear that @luisarriojas .  off-canvas as currently instantiated is a pain in the butt.  We're working on a new version that should be much better (cc @brettsmason).  That said, can you open a new issue with your example in a codepen that reproduces the problem you're seeing?  Thanks

brettsmason:
@luisarriojas as @kball said I am working on a complete overhaul of the off canvas component for the 6.3 release (RC should be out near the end of next month) so all user feedback would really be helpful to me to get it fit for purpose 😄

luisarriojas:
I created a new issue as @kball suggested. It is #9292. 
(cc @brettsmason)

Issue #7548, opened 2015-12-15
Reveal Scroll - fails on mobile


Ne-Ne:
I have just checked full screen reveal, on mobile.

Test Case:

iOS - 9.2
iPhone 6
Safari

This was the result:

![image1](https://cloud.githubusercontent.com/assets/2177851/11809895/f8a2fd6a-a321-11e5-9e3c-89f79269d000.PNG)

You can scroll, when the intention is to lock the screen. I think iOS is not respect overflow hidden. Could this be included to support touch devices?

http://stackoverflow.com/questions/4770025/how-to-disable-scrolling-temporarily

zurbrandon:
Oh man, that feels super dirty though. We'll take a look, but it feels like a really extreme option.

zurbrandon:
@gerardkcohen @Owlbertz any thoughts on locking the scroll for iOS here?

gerardkcohen:
Right, overflow:hidden on HTML will do it. A JS solution would be to trap
touch events to block "scrolling" of the viewport. Both have their issues,
with the CSS option my usual goto. Some gotchas for Firefox as well as
maintaining scroll position when overlay is dismissed. All as an FYI, I'm
sure you guys van work it out.

Let me know if you need anything else.

Gkc

zurbrandon:
@gerardkcohen looks like overflow: hidden doesn't work on iOS in this instance though, which is a huge bummer.

gerardkcohen:
@zurbrandon I use overflow:hidden here at work for that exact same thing with no issue. Interesting...

gerardkcohen:
Oh, positioned:fixed on the overlay too helps.

Gkc

zurbrandon:
Ah super odd, it's still scrolling. We may need to go another route.

Ne-Ne:
I think that if you are using a full screen class it should lock the viewport... However this leads to issues with locking child elements and the overflow potential and so on.

Typically I thought chrome dev tools would help me out, however it does not inherit bugs or touch events.

@gerardkcohen, may I ask what device you are using?

Ne-Ne:
@gerardkcohen It appears that the viewport meta tag affects the overflow property on the html & body. I don't think presuming all users will a surrounding class (for a overflow: hidden) the best idea.

What about a toggle function such as this: http://jsfiddle.net/3Ku26/

via - http://stackoverflow.com/questions/4770025/how-to-disable-scrolling-temporarily#answer-4770179

[...]

PR #7847, opened 2016-01-12
Update Reveal to support touch devices better

Ne-Ne:
Based off this issue (https://github.com/zurb/foundation-sites/issues/7548), @gerardkcohen css.

Hopefully this should sort scroll issues.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -333,12 +333,16 @@
      */
                  .trigger('open.zf.reveal');
 
-    $('body').addClass('is-reveal-open')
-             .attr({'aria-hidden': (this.options.overlay || this.options.fullScreen) ? true : false});
-    setTimeout(function(){
-      _this._extraHandlers();
-    }, 0);
-  };
+    // Remove class to html - supports mobile better
+    $('html').addClass('is-reveal-open');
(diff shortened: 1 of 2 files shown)


gakimball:
We'll need to do thorough browser testing on this, because the off-canvas CSS is fairly brittle and prone to cross-browser issues.

Would it be possible for you to take a look in Safari and Mobile Safari, if you can? Those two browsers tend to be the most problematic with off-canvas.

Ne-Ne:
I have checked through changes and they seem fine.

The CSS was actually recommended via @gerardkcohen.

zurbrandon:
Managed to pull this down and test it, but it appears to not function in safari at all. I'm getting the overlay for modals but not the modal it self. Are you seeing the same thing?

Ne-Ne:
I must have been in a different branch. 

Will re-check when I find time. @zurbrandon what browser version were you running?

kball:
Given the commentary above I'm going to close this.  Given the time lag I suspect @zurbrandon won't remember what the browser version was, but I'd check Safari 8 and Safari 9.  Please feel free to reopen after testing and bringing the branch up to date with develop.

PR #7871, opened 2016-01-13
Tab active color can be set

Grafikart:
I wanted to be able to change the color of an active tabs and there was no settings for that. 

```
/// Default text color for active items in a Menu.
/// @type Color
$tab-item-color-active: $tab-item-color !default;
```

I set the value as `$tab-item-color` to make it backward compatible

If I forgot to change some files please tell me

### DIFF
--- a/scss/components/_tabs.scss
+++ b/scss/components/_tabs.scss
@@ -114,6 +114,7 @@ $tab-content-padding: 1rem !default;
       &:focus,
       &[aria-selected='true'] {
         background: $tab-background-active;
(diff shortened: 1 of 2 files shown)


SalahAdDin:
:+1:

kball:
@Grafikart I'm going back through untended PRs and cleaning up.  As this is technically a new feature/api change, can you resubmit this against v6.3.0?  I'll close this one in favor of that as the expected new PR.  Thanks.

PR #7894, opened 2016-01-16
Moved closing div to wrap content on off-canvas

0xgdr:
The example had the closing `off-canvas` tag before the `off-canvas-content` so the toggle example would not work correctly.


gakimball:
The existing example is correct; the menus are siblings of the content wrapper. You can see in [our documentation layout](https://github.com/zurb/foundation-sites/blob/develop/docs/layout/default.html#L19) how this works.

PR #7898, opened 2016-01-16
use MouseEvent constructor to simulate touch events.

TvrtkoM:
simulating event with .initMouseEvent still available for browsers not supporting MouseEvent.


TvrtkoM:
.initMouseEvent is deprecated

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent

zurbchris:
Nice catch! Thanks :+1:

PR #7903, opened 2016-01-18
should fix #7836

### LINKED ISSUE #7836: [Responsive Menu] data-force-follow="true" resize
Setup is data-responsive-menu="accordion large-dropdown"

if you start from small and go to large the force follow not works
it gets lost equal you start, only by resizing

designerno1:
[Responsive Menu] data-force-follow="true" resize
Misspelling accordionMenu so the handler couldn't off on destroy

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -105,7 +105,7 @@
       var $submenu = $(this).children('[data-submenu]');
 
       if ($submenu.length) {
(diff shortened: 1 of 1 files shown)


designerno1:
or make it lowercase but then the off handler and so on has also to be lowercase to destroy and bind

gakimball:
@zurbchris I think all events are lowercase?

zurbchris:
grr, they should've been, but this one was missed.

PR #7912, opened 2016-01-18
Issue with responsive gutter

michalgurgacz:
Case:
screen shot 2016-01-18 at 22 00 37

Issue:
screen shot 2016-01-18 at 22 04 14
-- selector for current width breakpoint is loosing with previous selector for smaller resolution

After fix:
screen shot 2016-01-18 at 22 00 24
--it look like this selector is existing twice in line 1155 and 947, but those are selectors for medium and large breakpoint

### DIFF
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -28,7 +28,9 @@
 
     // Collapsing
     &.#{$collapse} {
-      > .#{$column} { @include grid-col-collapse; }
+      > .#{$column} {
+        @include grid-col-collapse;
+      }
     }
 
     // Nesting
@@ -39,8 +41,7 @@
             @include grid-row-nest($gutter);
           }
         }
-      }
-      @else {
+      } @else {
         @include grid-row-nest($grid-column-gutter);
       }
 
@@ -116,7 +117,9 @@
 
     // Responsive collapsing
     .#{$-zf-size}-#{$collapse} {
-      > .#{$column} { @include grid-col-collapse; }
+      > .#{$column} {
+        @include grid-col-collapse;
+      }
 
       .#{$row} {
         margin-left: 0;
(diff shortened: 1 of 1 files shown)


gakimball:
Can you explain in more detail what the changes do?

michalgurgacz:
Using 'X-uncollapse' css rule (in screenshot css line 737)  is always wining with rule '.column' rule (line 473). Because is more precise.

My idea was to get all smaller breakpoint css rules for gutter and override them with correct value of gutter (for current breakpoint).
Example:  (for large breakpoint we should override small and medium css)
```css
.small-uncollapse > .column, 
.medium-uncollapse > .column,
.large-uncollapse > .column{
      //and here correct value of responsive gutter for large screen
}
```

In code:
line 127: checks with kind of gutter we should perform responsive or static
lines 128-129: this is support for old not responsive gutter (this config will be used in lines 144-153)
lines 131-140: this is getting all smaller breakpoints names in to list $-zf-sizes
line 141: is getting gutter value for current breakpoint
lines 144-147: is generating .X-uncollapse (where X are names of smaller breakpoints from $-zf-sizes list)
lines 149-153: produce the final scss rule 

lines 144-153: are working for old static gutter and new responsive. I did it, to not duplicate logic of generating 'uncollapse' and 'column'

kball:
@andycochran @brettsmason @JeremyEnglert do you understand what this is trying to do?

andycochran:
There are conflicts in this PR. But I'm not sure it's still an issue. I don't see `[size]-uncollapse` selectors repeated in the compiled CSS. 

I do wonder though if `small-uncollapse` has any value. Why would you uncollapse the smallest breakpoint? 

I'm going to close this for now. @michalgurgacz, if you still find this to be a bug in the latest code, please fix the conflicts and reopen.

PR #7915, opened 2016-01-19
[6.2] increment jQuery version to ~2.2.0

tomByrer:
[faster & SVG support](http://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/)
bumped foundation-sites version

I'm building with this config now; let you know if there are problems if you want to wait.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "foundation-sites",
-  "version": "6.1.0",
+  "version": "6.2.0",
   "main": [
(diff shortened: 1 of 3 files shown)


gakimball:
Can you remove the version bump for Foundation itself? Our deploy process uses a find-and-replace to increment the version number, so you can leave it at 6.1.0 for now.

tomByrer:
OOps sorry, let me reopen.

Issue #7665, opened 2015-12-22
Visual quirk with button groups with a border radius


gakimball:
Reopening issue 172 by @ProLoser from the old foundation-sites-6 repo.

> So the fact that in a button group all the right borders are white (in order to create the gap) creates a weird visual quirk with the border-radius. I realize that inline-block is not normally super susceptible to margins, but in this case I think using a 1px margin instead of a white border would be smarter.

> I found a lot of frustrations with how borders were utilized in f5 and was required to completely rewrite the buttons mixins because borders are explicitly used to create buttons, even though they don't appear visually on the screen. The problem with this is that if you actually WANT borders to appear on buttons, you have to do a LOT of css overriding and it was just easier to write a lighter mixin.

> Any thoughts on removing border css from buttons? Should I file another ticket?

@ProLoser's suggestion for stripping radius properties from inner buttons:

```css
.button-group .button:not(:first-child) {
  border-top-left-radius: 0;
  border-bottom-left-radius: 0;
}
.button-group .button:not(:last-child) {
  border-top-right-radius: 0;
  border-bottom-right-radius: 0;
}
```

![image](https://cloud.githubusercontent.com/assets/1142001/11968365/717153c4-a8c0-11e5-940c-fb261622b461.png)

![image](https://cloud.githubusercontent.com/assets/1142001/11968380/9ae778a0-a8c0-11e5-97f8-74b9aa8f196a.png)

marcusconnor:
I agree with using margin rather than border as a separator. I also think the stacked group should follow the same convention with a single pixel margin on the bottom of each button. The current border separator doesn't look good in default because it causes weird visual behavior.

This image shows current default behavior:

![button-group-issue](https://cloud.githubusercontent.com/assets/11634985/12208456/d059c3dc-b61a-11e5-9985-dcccf9544684.png)

gakimball:
Our button groups are made with `display: table` and `display: table-cell`, and it appears that setting `margin-right` doesn't do anything—the margin doesn't apply, and the buttons sit right next to each other.

andycochran:
This is another reason to maybe rethink the `display:table` approach for button groups—in addition to #7844 (`<button>` element not working).

gakimball:
@andycochran Even with the old approach where we used floats and percentage widths, it still wouldn't work, because margin doesn't factor into width calculation. Is there another way to do even width?

andycochran:
[...]

Issue #7844, opened 2016-01-12
Expanded button-group does not work for <button> in non-IE browsers


arangas:
I'm using an expanded button-group. It works great using a button of form `<a class="button">` but not for `<button type="button" class="button">`.

I made a [CodePen here](http://codepen.io/alexangas/pen/JGJMqv) to demonstrate.

IE 11 works as expected. Firefox doesn't do the expanding. Chrome renders the buttons on top of each other.

andycochran:
Hmm. Looks like `display:table-cell` will not work on the `<button>` element because it has a "[special box type](https://bugzilla.mozilla.org/show_bug.cgi?id=960512)." 

The table cell technique for `.button-group` isn't gonna work for `<button>` elements. :disappointed:

arangas:
Great find! I could not work out what was going on using the dev tools.

Do you think a message could be added to the docs for Button Group so users are aware and don't waste their time?

gakimball:
@arangas We can add a temporary note to the docs, but we really want people to use `<button>` over `<a>` when it makes sense, so it may be worth going back to the old float model.

gakimball:
Fixed via #7933

PR #7933, opened 2016-01-19
Rework Button Group to not use borders

andycochran:
This PR implements margin for the space between buttons, instead of faking it with borders. It also nixes the use of table display & floats, instead using `inline-block`. 

Let me know if you have feedback. 

Woot!
-Andy

Related issues: #7665, #7844

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -99,7 +99,7 @@ $buttongroup-expand-max: 6 !default;
     @include button-group;
 
     // Sizes
-    @each $size, $value in $button-sizes {
+    @each $size, $value in map-remove($button-sizes, default) {
       &.#{$size} #{$buttongroup-child-selector} { font-size: $value; }
     }
 
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -12,7 +12,7 @@ $buttongroup-margin: 1rem !default;
 
 /// Margin between buttons in a button group.
 /// @type Border
-$buttongroup-spacing: 1px !default;
+$buttongroup-spacing: 10px !default;
 
 /// Selector for the buttons inside a button group.
 /// @type String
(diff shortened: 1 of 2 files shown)


andycochran:
Also makes button group spacing look good on backgrounds not matching `$body-background`.

andycochran:
Remedies these border-radius isses: 
screen shot 2016-01-20 at 12 39 01 pm
screen shot 2016-01-20 at 12 42 11 pm

gakimball:
Does `font-size: 0` get around the extra whitespace created by `display: inline-block`?

andycochran:
Exactly. The buttons have a space character's width between them. If the font size is zero, that space disappears.

gakimball:
One piece of feedback, the sizing classes shouldn't print inside the `button-group()` mixin—they should stay in the CSS output. Otherwise you're tightly coupled to those sizing class names (and class chaining) when using the mixin.

andycochran:
D'oh! I'll work on that. Thanks, @gakimball.

andycochran:
@gakimball Take a look. You'll notice I redid the button group sizes like this: 
```scss
@each $size, $value in $button-sizes {
  &.#{$size} #{$buttongroup-child-selector} { font-size: $value; }
}
```
This allows user-defined button sizes. 

A similar method could be used for all button sizing. For instance, lines 197–199 in `_button.scss` could be… 
```scss
@each $size, $value in $button-sizes {
  &.#{$size} { font-size: $value; }
}
```
instead of… 
```scss
&.tiny     { font-size: map-get($button-sizes, tiny); }
&.small    { font-size: map-get($button-sizes, small); }
&.large    { font-size: map-get($button-sizes, large); }
```

Caveat: It adds a `.default` button size class. 
Thoughts?

gakimball:
Definitely interested in that. You can remove `default` from the map on the fly using `map-remove()`.

```scss
@each $size, $value in map-remove($button-sizes, default) {
}
```

andycochran:
@gakimball Done. This'll close #7665, #7844, ~~and #7972~~.

gakimball:
@andycochran Solid, thanks so much for putting it together!

slavanga:
The font-size: 0 trick doesn't work in the android browser pre Jellybean.
http://codepen.io/stowball/pen/LsICH

andycochran:
@slavanga We decided in https://github.com/zurb/foundation-sites/issues/7665#issuecomment-172625174 that this was acceptable fallout in oder to fix a few other bugs in all browsers.

slavanga:
@andycochran Alright, that sounds reasonable.

PR #7939, opened 2016-01-20
the name is title bar

amosmos:
not only this is how the class is called, it is also how it is named in the customized download options of zurb...


amosmos:
it is also called title bar in the off-canvas documentation

gakimball:
Thanks for the fix :)

Issue #7907, opened 2016-01-18
[Top bar] Suggestion/feedback for more flexibility


brettsmason:
I've been trying to use the top bar in a WordPress theme I've just released (more to follow on that!) but am having problems using it. I wanted to see if it would be considered. If so I'll put together a PR.

Basically I would love to dscuss changing the top bar markup to:

```
<div class="top-bar">
	<div class="top-bar-branding"><!-- rather than having menu-text within a menu -->
		Logo or text here
	</div>
	<div id="menu-id-to-toggle" class="top-bar-menu"><!-- wrap the navigation -->
		<ul class="top-bar-left menu"> // or keep as divs like at the moment and nest the menu
			<ul class="menu">
			</ul>
		</ul>
		<ul class="top-bar-right menu"> // or keep as divs like at the moment and nest the menu
		</ul>
	</div>
</div>
```

Styling wise:

```
.top-bar-branding {
	float: left; // so that its displayed on the left of the top-bar-menu
}

.top-bar-menu {
	display: block; // displayed to the right of .top-bar-branding
}
```

The benefit of this is that the top-bar-menu can be toggled with the responsive toggle, but keep the branding intact. Also I find the markup a bit awkward at the moment when trying to use a logo (which I expect is fairly common) within the menu and thought this would be cleaner.

I'd love to hear thoughts on this - I can put up a Codepen if needed.

brettsmason:
Here's an idea of what I meant: [CodePen](http://codepen.io/brettsmason/pen/WrXoOe)
Just wondered if this was something that would be considered or altered or if its more a place for custom code on my part?

gakimball:
This is interesting, and like you said fits better with the responsive toggler plugin. We'll talk about it internally a bit.

zurbrandon:
This is actually really cool! I'd love a pull request with the top-bar-branding class you made if you have a chance.

brettsmason:
No problem I'll create a PR this evening. Before I do this though any preference on behaviour with the .stacked-for-xx code? This is what I thought:

```
/// makes sections stacked
@mixin top-bar-stacked {
  // Sub-sections
  .top-bar-branding {
    width: 100%;
  }

  .top-bar-right {
    width: 100%;
  }

  .top-bar-left {
    width: 100%;
  }
}

@mixin foundation-top-bar {
  // Top bar container
  .top-bar {
    @include top-bar-container;
  }

[...]

PR #7944, opened 2016-01-20
Add .top-bar-title class

brettsmason:
Adds .top-bar-title class to the title as per issue [7907](https://github.com/zurb/foundation-sites/issues/7907).

### DIFF
--- a/scss/components/_top-bar.scss
+++ b/scss/components/_top-bar.scss
@@ -42,6 +42,10 @@ $topbar-input-width: 200px !default;
 /// makes sections stacked
 @mixin top-bar-stacked {
   // Sub-sections
+  .top-bar-title {
+    width: 100%;
+  }
+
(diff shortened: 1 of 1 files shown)


gakimball:
Awesome, thanks! :) Just one bit of feedback.

By referencing `$grid-column-gutter`, top bar has a dependency on the grid. Usually this doesn't matter, but we avoid having components reference each others' variables. Instead, can you add a dedicated variable for this? Maybe `$topbar-title-spacing`? That would be great.

brettsmason:
@gakimball No problem I'll do that now. Do you think it should be padding or would margin make more sense?

gakimball:
@brettsmason I don't know that it will make a difference, assuming people are never setting a background color on the thing. I suppose do margin to be safe.

gakimball:
Thanks!

Issue #7962, opened 2016-01-22
Foundation._activePlugins object remains empty


manugill:
In 6.1.1

The _activePlugins object doesn't store any active plugins.

Comment on registerPlugins function in foundation.core.js:
* Creates a pointer to an instance of a Plugin within the Foundation._activePlugins object.

But I couldn't find it manipulating _activePlugins at any point.

zurbchris:
Yeah, that object is no longer being used. I'll pull it out.

gakimball:
Removed in d1f79fd515c719106e4d43c02865fe8b743b3468

PR #7963, opened 2016-01-22
Stored active plugins to Foundation._activePlugins

manugill:
Implemented to addition on registerPlugins and removal on unregisterPlugins. Corresponding to issue #7962.


gakimball:
Per @zurbchris's [comment here](https://github.com/zurb/foundation-sites/issues/7962#issuecomment-173997817), this is an unused property.

PR #7980, opened 2016-01-23
Update settings import in docs

petecoop:


gakimball:
You should have a `_settings.scss` in your own Sass files, separate from the one in the Foundation package. Then running `@import 'settings';` will work fine, assuming that file is next to your main Sass file.

PR #7989, opened 2016-01-25
Minor change to Advanced example...

LDigital84:
This is a very minor change. In the first example you had two `top-bar-left` classes. It seems more logical to have one left and one right as that example which will then match the example below that.


zurbchris:
Makes sense to me, thanks!

PR #8007, opened 2016-01-26
Update pagination.md

web2client:
Missing <a> tag for pagination-previous class


briantan888:
Turn out class pagination-next/previous disabled cannot use anchor, otherwise double left/right arrow. Close pr.

PR #8018, opened 2016-01-26
ES6 Template Literals

colin-marshall:
I went through all of the JavaScript files and searched for `+ '`, `+'`,  `+ "`,  `+"`, `' +`, `'+`, `" +`, and `"+` to find places where I could replace concatenation with template literals.

Be forewarned, this is my first time using ES6/Babel and I was not sure if template literals are allowed everywhere I used them. I also have not had a chance to go through every component to verify they still work after migrating to template literals.

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -100,7 +100,7 @@
      */
     validators: {
       equalTo: function (el, required, parent) {
-        return $('#' + el.attr('data-equalto')).val() === el.val();
+        return $(`#${el.attr('data-equalto')}`).val() === el.val();
       }
     }
   };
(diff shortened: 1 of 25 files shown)


andycochran:
@colin-marshall It looks like you've used backticks instead of straight quotes.

gakimball:
This is awesome!

A quick way to test this is by thumbing through the Kitchen Sink page in the documentation.

colin-marshall:
@gakimball good call on the Kitchen Sink. Found a typo in one of the templates in Tabs so I fixed that, and everything appears to work like it should. I also rebased the PR with the 6.2 branch to get the fix you just put in for one of the two merge conflicts, and now the build is passing.

gakimball:
Nice, thanks for that :)

PR #8031, opened 2016-01-28
Added button-color-alt variable to mixin

cucharadepalo:
The alternative font color for buttons wasn't present in the button-style mixin, and the button-style constructor for alternative colors was missing the `auto` parameter to get the foreground function working.
Every button, regardless of the lightness of the background color was getting white fonts…

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -99,7 +99,7 @@ $button-opacity-disabled: 0.25 !default;
   $color: $button-color
 ) {
   @if $color == auto {
-    $color: foreground($background);
+    $color: foreground($background, $button-color-alt, $button-color);
   }
 
   @if $background-hover == auto {
@@ -203,7 +203,7 @@ $button-opacity-disabled: 0.25 !default;
     @each $name, $color in $foundation-colors {
       @if $button-fill != hollow {
         &.#{$name} {
-          @include button-style($color, auto);
+          @include button-style($color, auto, auto);
         }
       }
       @else {


kball:
@andycochran or @brettsmason can one of you take a look at this and chime in on if you think if it is likely to cause any issues with existing sites or code?

It seems like a straightforward bugfix but I don't have a strong enough sense of the sass to know if there's likely to be many things depending on the old behavior.

andycochran:
Hmm. This change does nothing for me. All button text is still white. I'll dig deeper…

andycochran:
Ah ha! So this works for every button with an explicit color (e.g. `.secondary` or `.warning`) but not for default buttons. Default buttons still always have white text. 

Also, I think the threshold of the `foreground()` function needs to be adjusted, 'cause you can set the button color pretty light and still get white text: 
screen shot 2016-04-06 at 6 23 45 pm

andycochran:
Okay, I poked around with this and I'm gonna merge it. It fixes the bug (`$button-color-alt` applies correctly to light-colored buttons). 

Although it's not perfect, but this a probably a separate issue:  
* Default buttons are always styled with `$button-background` and `$button-color` :+1: 
* `$button-color-alt` is only used for light colors in `$foundation-palette` :+1: 
* The `foreground()` function isn't great and often applies the wrong color :-1: 

A better solution than `foreground()` might be a map of alt colors. A function can't replace a designer's eye.

PR #8038, opened 2016-01-29
Remove border when buttons are stacked

alamowo:
You can see the same false behavior in the foudation docs. I think my solution should do the trick.


andycochran:
The space between Buttons in Button Groups is no longer done with border (see #7933).

gakimball:
Yep, we refactored button groups to not need this fix.

Issue #7993, opened 2016-01-25
drilldown parent Link in Foundation 6


InsertCart:
I'm using foundation 6 everything works fine but in drilldawn menu how user can click in parent link, in Mobile device

```
<div class="title-bar" data-responsive-toggle="responsivehide" data-hide-for="medium">
  <button class="menu-icon" type="button" data-toggle></button>
  <div class="title-bar-title"><?php _e('Menu','TEXT-Domain'); ?></div>
</div>

<div class="top-bar" id="responsivehide">	
		
	<?php wp_nav_menu( array( 'theme_location' => 'top-menu', 'items_wrap' => '<ul class="vertical medium-horizontal menu" data-responsive-menu="drilldown medium-dropdown">%3$s</ul>', 'walker' => new myfun_walker( array( 'in_top_bar' => true, 'item_type' => 'li', 'menu_type' => 'top-menu' ) ), ) ); ?>
	
</div>
```

Here is walker 
```
<?php
/* ------------------------------------------------------------------------- *
 *  Navigation  
/* ------------------------------------------------------------------------- */
?>

<?php 

class myfun_walker extends Walker_Nav_Menu {

	
	/**
	 * Specify the item type to allow different walkers
	 * @var array
	 */
	var $nav_bar = '';
	function __construct( $nav_args = '' ) {
		$defaults = array(
			'item_type' => 'li',
			'in_top_bar' => false,
            'add_dividers' => false,
			'menu_type' => 'main-menu' //enable menu differenciation, used in preg_replace classes[] below
		);
		$this->nav_bar = apply_filters( 'req_nav_args', wp_parse_args( $nav_args, $defaults ) );
	}
	function display_element( $element, &$children_elements, $max_depth, $depth=0, $args, &$output ) {
        $id_field = $this->db_fields['id'];
        if ( is_object( $args[0] ) ) {
            $args[0]->has_children = ! empty( $children_elements[$element->$id_field] );
        }
        return parent::display_element( $element, $children_elements, $max_depth, $depth, $args, $output );
    }
	function start_el( &$output, $item, $depth = 0, $args = array(), $id = 0 ) {
		global $wp_query;
		$indent = ( $depth ) ? str_repeat( "\t", $depth ) : '';
		$class_names = $value = '';
		$classes = empty( $item->classes ) ? array() : (array) $item->classes;
		$classes[] = 'menu-item-' . $item->ID;
[...]

Issue #7770, opened 2016-01-05
[Accordion Menu/Drilldown] Feature request/suggestions


brettsmason:
I love the new menu system as a whole but would love to see a few enhancements to the accordion menu/drilldown menu for menus that have sub navigation to make it possible to click the sub-menu parent links - its the one thing stopping me using both of them.
### Drilldown

I would love to see an option (`data-options`?) that added the parent menu link as a link (copy?) in the sub-menu, so that it was clickable.
### Accordion Menu

Similar issue - it would be great if we could have an option to keep the link on the parent item and make the sub-menu toggle a separate element so that the parent link still works.
An example of what I mean is [here](http://samikeijonen.github.io/responsive-nav.js/demos/multiple-levels/index.html). There's also a writeup about it [here](https://foxland.fi/accessible-multi-level-dropdown-navigation/) if that helps.

I'd love to discuss this further and any other ideas! Unfortunately my JS skills are very limited so I can't do much to help in that area yet :disappointed:

brettsmason:
Wondered if this had ever been discussed at any point? (during planning for F6 or after release) I really wish I had the JS skills to write a PR! This would be a massive help for usability for me atleast (and I'm sure others too as they seem to be common patterns).

Baedda:
I am wondering that they dont have this feature for the drilldown/accordion menu in F6. They had in the old topbar menu since Foundation 4. This is the code from the topbar of F5:

```javascript
    if (!$dropdown.find('.title.back').length) {
          $titleLi = $('<li class="title back js-generated"><h5><a href="javascript:void(0)"></a></h5></li>');
  
      // Copy link to subnav
      if (settings.custom_back_text == true) {
        $('h5>a', $titleLi).html(settings.back_text);
      } else {
        $('h5>a', $titleLi).html('&laquo; ' + $link.html());
      }
      $dropdown.prepend($titleLi);
    }
```

I agree with you this feature is very essential. I am working on a new project atm where i wanted to use F6s menu plugins for the navigation, but because this feature is missing I have to watch for another solution :(

andycochran:
Hmm. Good idea. For now, you could roll your own with Toggler.

brettsmason:
Not a bad idea @andycochran, though I don't think it will particularly easy to do within WordPress (the context where I would be using it). I might explore this route if the Zurb team say this is a dead end though.

brettsmason:
[...]

PR #8051, opened 2016-02-02
Adds the parentLink option - false by default.

brettsmason:
Clones the parent menu link and prepends it to the submenu if the option is set to true.
Need some input on this for the correct way to achieve this, but it works correctly in my testing.

Also ideally should it should work with the Nest library to add the classes etc that way, rather than me adding them in the drilldown.js file.

I would love to hear thoughts and improve the code so we can get this added - I really need this in order to use the menu :smile:

See issues [7770](https://github.com/zurb/foundation-sites/issues/7770) and [7993](https://github.com/zurb/foundation-sites/issues/7993)

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -49,6 +49,12 @@
      */
     wrapper: '<div></div>',
     /**
+     * Adds the parent link to the submenu.
+     * @option
+     * @example false
+     */
+    parentLink: false,
+    /**
      * Allow the menu to return to top level menu on body click.
      * @option
      * @example false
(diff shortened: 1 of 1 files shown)


zurbchris:
Pretty sure this wasn't your desired effect, but this is what's happening on my local: 
screen shot 2016-02-02 at 9 47 42 am

It's duplicating the entire nested menu. You can fix this pretty easily, I think.

gakimball:
Also, could you target the `6.2` branch instead? Thanks :)

brettsmason:
@zurbchris sorry no idea how I managed to mess that up... Have a look now and let me know if its OK.
@gakimball No problem - whats the best way to do that, close this PR and open a new one?

zurbchris:
coolio, simple and functional. 
Also, yes, I don't know of an easy way to re-target a branch for a pr. Just easier to close this one and open a new one as 6.2 as the target.

brettsmason:
See [8060](https://github.com/zurb/foundation-sites/pull/8060) - hopefully thats OK!

PR #8065, opened 2016-02-03
Feature: added .alert, .warning and .success colors to switch input

foglcz:
|  |  |
| --- | --- |
| **Type** | Feature |
| **Fixes issues** | n/a |
| **Documentation** | updated |
| **BC Break** | no |

In our internal intranet, we have a shipping dispatch form built with Foundation.

In there, we have various checkboxes that update ERPs / suppliers interfaces / databases / etc. I wanted to bring one of those checkboxes to users attention when checked (since default is off, and when on, it does a lot of things _behind the scenes_).
Intuitively, I tried adding "warning" class to the checkbox label.... and nothing happened.

This fixes the "issue", if you agree that this might/should be part of standard.

Currently I'm using following in my app.scss tu spplement the feature:

``` scss
input:checked~.switch-paddle.alert { background: $alert-color; }
input:checked~.switch-paddle.warning { background: $warning-color; }
input:checked~.switch-paddle.success { background: $success-color; }
```

Let me know if this makes sense / you would want to somehow update this.

Thanks!
Pavel

### DIFF
--- a/docs/pages/switch.md
+++ b/docs/pages/switch.md
@@ -29,6 +29,37 @@ Give the `<input>` a unique ID and point the `<label>` to it with the `for` attr
 
 ---
 
+## Coloring
+
+Switch paddle can be colored, using the .success`, `.warning` or `.alert` classes.
+
+```html_example
+<div class="switch">
+  <input class="switch-input" id="exampleColorSwitch1" type="checkbox" name="exampleColorSwitch1">
+  <label class="switch-paddle success" for="exampleColorSwitch1">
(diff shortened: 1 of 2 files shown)


andycochran:
@gakimball, would this be better done with a mixin and app-specific CSS? Especially considering flexible color palettes https://github.com/zurb/foundation-sites/issues/7219#issuecomment-166990006, and to avoid hard-coded the classes.

gakimball:
Thanks for the PR! :) Unfortunately, we aren't interested in adding coloring classes to switches at the moment, as the use-case is fairly narrow.

PR #8075, opened 2016-02-03
[tooltip.js] Add tooltip as html

web2client:
Add tooltip as html rather than text, same behaviour as F5.

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -126,7 +126,7 @@
     this.template = this.options.template ? $(this.options.template) : this._buildTemplate(elemId);
 
     this.template.appendTo(document.body)
-        .text(this.options.tipText)
+        .html(this.options.tipText)
         .hide();
 
     this.$element.attr({


zurbchris:
That allows for script injection into a tooltip, which is definitely not cool.

rafibomb:
@zurbchris That's a valid concern. Is there a better way to do this you can suggest?

paxperscientiam:
What if the HTML were generated with PHP?

briantan888:
How is it different from dropdown, reveal, Ajax-reveal?

gakimball:
We just had a conversation about this internally, can we make it an option instead? Like `allowHtml`, set to `false` by default. That way if you're loading, say, user-generated content into tooltips, you aren't opening yourself up to XSS attacks.

kball:
@briantan888 I'd love to move this forward for 6.3, can you take a look at @gakimball's feedback, take a pass on it, and resubmit this as a PR against v6.3.0?  Thanks!

PR #8080, opened 2016-02-04
Features for drilldown - get visible height on a hidden drilldown-menu

monosize:
also extend Foundation.Box.GetDimensions

foundation.drilldown.js - new Option showParentLink, if true then the parent button is prepended to the submenu, only if the parent href is set.

foundation.util.box.js - Extend Foundation.Box.GetDimensions to use on an hidden element

drilldown.scss
- new vars: $drilldown-auto-width and $drilldown-auto-width-breakpoint
  to set the width to auto at a specific breakpoint
- new css class .is-drilldown-prepare - is used for calculating the max height and width of the menu during preparing

### DIFF
--- a/docs/pages/javascript-utilities.md
+++ b/docs/pages/javascript-utilities.md
@@ -15,6 +15,11 @@ One of the useful libraries is `Foundation.Box`, and it has a couple methods des
 ```js
 
 var dims = Foundation.Box.GetDimensions(element);
+
+// or to get dimension on an hidden element
+
+var dims = Foundation.Box.GetDimensions(element, true);
+
 ```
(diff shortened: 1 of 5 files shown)


zurbchris:
My only comment is that the Box util should be vanilla js.

monosize:
Box util is now in vanilla js.

monosize:
Shall I open a new PR for the 6.2 branch?

rafibomb:
@monosize That would be great! We're looking to do a release candidate on 6.2 this week for testing - thanks!

monosize:
Moved to 6.2 branch

PR #8085, opened 2016-02-04
topbar feature - set a diffrent background color of all submenus

monosize:
New option to set a different background-color for the submenus of a menu in the topbar.
For example if the topbar-background is transparent.
new var $topbar-background-submenu

### DIFF
--- a/scss/components/_top-bar.scss
+++ b/scss/components/_top-bar.scss
@@ -14,6 +14,10 @@ $topbar-padding: 0.5rem !default;
 /// @type Color
 $topbar-background: $light-gray !default;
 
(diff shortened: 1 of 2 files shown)


gakimball:
A few requests/suggestions:

- Can you open a new PR that targets the `6.2` branch instead? Variables are only changed in minor versions.
- Rename the variable to `$topbar-submenu-background`. We order the words by component name, then CSS property.
- You can also probably change `ul { ul {} }` to just `ul ul {}`.

Thanks :)

SalahAdDin:
:+1:

PR #8091, opened 2016-02-05
topbar feature - set a diffrent background color of all submenus

monosize:
New option to set a different background-color for the submenus of a menu in the topbar.
For example if the topbar-background is transparent.
new var $topbar-background-submenu


gakimball:
Thanks :)

SalahAdDin:
:+1:

PR #8106, opened 2016-02-07
Update jQuery dependency to 2.2

JoelESvensson:
Should be 100% backwards compatible

https://blog.jquery.com/2016/01/08/jquery-2-2-and-1-12-released/


gakimball:
Already implemented in the `6.2` branch, which was just merged into `develop`.

PR #8113, opened 2016-02-08
Typo in column classname (.column to .columns)

purplespider:


gakimball:
You can use `.column` or `.columns`, but the page wasn't clarifying this, so I updated it to do so: e1281f64e30900ead1dd2fcd896ae0c7be39d1e2

PR #8120, opened 2016-02-09
Features for drilldown - get visible height on a hidden drilldown-men…

monosize:
…u, also extend Foundation.Box.GetDimensions

foundation.drilldown.js - new Option showParentLink, if true then the parent button is prepended to the submenu, only if the parent href is set.

foundation.util.box.js - Extend Foundation.Box.GetDimensions to use on an hidden element

drilldown.scss
- new vars: $drilldown-auto-width and $drilldown-auto-width-breakpoint
  to set the width to auto at a specific breakpoint
- new css class .is-drilldown-prepare - is used for calculating the max height and width of the menu during preparing

### DIFF
--- a/docs/pages/javascript-utilities.md
+++ b/docs/pages/javascript-utilities.md
@@ -15,6 +15,11 @@ One of the useful libraries is `Foundation.Box`, and it has a couple methods des
 ```js
 
 var dims = Foundation.Box.GetDimensions(element);
+
+// or to get dimension on an hidden element
+
+var dims = Foundation.Box.GetDimensions(element, true);
+
 ```
 Will return an object that contains the dimensions of the `element` passed. The object return looks like:
 
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -78,13 +78,26 @@ export default class Drilldown {
     this.$submenus.each(function(){
       var $menu = $(this),
           $back = $menu.find('.js-drilldown-back');
+
+      if (_this.options.showParentLink) {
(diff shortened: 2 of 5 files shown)


rafibomb:
Thanks for the PR! What is the use case this solves?

monosize:
If you use a drilldown menu inside a hide-for-... element then the drilldown menu height isn't calculate correctly. Try it without this PR and change the window size of the browser from medium to small.

```html
<div class="hide-for-medium">
	<ul class="vertical menu" data-drilldown>
		<li>
			<a href="#">Item 1</a>
			<ul class="vertical menu">
				<!-- ... -->
			</ul>
		</li>
		<li><a href="#">Item 2</a></li>
	</ul>
</div>
```

## Second benefit
By using the help function Foundation.Box.GetDimensions following problem is solved:
With these changes, menu items can be defined with different sizes (paddings, margins, line-heihgt ... ), for example: the back element. The old version calculates the height of the menu based on the height of list item multiplied with the number of elements, and then the height of the menu is calculated incorrectly.

Owlbertz:
@monosize  Can you get rid of the conflicts?

kball:
Given the current conflicts and the fact this changes the settings file, I'm going to change this target to 6.3.  @monosize can you take a look at bringing the PR up to date with develop so we get rid of conflicts?

kball:
@monosize are you going to be able to take a look at this?

monosize:
yes, i will take a look now.

monosize:
@kball i have create a new PR for this feature, see here #9239

PR #8129, opened 2016-02-10
Don't prevent submitting forms with hidden inputs

koronowicz:
In my project I've dynamic forms, they can have 0 or N visible inputs. They always have one hidden input. I found a problem with submitting form where there is only a hidden input. 

Maybe you can change it?


kball:
The problem makes sense - Abide should not prevent you from submitting forms with a single hidden input... however, I'm not sure the solution is right... I don't know that applying validation to hidden inputs make sense.  I would prefer a solution that instead modifies the validate method to work appropriately even with 0 inputs.

kball:
Closing this based on the above comment.  I think applying validation to hidden inputs will cause other unexpected issues, lets instead allow the validate method to work with 0 inputs

PR #8143, opened 2016-02-12
Change for flex-block-grid AND Source Ordering

Gregor-Agnes:
new flex-block-grid breaks Source-Ordering. This fixed it.


gakimball:
Can you replace `flex-grid-order()` with `flex-order()`? The mixins for flexbox alignment and source ordering are being made into more generic mixins.

Gregor-Agnes:
Yes, I changed this, it still works.

PR #8152, opened 2016-02-12
Update clearfix mixin with flexbox fix

### LINKED ISSUE #8148: Problem with clearfix in flexbox
When using the clearfix mixin on an element with ``display: flex;`` you get a small 1px border left and right in Safari and Chrome.

The problem is [documented here](http://danisadesigner.com/blog/flexbox-clear-fix-pseudo-elements/) with a nice fix:

```css
.clear-fix:before,
.clear-fix:after {
  flex-basis: 0;
  order: 1;
}
```

Adding those two lines to [the mixin](https://github.com/zurb/foundation-sites/blob/develop/scss/util/_mixins.scss#L135) will fix the issue without harming current usage in non-flexbox layouts and only add a few lines to the compiled css.

maggo:
See http://danisadesigner.com/blog/flexbox-clear-fix-pseudo-elements/

Fixes #8148

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -129,12 +129,16 @@
 }
 
 /// Applies the micro clearfix hack popularized by Nicolas Gallagher. Include this mixin on a container if its children are all floated, to give the container a proper height.
+/// The clearfix is augmented with specific styles to prevent borders in flexbox environments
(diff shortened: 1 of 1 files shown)


gakimball:
Thanks!

PR #8175, opened 2016-02-16
character html via title

nokser:
Now accept character html via title


gakimball:
Duplicate of #8075, and I have [a suggestion](https://github.com/zurb/foundation-sites/pull/8075#issuecomment-196425337) for how to avoid XSS attacks with a feature like this.

PR #8192, opened 2016-02-18
Update MediaQueries to initialize if not initialized already.

### LINKED ISSUE #7554: [Tooltips] Instantiating tooltips recommended way doesn't seem to work (Meteor)
We are following the documentation on Atmosphere and not using `$(document).foundation()`, but instead are instantiating plugins using the Foundation plugins API. This works fine with our Reveals, Accordions and several other plugins, however, is not working with Tooltips. The tooltips will work when you use `$(document).foundation()`, but we need more control over the plugins. Below is how we are using tooltips:

``` html
<span id="messageTooltip" data-tooltip aria-haspopup="true" class="has-tip top" data-disable-hover='false' tabindex=1 title={{message}}>{{shorten message}}</span>
```

``` js
Template.generalTasksTable.onRendered(function () {
  this.tooltip = new Foundation.Tooltip($('#messageTooltip'));
});

Template.generalTasksTable.onDestroyed(function () {
  let tooltip = this.tooltip;
  if (tooltip) {
    tooltip.destroy();
  }
});
```

We have also tried targeting the `.has-tip` class as well, but no luck with that also. Just to make sure we aren't doing anything odd with Meteor, I have tried `var tooltip = new Foundation.Tooltip($("#messageTooltip"));` in the Chrome console, but that yielded no tooltips as well.

Thanks for any help!

merlinstardust:
Fixes #7554

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -4,6 +4,11 @@
 
 // Default set of media queries
 const defaultQueries = {
+  small : 'only screen and (min-width: 0em)',
+  medium : 'only screen and (min-width: 40em)',
+  large : 'only screen and (min-width: 64em)',
+  xlarge : 'only screen and (min-width: 75em)',
+  xxlarge : 'only screen and (min-width: 90em)',
   'default' : 'only screen',
   landscape : 'only screen and (orientation: landscape)',
   portrait : 'only screen and (orientation: portrait)',
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -76,6 +76,10 @@ var MediaQuery = {
    * @returns {String|null} - The media query of the breakpoint, or `null` if the breakpoint doesn't exist.
    */
   get(size) {
+    if (! this.queries.length) {
+      this._init();
+    }
+
     for (var i in this.queries) {
       var query = this.queries[i];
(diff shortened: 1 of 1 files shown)


merlinpatt:
Travis CI has 27 passing but is giving this as the error

```
/home/travis/build/zurb/foundation-sites/node_modules/gulp-scss-lint/node_modules/bluebird/js/main/promise.js:677
            throw e;
                  ^
Error: You need to have Ruby and scss-lint gem installed
```
How do I fix this?

gakimball:
Don't worry about that, it's on our end.

gakimball:
When the init function runs, it should be taking the user-provided `$breakpoints`, and merge them with `default`, `landscape`, `portrait`, and `retina`. By adding fallbacks for `small` through `xxlarge` in that defaults object, will it still work that way?

merlinpatt:
It will now :-)

SierraGolf:
+1

kball:
Looks good to me. Restarted the travis build.  If it passes will merge.

kball:
@merlinpatt I ended up having to revert this (see #8547) as it caused a bunch of issues with responsiveness, I think due to breaking assumptions in _getCurrentSize.  I didn't see a quick fix, but that doesn't mean there isn't one.  Please take a look at this again and see if you can resolve.  Thank you!

(cc @gakimball)

PR #8197, opened 2016-02-18
Some minor addings for feature

SalahAdDin:
Ok, i add some minor fixes for update features.

### DIFF
--- a/scss/components/_badge.scss
+++ b/scss/components/_badge.scss
@@ -26,6 +26,10 @@ $badge-minwidth: 2.1em !default;
 /// @type Number
 $badge-font-size: 0.6rem !default;
 
+/// Default border radius size for the badges.
+/// @type Number
+$badge-border-radius-size: ($global-radius * 2);
+
 /// Generates the base styles for a badge.
 @mixin badge {
   display: inline-block;
@@ -33,7 +37,7 @@ $badge-font-size: 0.6rem !default;
   min-width: $badge-minwidth;
   font-size: $badge-font-size;
   text-align: center;
-  border-radius: 50%;
+  border-radius: $badge-border-radius-size;
 }
 
 @mixin foundation-badge {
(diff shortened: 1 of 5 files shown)


andycochran:
:-1: I do not think this should be merged. 

It's a good idea to limit pull requests to one specific change/idea. There are quite a few disparate ideas here, most of which are use cases that are too specific. A lot of these can/should be handled in your specific app. 

For example, it's very easy to add this custom style to your app: 
```
legend {
  font-weight: $global-weight-bold;
}
```

And Foundation tries to stay away from `!important` declarations. 

And "*" is not a universal way to indicate "required." But you can easily do this in your app.

SalahAdDin:
Yes, the `!important` was solved change the components order.

gakimball:
Sorry for not getting to this sooner! I had a comment written for this but I'm not sure where it went.

Like @andycochran said, we don't encourage pull requests like this, that bundle multiple changes into one request. We want to be able to assess each change individually.

However, after reviewing this internally, unfortunately we won't be pulling in any of these changes right now. Let's break it down:

- Badge radius: Badges are meant to be round, so we don't need a variable to adjust this.
- `!important` on `.float-center`: this may be fixed by moving `@include foundation-float-classes` to the bottom of your `app.scss` file. In Foundation 6.2 we updated our recommended order for CSS exports to put utility classes last, lessening the need for `!important`. See [here](https://github.com/zurb/foundation-sites-template/blob/master/scss/app.scss).
- `<legend>` font weight: we don't add variables for CSS properties we don't set by default. We aim for as few settings variables as possible, so you can spend less time learning the framework and more time writing CSS.

The `.required` style is maybe a little specific, but it's no more harmless than `.help-text`, which is basically just small italicized text. If you want to submit a new PR with the `.required` class we can bring that in.

Let me know if you have any questions :)

PR #8198, opened 2016-02-18
Update _text.scss

piotros:
Improved modularity. Added missed $form-spacing value.

### DIFF
--- a/scss/forms/_text.scss
+++ b/scss/forms/_text.scss
@@ -66,6 +66,10 @@ $input-number-spinners: true !default;
 /// @type Border
 $input-radius: $global-radius !default;
 
+/// Size of form spacing.
+/// @type Number
+$form-spacing: rem-calc(16) !default;
+
 @mixin form-element {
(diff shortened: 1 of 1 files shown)


gakimball:
The `$form-spacing` variable is in the base forms file: <https://github.com/zurb/foundation-sites/blob/develop/scss/forms/_forms.scss#L11>

gakimball:
Sorry, I realize by saying "improve modularity" you mean to have each forms file be self-contained. However, `$form-spacing` is used by multiple form files, so you'd have to duplicate the variable between each one, but that doesn't work either because then you have multiple modules all trying to assign values to a variable with the same name.

piotros:
You're right that I wanted to use _text.scss as separated dependency.
(https://github.com/piotros/miblo/blob/8cb8aa8570daa377eb08c1e011147821ed7fb3c8/src/components/entry-add/_entry-add.scss#L4)

Now I realize that instead of:
`@import '~foundation-sites/scss/forms/text;'`
i can use:
`@import '~foundation-sites/scss/forms/forms;'`

Thank you! ;)

Issue #7812, opened 2016-01-08
DropdownMenu.defaults.clickOpen seems broken for Chrome


imaia:
Hello folks;

I have a dropdown menu and a clickOpen = true configuration here within my website. My development went on using firefox. Everything was fine, until I test it with chrome. It happens that, when I click in the dropdown anchor, using chrome, it won't follow the link unless I click once more, which is quite weird. Seems like a bug to me. I'm using firefox 43 and chrome Versão 47.0.2526.106 (64-bit).

zurbchris:
Can you post the options used for this menu or a link to a live page/codepen? Seems like you're getting the expected interactions on chrome and firefox is doing something strange.

The clickOpen option will retain the open state of the menu until you click outside it  (if closeOnClick = true, default settings) or the parent link again, on browsers with an `ontouchstart` event will cause a slightly different action, as a failsafe for mobile devices.

JeremyEnglert:
No reply. Closing.

IamManchanda:
@imaia Is it still an issue... please let us know? 
We are closing it for now but feel free to reopen!

Issue #7668, opened 2015-12-23
[Compatibility] Click behavior of the title-bar/Dropdown Menu in IE Mobile 


cogtail:
**How can we reproduce this bug?**
1. Use IE 11 in Windows Phone or Windows Tablet
2. Open Drop-down Menu or Title-bar
3. Click on a Link in this open sub menu

The problem is, that you have to "click" (or better: to touch) two times until the link works, I reproduced this bug not only with my Foundation 6 installation but also with http://jointswp.com/demo/title-bar/ 

The "two touch" bug could be used as a feature only for parent items, when I touch first to open the dropdown menu and then to choose a link in the open sub menu or choose the activated link of the parent item (extra doubleTapToGo-Feature?). It should behave like http://www.smartmenus.org/.

flobauer:
+1

flobauer:
I debugged a little -> the reason lies in the foundation.dropdownMenu.js File and there in the Eventlistener.

I am currently testing a solution, but the Event listener doesn't react on the touchstart event...

flobauer:
Ok, appearantly some IE Versions behave different when the aria-haspopup attribute is given: https://msdn.microsoft.com/en-us/library/jj152141(v=vs.85).aspx

I ended up checking for IE and if so, enforcing a redirect:
```php
if(hasSub){
    //...
}else{ 
    if (isIE) {
      window.location.href = $(e.target).attr('href');
    }
    return; 
}
```
I added this in the file foundation.dropdownMenu.js file in the Event Handling section on the `click.zf.dropdownmenu touchstart.zf.dropdownmenu` event

cogtail:
Unfortunately, there is currently no fix for the TOUCH twice to follow links bug.

flobauer:
I guess, this won't be fixed anytime soon...

ncoden:
@flobauer Hi, sorry for the late reply. We recently released `v6.3.0`, did you tried with this last version ? I remember we fixed some issues with the Dropdown menu.

Poke @kball. There is so many issues with the dropdown menu on touch devices. See #9484, #9464, #9322, #8843

kball:
Yeah, there are a ton.  I don't have a bunch of time for debugging this week and I'm off the following week but I'll take these on as a priority for 6.3.1 and tackle them first thing in Jan.

JeremyEnglert:
This is no longer an issue on devices I tested. Please re-open if anyone notices otherwise.

PR #8222, opened 2016-02-23
Fix for click twice to follow links #7812 #7668

joey-i:
I've experienced same issues like #7812 #7668.
I was able to fix it by adding following code to check event was made from mouse or not.

```
//prevent click twice to follow link for touche enabled PC browsers.
if(e.type === 'click'){
  hasClicked = true;
}
```

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -81,6 +81,11 @@ class DropdownMenu {
             hasClicked = $elem.attr('data-is-click') === 'true',
             $sub = $elem.children('.is-dropdown-submenu');
 
+        //prevent click twice to follow link for touche enabled PC browsers.
+        if(e.type === 'click'){
+          hasClicked = true;
+        }
+
         if (hasSub) {
           if (hasClicked) {
             if (!_this.options.closeOnClick || (!_this.options.clickOpen && !hasTouch) || (_this.options.forceFollow && hasTouch)) { return; }


gakimball:
Reviewing this now, do you know if it will affect any other touch browsers, like Mobile Safari or the Android Browser?

FFIN:
Well, I can say definitely no affect on mobile safari, and I confirm no affect on Sony Xperia OS 5.0.2 android browser.
But Android behavior may vary device to device, so if you have wide range of android devices for testing, please double check.

kball:
@colin-marshall it looks like you've spent some time in this code, can you take a look at this and see if it seems right?

colin-marshall:
@kball I just converted the code to ES6, I didn't get into the functionality of the plugin, however I will test this out later today.

kball:
@colin-marshall did you end up having a chance to check on this?

colin-marshall:
@kball sorry I forgot about this one! I can't reproduce this on any of my devices, which is weird because I have definitely seen this before.

@zurb/yetinauts can anybody else test this out?

designerno1:
@colin-marshall 
`can't` reproduce this on a w10 touch notebook in chrome, firefox, Safari
only thing is in edge and IE11 when touching the link (not when clicking) if i touch the first link the menu opens then if i touch the second link i have to touch it twice to follow
but with the PR its worse then the menu opens and close immediately

colin-marshall:
@designerno1 thanks for checking it out. I'll see if I can reproduce in a Windows VM.

designerno1:
@colin-marshall 
i think it's because hasTouch returns false on my device, and thusthe browser emulates a hover event

cogtail:
Unfortunately, there is currently no fix for the TOUCH twice to follow links bug. You have to touch twice on every windows phone and (I suppose) on every windows tablet.

kball:
It looks like from @designerno1 this PR breaks at least as much as it fixes, so closing for now.  Please feel free to reopen if you are able to extend it to not break edge/ie11

PR #8229, opened 2016-02-24
Add JSPM config

glen-84:
This will make it easier to install and use Foundation Sites with JSPM. Let me know if you have any questions.

http://jspm.io/

/cc @guybedford

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,9 +88,9 @@
   },
   "jspm": {
     "format": "global",
-    "meta": {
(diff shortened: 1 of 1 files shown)


gakimball:
This would be nice to slip into 6.2 on Friday, could you open a new pull request that targets the `6.2-stable` branch instead? That's where we're making final adjustments to the patch. Thanks :)

glen-84:
Done – closing.

PR #8233, opened 2016-02-24
Add JSPM config

glen-84:
Replaces #8229, as requested.


gakimball:
Thanks, appreciate it!

PR #8243, opened 2016-02-25
Update off-canvas.md

themestechnology:
Update shows a working example for Foundation 6.x


gakimball:
Is the intent here to show a full code sample with all of the pieces in one place?

themestechnology:
Yes that is correct When I transitioned my theme from 5.x to 6.x things were not quite strait forward and I really think anyone wishing to use the Off-Canvas in 6.x could benefit from this code.

gakimball:
Made some edits and then merged, thanks!

Issue #8264, opened 2016-02-27
change event for sliders


n0o0bSina:
How can i handle foundation sliders changes not moves?
I need this for ajax calls.

Deckluhm:
I'm not sure I understood your question but this might be what you are looking for: http://foundation.zurb.com/sites/docs/slider.html#js-events

n0o0bSina:
@Deckluhm You'r right but in that page there is only `moved.zf.slider` event i need `changed.zf.slider`.
`moved.zf.slider` is equal of http://api.jqueryui.com/slider/#event-slide
but i need http://api.jqueryui.com/slider/#event-change
this event is very important if you need sliders value for ajax calls.

gakimball:
Merged #8286 which implements this. Thanks @Owlbertz!

ZeeAzmat-qbx:
I have the same problem. I need **changed** event instead of **moved**, and tried using **changed** but it doesn't work. Can you please help me on this?

Owlbertz:
What version of Foundation are you using? PR #8268 should add the changed event. This is not in 6.2 yet I believe, but 6.2.1 will come up later this week iirc, which should have this implemented.

ZeeAzmat-qbx:
I am using version 6, and what i am looking for is the somewhat RELEASE
method for slider. So instead of "moved" method i can use "released"
method, whenever i release the slider event gets triggered and i can make
the AJAX request on that value.

Owlbertz:
The idea of what was merged with the PR mentioned is an event that will trigger when the slider has not been moved for a given time, by default 500ms. Would that fit your needs?

ZeeAzmat-qbx:
Yeah exactly i was looking for something like this. But that didn't fulfill
my needs.

ZeeAzmat-qbx:
The code i am using is:

$('#dropdown--price-slider').on('changed.zf.slider', function(e, handle) {
console.log ("changed"); });

  $('#dropdown--price-slider').on('moved.zf.slider', function(e, handle) {
    var currentAttribute = handle.attr('aria-valuenow');
    var parentValueLabel =
$(this).parent('.dropdown-pane').find('.slider-value');

    parentValueLabel.each( function() {
      $(this).text(currentAttribute);
    });
    console.log (currentAttribute);
  });

In second function i am updating the values as per the slider moves,  and
in the first function i am looking to get the value of slider when it stops
and make the AJAX request.

Owlbertz:
[...]

PR #8268, opened 2016-02-28
Added changed.zf.slider event

Owlbertz:
First idea to give the Slider something as requested in #8264.
The event fires when the value has not been changed for `changedDelay` milliseconds (default is 500).

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -532,7 +532,7 @@ Slider.defaults = {
    */
   invertVertical: false,
   /**
-   * Milliseconds before the `change .zf-slider` event is triggered after value change. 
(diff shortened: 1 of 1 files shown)


gakimball:
Can you clarify the difference between when the `moved` vs. `change` event fires? For example, would `change` fire if a slider was adjusted by hand *or* with a text field?

Owlbertz:
`changed` (I updated the variable description to match this) will fire regardless of the adjustment, whether it's via dragging, via text input or via these fancy little HTML5 number-input arrows. The timeout is used to only fire when the user has finished adjusting.

gakimball:
Thanks for the clarification :)

PR #8269, opened 2016-02-28
Meteor package - Foundation 6.2 adjustments

juliancwirko:
Meteor package changes:
- removed problematic Motion-UI dependency (there is also PR for Motion-UI Meteor package: https://github.com/zurb/motion-ui/pull/73 )
- added missing .scss files

There is needed `meteor publish` asap, because otherwise it will not work well. 

The new version for Meteor is `6.2.0_1` which is a proper versioning for Meteor packages. It is still foundation 6.2.0 but with fixes for Meteor.


gakimball:
Thanks for putting this together!

gakimball:
Oh actually, can you open a new PR that compares against `master` instead? That's our stable branch, so if I'm going to redeploy, it should be there.

juliancwirko:
done: https://github.com/zurb/foundation-sites/pull/8290

Issue #8270, opened 2016-02-28
[Orbit Doc] Code Example not HTML5 confirm


jdoubleu:
There are a few mistakes I found in the code example when I validated my site. I don't know if you want to be fully HTML5 confirm but this may should be fixed. Just have a look at it.

**Doc Version:** 6.2

The code example from http://foundation.zurb.com/sites/docs/orbit.html has two mistakes:
1. A figcaption must be placed inside a figure tag (see http://w3c.github.io/html/single-page.html#the-figcaption-element)
2. button tags can't be placed inside a ul (see http://w3c.github.io/html/single-page.html#the-ul-element)

**To reproduce:**
Check out the given w3c links and/or validate the code snippet with a validator (e.g. http://validator.w3c.org)

I created a pull request on this (see https://github.com/zurb/foundation-sites/pull/8271).
The orbit slide buttons still work but I'm not sure about the figure tag. The figure tag get's a default margin  of `1em 40px;` which could break the slides. Haven't tested it yet.

Thanks for your attention and your review.

jdoubleu:
Update after some discussion in the PR:

As mentioned [here](https://github.com/zurb/foundation-sites/pull/8271#issuecomment-207775399]) there's already a class `orbit-caption` which prevents the `figcaption` to break the visual.

DanielRuf:
The current docs seem to be fine.

![foundation zurb com_sites_docs_orbit html 1](https://user-images.githubusercontent.com/827205/36341091-2acb832c-13e9-11e8-9bdb-729f687322b7.png)

jdoubleu:
@DanielRuf you're right. For version 6.4.3 this seems to have changed (https://foundation.zurb.com/sites/docs/orbit.html).

I don't know if this issue should be solved for older version (specific 6.2).

DanielRuf:
> I don't know if this issue should be solved for older version (specific 6.2).

If you mean the (online) docs: there is just one docuentation for v5 and one for v6 afaik.

jdoubleu:
Okay.

So closing this issue because it is no longer relevant.

PR #8271, opened 2016-02-28
Updated Orbit documentation: corrected invalid HTML5 code examples

jdoubleu:
A given code example isn't HTML5 valid.

For more, detailed information have a look at https://github.com/zurb/foundation-sites/issues/8270 please.

### DIFF
--- a/docs/pages/orbit.md
+++ b/docs/pages/orbit.md
@@ -19,25 +19,25 @@ Orbit doesn't automatically generate any HTML for you, giving you the flexibilit
     <button class="orbit-next"><span class="show-for-sr">Next Slide</span>&#9654;&#xFE0E;</button>
   <ul class="orbit-container">
     <li class="is-active orbit-slide">
-      <figure>
+      <figure class="orbit-slide">
         <img class="orbit-image" src="assets/img/orbit/01.jpg" alt="Space">
         <figcaption class="orbit-caption">Space, the final frontier.</figcaption>
       </figure>
     </li>
     <li class="orbit-slide">
-      <figure>
+      <figure class="orbit-slide">
         <img class="orbit-image" src="assets/img/orbit/02.jpg" alt="Space">
         <figcaption class="orbit-caption">Lets Rocket!</figcaption>
       </figure>
     </li>
     <li class="orbit-slide">
-      <figure>
+      <figure class="orbit-slide">
         <img class="orbit-image" src="assets/img/orbit/03.jpg" alt="Space">
         <figcaption class="orbit-caption">Encapsulating</figcaption>
       </figure>
     </li>
     <li class="orbit-slide">
-      <figure>
+      <figure class="orbit-slide">
         <img class="orbit-image" src="assets/img/orbit/04.jpg" alt="Space">
         <figcaption class="orbit-caption">Outta This World</figcaption>
       </figure>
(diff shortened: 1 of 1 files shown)


gakimball:
Does it disrupt the existing styling of Orbit at all?

jshdoubleu:
Yes, as expected, the figure disrupts the styling (see screenshot).
By default the figure tag has a margin and so it breaks a slide (and every image).

The buttons still work.

bildschirmfoto 2016-03-01 um 19 17 14

Please compare:
This is how it is with figure:
bildschirmfoto 2016-03-01 um 19 20 30
This is how it should be/is without figure:
bildschirmfoto 2016-03-01 um 19 21 29

gakimball:
Hm, that makes it tricky then. The CSS would need some changes, but we don't want to break the visual styling for existing users.

If you can make the CSS work the same the old way and this new way, then we can work with that.

jshdoubleu:
Okay, I'll try to fix it. Which branch should I take? `6.2-dev` or `develop` ?

gakimball:
Do it on `develop`. We don't need the 6.2 branches now that 6.2 is out—I'll get rid of them soon.

kball:
@jshdoubleu did you end up taking on the css changes?  Thanks

jshdoubleu:
Sorry for the delay.

I found out, that the broken doc was caused by this commit 097677af1431f2f0616aa1de138bf9e90de1abc0.
There are already classes to prevent this "bug". Adding `orbit-figure` class to the figure tag removes the margin. But this information is still missing in the doc.

However this is not valid HTML5.

Owlbertz:
Hi,
I took a look and noticed two things. First, this PR targets `master`, but should go into `develop`. Second, the layout is broken for me:
image
The focus currently lies on the slide. Notice that it's far too big and that the buttons are pretty misaligned.
(Ubuntu 14.04, Chrome)

Can someone else double check that?

Owlbertz:
@jdoubleu Any update/comment on that?

kball:
@Owlbertz purely docs updates can target master. However I'm a little confused - is this only an update to the example?  Or are there updates that need to happen to the actual component scss as well?

jdoubleu:
To clarify I stumbled about the documentation of the orbit slider after my IDE marked the html invalid.

There were `button` tags inside a `ul` and the `figcaption` tags weren't placed inside a `figure` tag.

I cleaned up and corrected the markup and it broke. The buttons still worked properly but the `figcaptions` had a margin or padding which would crash the slider.

So I thought there must be a change in the scss. While trying to fix this problem I mentioned there was already a css class named `orbit-caption` which removes the margin. This class wasn't documented anywhere.

I updated the documentation to have valid html and use correct classes. I also tried out the changes locally and it worked for me.
Probably there were changes (e.g. the class was removed) while this issue has been reported.

Belonging issue: https://github.com/zurb/foundation-sites/issues/8270

I'm going to make further tests but haven't got an eye on this for a while.

kball:
@jdoubleu when I check this out, the docs example is visually broken as noted by @Owlbertz.  Can you take a look again and address?  Thanks

kball:
@jdoubleu were you able to take a look at the breakage noted above?  I'd still love to get this resolved

jdoubleu:
Okay for me it breaks too. But this has something todo with the `figcaption` being positioned absolute.
I'm investigating this and propose a fix/correction.

jdoubleu:
Closing because of https://github.com/zurb/foundation-sites/issues/8270#issuecomment-366439343 (https://github.com/zurb/foundation-sites/issues/8270#issuecomment-366457521)

PR #8280, opened 2016-02-29
proper loop iteration for javascript array ensures extensions to Array.prototype don't break us

pelly:


gakimball:
Can you explain in more detail what issue this fixes?

js/foundation.util.mediaQuery.js
```diff
@@ -84,7 +84,7 @@ var MediaQuery = {
   _getCurrentSize: function() {
     var matched;
 
-    for (var i in this.queries) {
```
pelly: So if Array.prototype is modified (I think either Ember or React does this), a for loop as so:

`for (var i in someArray) { ... }`

will have **i** iterate through the indices of the array, in addition to the property names added to the prototype. Then the `someArray[i]` will return undefined, so in this case `query` will be undefined, and calling `query.value` will fail. 

This occurred on me when I had a project with React, Ember and Foundation integrated. Not sure which (Ember or React) modifies Array.prototype, but it's probably good to be defensive and safe in iterating through the array.

pelly:
This might be affecting other code paths as well, but this was cropping up as a result of calling `$(document).foundation();` - the page wouldn't even load b/c of failure calling this method in mediaQuery

js/foundation.util.mediaQuery.js
gakimball: @pelly Got it, thanks :) A little annoying to have to work around other libraries that modify built-in prototypes, but it is what it is.

PR #8286, opened 2016-02-29
Util scss not well imported from settings

dfvalero:
We are not calling to the correct file.

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -40,7 +40,7 @@
 //  35. Tooltip
 //  36. Top Bar
 
-@import 'util/util';
+@import '../util/util';
 
 // 1. Global
 // ---------


gakimball:
The file assumes two things:

- It's sitting in your project-specific Sass files, not in the `foundation-sites/scss/settings` folder. That's just where we house it—but in a real project, a copy of that file should be with your project assets.
- You have a path to the Foundation Sass set as an import path in your Sass compiler. For example, in the ZURB Template, we configure `gulp-sass` to have an `includePaths` option of `bower_components/foundation-sites/scss`.

With that in mind, the existing path is correct.

kunKun-tx:
i guess it will be better to explain this in the doc, [(http://foundation.zurb.com/sites/docs/sass.html)]

PR #8290, opened 2016-02-29
zf 6.2 - meteor package adjustments

juliancwirko:
Meteor package changes:
- removed problematic Motion-UI dependency (there is also PR for Motion-UI Meteor package: https://github.com/zurb/motion-ui/pull/73 )
- added missing .scss files

There is needed `meteor publish` asap, because otherwise it will not work well. 

The new version for Meteor is `6.2.0_1` which is a proper versioning for Meteor packages. It is still foundation 6.2.0 but with fixes for Meteor.


gakimball:
Thanks!

gakimball:
Published Meteor package: 26698d8aed59787705ca4246bfd3e9e12e7b852b

PR #8317, opened 2016-03-02
cursor pointer need some love for disabled items

flexbox:
Hello there,

If there is `pointer-events: none;` for buttons we can't use tooltips on buttons :(
Here is an example [http://codepen.io/flexbox/pen/ONVVwa](http://codepen.io/flexbox/pen/ONVVwa)


andycochran:
:+1:

kball:
Looks great, thanks

Issue #8257, opened 2016-02-27
[6.2] breakpoint names hardcoded in _breakpoints.scss?


erutan:
I'm not using default classes for much, using F6 semantically with SASS. I had made some custom breakpoints that were (to me) more clearer than small, medium, etc. I'm now getting the following error when compiling F6 (via gulp-sass from the zurb template) that I didn't have in previous versions.

```
Error in plugin 'sass'
Message:
    bower_components/foundation-sites/scss/util/_breakpoint.scss
Error: Your list of breakpoints (defined in $breakpoints) must include one named "small" and one named "medium".
        on line 22 of bower_components/foundation-sites/scss/util/_breakpoint.scss
>>   @error 'Your list of breakpoints (defined in $breakpoints) must include one 
   --^
```

There's an easy fix around this obviously, but it'd be nice if things were a bit more agnostic like before. :)

gakimball:
We did this because there are many places in the framework where we use the word "small", and allowing that name to change was opening us up to bugs. So, we introduced a new rule that you have to have a breakpoint defined as `small: 0`.

If you think there's a way around that, we can look at a solution.

erutan:
I can live with small (I was using single-column) and medium (to replace my multi-column) - I generally don't just have typical device + wide breakpoints but create a new semantic one whenever I have a use for it (e.g. collapse-cards, breakout-info-to-columns, etc).

I'd rather not do it (and since I generally don't output a lot of CSS classes from Foundation itself aside I guess I hadn't run into those issues) but it's not a huge deal to be forced to use them. Just a little jarring to come across (on top of not being able to use breapoint+singularity for layout) from using it as a modular SASS library (generally extending or importing) vs toss a bunch of classes into markup perspective.

I haven't dug enough into the guts of F6 to give a technical solution - I've just been using it on one prototype and haven't had the need to.

andycochran:
@gakimball Would something like a `$zero-breakpoint-name` variable help? 

e.g.: 
```scss
/// The name of mobile-first styles (styles without a media query) and the class for styles on screens smaller than the lowest non-zero breakpoint. 
/// @type String 
$zero-breakpoint-name: small !default;
```

erutan:
I was thinking of having assignable values as well. There'd have to be a little explanation of them needing to also be in the SASS map, but it seems reasonable. Medium is probably the most perplexing one, as it's really more like "we're beyond needing a layout for a phone" but there are often some breakpoints between that and small.

[...]

PR #8318, opened 2016-03-02
possible solution for #8257

DaSchTour:
taking first breakpoint and check if smallest and define it as zero-breakpoint

### DIFF
--- a/scss/components/_dropdown-menu.scss
+++ b/scss/components/_dropdown-menu.scss
@@ -124,7 +124,7 @@ $dropdownmenu-border-width: nth($dropdownmenu-border, 1);
     }
 
     @each $size in $breakpoint-classes {
-      @if $size != small {
+      @if $size != $-zf-zero-breakpoint {
         @include breakpoint($size) {
           &.#{$size}-horizontal {
             @include dropdown-menu-direction(horizontal);
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -75,12 +75,12 @@ $mediaobject-image-width-stacked: 100% !default;
     }
 
     @if $global-flexbox {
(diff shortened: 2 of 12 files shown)


gakimball:
Thanks for putting this together! My feedback:

- Name the variable `$-zf-zero-breakpoint`. The prefix emphasizes that it's a private variable—we don't need developers using it in their own code.
- You'll need to replace *every* instance of `@include breakpoint(small)` in the framework with `@include breakpoint($-zf-zero-breakpoint)`, so that if someone renames `small` in their breakpoint list, all of the internals of the framework are able to switch over to the new name.

DaSchTour:
Thanks for the feedback. I hoped I already found all references to small, but I'll check again. I'll make the changes as soon as possible.

DaSchTour:
I change the variable following your feedback and replaced all usages of small breakpoint. I wasn't 100% sure about one place. It's _rows.scss:83

DaSchTour:
Is there anything more to do?

gakimball:
Nope, looks good. Thanks for putting it together!

proimage:
So what's the quick-and-dirty guide to implementing this change in our bleeding-edge code? Do we just add something like this to `settings.scss`:

```
$-zf-zero-breakpoint: 'small'; // or 'xsmall' or 'tiny' or 'smallest' or whatever
```

gakimball:
@proimage That will probably work.

proimage:
Thou assumest too muchly - I roll Foundation SCSS manually. :p

For the record, yes, after downloading the latest `develop` branch and using the `scss` folder therein, adding that one line above to my `settings.scss` appears to have brought things back to the way they were in 6.1.x. Much grass!

Issue #7070, opened 2015-11-20
[F6][Drilldown] Height consumption calc not working


aibrean:
When there are very long nested submenu (`menu vertical submenu is-drilldown-submenu is-drilldown-sub`) in the drilldown, the bottom gets cropped off.

Seems like the `_getMaxDims` from JS isn't working right or height needs to be adjusted elsewhere (it's set to 100% on `.is-drilldown-sub`) to account for long titles that may wrap multiple lines.

(stubbed due to sensitive information)
![image](https://cloud.githubusercontent.com/assets/6740252/11307356/a2aba1d2-8f85-11e5-9db8-301df11bd9be.png)

gakimball:
Hey @aibrean, sorry for not getting to this sooner! I'm working on recreating your issue, but I'm not able to. I made [this test case](https://github.com/zurb/foundation-sites/blob/develop/test/visual/drilldown/long-drilldown-menu.html), a simple drilldown with a ton of items within a sub-menu. The drilldown expands fine, no matter how many items are added.

It's possible the issue has been fixed in a newer version of Foundation, but just in case, I need your help recreating it. If you want to fiddle with the example I linked, you can clone the foundation-sites repo, set it up, and then run `npm run test:visual`. Then you can look at this page yourself and mess around with it.

Thanks for your help! :)

aibrean:
The issue is not the amount of items, but the _length_ of the items (i.e. item title).

gakimball:
Thanks for the clarification—I was able to recreate the issue. I got most of the way there but I've run into an issue with `.getBoundingClientRect()`, so we'll need to do a bit more testing. Here's the pull request for the fix: #8320

andycochran:
@aibrean Can you still reproduce this issue with the latest version? If so could you please make a [codepen](http://codepen.io/rafibomb/pen/xVVGOB) and reopen this issue?

PR #8320, opened 2016-03-02
Drilldown height calculation with long titles

gakimball:
The Drilldown plugin reads the height of each menu inside of it, and sets the height of the whole menu to match the tallest one. However, the calculated height will be wrong if any menu items differ in height, as reported in #7070.

This is because the total menu height is calculated by taking the height of the _first_ item in the list, and multiplying it by the number of items in the menu.

This change instead just calculates the height of the entire menu all at once, so it will be accurate.

However, the call to `.getBoundingClientRect()` manages to be inaccurate. In the test case I put together, the height calculated is always off by 10 pixels, and I'm not sure why.

To finish this fix, we need the height calculated accurately.


scss/components/_drilldown.scss
```diff
@@ -39,7 +39,7 @@ $drilldown-background: $white !default;
     top: 0;
     #{$global-left}: 100%;
     z-index: -1;
-    height: 100%;
+    // height: 100%;
```
kball: Yeah, this appears to be responsible for https://github.com/zurb/foundation-sites/issues/9368.  Looking into 9239 now.

guyvanbael:
Not really working correctly when one of the submenus are longer.
schermafbeelding 2016-11-23 om 13 02 48

PR #8336, opened 2016-03-04
Doc: updated flexbox setup

tomByrer:
Seems switch was added a version or 2 ago: https://github.com/zurb/foundation-sites/blob/develop/scss/foundation.scss#L52
Worked for me!

I made the old method more terse; still has use-case.


tomByrer:
I also wonder if the [WiKi](https://github.com/zurb/foundation-sites/wiki/Upgrading-to-Foundation-6.2) should be updated also?

PR #8369, opened 2016-03-08
Fixed keyboard access for accordion menu.

### LINKED ISSUE #8352: [Accordion Menu] not tabbable/keyboard navigatable

1. Tab through webpage tab sequence
2. focus will hit first item of accordian-menu, at which point, tab (and backwards tab/alt+tab) no longer work.
3. left/right arrow keys will open and close that first level item dropdown, however up/down arrows do not work, so keyboard can never access the submenu items, by either up/down or tab

functionality of keyboard should match the standard accordian. Aka, tab sequence should not be interrupted, and tab key should continue to navigate to the (other) visible links/focus points. up/down arrows should also enable user to navigate the menu items, in same manner that a user an do so in standard accordian.

**Test case**

http://foundation.zurb.com/sites/docs/accordion-menu.html
http://foundation.zurb.com/sites/docs/accordion.html

Owlbertz:
Set tabindex before setting focus.
Fixes #8352.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -127,6 +127,7 @@ class AccordionMenu {
           return;
         }
       });
+        console.log('Keydown!');
       Foundation.Keyboard.handleKey(e, 'AccordionMenu', {
         open: function() {
           if ($target.is(':hidden')) {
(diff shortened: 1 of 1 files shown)


gakimball:
Going to put together a test case and then I'll merge.

gakimball:
I'm not able to trigger actual links in the menu with the Enter or Space keys, can you take a look?

Owlbertz:
Yeah, I noticed this a while back already. I think this is related on how the Menus are designed, since the trigger is the `li` and not the `a`. I adjusted the selectors and this should be fine now.

gakimball:
Thanks!

Issue #8204, opened 2016-02-19
[Top Bar] Menu Icon (span v button)


andycochran:
The docs recommend marking up a hamburger menu with… 
```html
<span data-responsive-toggle="responsive-menu" data-hide-for="medium">
  <span class="menu-icon dark" data-toggle></span>
</span>
```

Semantically, shouldn't this be a `<button>` element? 

Downside to using a `<button>`, it gets an outline on click focus. 

http://codepen.io/andycochran/pen/EPzdvG

Lynda333:
They do use `<button>` in the [Responsive Toggle ](http://foundation.zurb.com/sites/docs/responsive-navigation.html#responsive-toggle) section. In the [Top Bar](http://foundation.zurb.com/sites/docs/top-bar.html#advanced) section, it's as you have it. 

Also, the outline doesn't happen in Firefox but does in Chrome.

gakimball:
Makes sense, all of our `.menu-icon` examples should use `<button>`.

You can add the `disable-mouse-outline()` mixin to the CSS to get rid of the outline for mouse users.

gakimball:
@andycochran Are any changes required in the CSS, or just in the examples? And if you want to implement it we'll bring it in :)

andycochran:
@gakimball Should I target 6.2-dev?

andycochran:
@gakimball I think we need to add the `disable-mouse-outline` mixin to the menu's `button` element. Generic `button` elements don't get the What-Input treatment unless they also have the `.button` class. Would it be bad to add What-Input to all `button` elements globally? I can't think of any issues with that, but it's a hefty change I thought might need discussion.

gakimball:
@andycochran Shouldn't be a hefty change, and because `<button>`s are interactive elements it makes sense.

gakimball:
@andycochran If you want to implement this you can target `develop`.

gakimball:
Merged #8370

PR #8370, opened 2016-03-08
.menu-icon & buttons elements

andycochran:
This PR fixes the Top Bar documentation, recommending `<button>` instead of `<span>` for a `.menu-icon`. It also applies the `disable-mouse-outline()` mixin to default button elements—so they don't need the `.button` class in order to make use of [What Input?](https://github.com/ten1seven/what-input).

https://github.com/zurb/foundation-sites/issues/8204#issuecomment-193515849


gakimball:
Awesome, thanks!

SalahAdDin:
:+1:

PR #8387, opened 2016-03-10
[Bug Fix] Incorrect calculation for container right overlap

tjhiggins:
This probably warrants adding some test cases for the hit detection.


kball:
"This probably warrants adding some test cases for the hit detection."  Indeed it does.  @tjhiggins do you want to tackle those tests?  I think this would be a good set of unit tests.

Alternatively, @Owlbertz this might be a good area to focus on some of your unit testing work, as the box utility impacts a ton of things.

PR #8399, opened 2016-03-11
clarify loadPath for node-sass wrt to npm v3

akrawchyk:
This PR updates the Foundation SASS docs to clarify `loadPath` for the grunt example.

Due to the changes npm v3 introduced regarding dependency resolution, I'm not sure we can guarantee the `foundation` module will live in the root of the `node_modules` folder.

Therefore, this update suggests using node's builtin require (https://nodejs.org/dist/latest-v4.x/docs/api/modules.html#modules_all_together) and path (https://nodejs.org/dist/latest-v4.x/docs/api/path.html) modules for determining the path to Foundation.

### DIFF
--- a/docs/pages/sass.md
+++ b/docs/pages/sass.md
@@ -48,11 +48,13 @@ Next, add the framework files as an import path. How you do this depends on your
 Here's an example using grunt-contrib-sass:
 
 ```js
+var path = require('path');
+
 grunt.initConfig({
   sass: {
     dist: {
     options: {
-        loadPath: ['node_modules/foundation-sites/scss']
+        loadPath: [path.join(path.dirname(require.resolve('foundation-sites')), '../scss')]
       }
     }
   }


gakimball:
Assuming Foundation is a top-level dependency, it's always going to be in the same place, right?

akrawchyk:
> Assuming Foundation is a top-level dependency, it's always going to be in the same place, right?

Yes I believe this is correct.

> Also, going from npm v2 to v3, isn't it more likely that foundation-sites will be a top-level folder in node_modules?

Come to think of it, that's probably true too. Only conflicts would be between different versions of `foundation-sites` in a package's sub-dependencies and therefore installed in a sub-folder of a dependency. Probably not an ordinary use-case though!

gakimball:
For sure. [Eyeglass](https://github.com/sass-eyeglass/eyeglass) is going to be a good solution to this problem once it sees wider adoption, because it makes Sass's `@import` basically work like Node's `require()`, which removes the need to specify manual load paths.

Foundation works as an Eyeglass module right now, and once that thing hits 1.0 I'd like to integrate it into our build processes.

akrawchyk:
Great, thanks for the consideration! Glad that Eyeglass is getting attention here!

PR #8400, opened 2016-03-11
Customizer Updates

gakimball:
@kball and I are working on some updates to the customizer. The big thing is integrating the build process directly into the foundation-sites Gulpfile. This makes it easier to test various configurations, and easier to test the custom builds with the framework before we deploy a new version.

On the server end, the customizer holds a specific tagged version of the repository, and can generate custom builds by running `gulp customizer --modules path/to/config.json`, where `config.json` is a file built from the customizer's form data.

The server end is still being worked on (and will probably be closed source).


gakimball:
For @kball, this is the spec for the configuration.

```json
{
  "modules": [],
  "variables": {}
}
```

`modules` is a list of strings—they correspond to the modules the user has enabled/disabled in the form. Each string should be a top-level key found in the [master modules list](https://github.com/zurb/foundation-sites/blob/customizer/customizer/config.yml).

`variables` is an object of Sass variables. The key is the name of the variable *without the dollar sign*, and the value is the value of the variable. All values should be strings, including booleans.

```json
"variables": {
  "primary-color": "#ffffff",
  "global-flexbox": "true"
}
```

gakimball:
@kball I've made the final needed changes on my end:

- Ensuring all compressed/uncompressed assets are properly copied.
- Combining color variables into `$foundation-palette` instead of including them as separate variables.

kball:
This is now working & the web interface is live.

PR #8402, opened 2016-03-14
Update foundation-flex.css

paxperscientiam:
OK, I'm not entirely sure what's going on, but there's some interaction between nested flex-grid and callout that causes the final box to have incorrect dimensions.

example: http://imgur.com/sJTMYpU

Changing....

```
.callout > :first-child {
    margin-top: 0; }
```

to...

```
.callout > :first-child {
    margin-top: inherit; }
```

fixes it. However, I'm not sure if this will screw something else up.

### DIFF
--- a/dist/foundation-flex.css
+++ b/dist/foundation-flex.css
@@ -2163,7 +2163,7 @@ select {
   .callout > :first-child {
     margin-top: 0; }
   .callout > :last-child {
-    margin-bottom: 0; }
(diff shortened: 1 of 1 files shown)


kball:
Hi @paxperscientiam the dist file is generated from the scss, so this change would not persist... we should get to the bottom of the issue and fix it in the root location.    Please file an issue with sample code that reproduces the issue, and we'll go from there.

PR #8410, opened 2016-03-15
Address the FAQ regarding overriding styles

MainelyWeb:

Happy coding! :)

Directly after the discussion of the ZURB Template directory structure, added the preferred method of overriding and adding to the default styles. Also addressed the question of whether _settings.scss is overwritten during update.

### DIFF
--- a/docs/pages/starter-projects.md
+++ b/docs/pages/starter-projects.md
@@ -35,7 +35,7 @@ The biggest difference between this and the basic template is the folder structu
 
 To override or add to the default styles of the ZURB Template, in your project's `src/assets/scss/` folder
  - Change Sass variables in `_settings.scss`
- - Add custom scss and css at the bottom of `app.scss`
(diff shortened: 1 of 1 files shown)


rafibomb:
Love it! I would suggest one change to pull it in.

Add custom SCSS imports at the bottom of `app.scss`. New component SCSS partails can be added in the included `components` folder.

MainelyWeb:
I agree. I think the wording below may be more understandable to someone new to Foundation. It does not require a new SASS programmer to understand the phrase "component SCSS partails".

Add custom SCSS and css to new files in the `components` folder then import those files at the bottom of `app.scss`.

I don't think it should turn into a tutorial, so I would stop short of explaining that multiple files can be imported with each file concentrated on a specific topic or area, for example, buttons, images, typography, etc.

I think such a tutorial has value, but my goal is to relieve you from constantly having to answer the style override question.

PR #8421, opened 2016-03-16
Update CONTRIBUTING.md

designerno1:
i stopped the work on Clearing, please change it to the equalizer extension i wrote thanks


gakimball:
We'd like to bring back that Clearing example eventually, but we can talk about that later. :)

Issue #8424, opened 2016-03-17
Request/ Proposal: Add SCSS variables for customizing margins for checkboxes, radiobuttons, inputs and select


uranuz:
I propose to add seperate SCSS variables for specifying margins for elements like checkbox, radio, file inputs, text inputs and 'select'. So this is where variable $form-spacing is currently used for setting margins. By default these proposed variables would still be equal to $form-spacing, but user could change them in order to customize margin-bottom. For example I hate that margin-bottom is set for checkboxes. I have checkboxlists that are placed verticaly one by another. So the have ugly huge space in between, because of these margins. I think that if someone needs it it's simplier to add it in place where it needed than fight with these margins everywhere, because I need them only in 20% of cases. I could make PR myself if someone interested in this improvement. What do you think?

colin-marshall:
Thanks for this feature request. We’re going to close out this ticket, as we will not be adding more features to Version 6, but your request is being considered for Version 7.

PR #8431, opened 2016-03-18
Added variables to customize margins for checkbox, radio, select and …

uranuz:
My proposal for additional scss variables for margins customizations. Details are there:
https://github.com/zurb/foundation-sites/issues/8424

### DIFF
--- a/scss/forms/_checkbox.scss
+++ b/scss/forms/_checkbox.scss
@@ -6,11 +6,15 @@
 /// @group forms
 ////
 
+/// Margin around checkboxes, radios and file inputs.
+/// @type List
+$form-checkbox-margin: 0 0 $form-spacing !default;
+
 @mixin foundation-form-checkbox {
   [type='file'],
   [type='checkbox'],
   [type='radio'] {
-    margin: 0 0 $form-spacing;
+    margin: $form-checkbox-margin;
   }
 
   // Styles for input/label siblings
--- a/scss/forms/_label.scss
+++ b/scss/forms/_label.scss
@@ -24,7 +24,7 @@ $form-label-line-height: 1.8 !default;
 
 /// Margin around label-tag elements.
 /// @type List
-$form-label-margin: 0 0 $form-spacing;
+$form-label-margin: 0 0 $form-spacing !default;
 
 @mixin form-label {
   display: block;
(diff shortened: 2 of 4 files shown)


uranuz:
Maybe also needed some changes to *form-error* and *foundation-form-fieldset* or else, but I'm not sure.

SalahAdDin:
I think that this PR will go againt's new Foundation philosofies, but, :+1:

uranuz:
I'm not familiar with new Foundation philosofy, so it's up to community responible people decide whether this approach appropriate. Maybe I should solve these problems another way. I would be interested to consider for alternatives.

kball:
This is an API change, so the PR should be against v6.3.0 rather than develop.  Other than that, @andycochran or @brettsmason what are your thoughts on making this change in the scss?

brettsmason:
@kball I understand the reasoning for wanting to implement additional control over this (though I havent had the need for this myself other than labels). The only thing I wondered is if the variables should be prefixed with form (so $form-checkbox-margin etc) just to make them really clear, but I'm not sure.

andycochran:
Shouldn't these also be exposed in `_settings.scss`?

brettsmason:
I'm pretty sure the `_settings.scss` file is generated by Octophant on each release so no need to add them to settings, but I could be wrong (I'm sure I remember Geoff saying this).

rafibomb:
@brettsmason That's right! Octophant will update the settings file when we deploy a version or if you run `npm run deploy:settings` - so you're good!

rafibomb:
@uranuz Can you point this to branch `v6.3` so we can merge it? Thanks!

kball:
@andycochran @brettsmason how did you feel about this change?

andycochran:
Only issue I see is the inconsistent variable prefixes.

zurbrandon:
I'm not a fan of adding these, mainly since it's something that can be nearly as easy to change in css as it would be in the settings. Variables are great when a single variable changes many many things or some math occurs to make them more useful. 

This seems to just be moving code from one place to another.

kball:
Based on @zurbrandon's comment I'm going to close this without merging.  If you'd like to reopen with an argument as to why this is adding more value/needed please do.

Issue #8377, opened 2016-03-09
Dropdown Menu Arrows "Flash" on Load


JeremyEnglert:
Not a huge deal, but slightly annoying. 

**How can we reproduce this bug?**

- Create a dropdown menu with Foundation 6.2

**What did you expect to happen?**

- Menu to load without a flash of unstyled content

**What happened instead?**

- Slight flash of unstyled content

**Test case:**

- JointsWP Demo Site: http://jointswp.com/demo/

rafibomb:
Hey Jeremy! There is a spot on the docs with a solution for this: http://foundation.zurb.com/sites/docs/responsive-navigation.html#preventing-fouc

Let us know how this helps or if you're seeing a different issue.

JeremyEnglert:
Hi Rafi,

Unfortunately, I think this is a separate issue. The arrows for the dropdown menus are added by JS - which causes them to "flash" as the page loads. 

See attached. 

![original](https://cloud.githubusercontent.com/assets/6110968/13656235/720867ce-e622-11e5-939b-2247680c38bd.gif)

gakimball:
@JeremyEnglert Is a possible solution to show down-facing arrows by default (with no class attached), and then the JavaScript is able to correct it when it initializes the menu?

JeremyEnglert:
@gakimball I wonder if that would be something that could be added to Foundation (it would be easy enough to do manually, but would be helpful it included).

But that may require a class to be added to the markup, Let me play around and see if anything sticks.

gakimball:
There's `.dropdown.menu` you could attach to, although I suppose each menu won't know if it has a submenu or not.

JeremyEnglert:
It looks like F5 used a `.has-dropdown` class to prevent this. 

But that would require the markup to be tweaked. Which may not be the worst thing, as the menu would still function without the extra markup (but have the flash).

gakimball:
Hm, in that case, we already have `.is-dropdown-submenu-parent`, which is the same thing. I don't necessarily want to put those in the markup examples, since it makes them a bit more complicated, and the idea is that the plugin adds those classes for you.

We could make a note of this in the documentation, i.e., if the FOUC is problematic for you, here's how you can get around it.

JeremyEnglert:
> We could make a note of this in the documentation, i.e., if the FOUC is problematic for you, here's how you can get around it.

LOVE that idea.

[...]

PR #8436, opened 2016-03-21
Preventing FOUC

JeremyEnglert:
Shows how to prevent dropdown menu FOUC, as noted here: https://github.com/zurb/foundation-sites/issues/8377


gakimball:
Looks hot, thanks for putting it together :)

1forh:
There is an error in the docs where this is concerned.
`class=".is-dropdown-submenu-parent"`
Should be:
`class="is-dropdown-submenu-parent"`

1forh:
There is an error in the docs where this is concerned.
`class=".is-dropdown-submenu-parent"`
Should be:
`class="is-dropdown-submenu-parent"`

Issue #8444, opened 2016-03-22
[Button group] the separator position isn't correct in the stacking mode


abdullahsalem:
The margin (separator) of the stacked button group isn't in the expected position.

**How can we reproduce this bug?**
Add `.stacked` to a button group markup.

**What did you expect to happen?**
The margin (separator) should be from the bottom.
**What happened instead?**
The margin (separator) from the right side.

**Test case:**
- [CodePen with Foundation 6.2 ](http://codepen.io/abdullahsalem/pen/mPWZoy)

PR #8445, opened 2016-03-22
Fixed the margin (separator) of the stacked button group #8444

imabdulsal:
This PR fixes the margin of the stacked button group, where the margin should be from the bottom not the right (#8444).
Current state: http://codepen.io/abdullahsalem/pen/mPWZoy
button-group-stacked-issue-zoom-in

But it should look like this
button-group-stacked

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -109,13 +109,20 @@ $buttongroup-expand-max: 6 !default;
   #{$selector} {
     @if $global-flexbox {
       flex: 0 0 100%;
+      margin-#{$global-right}: 0; 
     }
     @else {
       width: 100%;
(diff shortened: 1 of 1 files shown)


kball:
This looks good, and works like a charm.  Thanks @abdullahsalem

andycochran:
I think this merge caused #8525 to fail. It was passing before. :(

kball:
hmm... not sure.  It looks like the travis fail was an npm install issue. It's been super flaky.  I just restarted it.

andycochran:
Word. I assumed the conflicts were 'cause the same file (`_button-group.scss`) was being changed. Shoulda looked at the details. Thanks, @kball. :blush:

Issue #8447, opened 2016-03-22
[Button group] radius doesn't wrap the whole group, but each one!


abdullahsalem:
I think that when the radius is applied on the button group, it should appear as a wrapper for the whole group not each one!

**How can we reproduce this bug?**
- Change the global radius in the settings
- Build your code
- Create your button group's markup ([Get the basic markup from here](http://foundation.zurb.com/sites/docs/button-group.html#basics))

**What did you expect to happen?**
![button-group-container-radius](https://cloud.githubusercontent.com/assets/177020/13958084/d7206a84-f05f-11e5-9961-ae49585d3f6c.png)

**What happened instead?**
![button-group-container-radius-issue](https://cloud.githubusercontent.com/assets/177020/13958597/bf4e8236-f061-11e5-8b23-bfc450167991.png)

g12n:
Perfect! I was just fixing this bug locally! 
Thanks @abdullahsalem! Hope the pull request is integrated soon!

andycochran:
Hmm… I'm not sure if this is a bug. 

This makes an assumption that you don't want each button to have its own radius. That definitely makes sense for Split Buttons (w/ the arrow). But if `$buttongroup-spacing` is set to a larger value, it might look better to have the radius on each button. 

I suppose having the radius only on the group does help to associate the buttons with each other as a group. But I'm still not sure if the radius should go on the group or on each button. Maybe what's needed is a `$button-group-radius` variable. Then you could have it either way. 

Honestly, I'm for dropping `$button-radius` altogether and making it require app-specific customization.

g12n:
@andycochran I guess we could add another variable, to make this behavior optional. Though personally I think, the given Example of a split button shows quite clearly, that removing the radius inside the group is the typically expected behavior.
  
Just for Illustration purposes. This is a split button: 
![split button](https://cloud.githubusercontent.com/assets/57556/14227048/be92423e-f8f3-11e5-88a4-9e6cba7ff942.png)
The current result is merely two buttons real close to each other: 
![two buttons](https://cloud.githubusercontent.com/assets/57556/14227059/de70c382-f8f3-11e5-942f-3560a0d2f540.png)

[...]

PR #8448, opened 2016-03-22
Fixed the button group's radius to be as a wrapper for the whole group #8447

imabdulsal:
#8447


rafibomb:
@abdullahsalem Since this does change the default behavior, it would be better to target to branch `v6.3`. Can you point it there and we'll merge in for 6.3?

abdullahsalem:
Sure, @rafibomb.

PR #8449, opened 2016-03-22
Fixes touch interactions for slider

Owlbertz:
Fixes the general touch support for slider plugin and I guess also for other plugins since the error seemed to affect all plugins there.

However, sliding still kinda bugs when vertical slider is used on touch devices (doesn't happen in Chrome emulation...) - I will fix that within the next days. 
Edit: Clicking the slider's track actually works well, just the dragging part needs some work.

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -387,6 +387,10 @@ class Slider {
 
             $body.off('mousemove.zf.slider mouseup.zf.slider');
           });
+      })
+      // prevent events triggered by touch
+      .on('selectstart.zf.slider touchmove.zf.slider', function(e) {
+        e.preventDefault();
       });
     }
 
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -271,7 +271,9 @@ class Slider {
           pageXY = vertical ? e.pageY : e.pageX,
           halfOfHandle = this.$handle[0].getBoundingClientRect()[param] / 2,
           barDim = this.$element[0].getBoundingClientRect()[param],
-          barOffset = (this.$element.offset()[direction] -  pageXY),
+          // touch events emulated by the touch util give position relative to screen, add window.scroll to event coordinates...
+          windowScroll = vertical ? $(window).scrollTop() : $(window).scrollLeft(),
+          barOffset = this.$element.offset()[direction] - (this.$element.offset()[direction] < pageXY ? pageXY : (pageXY + windowScroll)),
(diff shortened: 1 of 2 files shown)


Silvester20:
Damn this is a nice one! Thanks for commiting!

Just curious, are you contracted by Zurb Foundation? If not they should :)

Owlbertz:
Should be touch-usable now. Although the slider handle is quite hard to hit on touch. Maybe there should be some CSS making it bigger on small screens by default?

Owlbertz:
@Silvester20 would you be able to test this one?

rafibomb:
Thanks @Owlbertz We'll be testing this out! Would like to to get this into the next patch

Silvester20:
@Owlbertz Yeah sure, but don't know how to test this. My site is using the foundation.min.js & foundation.min.css from the complete download package.

And I assume when I redownload the complete package the files doesn't have changed, still the same foundation.min.js as the release of 6.2.0, without your commits right?. So how can I test this?

rafibomb:
@Silvester20 You can checkout the branch here: git checkout -b Owlbertz-touch-slider develop and pull those changes to test it. We'd love to see more people confirming the fix so we can merge it in faster.

andycochran:
@Owlbertz—regarding "hard to hit on touch"—since small screen ≠ touch screen, we could use `what-input.js` to know when the user is trying to touch the slider and adjust the style with something like this: 
http://codepen.io/andycochran/pen/VaQzzz

Owlbertz:
@andycochran Yeah, I had something like this on my mind as well. Looks promising.

kball:
@Silvester20 were you able to test this out?  Alternatively, @andycochran can you try it out and see if it works as expected?

andycochran:
@kball This is only a JS change, no CSS. So you'd be a better judge. Did we want this to address the "hard to hit on touch" problem, or should we document that in a separate issue?

Owlbertz:
Since this fixes an actual issue I think the "convenience stuff" could be done later on, in a separate issue.

kball:
k tested it out, seems good.  Merging.

kball:
@Owlbertz does this fix any outstanding filed issues that you are aware of?

Owlbertz:
At least #8666 which I recently referenced. I do believe there was something else reported by @Silvester20

Silvester20:
I can only test this with a compiled javascript (*.js) file. But I trust you guys :)

sjransom:
This seems to fix some touch issues I've been having. Have tested on latest iOS and Android, many thanks @Owlbertz

PR #8453, opened 2016-03-24
Fixes issue when putting focus to an input outside of the modal.

### LINKED ISSUE #8442: [Reveal] steals focus and key events
First, thank you for Foundation, I appreciate all the work that you've put into this!

I am using a Reveal modal with the intent to show search results. If you are familiar with search in iOS' calendar then you'll have a good sense of the goal. If you that doesn't clarify then it's sufficient to know that the goal is to open a modal while retaining focus in a text field.

Surprisingly, opening the reveal modal removes the focus from a text input. I've tried to handle the 'open.zf.reveal' event to take focus back, but that results in a warning 'Component not defined!' inside of foundation.util.keyboard.handleKey and causes key events to be lost. Specifically the text input appears to have focus, but all typing is lost.

I understand that taking focus is intended for screen readers, however, I'm wondering if this could be made optional behaviour, or if somehow both goals could be achieved.

To demonstrate see http://codepen.io/anon/pen/LNWaQX and start typing in the text field. The first key press causes a modal to appear after which the text input appears to have the focus, but all key events are lost.

Owlbertz:
Fixes #8442.
It doesn't make sens to have these event handles attached to the whole document at all.

See [this pen](http://codepen.io/Owlbertz/pen/qZmyoK) using the updated version.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -292,9 +292,6 @@ class Reveal {
             }
           }
         });
-        if (_this.focusableElements.length === 0) { // no focusable elements inside the modal at all, prevent tabbing in general
-          e.preventDefault();
-        }
       });
     }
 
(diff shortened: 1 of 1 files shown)


Owlbertz:
As @josh-h confirmed this is working, I will merge the PR.

einfallstoll:
I request to revert the pull request, since it introduces an issue with inputs within the reveal. Because of the `preventDefault()` it's not possible to type into the inputs within the reveal at all.

PR #8456, opened 2016-03-24
Include $body-font-family as a default

cortopy:
Currently, $body-font-family is not included in output css. This is because $body-font-family is not referenced anywhere except for $header-font-family. Even <p> would not be styled with $body-font-family

By adding a body tag, we are defaulting to $body-font-family but this can be overriden by any other selector, including headers (as it will be more specific)


andycochran:
This style is already applied in `_global.scss`.

PR #8457, opened 2016-03-24
Fix typos in Mixins and Functions docs

kaytcampbell:
- Replace commas at the ends of sentences with periods (Sass Mixins documentation)
- Change top/right/bottom/right to be top/right/bottom/left (Sass Functions documentation)


kball:
Please resubmit against develop

PR #8458, opened 2016-03-24
Text inputs function with modifier (for classes and pseudo-classes)

michelts:
Permit to define a class or a pseudo-class for text inputs generated through the text-inputs sass function.

### DIFF
--- a/scss/util/_selector.scss
+++ b/scss/util/_selector.scss
@@ -9,7 +9,8 @@
 /// Generates a selector with every text input type. You can also filter the list to only output a subset of those selectors.
 ///
 /// @param {List|Keyword} $types [()] - A list of text input types to use. Leave blank to use all of them.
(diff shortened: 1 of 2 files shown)


kball:
Is this addressing a particular issue?  What's the driver?

@brettsmason and @andycochran as our sass expert yetinauts, what are your thoughts?

michelts:
We wanted to define a custom shadow for focused text inputs but I can imagine a better situation. We could use this approach to show a hidden help text positioned after an input, like:

```
input:focus + .help-text {
    display: block;
}
```

kball:
This seems reasonable to me. @brettsmason and @andycochran what do you think?

brettsmason:
Yep this seems fine to merge to me 👍

andycochran:
Nice. 👍

Issue #8428, opened 2016-03-18
Callout makes reveal close (bug) (again)


Silvester20:
And this bug ( https://github.com/zurb/foundation-sites/issues/7833 ) is there again in Foundation 6.2.0.

CodePen: http://codepen.io/anon/pen/xVgmqx

Owlbertz:
I did something to prevent this in this commit https://github.com/Owlbertz/foundation/commit/4e01cecca10ba4f4dee3e39b504c6567ef1258c1. But as mentioned the solution may not be ideal. I don't know why the `stopPropagation()` doesn't work as intended...
Maybe someone has some feedback or solution to this.

Silvester20:
Will be very nice if this is the next release, thank you so much!

kball:
Fixed by #8471, will be in 6.2.2

Silvester20:
Amazing, a big thanks to @Owlbertz and ofourse @kball !

rui-castro:
@Owlbertz @kball  This bug still exists in foundation 6.4.3.

Using something like the example bellow; clicking callout close button closes the reveal.

```html
<div class="reveal reveal-animated">

        <div class="callout warning" data-closable>
            <button class="close-button" aria-label="Close warning" type="button" data-close>
                <span aria-hidden="true">&times;</span>
            </button>
            Callout text
        </div>

    <a aria-label="Close" data-close>&#215;</a>
</div>
```

PR #8471, opened 2016-03-26
Make callouts not close the reveal they are in.

Owlbertz:
This solution is kinda sloppy, maybe there is a more elegant way to solve this...
Hope someone else can review this soon, really would like to see this in the upcoming release.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -133,8 +133,8 @@ class Reveal {
 
     this.$element.on({
       'open.zf.trigger': this.open.bind(this),
-      'close.zf.trigger': (event, element) => {
-        if (typeof element === 'undefined') { // only close reveal when it's explicitly called
+      'close.zf.trigger': (event, $element) => {
+        if ($(event.target).parents('[data-closable]')[0] === $element) { // only close reveal when it's explicitly called
           return this.close.apply(this);
         }
       },
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -133,7 +133,11 @@ class Reveal {
 
     this.$element.on({
       'open.zf.trigger': this.open.bind(this),
-      'close.zf.trigger': this.close.bind(this),
+      'close.zf.trigger': (event, element) => {
+        if (typeof element === 'undefined') { // only close reveal when it's explicitly called
+          return this.close.apply(this);
+        }
+      },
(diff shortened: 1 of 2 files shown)


Silvester20:
This will fix https://github.com/zurb/foundation-sites/issues/8428, so many thanks for fixing this!

kball:
Hmm... this approach feels like it could have unintended side effects.  What if instead we have the callout event handler call event.stopPropagation() to prevent it bubbling up?

Owlbertz:
Well there is no callout event handler because it's all done in the `Foundation.util.triggers.js`:
```js
(document).on('close.zf.trigger', '[data-closable]', function(e){
  e.stopPropagation();
  let animation = $(this).data('closable');

  if(animation !== ''){
    Foundation.Motion.animateOut($(this), animation, function() {
      $(this).trigger('closed.zf');
    });
  }else{
    $(this).fadeOut().trigger('closed.zf');
  }
});
```
There is already a `stopPropagation()`, but that doesn't seem to work well here. That's why I built it this way.

kball:
Ah interesting... the reveal handler is probably getting called first, given that the callout doesn't have its own handler but is depending on that base handler.

Hmm... so I still don't think this is the right fix, because it will break the ability to simply include a 'data-close' element inside a reveal... but I'm not sure what the right fix is.  I'm also not 100% sure that element is going to get passed along in all click cases...

What we're really trying to do is handle nesting... so what if instead of looking at element we did a check something like this

```
if ($(event.target).parents('[data-closable]')[0] !== this.$element[0])
```

Basically looking to see if there's another data-closable element in the tree before you get to the reveal event handler?

Silvester20:
Any updates on this?

kball:
@Owlbertz can you take a look at the concern above and rework this fix to still allow using data-close inside of a reveal?

kball:
Great.  Nice work!

rui-castro:
This bug still exists in foundation 6.4.3.

Using something like the example bellow; clicking callout close button closes the reveal.

```html
<div class="reveal reveal-animated">

        <div class="callout warning" data-closable>
            <button class="close-button" aria-label="Close warning" type="button" data-close>
            </button>
            Callout text
        </div>

    <a aria-label="Close" data-close>&#215;</a>
</div>
```

PR #8473, opened 2016-03-27
Fix issue with Reveal#destroy().

Owlbertz:
Noticed that Reveal's `destroy()` runs into an error because the `$element` is appended to the `$overlay` and the `$overlay` is then removed before `unregisterPlugin()` gets called. Therefore `plugin.$element.data('zfPlugin')` failed because `$element` as already removed with `$overlay`.

Moved $element outside of `$overlay` to prevent error `unregisterPlugin()`.


kball:
Looks good

Issue #7392, opened 2015-12-04
[Dropdown] Alternative alignment


foodgy:
There is positioning option in dropdown plugin, but there is no alignment option.
Dropdown is almost always anchored to the left edge of the trigger button.
Adding option to anchor dropdown-pane to the right would be good adjustement.

![Dropdown](https://cloud.githubusercontent.com/assets/706377/11583850/7d674850-9a6f-11e5-979f-1ae5977af552.jpg)

rafibomb:
Not sure if such fine control of the positioning is on our roadmap. We'll discuss it with the team. A PR would probably go a long way into swaying us ;)

Owlbertz:
I made a suggestion here: https://github.com/Owlbertz/foundation/commit/9206675395a85e71b878bd3fa8a2a8f65b573152
Not sure if this fits your needs, but I felt your issue was related to floating the anchor... Otherwise the dropdown itself would need to get some additional information for even more fllexibility.

andrewbiddinger:
:thumbsup: This would be very helpful!

jlgreene2:
Has there been any further discussion on this topic?  I'm running in a similar issue implementing a dropdown panel, displayed from a very small button/icon located on right hand side of my title bar.  Ideally, I'd like to see the panel right aligned across the bottom of the button.  But the current behavior is for the content to default to the left side of my button when where isn't enough room for the panel to fit between the left edge of the button and the right edge of the viewport.  It would be very helpful to be able to provide the dropdown plugin an "align" hint to permit the panel being displayed to be right aligned across the bottom of the triggering button.

kball:
Addressed in #8475.  Will be in 6.2.1

PR #8475, opened 2016-03-28
Enabling floated dropdown alignment

Owlbertz:
As requested in https://github.com/zurb/foundation-sites/issues/7392.
Makes dropdown play nice with floating.

Basically my former solution merged into the current state of development.


ncoden:
:+1: need

kball:
Looks good to me.

ncoden:
Thanks

dobromir-hristov:
So if my trigger is NOT floated, I cannot horizontally align the dropdown? Why on earth would you hard code the horizontal positioning to a float class.... a simple check for a horizontal align class on dropdown after the anchor would have been enough. That way it would have still be compatible with float classes, but a user could align it to his needs without floating the trigger. Yes I know i can override the float class in my own css, but is it necessary?

Issue #8072, opened 2016-02-03
[Drilldown] Accessibility/Keyboard controls issues


anthonysimone:
Keyboard control in the drilldown menu component seems to have some issues. This is using the drilldown component in a new project with the basic example markup provided in the documentation.  Focus doesn't move off of the first item. This appears to be because once you are focused inside the component, the keyboard actions are being handled by the component, and they seem to be applying focus to the `<li>` instead of the `<a>`.

Adjusting this in each of the key handlers seems to fix the focus issue. For example, adjust the down method to
`down: function() {
          $nextElement.children('a').focus();
},` 
and adjusting the others accordingly seems to take care of the focus issue. However, there still seems to be some issues with navigating through menu items that have children and the js added back button, along with moving the focus back up a layer.

I'll be looking into this a bit more when I have some time, I like the component, but need a completely accessible menu for the project I'm currently working, so hopefully I can get this sorted.

rafibomb:
It sounds like that is the issue. Were you able to tinker with it more? Any suggestions?

kball:
Fixed in https://github.com/zurb/foundation-sites/commit/ab0f267608e95a9a9ee9d1bbc85746577b756493

PR #8477, opened 2016-03-28
Fix keyboard access for drilldown

Owlbertz:
Adresses #8072.
Mainly I changed the focused elements to be the `a` tags rather than the `li` tags.
I also added anchors to the menu links in the docs, so it is possible to actually click/space/enter the links to open a new page (or at least change their window's location).
The fact that the links within the menus had only pure `#`es in their `href` did not cover the real-life menu purposes and made proper testing a bit hard.


kball:
Looks good.  Merging

PR #8478, opened 2016-03-29
[Orbit] Bugfix: Remove image-loaded checking with this.naturalWidth

milose:
When serving properly cached images, this method of checking weather an image has been loaded breaks the height calculation resulting in "thin" orbit boxes.
<br />
<br />Reloading the page and ignoring cached content (Cmd + Shift + R) everything loads as it should be, however, on subsequent normal refreshes (Cmd + R) orbit galleries "shrink".
<br />
<br />See image for details.
<br />
<br />#### Preview:
<br />break2
<br />
<br />#### DOM:
<br />break
<br />

### DIFF
--- a/js/foundation.util.timerAndImageLoader.js
+++ b/js/foundation.util.timerAndImageLoader.js
@@ -62,9 +62,9 @@ function onImagesLoaded(images, callback){
     if (this.complete) {
       singleImageLoaded();
     }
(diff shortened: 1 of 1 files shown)


rafibomb:
I wonder if there is a way to do this more pragmatically?

kball:
This fix will break other scenarios, so going to close this without merging.  We need a better way to fix this bug.

milose:
What other scenarios?

Issue #8460, opened 2016-03-24
[Feature Request] Add javascript function showAll() to Accordion Menu


crbranch:
Accordion Menu currently has a `hideAll()` function, but there is no `showAll()` function.  This would be helpful when implementing Collapse All / Expand All options.

Current workaround:

Initially I tried to implement `showAll()` using the existing `hideAll()` function as inspiration and simply doing the opposite by calling `slideDown()` on all of the submenu items.  This appeared to work at first glance, but after more careful examination I realized that some of the attributes end up in an inconsistent state.  In particular, `aria-hidden=` is not always reset properly after calling `slideDown()`.

I ended up with the following solution that calls `up()` and `down()` instead of `slideUp()` and `slideDown()`.  I'm not 100% sure this is the best approach, but I haven't run into any issues so far:

            var menu = new Foundation.AccordionMenu($('#myMenu'));

            $("#expandAll").click(function () {
                menu.down(menu.$element.find('[data-submenu]'));
            });

            $("#collapseAll").click(function () {
                menu.up(menu.$element.find('[data-submenu]'));
            });

So, there are two takeaways:

1. It would be helpful to include a `showAll()` in the plugin.

2. The current `hideAll()` function calls `slideUp()` instead of `up()`.  Someone may want to confirm that using `slideUp()` in this situation is not causing any issues that haven't been discovered yet since doing the reverse does seem to cause trouble.

kball:
@Owlbertz @crbranch with #8743 merged into 6.3, can we close this issue?

Owlbertz:
From my understanding, yes.

crbranch:
Looks good to me.  Thanks!

PR #8493, opened 2016-03-31
Added showAll.

Owlbertz:
Added showAll using `down()` for all elements.
Rewrote hideAll to use `up()` for all elements.

Like in #8460 purposed by @crbranch the hideAll method should use the `up()` method.
Also `showAll()` could be useful.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -170,7 +170,15 @@ class AccordionMenu {
    * @function
    */
   hideAll() {
-    this.$element.find('[data-submenu]').slideUp(this.options.slideSpeed);
+    this.up(this.$element.find('[data-submenu]'));
+  }
+
+  /**
+   * Opens all panes of the menu.
+   * @function
+   */
+  showAll() {
+    this.down(this.$element.find('[data-submenu]'));
   }
 
   /**


ncoden:
it's a feature purpose, if I undestand correctly ?
It should be better to merge it in `v6.3` than `develop`

kball:
Agree with @ncoden.  Please resubmit against v6.3.  Also please add a unit test for this case in `test/javascript/components/accordionMenu.js`.  You can use the tests in `test/javascript/components/toggler.js` as an example of how these can be structured. See: https://github.com/zurb/foundation-sites/blob/develop/test/javascript/components/toggler.js

Thanks!

Owlbertz:
Well this actually solves an issue with the ARIA attributes when using `hideAll()`, because as mentioned in #8460 the current implementation of `hideAll()` doesn't use `up()` which causes the ARIA attributes to not match the element state.
I can add this PR to v6.3 if you want, though.

kball:
I'm happy to have the hideAll fix for the patch release, but we want to be very careful to only have api changes at minor releases.  I'll leave it up to you if you want to split one to develop and one to v6.3 or just send them all to v6.3

PR #8503, opened 2016-04-01
Add possibility to disable slider after initialization.

### LINKED ISSUE #8497: Problem disabling slider with jquery
I have an issue with the foundation slider described here:
http://foundation.zurb.com/sites/docs/slider.html
When I add the "disabled" class to the slider directly in HTML it gets disabled as expected. No dragging or clickSelecting of the handle is possible.

But when I do this: `$("#my_slider").addClass("disabled");` the slider gets disabled but it's still possible to drag and clickSelect the slider handle.

Anyone?

Jimmy

Owlbertz:
Slider will check for disabled state on every event now rather than only at its initialization.

Fixes #8497.

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -106,7 +106,11 @@ class Slider {
    * @fires Slider#changed
    */
   _setHandlePos($hndl, location, noInvert, cb) {
-  //might need to alter that slightly for bars that will have odd number selections.
+    // don't move if the slider has been disabled since its initialization
+    if (this.options.disabled || this.$element.hasClass(this.options.disabledClass)) {
+      return;
+    }
+    //might need to alter that slightly for bars that will have odd number selections.
     location = parseFloat(location);//on input change events, convert string to number...grumble.
 
     // prevent slider from running out of bounds, if value exceeds the limits set through options, override the value to min/max
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -107,7 +107,7 @@ class Slider {
    */
   _setHandlePos($hndl, location, noInvert, cb) {
     // don't move if the slider has been disabled since its initialization
-    if (this.options.disabled || this.$element.hasClass(this.options.disabledClass)) {
+    if (this.$element.hasClass(this.options.disabledClass)) {
       return;
     }
     //might need to alter that slightly for bars that will have odd number selections.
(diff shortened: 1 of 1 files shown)


ncoden:
Thanks. Should we check the `[disabled=false]` case ?

Owlbertz:
Yeah, you're absolutely right. It only makes sense if you can also enable a disabled slider... Did another commit for this.

ncoden:
I mean : should we consider the slider as enabled when it has a `disabled="false"` or `disabled="enabled"` attribute ?

ncoden:
I don't know if `this.options.disabled` care about this cases, or if it is only checking that a `disabled` attribute has been set, regardless of its value

Owlbertz:
I guess you mean `data-disabled`, since `disabled` is only for `input`s. The options are only initialized once based on this attribute, so for now only checking the class is possible.

js/foundation.slider.js
```diff
@@ -106,7 +106,11 @@ class Slider {
    * @fires Slider#changed
    */
   _setHandlePos($hndl, location, noInvert, cb) {
-  //might need to alter that slightly for bars that will have odd number selections.
+    // don't move if the slider has been disabled since its initialization
+    if (this.$element.hasClass(this.options.disabledClass)) {
```
ncoden:
```js
var disabledAttr = this.$element.data('disabled');
var disabledClass = this.$element.hasClass(this.options.disabledClass);

if ((disabledClass || disabledAttr) && !(disabledAttr && disabledAttr == "false")) {
```

So :
- `disabled` class or attribute (not `false`) -> **Disabled**
- `disabled` option in initialization -> **Disabled**
- `disabled` option in initialization + `disabled` class or attribute (not `false`) -> **Disabled**
- `disabled` option in initialization + `disabled` class + `disabled` attribute equal to `false` -> **Enabled**

kball: I'd prefer there be a single source of truth, and it looks like we're currently using the class as the way to do that.  @ncoden did you see someplace in particular that is setting data-disabled?  I'm not seeing that mechanism being used.

kball:
@Owlbertz this looks good.  Can you set up a unit test for this case in test/javascript/components/slider.js? You can use the tests in test/javascript/components/toggler.js as an example of how these can be structures. See: https://github.com/zurb/foundation-sites/blob/develop/test/javascript/components/toggler.js

Thanks!

js/foundation.slider.js
ncoden: @kball `this.options`*`.enabled`* is initialized with `$element.data`*`.enabled`*

kball:
Gonna move forward with this

Issue #8496, opened 2016-03-31
[Slider] data-start is breaking the slider drag


christianallred:
1. Any slider with a data-start value doesn't appear to build properly
2. Try to drag slider it only moves a small amount in the middle of the slider. 

http://codepen.io/EvilEeyore/pen/GZMgvb

whokilledtheelectricmonk:
You need to load jQuery over HTTPS - it is being blocked in your codepen

briantan888:
created PR #8506  to fix the bug. 2 lines fix but it is failing the Travis CI build.

kball:
Fixed in #8506 and will be in 6.2.1

PR #8506, opened 2016-04-01
Fix slider bug when start is not 0

web2client:
Address issue #8496 

Line 141: pctOfBar should take account of start value
Line 278: value should have start offset.


kball:
Hi @briantan888 this looks good.  Can you set up a unit test for this case in `test/javascript/components/slider.js`?  You can use the tests in `test/javascript/components/toggler.js` as an example of how these can be structures.  See:  https://github.com/zurb/foundation-sites/blob/develop/test/javascript/components/toggler.js

Thanks!

briantan888:
@kball  I have no idea how to test the case asynchronously as bug manifest itself interactively.

PR #8510, opened 2016-04-02
Various improves on flex-grid

ncoden:
**Add flex-grid wrap option**

Add a flex-grid-row wrap option, allowing to create a row which wrap (by default) or not.

**Use "expand" explicit value for default flex-grid-row size**

I often have in my semantic Scss many column definition aligned, and I think its behaviours could be more understandable with an explicit value (like `expand`) for columns which take the remaining place, instead of `null`.
_`expand` would be always the default behaviour_

Something like this, similar to what `foundation-apps` purpose :

``` scss
.my-row {
  .col-a { @include flex-grid-column(1); }
  .col-b { @include flex-grid-column(expand); }
  .col-c { @include flex-grid-column(shrink); }
}
```

Instead of

``` scss
.my-row {
  .col-a { @include flex-grid-column(1); }
  .col-b { @include flex-grid-column; }
  .col-c { @include flex-grid-column(shrink); }
}
```

### DIFF
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -19,9 +19,11 @@
   $width: $grid-row-width,
   $columns: null,
   $base: true,
-  $gutter: $grid-column-gutter
+  $gutter: $grid-column-gutter,
+  $wrap: true
 ) {
   $margin: auto;
+  $wrap: if($wrap, wrap, nowrap);
 
   @if index($behavior, nest) != null {
     @include grid-row-nest($gutter);
@@ -39,7 +41,7 @@
 
   @if $base {
     display: flex;
-    flex-flow: row wrap;
+    flex-flow: row $wrap;
   }
 
   @if $columns != null {
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -129,7 +129,9 @@
 
 /// Changes the width flex grid column.
 /// @param {Mixed} $columns [expand] - Width of the column. Refer to the `flex-grid-column()` function to see possible values.
-@mixin flex-grid-size($columns: expand) {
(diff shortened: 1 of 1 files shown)


kball:
@andycochran @brettsmason Can you take a look at this?  Do you think this change of defaults makes sense?

kball:
ping @andycochran @brettsmason thoughts on this change of defaults?

brettsmason:
I've personally never used the flex grid as I need to support ie9 for my projects, so I'm not familiar with it. I'll try and take a look but hopefully @andycochran will have more knowledge on this.

andycochran:
The changes seem logical, but I too have little experience with Flex Grid (also thanks to ie9). @kball, do we have a Yetinaut with flexbox expertise? @JeremyEnglert? 

My one question is the change from `null` to `expand`. Is this just semantics? What happens when it's left `null`—e.g. if I use this mixin: 

```scss
.col-b { @include flex-grid-column; }
```

ncoden:
@andycochran Yes, it is only semantical.

With this PR, we have :
```scss
// .col-a is equal to .col-b
.col-a { @include flex-grid-column(expand); }
.col-b { @include flex-grid-column; }

// but .col-c is not accepted
.col-b { @include flex-grid-column(null); } // because a "null" value is meaningless
```

Should we accept when a `null` value is explicitly passed to the mixin ?

andycochran:
If a `null` value can currently be passed to the mixin, we shouldn't drop support for that. Don't want to break apps.

ncoden:
@andycochran Fixed (and rebased)

kball:
@andycochran can you take a look at this and merge if it seems good to you?

kball:
Having used flexbox and flex grid more now, digging into this it makes perfect sense and I think is good to go.

PR #8515, opened 2016-04-03
Add Option for Collapsing Foundation Tabs

rspraymond:
Collapsing tabs was a feature I needed on a current project. I end up writing custom js to do it. But I feel like collapsing tabs should be an option in any tab library. 

I'm new to contributing so if you have any feedback/questions please let me know. I tried to follow coding standards I found with brief research and following structure of existing codebase.


rspraymond:
Not sure I did that right.

PR #8516, opened 2016-04-03
Add Option for Collapsing Foundation Tabs

rspraymond:
Collapsing tabs was a feature I needed on a current project. I end up writing custom js to do it. But I feel like collapsing tabs should be an option in any tab library. 

I'm new to contributing so if you have any feedback/questions please let me know. I tried to follow coding standards I found with brief research and following structure of existing codebase.

### DIFF
--- a/docs/pages/tabs.md
+++ b/docs/pages/tabs.md
@@ -109,3 +109,33 @@ Add the `.vertical` class to a tabstrip to stack tabs vertically. You can also p
   </div>
 </div>
 ```
+
+---
+
+## Collapsing Tabs
+
+Add the attribute `data-active-collapse="true"` to a tabstrip to collapse active tabs.
+
+```html_example
+<ul class="tabs" data-active-collapse="true" data-tabs id="collapsing-tabs">
+  <li class="tabs-title is-active"><a href="#panel1c" aria-selected="true">Tab 1</a></li>
+  <li class="tabs-title"><a href="#panel2c">Tab 2</a></li>
+  <li class="tabs-title"><a href="#panel3c">Tab 3</a></li>
+  <li class="tabs-title"><a href="#panel4c">Tab 4</a></li>
+</ul>
+
+<div class="tabs-content" data-tabs-content="collapsing-tabs">
+  <div class="tabs-panel is-active" id="panel1c">
+    <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
+  </div>
+  <div class="tabs-panel" id="panel2c">
+    <p>Vivamus hendrerit arcu sed erat molestie vehicula. Sed auctor neque eu tellus rhoncus ut eleifend nibh porttitor. Ut in nulla enim. Phasellus molestie magna non est bibendum non venenatis nisl tempor. Suspendisse dictum feugiat nisl ut dapibus.</p>
+  </div>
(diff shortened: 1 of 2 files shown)


kball:
Seems reasonable to me.  @Owlbertz @colin-marshall @designerno1 can one of you take a look at this?

rspraymond:
@Owlbertz @colin-marshall @designerno1 @kball Any feedback on this request?

Thanks

Owlbertz:
I will have a look tomorrow or during the weekend.

Owlbertz:
Looks good to me. Only thing I would like to add is if it would be good to have a private method to handle showing and hiding the tabs so all the class adding/removing and setting of ARIA-attributes needs less repeating. Whats your though? @rspraymond @kball @colin-marshall @designerno1

colin-marshall:
@rspraymond looks good! One thing I noticed is that when a tab is collapsed, that tab stays focused until you click somewhere else. Is this the way we want it to behave?

@Owlbertz agreed.

designerno1:
great feature looks good for me
i also think for usability reasons, focus should go away when tab is collapsed

colin-marshall:
Couple more notes regarding focus:
- If the tabs are collapsed, hitting enter will uncollapse the last active tab, because the focus is retained after collapsing
- An uncollapsed tab that's in focus will not collapse when you hit enter 

Keeping focus after collapsing a tab might actually be the more accessible way to go, but if we do that, I think we should make it so that hitting enter will collapse a tab, as well as uncollapse it.

rspraymond:
@Owlbertz @colin-marshall @designerno1 @kball My last commit addresses your concerns. If the consensus is to keep the tab focused and open and collapse with enter then I can take a look at that.

Thanks for your feedback!

Owlbertz:
For the latest version it is not possible to collapse the tabs using the keyboard, I think that's what @colin-marshall meant. In my opinion using the keyboard should allow the same interactions as using the mouse, so I would vote for a support to collapse/open via enter and space. Would need some check for the `activeCollapse` in the `Foundation.Keyboard.handleKey#open` call I guess.

Also I'd like to see the events aligned. The `collapse.zf.tabs` is now fired in the `_handleCollapse` function and the `_openTab()` function fires no event. I'd suggest having the events fired in the functions that actually do stuff, i.e. `_collapseTab()` and `_openTab()`.

rspraymond:
@Owlbertz @colin-marshall @designerno1 @kball

I have changed code back to keep tab focused. Using keyboard to open and close should work when pressing space or enter.

To allow us to use the same function to close the old tab and collapse the current tab I elected to keep the events triggered inside the _handleTabChange function as it was before collapsing functionality was added.

Let me know what you think!

Raymond

kball:
Awesome, @colin-marshall @Owlbertz @designerno1 can you take a look at the current state of this, and if at least 2 of you chime in OK go ahead and merge

rspraymond:
Hey @colin-marshall @Owlbertz @designerno1,

If you need me to do anything else to get this feature moving forward, let me know.

Thanks

Raymond

rspraymond:
@kball Any word back on this?

Owlbertz:
Hi all,
sorry that it took me so long to check this out. This looks good to me from a technical perspective and I also understand that this might be a requirement, so I'd say merge this to 6.3.

rspraymond:
@kball @Owlbertz Sounds good. Let me know if you need anything from me.

rspraymond:
@kball Anything else we need to happen?

PR #8518, opened 2016-04-04
Add dropdown background

ncoden:
Add `$dropdown-background` config to allow to customize dropdown background color.

The default background value is the same (`$body-background`).

### DIFF
--- a/scss/components/_dropdown.scss
+++ b/scss/components/_dropdown.scss
@@ -10,6 +10,10 @@
 /// @type List
 $dropdown-padding: 1rem !default;
 
+/// Background for dropdown panes.
+/// @type Color
+$dropdown-background: $body-background !default;
+
 /// Border for dropdown panes.
 /// @type List
 $dropdown-border: 1px solid $medium-gray !default;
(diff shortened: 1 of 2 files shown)


ncoden:
@colin-marshall @kball

kball:
Seems reasonable to me.  What do you think @andycochran @brettsmason ?

ncoden:
BTW should I always tag someone on PR ? If yes, who, `@random()` ?

kball:
Good question @ncoden.  I don't think you should need to, we've just had a big backlog build up recently with the foundation for emails launch, some other ZURB projects, and figuring out some transitions in personel.  One of the hopes of expanding the Yetinaut program is to reduce this type of volatility.

brettsmason:
@kball no problems with it here, it makes sense. One question, do we need to add variables to the settings file? I'm sure I remember Geoff saying that the file is generated each patch. Would be good to know for future issues/Pls.

andycochran:
Agreed. :+1: This looks good / would also like clarification on `_settings.scss` (this also came up in #8431).

kball:
@brettsmason and @andycochran yes the _settings.scss is autogenerated as a part of our packaging/deploy process.  I don't think it hurts to update it as a part of the PR, but it shouldn't be necessary.

PR #8519, opened 2016-04-04
Update tooltip.md

iamrobert:
I think this will help clear up that in former versions we could use html for tooltips - and now we can't.


Owlbertz:
I wouldn't put this information in the component description, but rather put it into an info box like it's done for the [dropdown documentation](http://foundation.zurb.com/sites/docs/dropdown.html).

kball:
I agree with @Owlbertz. Also, bear in mind that this has been addressed in https://github.com/zurb/foundation-sites/pull/8527 so in v6.3 an 'allowHtml' option will be available on tooltips.

PR #8522, opened 2016-04-04
Update panini.md

EHackettTriCore:
I was working on adding a new Panini helper but I noticed the missing step to update your gulpfile.babel.js so that it looks for the helper directory.

### DIFF
--- a/docs/pages/panini.md
+++ b/docs/pages/panini.md
@@ -183,6 +183,22 @@ module.exports = function(options) {
   return bolder;
 }
 ```
+
+```
+// Example  gulpfile.babel.js
+function pages() {
+  return gulp.src('src/pages/**/*.html')
+    .pipe(panini({
+      root: 'src/pages',
+      layouts: 'src/layouts',
+      partials: 'src/partials',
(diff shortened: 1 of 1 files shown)


kball:
Hi @EHackettTriCore thanks for filling in the missing step.  Can you actually add a snippet of what this looks like in the gulpfile for users who are unfamiliar with gulp?  Thanks.

EHackettTriCore:
@kball I added the default panini snippet with the helper line included. I'm not sure if that's more or less confusing. It almost might be worth rewriting it into step 1. create helper file, step 2 update gulp file, step 3. call in file.

Also after looking at this a little more I don't believe there is any conflict or error if you by default include the `helpers: 'src/helpers',` line in the default foundation configuration but no helper directory exists in src. I.e. I don't if it's worth making this part of the documentation and having users add it themselves or just putting it into the default setup for foundation.

kball:
@EHackettTriCore  I think this looks good, and I like having it in the docs in case folks aren't coming straight from a ZURB template.  If you can resubmit this against the develop branch I'll merge it.  Thanks!

PR #8523, opened 2016-04-04
Updating wording for mixin.scss regarding breakpoint and adding zero breakpoint to settings

DusanMilko:
Updating wording for mixin.scss regarding breakpoint and adding zero breakpoint to settings

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -86,6 +86,7 @@ $breakpoints: (
   xxlarge: 1440px,
 );
 $breakpoint-classes: (small medium large);
+$-zf-zero-breakpoint: small;
 
 // 3. The Grid
 // -----------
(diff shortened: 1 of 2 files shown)


andycochran:
Variables prefixed with `$-zf-` are intentionally not exposed in `_settings.scss`. Also, small/smallest/etc has been debated and the default zero breakpoint is gonna remain `small`.

DusanMilko:
Shouldn't the smallest/zero breakpoint be customizable and not fixed to small?

andycochran:
Yup. It is customizable. It's set in `_breakpoint.scss`, [line 26](https://github.com/zurb/foundation-sites/blob/21b75adf577a4da3ed1a08228abba4eee8bfc603/scss/util/_breakpoint.scss#L26).

PR #8524, opened 2016-04-04
Update Panini Docs : Custom Helpers

EHackettTriCore:
I was working on adding a new Panini helper but I noticed the missing step to update your gulpfile.babel.js so that it looks for the helper directory.

Update Panini docs https://github.com/zurb/foundation-sites/pull/8522


EHackettTriCore:
@kball I believe this is now associate with the development branch.

PR #8525, opened 2016-04-04
Redo clobbered non-flex Button Group styles

### LINKED ISSUE #8520: $buttongroup-spacing implemented differently for $global-flexbox true and false
The spacing within a `.buttongroup` is implemented differently, depending weather the variable `$global-flexbox` is set `false` or `true`. 
In a flexbox context, the spacing is implemented by a `margin-right` 
In a non-flexbox context the spacing is implemented by a white `border-right`

The border-method causes _funky_ results, when `$buttongroup-spacing` ramped up to values bigger, than the default button-border-width

1. Set the value of `$buttongroup-spacing` to a high value. e.g. 10px; 
2. Change the value of `$global-flexbox` from false to true and vice versa. 
3. Compare the results

**What happened?**
The Border-Method the color of the top border of the button to _leak_ into the spacing:

buttongroup-spacing

The problem of the border-method becomes most visible, when placing a button-group on a colored element, like a top bar. The spacing is running in the body-color:

buttongroup-spacing-bg

**Source of this Problem**

This difference is caused by this code within the[ button group](https://github.com/zurb/foundation-sites/blob/develop/scss/components/_button-group.scss#L52). Starting at line 51:

```scss
     @if $global-flexbox {
        margin-right: $buttongroup-spacing;
      }
      @else {
        border-#{$global-right}: $buttongroup-spacing solid $body-background;
      }    
```

** Suggestion **

If the border-method is not absolutely necessary for browser compatibility, i suggest to remove the conditional completely and use for both cases the margin-method. 

----

This Issue is related to the discussion in issue https://github.com/zurb/foundation-sites/issues/8447

andycochran:
Closes #8520.

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -29,12 +29,14 @@ $buttongroup-expand-max: 6 !default;
 ) {
   @include clearfix;
   margin-bottom: $buttongroup-margin;
-  font-size: 0;
 
   @if $global-flexbox {
     display: flex;
     flex-wrap: nowrap;
     align-items: stretch;
+  }
+  @else {
+    font-size: 0;
   }
 
   #{$child-selector} {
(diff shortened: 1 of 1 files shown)


kball:
Looks good to me

Issue #8521, opened 2016-04-04
[6.2 Equalizer] Setting up an empty equalized list produces an error


acarnwath:
When setting up a dynamically-populated list, there is nothing within to be watched. This causes an error which halts initialization:

> foundation-custom.js:199 TypeError: Cannot read property 'top' of undefined(…)

This is caused by the way that `lastElTopOffset` is set:

```
 getHeightsByRow(cb) {
    var lastElTopOffset = this.$watched.first().offset().top,
        groups = [],
        group = 0;
```

By first checking to see that there is something being watched, this error will not be triggered.

```
getHeightsByRow(cb) {
    var lastElTopOffset = this.$watched.length ? this.$watched.first().offset().top : 0,
        groups = [],
        group = 0;
```

ncoden:
:+1: Thanks. Pull request: https://github.com/zurb/foundation-sites/pull/8526

PR #8526, opened 2016-04-05
Fix setting up of an empty equalized list #8521

ncoden:
Fix https://github.com/zurb/foundation-sites/issues/8521


kball:
Looks good, thanks @ncoden.

Issue #8508, opened 2016-04-01
[Grid] Responsive Gutters with Unresponsive Gutters


andycochran:
Would it be possible to use responsive gutters with the exception of some rows/columns which you'd always want to use a specific gutter size? 

**Example use case:** 
You want responsive gutters to add loads of space between your content and sidebar on large screens. And your content is two columns. It'd be nice if the content columns were separated by the `small` value in `$grid-column-gutter`, but they're spaced just as far apart as the content is from the sidebar. So you basically just get a three-column layout. 

**Possible solution:**

``` html
<div class="row">
  <div class="large-8 columns">
    <div class="row small-gutters"><!-- .[size]-gutters could force unresponsive gutters -->
      <div class="large-6 columns">Content A</div>
      <div class="large-6 columns">Content B</div>
    </div>
  </div>
  <div class="large-4 columns">
    ...Sidebar...
  </div>
</div>
```

Maybe this is already possible through semantic classes + mixins?

ncoden:
:+1: Can be usefull.
But I prefer the `.gutter-{size}` syntax. What do you think ?

Pull request: https://github.com/zurb/foundation-sites/pull/8528

christianmagill:
I wrote a small mixin that does variant gutters based on multipliers. Right now I am just overwriting the css, but could probably be better implemented within Foundation I am sure.

The idea is that your gutters are still responsive based on breakpoint, however you also have variants that you can select within each breakpoint.

https://github.com/christianmagill/foundation--variant-gutter-classes

andycochran:
@ncoden It looks like you resolved this in https://github.com/zurb/foundation-sites/pull/8528?

andycochran:
Although, there's no mention of `.gutter-{size}` in the docs.

kball:
@andycochran can you verify if this is working and perhaps add to the docs?

PR #8528, opened 2016-04-05
Add static grid gutters #8508

ncoden:
**Add static (unresponsive) grid gutters**
See: https://github.com/zurb/foundation-sites/issues/8508
Use: `.gutter-{size}` class on a row

Example:

``` html
<div class="row gutter-small"> <!-- Gutters always have a "small" width -->
  <div class="large-6 columns">...</div>
  <div class="large-6 columns">...</div>
  ...
</div>
```

**Standardize grid gutters functions, mixins and arguments.**

Adds/Removes:
- Add `@function grid-column-gutter( {breakpoint} )`: Get a gutter size for a given breakpoint.
- Add `@mixin grid-column-size( {px or breakpoint} )`: Set the gutters on a column.
- Depreciate `@mixin grid-column-uncollapse`. Use `grid-column-gutter` instead.
- Use these new features in all `(flex-)grid` mixin & functions. Use `@function grid-column-gutter` instead of modifying `$grid-column-gutter` to support its possible single value.

Other changes:
- Argument naming: use `$gutter` for a single size, and `$gutters` for a size map.
- Fix `$gutter` (now `$gutters`) argument descriptions.

_Required for the previous feature.
Anyway, all the "groups of properties" of a component (i.e. `size`, `align`, `color`, or specific behavior like `collapse`, `expand`...) should have their own `@mixin`, to allow a component customization in responsive/semantic cases_

### DIFF
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -63,18 +63,13 @@
   // Gutters
   @if type-of($gutter) == 'map' {
     @each $breakpoint, $value in $gutter {
-      $padding: rem-calc($value) / 2;
-
       @include breakpoint($breakpoint) {
-        padding-left: $padding;
-        padding-right: $padding;
+        @include grid-col-gutter($value);
       }
     }
   }
   @else if type-of($gutter) == 'number' and strip-unit($gutter) > 0 {
-    $padding: rem-calc($gutter) / 2;
-    padding-left: $padding;
-    padding-right: $padding;
+    @include grid-col-gutter($value);
   }
 
   // Last column alignment
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -52,24 +52,17 @@
 /// Creates a grid column.
 ///
 /// @param {Mixed} $columns [$grid-column-count] - Width of the column. Refer to the `grid-column()` function to see possible values.
-/// @param {Number} $gutter [$grid-column-gutter] - Spacing between columns.
+/// @param {Mixed} $gutters [$grid-column-gutter] - Spacing between columns. Refer to the `grid-column-gutter()` function to see possible values.
 @mixin grid-column(
   $columns: $grid-column-count,
-  $gutter: $grid-column-gutter
+  $gutters: $grid-column-gutter
 ) {
   @include grid-column-size($columns);
   float: $global-left;
 
   // Gutters
-  @if type-of($gutter) == 'map' {
-    @each $breakpoint, $value in $gutter {
-      @include breakpoint($breakpoint) {
-        @include grid-col-gutter($value);
-      }
-    }
-  }
-  @else if type-of($gutter) == 'number' and strip-unit($gutter) > 0 {
-    @include grid-col-gutter($value);
+  @include -zf-each-breakpoint {
+    @include grid-column-gutter($-zf-size, $gutters);
   }
 
   // Last column alignment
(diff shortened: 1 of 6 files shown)


kball:
@ncoden I like the direction this is going. You should be aware though that $grid-column-gutter currently supports setting it to a single (scalar) value as well as the responsive approach... I think this would break in that case.  Probably in that situation these should just degenerate to the same gutter, yes?

andycochran:
@kball Ah, right! This'll need to account for if you're not using responsive gutters (e.g. `$grid-column-gutter` is a single value).

ncoden:
:+1: hum. Actually, if `$grid-column-gutter` is a single value, it is transformed to a `(small: $value)` map.
So it would generate a useless `.gutter-small` class.

ncoden:
It's dirty because we're assuming there is always a smallest breakpoint called "small".

So I purpose to not modify `$grid-column-gutter`, and create:
```scss
@function grid-column-gutter( $breakpoint )
```
that return *the gutter* if it is a single value, or *the gutter for a given breakpoint* if it is a map.

andycochran:
You can use `nth(map-keys($breakpoints),1)` to get the smallest (zero) "breakpoint."

ncoden:
> You can use nth(map-keys($breakpoints),1) to get the smallest (zero) "breakpoint."
Can there be no breakpoint ? Does the "zero breakpoint" always exists and be zero ?

I also see a problem with `grid-column-uncollapse`:
```scss
@mixin grid-column-uncollapse($gutter: $grid-column-gutter) { ... }
```
- `$grid-column-gutter` is always a map, and is used as padding. We must remove this bugged default value`
- So `$grid-column-uncollapse` is an alias of `$grid-column-gutter`
- And `$grid-column-collapse` is an alias of `$grid-column-gutter(0)`
- So it would be better to only have:

```scss
@include grid-column-gutter( {size_in_px} );
@include grid-column-gutter( {breakpoint} ); // "small", "medium"...
```

And maybe
```scss
@include grid-column-gutter( collapse ); // alias for 0
```

*CSS classes keep unchanged*

andycochran:
@ncoden Yeah. A zero breakpoint is required as the first key/value of the `$breakpoints` map. 

But [the error](https://github.com/zurb/foundation-sites/blob/12317b1854283b3a2708d586f1141d523684f8cd/scss/util/_breakpoint.scss#L24) is misleading. 

This:
```scss
@error 'Your smallest breakpoint (defined in $breakpoints) must be set to "0".';
```

Should probably be:
```scss
@error 'The first key in the $breakpoints map must have a value of "0".';
```

And a note should also be added to [the docs page](http://foundation.zurb.com/sites/docs/media-queries.html#changing-the-breakpoints).

DaSchTour:
BTW, since #8318 is merged you can use $-zf-zero-breakpoint to get the smallest breakpoint.

ncoden:
@andycochran @DaSchTour @kball Ok, I will add theses changes.

ncoden:
I done these changes, and standardize/factorized all the gutter functions/mixins/uses. See all the changes in this PR description.

I have two questions:
- Should I update the documentation ? (docs/pages/grid.md)
- Should I keep `@mixin grid-column-uncollapse` deleted, or would it be better to depreciate it (redirection to `@mixin grid-column-size` and throw a warning) ?

kball:
@ncoden I prefer depreciations than outright deletions, so that sounds like a good step.  Updating the documentation would be great.  @andycochran can you try this out and see if the code changes are good to merge?

kball:
@ncoden looks like this branch has gotten stale and needs some conflicts resolved before it can be merged.  @andycochran can you give this trial run?

ncoden:
Ah sorry, i forgot this PR ;)
I rebase it...

ncoden:
@kball @andycochran  It's rebased (and fixed). :)

scss/grid/_column.scss
```diff
@@ -87,12 +75,12 @@
 
 /// Creates a grid column row. This is the equivalent of adding `.row` and `.column` to the same element.
 ///
-/// @param {Number} $gutter [$grid-column-gutter] - Width of the gutters on either side of the column row.
+/// @param {Mixed} $gutter [$grid-column-gutter] - Width of the gutters on either side of the column row. Refer to the `grid-column-gutter()` function to see possible values.
```
kball: comment @param name should prob get updated to $gutters

kball:
@andycochran @brettsmason @JeremyEnglert as our scss expert yetinauts, does this seem ready to you?

scss/grid/_column.scss
ncoden: Thanks

ncoden: @kball Fixed

PR #8534, opened 2016-04-06
add specificity for .expanded.row .[size]-collapse.row

### LINKED ISSUE #8533: [Grid] responsive collapsed row not flush left/right
Responsive collapsed rows spill into the padding of their parent column. A `.row.collapse` works fine. But `.row.[size]-collapse` doesn't. 

Here's an example of the spillage [in the docs](http://foundation.zurb.com/sites/docs/grid.html#collapse-uncollapse-rows): 

screen shot 2016-04-05 at 1 57 58 pm

And here's a working example: http://codepen.io/andycochran/pen/WwXKxJ

andycochran:
Fixes #8533. 

The `.expanded` class had more specificity than the `.[size]-collapse` classes, so they weren't getting the zero side margins and were spilling past the edges. So I added an explicit selector for responsive collapsed rows that are within expanded rows.

### DIFF
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -116,7 +116,8 @@
     .#{$-zf-size}-#{$collapse} {
       > .#{$column} { @include grid-col-collapse; }
 
-      .#{$row} {
+      .#{$row},
+      .#{$expanded}.#{$row} &.#{$row} {
         margin-left: 0;
         margin-right: 0;
       }


ncoden:
:-1: We shouldn't play with specificity. In a perfect approach, all classes should have the same specificity, with the standard composants declared first and the modifiers declared last.

*to Zurb members: we should really switch to BEM*

Is it possible to move `.expanded` **after** `.*-collapse` ? It should be by this way we define priority.

andycochran:
@ncoden, I totally understand not wanting to mess w/ specificity. But unfortunately, it's not a matter of cascade. 

Expanded classes already have a relatively high level of specificity to handle nesting: 
```
.expanded.row .row
```
All I've done is add in the responsive classes to work with the existing level of specificity: 
```
.expanded.row .[size]-collapse.row
```
My commit message was probably scarier than in shoulda been.

kball:
👍

PR #8542, opened 2016-04-07
Add checkbox-group validation

web2client:
Propose adding checkbox group for required validation. Wrap the icheckbox fields in container with class `checkbox-group` with`data-min-required="n"` attribute. `data-min-required` is default to 1. In field `required` attribute check is still supported.

Example for checkbox-group required validation

```

<div class="row">
  <div class="medium-4 column">
    <label for "hobby-group">Select two or more hobbies</label>
  </div>
  <div id="hobby-group" class="medium-8 column checkbox-group" data-min-required="2">
    <input name="hobby" type="checkbox" value="reading">Reading
    <input name="hobby" type="checkbox" value="writing">Writing
    <input name="hobby" type="checkbox" value="music">Music
    <input name="hobby" type="checkbox" value="sport">Sport
    <br />
    <div class="form-error">At lease two hobbies must be selected</div>
  </div>
</div>
```

If no checkbox-group is found, required check on checkbox field still holds.

Please provide feedback and comments.

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -83,11 +83,6 @@ class Abide {
     var isGood = true;
 
     switch ($el[0].type) {
-      case 'checkbox':
-        isGood = $el[0].checked;
-        break;
-      case 'radio':
-        break;
       case 'select':
       case 'select-one':
       case 'select-multiple':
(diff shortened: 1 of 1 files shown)


justrhysism:
I can see what you're trying to achieve here, but I worry that the solution here steps _away_ from W3C's recommended approach for required radio buttons.

I feel like this feature would be better suited as a plugin leveraging Abide's custom validator system.

briantan888:
@justrhysism I removed radio-group as I now think the new 6.2.1 validateRadio would have taken care of it.

Anyway I find that `required` attribute on single radio button meaningless. It is something like: please choose a color, but you must choose Red. `required` attribute should be in all radio button per W3C recommendation, and it means must choose one, any one.

kball:
I kinda like this direction, though I think as it is a feature change it should be against v6.3 rather than develop.  Lets get some more feedback from JS-focused yetinauts.  @colin-marshall @Owlbertz what do you think of this proposal?

colin-marshall:
I'm all for it if everybody else is on board. Nice work @briantan888. I agree with @kball that this should go on v6.3.

This should probably have unit tests added as well as documentation before it gets merged.

justrhysism:
Yeah now that the radio button stuff has been removed I think that this is a good feature.

kball:
Closing as this is against develop.  Please reopen against v6.3

PR #8547, opened 2016-04-07
Revert "Update MediaQueries to initialize if not initialized already."

kball:
Reverts zurb/foundation-sites#8192

Unfortunately, upon testing this PR turned out to cause a number of issues related to responsiveness... in particular, it breaks _getCurrentSize(), which caused all sorts of issues.  I looked for a quick fix and didn't find one, so I'm going to revert this for now.

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -4,11 +4,6 @@
 
 // Default set of media queries
 const defaultQueries = {
-  small : 'only screen and (min-width: 0em)',
-  medium : 'only screen and (min-width: 40em)',
-  large : 'only screen and (min-width: 64em)',
-  xlarge : 'only screen and (min-width: 75em)',
-  xxlarge : 'only screen and (min-width: 90em)',
   'default' : 'only screen',
   landscape : 'only screen and (orientation: landscape)',
   portrait : 'only screen and (orientation: portrait)',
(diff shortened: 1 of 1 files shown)


merlinpatt:
@kball how is it breaking `_getCurrentSize()`? I'm not running into that.

kball:
@merlinpatt _getCurrentSize is occasionally returning things like 'portrait' and 'landscape' rather than the size related queries... this was resulting in issues wrt things like the responsive menu upon resizing.

merlinpatt:
@kball So I see the issue now. The issue is with the `defaultQueries` object which I had assumed should be used. I write a quick fix for this.

But before I do, do you know what the `defaultQueries` object is supposed to be for? Prior to my pull request, it looks like it wasn't used at all.

kball:
Good question, I'm not sure.  It looks like it was originally being used to help fill out the queries list so that the utility can make it easy for the JS to see all of the media queries that are available.

Now that I think about it, the fix might be as easy as moving the non-size related queries ahead of the size ones in the `defaultQueries` object, not sure.

merlinpatt:
I don't know if object order is always guaranteed so if it's guaranteed, then that would work.

Alternatively, we could just remove the non-size queries or move them into a separate object. I'm in favor of just removing them entirely.

kball:
I think there is value in having some way to get access to the full list (including non-size) via javascript (ie in the .queries attribute)... in my mind this module is about exposing the queries that are otherwise just in sass/css into javascript.  That said, I'm not super opinionated about where/how they are exposed, so long as it makes sense and is straightforward to access.

merlinpatt:
What about having two attributes: `queries.size`, `queries.orientation`?

kball:
@merlinpatt I like that... however that would be an api change so would need to targeted at 6.3.

merlinpatt:
For a more transitional change, what about keeping the `queries` object like it is and adding on those additional properties? So we'd get something like

```
queries = {
  small: 'only screen and (min-width: 0em)',
  ...
  size: {
    small: 'only screen and (min-width: 0em)',
    ...
  },
  orientation: {
    landscape : 'only screen and (orientation: landscape)',
    ...
  }
}
```

PR #8550, opened 2016-04-07
[Bugfix] Drilldown - Cleaning attached events correctly on destroy() call

gehasia:
Having a responsive drilldown/dropdown clickable only menu configured like this :
`data-responsive-menu="drilldown medium-dropdown" data-click-open="true"  data-disable-hover="true"`
If you load the page (dropdown mode), then resize to small to get your drilldown, then go back to medium/large to get back your dropdown the event handler for drilldown is still registered in handlerQueue.handlers[0] while the dropdown one is in handlerQueue.handlers[1], thus the drilldown handler catch the request (first array item...) and the dropdown handler never got it...
In this specific case you can't click your dropdown menu anymore after resizing (click event is still registered to the destroyed drilldown menu...). This bug didn't appeared in dropdown hover mode because drilldown is click only mode, and then events does not "collide".

This patch try to unregister every event handlers attached to submenuAnchors on Drilldown, the same way as it is constructed in _prepareMenu in place of the
`.off('.zf.drilldown').end().off('zf.drilldown')`
which seems to be inconsistent.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -315,10 +315,8 @@ class Drilldown {
     this.$element.unwrap()
                  .find('.js-drilldown-back, .is-submenu-parent-item').remove()
                  .end().find('.is-active, .is-closing, .is-drilldown-submenu').removeClass('is-active is-closing is-drilldown-submenu')
-                 .end().find('[data-submenu]').removeAttr('aria-hidden tabindex role');
-    this.$submenuAnchors.each(function() {
-      $(this).off('.zf.drilldown');
-    });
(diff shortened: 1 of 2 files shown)


gehasia:
Sorry, i messed up with my commit, mixing a sticky bugfix with this branch...bugfix which has nothing to do with this PR... And i reverted the good patch instead of the bad at first...
Sorry for the mess...it's the first time i create a PR...

kball:
The change seems reasonable, but can you please resubmit without the extra changesets & revisions?  You can probably just cherrypick the commits in question into a new branch.

PR #8558, opened 2016-04-08
tabs:destroy fixed; reponsiveToggle:destroy

dragthor:
foundation.tabs::destroy() fixed to unbind itself from global $(window).on('changed.zf.mediaquery'), rather than unbinding everyone.

Implemented foundation.responsiveToggle::destroy().

Added a couple of asserts to core.js test.


kball:
Sweet, good fixes

PR #8567, opened 2016-04-11
Update foundation.tooltip.js to enable to parse a postionClass when c…

JamesJGarner:

Happy coding! :)

…reating Tooltips using js

Changed code so the positionClass checks inside this.options rather than just looking at the elements css


kball:
Seems reasonable to me

PR #8570, opened 2016-04-11
Add (flex-)grid-row size modifier

ncoden:
Add a `grid-row` and `flex-grid-row` width modifier

``` scss
@mixin grid-row-width($width)
// Set a grid row width

// $width : [$grid-row-width] Maximum width of the row. Set to `expand`
//          to make the row taking the full width.
```

Examples with a semantic sass:

``` scss
.my-row {
  @include grid-row($width: 900px);
  // Or
  @include grid-row($width: expand);
}
```

``` scss
.my-row {
  @include grid-row;

  @include breakpoint(medium) {
    @include grid-row-width(expand);
  }
}
```

**Note**
There is lot of similar work to do in `grid` and `flex-grid`. As said in https://github.com/zurb/foundation-sites/pull/8528 :

> Anyway, all the "groups of properties" of a component (i.e. size, align, color, or specific behavior like collapse, expand...) should have their own `@mixin`, to allow a component customization in responsive/semantic cases

So if you searched what to do to improve foundation, you know what to do ;) !

### DIFF
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -45,7 +45,7 @@
 
     // Expanded (full-width) row
     &.#{$expanded} {
-      max-width: none;
+      @include grid-row-size(expand);
 
       .#{$row} {
         margin-left: auto;
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -10,13 +10,13 @@
 ///
 /// @param {Keyword|List} $behavior [null]
 ///   Modifications to the default grid styles. `nest` indicates the row will be placed inside another row. `collapse` indicates that the columns inside this row will not have padding. `nest collapse` combines both behaviors.
-/// @param {Number} $width [$grid-row-width] - Maximum width of the row.
+/// @param {Keyword|Number} $size [$grid-row-width] Maximum size of the row. Set to `expand` to make the row taking the full width.
 /// @param {Number} $columns [null] - Number of columns to use for this row. If set to `null` (the default), the global column count will be used.
 /// @param {Boolean} $base [true] - Set to `false` to prevent basic styles from being output. Useful if you're calling this mixin on the same element twice, as it prevents duplicate CSS output.
 /// @param {Number} $gutter [$grid-column-gutter] - Gutter to use when inverting margins, in case the row is nested.
 @mixin flex-grid-row(
   $behavior: null,
-  $width: $grid-row-width,
+  $size: $grid-row-width,
   $columns: null,
   $base: true,
   $gutter: $grid-column-gutter
@@ -32,7 +32,7 @@
     }
   }
   @else {
-    max-width: $width;
+    @include grid-row-size($size);
     margin-left: auto;
     margin-right: auto;
   }
@@ -172,7 +172,7 @@
 
     // Expanded row
     &.expanded {
-      max-width: none;
+      @include grid-row-size(expand);
     }
 
     &.collapse {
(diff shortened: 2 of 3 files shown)


ncoden:
poke @kball 
See also : https://github.com/zurb/foundation-sites/pull/8528, https://github.com/zurb/foundation-sites/pull/8510

kball:
I think this direction makes sense, but I'd like to hear from our scss experts @andycochran @brettsmason

brettsmason:
Yeah that definately makes sense to me and seems a good approach - it feels very similar to `grid-column-size` usage.

ncoden:
@brettsmason It is intentional :)
Should I rename this to `grid-row-size`, `flex-grid-row-size` and `$size` ?

kball:
@brettsmason from your perspective is this good to go?

brettsmason:
@ncoden Sorry for the delay. I think for consistency we should probably name these `grid-row-size` / `flex-grid-row-size` and `$size` so they match with `grid-column-size` as you suggested.

Could you change that and I'll merge it :+1: 

Also looking forward to any other suggestions you have like this!

ncoden:
@brettdewoody I renamed to `grid-row-size`, `flex-grid-row-size` and `$size`. But we still have the configuration variable `$grid-row-width`.
I let it unchanged or I rename it and add a deprecation warning ?

brettsmason:
@ncoden Sorry I forgot there was an existing `$grid-row-width` variable. I better check on this with some others: @andycochran @kball could you take a look and see what you think is the best approach or if we should keep them as `grid-row-width`?

andycochran:
Isn't `$grid-row-width` more semantic than `$grid-row-size`. Maybe `$grid-column-size` is the one that's wrong. 😉

brettsmason:
@andycochran Yeah I was going to suggest that :+1:  I think `grid-row-width` / `grid-column-width` make more sense.

If @kball agrees maybe changing that would make more sense? Its difficult though because of backwards compatibility... I think a Zurb decision on the correct approach might be best on this.

kball:
I tend to bias for backwards compatibility to allow folks to upgrade with the minimal pain possible without having to dig through all of their settings, but one way to handle that is to make the change, and then include some additional code that checks for the old version.  For example could change to $grid-column-width but then in global.scss have something like 
```
@if variable-exists($grid-column-size) {
   $grid-column-width = $grid-column-size;
}
```

Thoughts?

brettsmason:
I like the sound of that approach, and I think the grid/row mixins should be the same (size or width, not both). :+1:

ncoden:
@kball `$grid-column-width` is actually named as it. Did you mean *could change to $grid-column-size* ?

@andycochran @brettsmason I prefer the `size` term:
- a *column size* is explicit enough about finally changing the column *width*.
- a *column width* can lead to think it is always affecting the specially `width` CSS property

ncoden:
@brettsmason 
> I like the sound of that approach

Semantic sass + BEM = ❤️ 

*As __often__ said*
> All the "groups of properties" of a component (i.e. size, align, color, or specific behavior like collapse, expand...) should have their own %placeholder and @mixin, to allow a component customization in responsive/semantic cases

kball:
I'll let ya'all figure out what the final set of names should be.  My key concern is that settings files generated in the old way continue to work in the new way, ie things have backwards compatibility built in.

andycochran:
Didn't mean to take the conversation away from the PR's goal. Honestly, I'm fine with either "size" or "width." It's probably a bigger issue to take inventory of all size/width variables and make 'em consistent throughout the framework. But this _is_ targeting 6.3, so it's the right time to make such a change.

ncoden:
Agree with @andycochran

Issue #8569, opened 2016-04-11
[Tabs] Prevent default browser behavior only if keyboad key press is handled by the plugin


dchekanov:
Currently Tabs plugin [prevents](https://github.com/zurb/foundation-sites/blob/develop/dist/plugins/foundation.tabs.js#L151) browser from reacting to all keyboard key presses (except for the "Tab") when any tab link has focus.

This negatively affects user experience as it is not clear why nothing happens when I try, for example, to reload a page after clicking on a tab using a keyboard key.

How to reproduce: http://foundation.zurb.com/sites/docs/tabs.html, click on any rendered tab link ("Tab 1", for example) and press F5 (on Windows, Mac users can press whatever button/combo is used to reload a page).

I'd say this plugin (and any other in Foundation) shouldn't prevent default browser reaction to key presses unless that plugin handles those exact events.

Owlbertz:
Hi @dchekanov. Thanks for reporting this. I am currently enhancing the keyboard util to reduce the events handled with `preventDefault()`, so this will be fixed soon. In the meantime I will assign this issue to me to prevent someone else from putting effort into this.

kball:
Fixed in 5a1450d, will be in 6.2.2

Issue #8535, opened 2016-04-06
Input text or search is not working in dropdown top bar.


ve3:
Link to demo: [demo](http://yenped.com/pen/test-foudnation-search-topbar/)
I'm also try it in codepen but I think Foundation js does not work properly in codepen. [codepen link](http://codepen.io/anon/pen/EKbOwx)

The problem is I cannot use **space**, **enter** in the input field.

```html
<div class="top-bar">
  <div class="top-bar-title">
    <span data-responsive-toggle="responsive-menu" data-hide-for="medium">
      <span class="menu-icon dark" data-toggle></span>
    </span>
    <strong>Site Title</strong>
  </div><!--.top-bar-title-->
  <div id="responsive-menu">
    <div class="top-bar-left">
      <ul class="dropdown menu" data-dropdown-menu>
        <li><a href="#cat1">Category 1</a></li>
        <li><a href="#cat2">Category 2</a></li>
        <li><a>Search icon</a>
          <ul class="menu">
            <li>
              <form class="search-form">
                <div class="input-group">
                  <input class="input-group-field input-search" type="search" name="q" value="" placeholder="Search...">
                  <div class="input-group-button">
                    <button type="submit" class="button search-button">Search</button>
                  </div>
                </div>
              </form>
            </li>
          </ul>
        </li>
        <li><input type="search" placeholder="Search"></li>
        <li><button type="button" class="button">Search</button></li>
      </ul><!--.menu-->
    </div><!--.top-bar-left-->
  </div><!--#responsive-menu .top-bar-menu-items-container-->
</div><!--.top-bar-->
```

If I remove `data-dropdown-menu` from dropdown menu (`ul`) then it works but I have to use this dropdown menu in real site.
How to fix this?

brettsmason:
The Codepen example does work, you just have a typo in the foundation call :smile: 
From a quick investigation you are right though, it looks like the dropdown menu JS may be causing the issue. Hopefully someone else may have a better idea of the cause.

ve3:
The codepen example is really not working. Please carefully check again. If it does work the hamburger icon (three horizontal lines) must not appears.
The html and js are same on both codepen and example site.

However, thanks to take care of this bug. :)

brettsmason:
[...]

PR #8571, opened 2016-04-11
Enhance keyboard util.

Owlbertz:
Enhanced keyboard util by adding a parameter that is passed to the `handled` function based on the evaluation of the function that was handled.
It is now easier to only do certain actions (e.g. calling `preventDefaul()`) in the `handled` handler by letting the other function (e.g. `next`, `up`, ...) return some value, which is directly passed into `handled` as the first parameter.
I also adjusted the calls to the keyboard util for some plugins to fix issues and prevent bad behavior.

Since this rework fixes some bugs (#8535, #8569) I will target the develop branch.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -144,11 +144,11 @@ class AccordionMenu {
         },
         up: function() {
           $prevElement.attr('tabindex', -1).focus();
-          e.preventDefault();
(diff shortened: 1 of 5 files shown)


kball:
This looks good, I'll try to find some time to test it out in the next day or two

kball:
Checks out, looks good

PR #8576, opened 2016-04-11
Fix missing checkbox requiredCheck

web2client:
requiredCheck for checkbox was removed from 6.2.1. Reinstating.

Label set/clear error highlight only for associated checkbox input, not for the entire group.


kball:
Yep, looks like that was accidentally removed.  Good fix.

PR #8578, opened 2016-04-11
Remove duplication of max-width property in row mixin

### LINKED ISSUE #8566: grid-row-nest mixin output duplicates max-width property per breakpoint
[This declaration](https://github.com/zurb/foundation-sites/blob/develop/scss/grid/_row.scss#L88) duplicates the `max-width` property. It should be moved above `@each $breakpoint, $value in $gutter`.

alexej-d:
Fixes #8566 
As a result of the `max-width` property being inside the `@each` statement, the property was needlessly duplicated while compiling.


kball:
Looks good

Issue #8572, opened 2016-04-11
[Dropdown] Horizontal position does not work with additional classes


maggo:
When adding classes after float-left/right the regex in [foundation.dropdown.js#L74](https://github.com/zurb/foundation-sites/blob/7133cd464357e00566efbc2df8549209755f0a8c/js/foundation.dropdown.js#L74) will no longer read the horizontal position correcly. The matched string will be everything till the last space, see this [RegexPal example](http://www.regexpal.com/?fam=94612)

Testing for float-left and float-right explicitly or writing the regex like ```/float-(\S+)\s/``` should fix this.

ncoden:
Pull request: https://github.com/zurb/foundation-sites/pull/8579

PR #8579, opened 2016-04-11
[Dropdown] Fix horizontal position with additional classes #8572

ncoden:
Fix https://github.com/zurb/foundation-sites/issues/8572

> [Dropdown] Horizontal position does not work with additional classes

Thanks @maggo :clap: . _BTW beautiful profile picture_


maggo:
Awesome! Thank you :tada:

PR #8581, opened 2016-04-12
Remove margin from .menu-icon

ghost:
Hello,

if you have for example a logo in .title-bar-left (or right) and the .menu-icon on the other side the gaps to the sides are not equal because of this margin.

Olli


andycochran:
Could you make a [codepen](http://codepen.io/pen?template=NNyXxB) to illustrate the bug?

PR #8589, opened 2016-04-13
Can it be height instead of max-height?

ghost:


kball:
This is a deceptively large behavior change for a single line patch; changing the behavior of any off-canvas that is less than 100% in content.  Any sort of info why you'd want to make this change or why you think this is the correct behavior at a framework level?  For an individual site implementation you could always just add `height: 100%` to the .off-canvas class...

A change with this impact should be submitted against a minor point branch (e.g. v6.3), so I'm going to close this.  You are welcome to resubmit against v6.3 but please without more context about why you believe this is the correct behavior.

tyrion273:
Sorry!
I thought with 100% height it would more look like a sidebar.
Do you think this is a good idea?

PR #8597, opened 2016-04-13
fixes wrong calc on first menuitems height #8594 for me

### LINKED ISSUE #8594: [Drilldown Menu] wrong height _getMaxDims()
this.$menuItems[0].getBoundingClientRect().height

gets wrong height

see foundation docs the function gets 54.39px height but the height is 38px

this.$menuItems.first().height() 
gets 31.6px height maybe a timing problem?

designerno1:
don't know why but setting the width temp to auto fixes the wrong calc on first element of menu items fixes #8594 
#7070

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -293,13 +293,16 @@ class Drilldown {
    * @private
    */
   _getMaxDims() {
-    var max = 0, result = {};
+    var max = 0, result = {}, oneHeight, tmpWidth = this.$element.css('width');
+    this.$element.css({width:'auto'});
+    oneHeight = this.$menuItems[0].getBoundingClientRect().height;
+    this.$element.css({width:tmpWidth});
     this.$submenus.add(this.$element).each(function(){
       var numOfElems = $(this).children('li').length;
       max = numOfElems > max ? numOfElems : max;
     });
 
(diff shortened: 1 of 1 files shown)


js/foundation.drilldown.js
```diff
@@ -298,13 +298,14 @@ class Drilldown {
...
-    var max = 0, result = {};
-    this.$submenus.add(this.$element).each(function(){
-      var numOfElems = $(this).children('li').length;
-      max = numOfElems > max ? numOfElems : max;
+    var $uls = this.$element.find('ul'), max = 0, result = {};
+    $uls.add(this.$element).each(function(){
+      var $elem = $(this);
+      $elem.hasClass('is-drilldown-submenu')?$elem.css({height:'auto'}):'';
+      max = Math.max(this.getBoundingClientRect().height, max);
+      $elem.hasClass('is-drilldown-submenu')?$elem.css({height:'100%'}):'';
```
kball: This doesn't seem right - what if someone has overridden the styles on the submenus to have a different height?

This loop approach seems right - the bug seems to be due to assuming all of the menu items will always be the same size, where actually should take a look - but this auto/100% hardcoding feels wrong.  When these are actually shown, its done via adding a class, no?  Can we use that as a way to get them to expand to their auto height instead, and find from that what the height should be?

PR #8598, opened 2016-04-13
fixes this.$wrapper return

designerno1:
fixes this.$wrapper is the created element not the dom element
if we want to use it in code for example
var $body = $('body').not(_this.$wrapper);


kball:
This change looks reasonable, the commit history is messy.  Can you please resubmit with just the change in question?

PR #8599, opened 2016-04-13
fixes closeOnClick

designerno1:
var $body = $('body').not(_this.$wrapper);
not ignores clicks on links in the drilldown menu
fixes links with no sub and closeOnClick true doesn't fire


kball:
Looks good, going to merge.  Does this address any currently filed issues that you are aware of?

designerno1:
i found no issues addressed but when i was dev i ran into issues with that

PR #8605, opened 2016-04-14
fixes #8603 removes e.stopImmediatePropagation(); from mouseenter event

### LINKED ISSUE #8603: [Dropdown Menu] hoverDelay leave pane reenter pane issue
an a dropdownmenu the first level has a submenu so the pane opens the items in the pane has no submenu.
if you leave the pane and reenter it within the hover delay the pane gets closed

designerno1:
fixes #8603 for me
removes e.stopImmediatePropagation(); from mouseenter event to provide mouseenter can bubble through and clear the timeout on re enter a pane

please take a look and discuss and explain the need of the e.stopImmediatePropagation();

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -96,7 +96,6 @@ class DropdownMenu {
 
     if (!this.options.disableHover) {
       this.$menuItems.on('mouseenter.zf.dropdownmenu', function(e) {
-        e.stopImmediatePropagation();
         var $elem = $(this),
             hasSub = $elem.hasClass(parClass);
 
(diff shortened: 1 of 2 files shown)


kball:
Good question.  I don't know why stopImmediatePropagation would be needed, and looking at the commit history isn't helpful because it was added as part of a massive commit without a useful commit message.  @colin-marshall @Owlbertz any ideas why this might be needed?  If not I say we go ahead and merge this.

colin-marshall:
@designerno1 can you please reproduce #8603 in a CodePen? Just want to make sure I have the same syntax as you.

@kball is this worthy of a visual test?

designerno1:
@colin-marshall 
https://codepen.io/designerno1/full/QNZVNw/

kball:
@colin-marshall we're light on dropdown menu visual tests anyway, with all the current ones focused on rtl... this seems like it would be a good start to a basic set highlighting different option behaviors.

colin-marshall:
@designerno1 thank you for the CodePen, it helped me understand the problem and why this fix works.

Only thing I would recommend before merging is converting your CodePen to a visual test in the repo. The only thing you need to change is the instructions for how it should behave normally. Something like `test/visual/dropdown-menu/submenu-mouse-reenter.html` would be great. Thanks!

TejasQ:
+1! This is awesome!

Issue #7821, opened 2016-01-08
[Drilldown Menu] Menu height is not dynamic


mitchartifextech:
**How can we reproduce this bug?**
1. Add a nested section. 
2. Ensure the nested UL has more children than on the parent menu. 

**What did you expect to happen?**
I assumed the nested menu that animates in would grow tall enough to show all of the children.

**What happened instead?**

The root menu now has extra space below it, to accommodate the taller nested menu. For UI, this is not great, as it pushes content way below the menu. Ideally, the menu would grow if the nested section is taller. 

**Test case**

http://lawyerb2b.atcdemo.com/ has the menu. Just shrink the screen below 1024px to show the Offcanvas menu. 

See Screenshots here. http://screencast.com/t/4ap8MnzcfS http://screencast.com/t/8iAGj9C93mX

zurbrandon:
This has been mentioned before, but there was never a clear idea as to what should happen on transition  for a drill down. It felt far to jarring to have the menu just get taller and push the page down or up every time an item was selcted. 

Thoughts?

mitchartifextech:
Thank you so much for the reply. I really appreciate it. 

I agree, that having the menu get taller would be janky, but there was a version that worked well in F5. It wasn't called drilldown, but topbar had a function that would do something "similar". I have it working on a site now. http://www.woundhealingfoundation.org/

It just slides the whole menu over to the left, revealing the submenu. That way, height isn't an issue at all.

andycochran:
There are a few options:

1. Keep it as-is. Con: extra whitespace below top-level menu.
2. Set the Drilldown to `overflow-y:visible` and keep its height equal to the top-level menu. In this case, tall nested menus would overlap any content beneath (kinda like a Dropdown). Con: could be trouble on screens shorter than the nested menus. 
3. Animate the height of the Drilldown as you transition between nested menus (kinda like an Accordion). Con: content will reflow.

The Drilldown menu pattern probably isn't the best choice if any of those caveats don't work for you. I imagine Drilldown is best used with Off-canvas or in a layout where there's nothing underneath it. If you're using Drilldown in the middle of content, you're probably better served by a Dropdown or Accordion.

zurbrandon:
Thanks @andycochran and @mitchartifextech for weighing in. 

[...]

PR #8607, opened 2016-04-15
Drilldown feature autoheight

designerno1:
ads features to drilldown
data-auto-adjust-height and data-animate-height
#7821


kball:
This seems like a reasonable feature, but as it is a new feature please resubmit this PR against v6.3

PR #8614, opened 2016-04-15
remove margin bottom from :last-child of .media-object-section

### LINKED ISSUE #8604: [Media Object] margin bottom within Callout
In order to keep the padding even on all four sides of a Callout, the last child has its bottom margin removed. But if that last item is a Media Object, the margin only gets partially removed, 'cause a Media Object's bottom margin is actually a combination of the Media Object's margin plus the last item's margin in tallest `.media-object-section`. 

What does everyone think about this? 
```scss
.media-object-section > :last-child {
  margin-bottom: 0;
}
```

Here's a pen showing the issue and fix: 
http://codepen.io/andycochran/pen/aNYRJW

andycochran:
Closes #8604.

### DIFF
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -46,6 +46,10 @@ $mediaobject-image-width-stacked: 100% !default;
   &:last-child:not(:nth-child(2)) {
     padding-#{$global-left}: $padding;
   }
+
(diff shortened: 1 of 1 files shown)


kball:
Looks good and based on feedback #8604 is desired.  Merging

PR #8632, opened 2016-04-19
s.no_pip vs. self.settings.no_pip

dragthor:
https://github.com/zurb/foundation-sites/commit/48aa4843c51f7a95a31ff59db146c55bb5f68b03

We would like this change in our project. However, `!s.no_pip` was accidentally over written. In both `top:` and `bottom:` it should be `!s.no_pip` and **not** `!self.settings.no_pip`.

### DIFF
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -331,7 +331,7 @@
           p.left = p.left - thisOuterWidth + tOuterWidth;
         }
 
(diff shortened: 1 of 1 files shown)


kball:
@coreysyms can you take a look and confirm this is correct, as you were the originator of the change referenced?

coreysyms:
@kball This is correct, the PR should be merged, it was a late night, sorry for the overwrite @dragthor

dragthor:
Thanks for the validation @coreysyms 

We rely heavily on the Zurb 5 dropdown and our test cases failed immediately.

Issue #8617, opened 2016-04-15
[Drilldown Menu] feature suggestion scrollTop


designerno1:
for drilldown menu with very long submenu's it would be nice to have a function to jump after show,hide,closed to the top of the drilldown menu
i wrote an quick js to do this.
should we take this in the drilldown plugin, or is this a thing to do with the callbacks
```js
$('[data-drilldown]').on('open.zf.drilldown hide.zf.drilldown closed.zf.drilldown',function(){
  var animationDuration = 500, animationEasing = 'linear', scrollPos = $(this).offset().top;
  $('html, body').stop(true).animate({ scrollTop: scrollPos }, animationDuration, animationEasing);
});
```

simonrcodrington:
Hey there @designerno1. I would like that would be something you would handle in your own callbacks (as it might be unwanted for people who don't need that functionality). Be easier to visualize if you made this into a CodePen example so everyone could see.

designerno1:
just created a pen
http://codepen.io/anon/pen/jqxZxO
if you got a very long level 2 menu and there is a very short level 3 menu it is nice if it scrolls to the top of each drilldown element

brettsmason:
@designerno1 I checked the CodePen and that's a really nice user experience enhancement in my opinion. I guess the best way to implement it would be via an data-scroll-top option.

I think this requires some other opinions though.

designerno1:
ok please zurb let me know if i should code that feature into data-options feature

Silvester20:
I like this one too +1

kball:
Having this as a new data-option behavior sounds like a great addition for v6.3

simonrcodrington:
I think having this an option is probably the best course of action, I like the way it works. 

This should make it's way eventually to one of the zurb guys and they will be able to let us know if this is something they are interested in.

designerno1:
just added a PR #8633

Owlbertz:
Can be closed since the PR has been merged, will be fixed in 6.3 then.

PR #8633, opened 2016-04-19
Drilldown feature scroll top

designerno1:
- adds feature ScrollTop to Drilldown Menu
  scrollTop
  scrollTopElement
  scrollTopOffset
  animationDuration
  animationEasing
#8617

### DIFF
--- a/docs/pages/drilldown-menu.md
+++ b/docs/pages/drilldown-menu.md
@@ -73,6 +73,62 @@ Any `<a>` without a submenu will function like a normal link.
   <li><a href="#Item-4"> Item 4</a></li>
 </ul>
 
+## ScrollTop Drilldown
+
+<div class="callout">Scroll to the top on change the menu - Option scrollTop</div>
+
+<ul class="menu" data-drilldown data-scroll-top="true" style="width: 200px" id="m2">
+  <li><a href="#">Item</a></li>
+  <li><a href="#">Item</a></li>
+  <li><a href="#">Item</a></li>
+  <li><a href="#">Item</a></li>
+  <li><a href="#">Item</a></li>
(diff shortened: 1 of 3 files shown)


mesnitu:
Thanks. With the auto height feature , this drilldown is getting quite awesome.

mesnitu:
However ( but I' quite new to foundation ) , when using a responsive navigation, say from drill down to dropdown, I'll get  

`Uncaught TypeError: Cannot read property 'trigger' of null`

But probably I'm doing it in the wrong way : 

`<ul id="mainMenuLeft" class="drilldown menu" data-responsive-menu="drilldown medium-dropdown" data-auto-adjust-height="true" data-animate-height="true" data-scroll-top="true" data-scroll-top-offset="-30">`

This is related to
` if (this == $('html')[0]) _this.$element.trigger('scrollme.zf.drilldown');`

And also the max-with. Rotating the device, the menu gets whatever first value he got.

designerno1:
ce42487 fixes problem with destroy() thanks to @mesnitu 
havent tested it with responsive menu, this commit fixes the destroy problem on the attached handlers.
But i'am thinking of recoding the feature without attaching to the trigger handlers

mesnitu:
That works. The only thing I had to do  was to add a 
`this.$submenus.css('height', 'auto'); `
to the destroy
For some reason, having the submenu with style height 100%, only the first <li> gets the background color. 
Not sure if this is the correct way to do it,  ( I'm not a developer ) , but now switching from drill to drop down, the submenu <ul> gets the proper background. 
Thanks

kball:
This seems like a reasonable feature, but as it is a new feature please resubmit this PR against v6.3

PR #8640, opened 2016-04-20
A second option to handle the setting file

OndrejVasicek:
I have been using this approach since Foundation 4 and it’s been really paying off.
If you decide to merge my lines, please check the gramma. English isn’t my native language. Thanks.

### DIFF
--- a/docs/pages/sass.md
+++ b/docs/pages/sass.md
@@ -123,7 +123,17 @@ Every component includes a set of variables that modify core structural or visua
 
 <div class="callout warning">
   <p>Once you've set up a new project, your settings file can't be automatically updated when new versions change, add, or remove variables. Keep tabs on new <a href="https://github.com/zurb/foundation/releases">Foundation releases</a> so you know when things change.</p>
+  <p>
+  A second handy approach is to set up a new setting file (e.g. `_setting-custom.scss`) and add only settings you wish to overwrite. In your `app.scss` include your custom setting right after the default setting. It may look like this:
+  </p>
 </div>
+
+```scss
+@import 'settings';
+@import 'settings-custom';
+```
+
+
 
 Here's an example set of settings variables. These change the default styling of [buttons](button.html):
 


MainelyWeb:
Ondřej, that is the way that I used to do it as well, however, the new method used in the Zurb template, for example, is more flexible. With more complex customizations, it helps to organize everything by functional area.
See https://zurb.com/expo/lessons/avoid-a-cluttered-mess-sensible-sass-file-structure for more information.

OndrejVasicek:
II’m afraid I don’t get your point. I read the whole article. It’s great and more or less I knew everything, but there wasn’t anything about the issue I was writing about.

My pull request is about “how to accomplish an update of Foundation without overwriting my settings”. I responded to the sentence:

> Once you've set up a new project, your settings file can't be automatically updated when new versions change, add, or remove variables. Keep tabs on new Foundation releases so you know when things change.

With my approach – you CAN update automatically. Or maybe I'm missing something?

andycochran:
The first settings file is unnecessary if it has no changes. Every component begins by declaring the same `!default` variables as `_settings.scss`. So you can delete everything from `_settings.scss` and still get all the default settings. All you need is a settings file that includes the variables you want to override.

OndrejVasicek:
Two things
- I have to make an apologize, cause my pull request doesn’t even work properly. There is a complication when trying to change a colour from foundation pallete –  a change one of those colour in _setting-custom (even with using the add-foundation-colors mixin) doesn’t affect already sett settings from _settings file.
- Your answer has a point. But then we should ask ourselves – why to have default setting file with all those active settings? I now, it’s there for users – to have every settings at hand. But wouldn’t be better to deliver this file with all those settings at least commented?
There is one big danger – if foundation change some default setting in a component, any user will reflect that – because it will be sett in his _setting file from the beginning. And we have to be honest here – nobody will go through full changelog and check if some of the setting has changed…for every project. 
But with every setting commented, only setting uncommented and changed by user will have priority against the foundation setting. What do you think about that?

PR #8644, opened 2016-04-20
changed.zf.mediaquery improvements

designerno1:
this.current = newSize; (should be set before window triggered because if something goes wrong changed.zf.mediaquery fires constantly, because current is neer set

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -106,14 +106,14 @@ var MediaQuery = {
    */
   _watcher() {
     $(window).on('resize.zf.mediaquery', () => {
-      var newSize = this._getCurrentSize();
+      var newSize = this._getCurrentSize(), currentSize = this.current;
 
-      if (newSize !== this.current) {
+      if (newSize !== currentSize) {
         // Change the current media query
         this.current = newSize;
 
         // Broadcast the media query change on the window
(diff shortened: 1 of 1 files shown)


js/foundation.util.mediaQuery.js
```diff
@@ -109,11 +109,11 @@ var MediaQuery = {
...
 
       if (newSize !== this.current) {
-        // Broadcast the media query change on the window
-        $(window).trigger('changed.zf.mediaquery', [newSize, this.current]);
-
         // Change the current media query
         this.current = newSize;
+
+        // Broadcast the media query change on the window
+        $(window).trigger('changed.zf.mediaquery', [newSize, this.current]);
```
kball: Triggering it here means this.current is newSize, rather than the old size.  I think moving this to after the set is reasonable, but if we do we should capture the previous value of this.current and pass it as the second value.

designerno1:
ah ok i see, the only thing this PR was for i ran into an error on my `.on('changed.zf.mediaquery',function(){})`
and wondering afer the error why changed.zf.mediaquery is fired several times, because this.current didn't gets a update because of the error

kball:
@designerno1 do you have time to clean up this PR to address the problem raised in the comment?  I'd love to get this fix in place

kball:
👍

PR #8649, opened 2016-04-21
Add inline-svg icon support

artursopelnik:


denisahac:
Not everyone would like to have an SVG tangled up with their HTML. If someone wants to have an SVG supported in the menu, then they can just add their own styles without messing up the core ZF.

What are those other commits in this pull request? I think they're unrelated to this **Add inline-svg icon support** commit you'd like to address.

artursopelnik:
We use [svg-icon-system ](http://danburzo.ro/svg-icon-system/) for our client project. Currently we add (svg-support) for menu-icons in own styles but why not make it global ...
---
You're right. Only the **Add inline-svg icon support** commit is important.

kball:
The addition of SVG seems reasonable to me, but please resubmit with a clean branch that contains only that commit difference.

artursopelnik:
Look at this: https://github.com/zurb/foundation-sites/pull/8683

PR #8650, opened 2016-04-21
prevent hover/focus effects on disabled buttons

### LINKED ISSUE #8562: 6.2.1 - Disabled buttons has background-color on hover
When an button is disabled with the 'disabled' class & 'disabled' attribute present, there should not be an background-color change on hover, like it was in 6.2.0.

andycochran:
Closes #8562.


Silvester20:
Awesome, thanks!

AxelKhaldeev:
It sets button background anyway to $button-background. So if we have .button.success the background on hover on disabled button will be $button-background, but not .success color.

PR #8655, opened 2016-04-22
Table stripes

### LINKED ISSUE #8237: [Tables] row stripes via class
Should the default style of table rows be striped? Seems like it'd be more useful as a `.striped` class, so it'd work just like `.hover`. 

As-is, if you want some (not all) tables striped, it takes a bit of custom work. 

I propose we ~~add `border-bottom: $table-border` to rows, remove the `border-bottom` on the table, and move the `:nth-child` styles to a `.striped` class~~ add a `$table-is-striped` variable that, if set to `true` creates an .unstriped class, and if set to `false` creates a .striped class. 

If :+1:, I'll PR.

andycochran:
Closes #8237

### DIFF
--- a/docs/pages/reveal.md
+++ b/docs/pages/reveal.md
@@ -186,3 +186,22 @@ $.ajax('/url')
     $modal.html(resp.html).foundation('open');
 });
 ```
+
+
+---
+
+## Accessibility
+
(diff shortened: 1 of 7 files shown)


andycochran:
Ack! I worked off of develop instead of v6.3. I'll redo this…

PR #8656, opened 2016-04-22
Add options to unstripe Tables

### LINKED ISSUE #8237: [Tables] row stripes via class

andycochran:
Closes #8237.

### DIFF
--- a/docs/pages/table.md
+++ b/docs/pages/table.md
@@ -94,7 +94,7 @@ By default, table rows are striped. There's an `.unstriped` class to remove the 
 </table>
 ```
 
-<table class="unstriped hover">
+<table class="unstriped">
   <thead>
     <tr>
       <th width="200">Table Header</th>
--- a/docs/pages/table.md
+++ b/docs/pages/table.md
@@ -85,6 +85,48 @@ Need to spiff up the table just a tad? Just add the class `.hover` to lightly da
 
 ---
 
+## Stripes
+
+By default, table rows are striped. There's an `.unstriped` class to remove the stripes. If you change `$table-is-striped` to `false` to remove stripes from all tables, use the `.striped` class to add stripes.
+
+```html
+<table class="unstriped">
+</table>
+```
+
+<table class="unstriped hover">
+  <thead>
+    <tr>
+      <th width="200">Table Header</th>
+      <th>Table Header</th>
+      <th width="150">Table Header</th>
+      <th width="150">Table Header</th>
+    </tr>
+  </thead>
+  <tbody>
+    <tr>
+      <td>Content Goes Here</td>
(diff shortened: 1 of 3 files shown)


andycochran:
@brettsmason Do you have any feedback on this? I'm not sure I got the mixins right to be able to include them with `@include table($stripe);`.

brettsmason:
Nice job @andycochran!
I've just pulled down this PR to check it over. One thing I noticed in the docs - you have `unstriped` and `hover` on the unstriped example. I'd suggest removing the `hover` class to avoid any confusion of what it should look like.

I also noticed this in the default css:
```
table.unstriped tbody {
    background-color: Blue;
```
Is the blue colour supposed to be there?

Finally this is the first time I've actually looked at the table Sass file. Wouldn't the internal mixins be better as public ones, so we could apply `-zf-table-stripe` (or just `table-stripe`) semantically to any table we wanted?

I might be wrong on all this as this is the first time I've really looked over this file.

andycochran:
@brettsmason Ack! That `blue` is a holdover from some debugging I was doing. I'll nix that. The reason I put hover on the unstriped table was to make sure they still worked together, also checking it w/ ``$table-is-striped:false` and `striped`.

andycochran:
@brettsmason Maybe I should just add the hover class to the code snippet so it matches the example.

brettsmason:
Yeah as long as they match that's fine by me

andycochran:
I decided to remove hover from .unstriped table example. Folks might use the copy button and unintentionally get the hover class. But hover and stripes do play nice.

andycochran:
I wonder who wrote the table mixins. They're probably internal for a reason. @kball, any insight on this?

kball:
@andycochran this appears to be the background: https://github.com/zurb/foundation-sites/pull/7454

Looks like it was more of a 'this is pretty messy and will probably change' rather than a more fundamental issue.

If you and @brettsmason are happy with this solution and think it is forward facing, we can absolutely make them public and document

andycochran:
I'm atually not sure I see much value in any of the table mixins being public.

kball:
ok... @brettsmason @andycochran are you happy with the state of this?  Is it ready to go?

brettsmason:
I think the `_table.scss` file could use a bit of love in general, but the changes @andycochran made look good to go.

PR #8657, opened 2016-04-22
fixes #8643 do not reset val of radio and checkbox on reset

### LINKED ISSUE #8643: 6.2.1 - Abide fails with radio button after resetting form
1. Create a abide form with a radio-group with all of the radio's required which is preffered by W3.
  ( or at least 1 )
2. Fill in the form correctly and submit the form. Now abide says form is valid.
3. Reset the form.
4. Fill in the form correctly and submit the form again.

The form is valid so "formvalid.zf.abide" should be triggered

**What happened instead?**
"form**in**valid.zf.abide" triggered

**Test case:**

http://codepen.io/anon/pen/mPKQdG

designerno1:
fixes #8643
do not reset val of radio and checkbox on reset the form reset prop checked false instead
maybe a saving of the initial markup state of radio checked and checkbox checked would be great


Silvester20:
You rock, thanks!

kball:
Looks good to me

Issue #8435, opened 2016-03-20
[Sticky] Sticky element loses height when anchored


KharamanV:
The thing is that when the page is scrolled down - everything is fine, but when then scroll back to top, the sticky element loses its height. Help please, what's the problem?
![image](https://cloud.githubusercontent.com/assets/4006620/13907183/f2d45b48-eef0-11e5-8f79-ba2be846ced4.png)
![image](https://cloud.githubusercontent.com/assets/4006620/13907185/f9672b70-eef0-11e5-86e6-1e14b4d1f717.png)
![image](https://cloud.githubusercontent.com/assets/4006620/13907187/fdb3f8d4-eef0-11e5-9279-d9113a4a1585.png)

lukejoyce:
I encountered this also - it may be a bug unless I configured something wrong. I fixed it with:

`.sticky.is-anchored {
    bottom: auto !important;
}`

[http://stackoverflow.com/questions/35897558/foundation-6-sticky-menu-left-menu-div-pushes-up-when-page-is-scrolled-back-to/35926403#35926403](http://stackoverflow.com/questions/35897558/foundation-6-sticky-menu-left-menu-div-pushes-up-when-page-is-scrolled-back-to/35926403#35926403)

iammichiel:
I have solved the issue by setting the background on sub-element. 

```
<div class="sticky">
    <div style="background-color: blue">...</div>
</div>
```

It seems to be caused by the `position` property. This work-around however works perfectly.

colin-marshall:
@iammichiel @lukejoyce thanks for the solutions! The problem I found with both of those was that a sticky element with `height: auto` will lose its height when it becomes unstuck because of `position: absolute`.

If you have a minute, please check out my PR #8663 to see a visual test I added, along with a new option called `staticHeight` that helps with the sticky element losing its height. I will probably try to refactor it so that no option is necessary and the static/variable height detection is done automatically by Sticky. The visual test is also in the develop branch so you can see how it performs before my fix.

Thanks guys!

PR #8663, opened 2016-04-24
Sticky Experiment: staticHeight option added

colin-marshall:
Added a `staticHeight` option to Sticky which defaults to `true`. If the sticky element has a height that will change between sticky states, set this to false. Otherwise, the default plugin setting of `true` will get the height of the element before sticky is removed, and set the element to that height after sticky is removed. This is so when `position` is changed to `absolute` when it "unsticks" the container doesn't lose it's height.

Resolves: #8435 

@gehasia please check out what I have done when you get a chance.

### DIFF
--- a/docs/assets/partials/sticky-nav.html
+++ b/docs/assets/partials/sticky-nav.html
@@ -32,7 +32,7 @@
 </head>
 <body>
   <div data-sticky-container>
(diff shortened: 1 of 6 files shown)


colin-marshall:
Not the best solution, closing.

Issue #8671, opened 2016-04-26
[Equalizer] equalizeOnStack callout issue


designerno1:
equalize on stack not working if watched in callout

as of:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetTop
https://developer.mozilla.org/en-US/docs/Web/API/HTMLelement/offsetParent

```js
_isStacked() {
  return this.$watched[0].offsetTop !== this.$watched[1].offsetTop;
}
```
`offsetTop ` is the value related to the offsetParent and because callout has `position: relative;` ist the offsetTop related to the callout in my case the same on the first and second

maybe using `getBoundingClientRect()` or `offset().top`?

https://codepen.io/designerno1/full/JXaMaw/

PR #8673, opened 2016-04-26
possible fix for #8671

designerno1:
possible fix for #8671
use getBoundingClintRect().top to detect stack


kball:
so I know getBoundingClientRect is impacted by scroll offset, but I think for this case that doesn't matter, so we should be ok to merge this.

Issue #8587, opened 2016-04-12
[Sticky] Sticky Container not Re-flowing on Resize


merlihan:
**How can we reproduce this bug?**

1. Step one - Visit this page: http://foundation.zurb.com/sites/docs/sticky.html
2. Step two - Resize the browser (1300 seems to be consistent)
3. Step three - Scroll down to "Advanced"
4. Step four - Scroll up and down repeatedly

**What did you expect to happen?**
I expected the sticky elements to reflow without overlapping.

**What happened instead?**
The Orbit photo overlaps the text above it.

NOTE - This happens intermittently.  

Give us a link to a CodePen or JSFiddle that recreates the issue.
N/A this issue is recreated on the doc page here: http://foundation.zurb.com/sites/docs/sticky.html

- [CodePen with Foundation 6 and MotionUI](http://codepen.io/rafibomb/pen/xVVGOB)
- [CodePen with Foundation 6, Flexbox grid and MotionUI](http://codepen.io/rafibomb/pen/jqqPra)

![image](https://cloud.githubusercontent.com/assets/4990374/14475465/1219e2a2-00d0-11e6-99d9-bf3ad76e103d.png)

merlihan:
Hi there,  Just checking if this ticket is now stable. Looks like it was closed then reopened. :(

andycochran:
@merlihan Did #8715 take care of this?

merlihan:
@andycochran Yes, it did!  Thank you!

Issue #8645, opened 2016-04-21
[Sticky] Transitions when adjusting height 


colin-marshall:
@gehasia you recently introduced commit https://github.com/zurb/foundation-sites/commit/26e716e404f4408f194b614114a88a9f4c93f3b7. I think the lines about transitions have messed up the behavior in the sticky-nav example at the bottom of the Sticky docs. The transition is now jerky and the offset is messed up too. You can only see the bug in local docs, so compare to the live version of the [Sticky docs](http://foundation.zurb.com/sites/docs/sticky.html) for normal behavior.

@gehasia please have a look when you get a chance, we are working on polishing up the Sticky plugin this week. Thanks!

gehasia:
Hi, i've done some more work on it last week, and changed the way container resizing is fired up. But I didn't PR for it yet.
I'll try to test it against the sticky-nav at the bottom of the sticky docs tomorrow (i won't be there for today) and try to debug it quickly. 
Does something like an option to activate/deactivate the container resize automation seems reasonable to introduce too (autoresizeContainer=false by default ?) ? So it'd keep the stable behavior by default (the container does not resize) and people willing to have their sticky container resizes (for the rest of content to move up/down following the new size) could just activate the behavior.
(sorry for my bad english...).

Thanks for the report !

colin-marshall:
@gehasia Thanks for the quick response. I am working on a fix for #8435 that is different from the suggested solutions. This may be the same thing you are trying to achieve.

I was going to see what people think about adding an option to Sticky, named something like `staticHeight` which defaults to `true`. If it's set to the `true`, Sticky will get the height of the `.sticky` element right before "sticky" gets removed (like when you scroll to an anchor set in the options), and then it will explicitly set the height of `.sticky` to what it was before the `position` property was changed to `absolute`. 

You would set the option to false if the sticky element will change height when it becomes stuck/unstuck. When it's false, it won't do anything with the height (which is the current behavior), allowing you flexibility to alter the height in CSS or JS between sticky states.

gehasia:
@colin-marshall  I made some videos to show different behaviors.
In each videos, i start by scrolling down and up with a mouse-click on the scroll bar. Then i scroll down and up with the mouse wheel. The sticky-container background is set to red, and the sticky element is the nav bar with black background. I made videos because i think it's more explicit than tons of textual explanations (in bad english...)

The current foundation "stable" behavior, as seen in the current online documentation :
[...]

PR #8676, opened 2016-04-26
[Bugfix] Sticky height when anchored + Position when anchors are outs…

gehasia:
Fixes https://github.com/zurb/foundation-sites/issues/8435
Fixes bad position of a sticky element when it's container is in another div than his top anchor.
Fixes anchored element not following the flow when resizing the window and going over/behind other elements (see https://github.com/zurb/foundation-sites/issues/8587 )
I made a video showing this :
[Reflow bug](http://sendvid.com/52a7xxmg)

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -215,9 +215,6 @@ class Sticky {
                   * @event Sticky#stuckto
                   */
                  .trigger(`sticky.zf.stuckto:${stickTo}`);
-    this.$element.on("transitionend webkitTransitionEnd oTransitionEnd otransitionend MSTransitionEnd", function() {
-      _this._setSizes();
-    });
   }
 
   /**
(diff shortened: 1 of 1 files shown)


js/foundation.sticky.js
```diff
@@ -215,9 +215,6 @@ class Sticky {
                   * @event Sticky#stuckto
                   */
                  .trigger(`sticky.zf.stuckto:${stickTo}`);
-    this.$element.on("transitionend webkitTransitionEnd oTransitionEnd otransitionend MSTransitionEnd", function() {
-      _this._setSizes();
-    });
```
kball: We are we removing _this._setSizes after transition ends?

kball:
@colin-marshall you've been digging around with Sticky, does this change make sense?

gehasia:
@kball We are talking about various issues here :
https://github.com/zurb/foundation-sites/issues/8645

The first patch you merged was not good, i made a better patch since then, but in the same time @colin-marshall was working on another bugfix which was close ( https://github.com/zurb/foundation-sites/issues/8435 ) 
His solution was based on a data-static-height option, and has some bugs remainings. 
I made another fix (wich is this PR), which fix the same thing than the colin experiment, but without the need of the data-static-height option, and fixing other bugs in the same time.
I will redo a PR on the fluid height transition (the first patch, the one you marked in this comment) after all the sticky bugfixing stuff merged. 
But yes, this code is not valid anymore, waiting for transitionEnd make some glitchy artifacts.

js/foundation.sticky.js
colin-marshall: @gehasia is removing this technically a part of the fix for #8435 or is it just being removed in the same commit? Can we please separate these into two different commits if they're not dependent on each other? It will make tracking progress easier as we tackle Sticky's bugs. Thanks!

colin-marshall:
@gehasia I think you have found the fix for #8435! I tried the `bottom: auto` but the offset was off and for some reason I didn't think of just fixing it in the JS. Nice work!

Please see my diff comment upthread about separating the issues out to different commits. I'm going to get us organized tomorrow with a "master" issue that lists all outstanding sticky issues with their progress. That should make the road ahead clearer for all of us. Thanks!

js/foundation.sticky.js
gehasia: It was just removed in the same commit, it's not linked. I will separate it, and include those lines deletion when i'll do the correct PR for fluid height transition (which will replace this 3 lines)

gehasia:
Hi i'll work on it tomorrow. I'll submit clean PR for each bugs.

gehasia:
Ok i've separated it into different PR :
https://github.com/zurb/foundation-sites/pull/8713
https://github.com/zurb/foundation-sites/pull/8714
https://github.com/zurb/foundation-sites/pull/8715

There is also a Doc + Test PR which need to be applied in order to test correctly :
https://github.com/zurb/foundation-sites/pull/8674

I close this one.

PR #8678, opened 2016-04-26
Is7458 reveal animation speed

### LINKED ISSUE #7458: [Reveal] Cannot change the animation speed of Reveal open/close
We are having a hard time figuring out how to change the animation speed of the Reveal open/close events. Added a "fast" class to the element does not change the speed, and we cannot find documentation on it as well.

I looked through the source and noticed that in the `foundation.util.motion.js` file, line 57, the duration options are commented out:

```javascript
var Motion = {
  animateIn: function(element, animation, /*duration,*/ cb) {
    animate(true, element, animation, cb);
  },

  animateOut: function(element, animation, /*duration,*/ cb) {
    animate(false, element, animation, cb);
  }
};
```

Is this a planned feature for the future?

kball:
Fix #7458 for the base cases of fast/slow.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -209,6 +209,11 @@ class Reveal {
 
     if(this.$overlay) {
       this.$overlay.css({'visibility': ''}).hide();
+      if(this.$element.hasClass('fast')) {
+        this.$overlay.addClass('fast');
+      } else if (this.$element.hasClass('slow')) {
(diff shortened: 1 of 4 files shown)


designerno1:
visual test works for me
only thing the assets and motion-ui symlinks don't work on a windows machine, have to mklink them for me

kball:
Based on @designerno1 's assessment that it works, going to merge.

PR #8688, opened 2016-04-27
Toggle aria-expanded on all offcanvas triggers

gapple:
If multiple elements are triggers for an off-canvas element, they all receive `aria-expanded="false"` when initialized, but only the actioned trigger is updated when the menu is opened.  This causes screen readers to read the wrong state of additional triggers.


kball:
Makes sense to me.  @Owlbertz do you agree?

Owlbertz:
Yup. I will pull this down and have a look.

kball:
@Owlbertz did you get a chance to look at this?

Owlbertz:
Huh sorry, I am not assigned and therefore did not see this while filtering the issues. Doing this asap.

Issue #8300, opened 2016-03-01
[Tooltips] Tooltip remains open despite option data-click-open is set to false


rianne-o:
The documentation of tooltip's data-click-open option says:

> Allows the tooltip to remain open if triggered with a click or touch event.

The default value is set to true.

I've set up a Fiddle to reproduce this issue for you.
https://jsfiddle.net/oagy66s8/5/

As you can see, the anchor has the attribute `data-click-open="false"`. I'd expect the tooltip **not** to remain open after I've clicked on the element, but that's not what happens.

Despite of my false value, the tooltip still remains open after clicking on the element.

colin-marshall:
Closing this out, as I believe this was resolved by @kball in #8691.

PR #8691, opened 2016-04-27
Fix data-click-open tooltip behavior

kball:
Fixes issue #8300.  Test added in tooltip docs

### DIFF
--- a/docs/pages/tooltip.md
+++ b/docs/pages/tooltip.md
@@ -30,6 +30,21 @@ At the spot thus attained a second peg was driven, and about this, as a centre, 
 
 ---
 
+## Tooltip clicking
+By default, clicking on a tooltip will leave it open until you click somewhere else.  However, you can disable that by adding data-click-open="false"
+
+```html_example
+<p>
+this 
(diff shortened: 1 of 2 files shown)


designerno1:
if i click on the `tooltip will stay open` and then click again on the `tooltip will stay open` and then again on the `tooltip will stay open` it stays close.
On touch device if you hit `tooltip will stay open` multiple times it gets a weired behavor of sometimes show it sometimes close it or sometimes show and close it.
But i don't know if these multiple clicks are commen use

kball:
interesting... I'll take a look

kball:
@designerno1 I'm able to reproduce... but I can also reproduce in develop branch, so I think it is a different bug.  I'll see if I can find a quick fix

designerno1:
@kball forget my `touch issue` checked again and it works. but the first happend again on click

kball:
@designerno1 Yeah I see the click issue as well.  Its only there if you never 'defocus' the tooltip... clicking elsewhere and then back on the tooltip always seems to work.  I'm not seeing any obvious ways to address it, so since it isn't a regression my inclination is to file it as a separate issue and not have it be a blocker for this PR.  Thoughts?

colin-marshall:
Agreed on merging this and opening a new issue for tooltip staying closed. I'll go ahead and do that since 3 of us seem to be in consensus on this.

PR #8693, opened 2016-04-28
Make grid-column-gutter define responsive gutters by default #8259

### LINKED ISSUE #8259: The grid-col-uncollapse mixin doesn't work
It looks like `grid-column-uncollapse` uses not defined variable `$grid-column-gutter` by default. When I don't pass any value `rem-calc` try to calculate gutter from null:
```
Running "sass:default" (sass) task
WARNING: null was passed to rem-calc(), which is not a number.
        on line 75 of node_modules/foundation-sites/scss/util/_unit.scss
        from line 44 of node_modules/foundation-sites/scss/util/_unit.scss
        from line 19 of node_modules/foundation-sites/scss/grid/_gutter.scss
        from line 33 of node_modules/foundation-sites/scss/grid/_gutter.scss
        from line 22 of saleor/static/scss/store/partials/product_page.scss
        from line 140 of node_modules/foundation-sites/scss/util/_breakpoint.scss
        from line 21 of static/scss/page.scss

>> Invalid null operation: "null plus "2"".
>>   Line 19  Column 22  node_modules/foundation-sites/scss/grid/_grid.scss
Warning:  Use --force to continue.
```
I am using foundation-sites 6.1.2, my code:
```sass
  .content {
    padding: 0 $global-padding * 4;

    @include breakpoint(medium) {
      @include grid-col-uncollapse();
    }
  }
```

My solution for this problem:
```sass
  .content {
    padding: 0 $global-padding * 4;

    @include breakpoint(medium) {
      @include grid-col-uncollapse(map_get($grid-column-responsive-gutter, 'medium'));
    }
  }
```

ncoden:
Resolve #8259

**Changes**
Make `grid-column-gutter` (and also `grid-col-gutter`) define
responsive gutters when no gutter value (`$gutter`) is given, instead
of defining gutters from the zero breakpoint.

If only a `$gutters` map is given, use this map to generate responsive
gutters.

If only a `$gutters` single value is given, use this value to generate
no-responsive gutters.

---

**:warning: Require**
- [x] https://github.com/zurb/foundation-sites/pull/8528

### DIFF
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -25,19 +25,35 @@
 }
 
 /// Set the gutters on a column
-/// @param {Number|Keyword} $gutters [small] - Spacing between columns or breakpoint name.
+/// @param {Number|Keyword} $gutter [auto]
+///   Spacing between columns, accepts multiple values:
+///   - A single value will make the gutter that exact size.
+///   - A breakpoint name will make the gutter the corresponding size in the $gutters map.
+///   - "auto" will make the gutter responsive, using the $gutters map values.
 /// @param {Number|Map} $gutters [$grid-column-gutter] - Gutter map or single value to use. Responsive gutter settings by default.
 @mixin grid-column-gutter(
-  $gutter: $-zf-zero-breakpoint,
+  $gutter: auto,
   $gutters: $grid-column-gutter
 ) {
-  @if type-of($gutter) == 'string' {
-    $gutter: grid-column-gutter($gutter, $gutters);
-  }
-  $padding: rem-calc($gutter) / 2;
+  @if $gutter == auto and type-of($gutters) == 'map' {
+    // "auto"
+    @each $value, $breakpoint in $gutters {
+      @include breakpoint($breakpoint) {
+        @include grid-column-gutter($value);
+      }
+    }
+  } @else {
+    // breakpoint name
+    @if type-of($gutter) == 'string' {
+      $gutter: grid-column-gutter($gutter, $gutters);
+    }
 
-  padding-left: $padding;
-  padding-right: $padding;
+    // single value
+    $padding: rem-calc($gutter) / 2;
+
+    padding-left: $padding;
+    padding-right: $padding;
+  }
 }
 
 /// Collapse the gutters on a column by removing the padding. **Note:** only use this mixin within a breakpoint. To collapse a column's gutters on all screen sizes, use the `$gutter` parameter of the `grid-column()` mixin instead.
@@ -55,8 +71,11 @@
 
 /// Shorthand for `grid-column-gutter()`.
 /// @alias grid-column-gutter
-@mixin grid-col-gutter($gutters) {
-  @include grid-column-gutter($gutters);
+@mixin grid-col-gutter(
+  $gutter: auto,
+  $gutters: $grid-column-gutter
+) {
+  @include grid-column-gutter($gutter, $gutters);
 }
 
 /// Shorthand for `grid-column-collapse()`.
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -6,19 +6,31 @@
 /// @group grid
 ////
 
+/// Set the gutters on a column
+/// @param {Number} $gutter - Spacing between columns.
+@mixin grid-column-gutter($gutter) {
+  $padding: rem-calc($gutter) / 2;
+
+  padding-left: $padding;
+  padding-right: $padding;
+}
+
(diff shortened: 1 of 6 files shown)


scss/grid/_gutter.scss
```diff
@@ -6,19 +6,76 @@
...
+    // "auto"
+    @each $value, $breakpoint in $gutters {
+      @include breakpoint($breakpoint) {
+        @include grid-column-gutter($value);
+      }
+    }
+  } @else {
+    // breakpoint name
+    @if type-of($gutter) == 'string' {
+      $gutter: grid-column-gutter($gutter, $gutters);
```
kball: Hmm... it took me a while to realize this was actually calling a function by the same name, rather than infinitely recursing through the mixin... hmm.  I see looking at our codebase that this isn't that uncommon of a pattern, but it struck me as unintuitive at first glance.  Is this a common scss practice? (cc @brettsmason @JeremyEnglert @andycochran @ncoden @zurbrandon @rafibomb). Alternatively, my inclination would be to have some sort of consistent scheme for naming functions that are connected to particular mixins, for example by prefixing with a '-' or something similar.

A quick look around at some existing scss styleguides doesn't seem to highlight function naming... interested in  your thoughts

kball: So there's 2 aspects to this... there's having a consistent naming scheme in general, and there's the question of how to handle interconnected functions and mixins.

With regards to the first, I've heard arguments for and against BEM - generally it feels unnecessarily verbose to me, though I'd be open to arguments why that verbosity is helpful - but I think in most cases our naming conventions have at least been reasonably consistent when it comes to component structure and final class name.

On the second though, I'd personally like some sort of immediate visual cue as to what role the thing I'm looking at is... for example, in ruby modules and class names are CamelCase, constants are CAPITALIZED, and variables and method names are snake_case.  In scss we have an immediate visual cue of something being a $variable, but no naming distinction between a @mixin (which must be @include'd in) and a @function which is called directly...

I'd like to propose we adopt some sort of convention that gives that immediate visual cue at least within foundation.  I don't have a strong opinion on what it is...

ncoden: > A quick look around at some existing scss styleguides doesn't seem to highlight function naming... interested in your thoughts

I didn't found scss naming convention for functions, variables and mixins while I was working on [CaliOpen](https://github.com/CaliOpen/caliopen.web-client-ng), so I used an adapted **BEM** :

Here an exemple : https://gist.github.com/ncoden/171b26636f052c41739aff3c26ab8357

ncoden: > I'd like to propose we adopt some sort of convention that gives that immediate visual cue at least within foundation. I don't have a strong opinion on what it is...

CSS is case-insensitive, so Sass repeated the ~~bug~~feature. But sometimes I see code which break the `always-snipal-case` for a `CamelCase` alternative version of BEM, because case errors never happen.

Also, everything is declared in the global scope, so we have to almost always deal with a namespace. But maybe we can write something like `NameSpace_CONST` ?

To distribute the different cases to the differentes Sass things, we only have to deal between `@mixin` and `@function`.
- NameSpace, `@mixin` (and the `.class` involved) is more a OO Class, so I would say `CamelCase`
- `@function` (with `$variable`, unless for const) takes the rest, `snipal-case` or `snake_case`.

I made the following exemple : https://gist.github.com/ncoden/a1cef157c7ebe236fe115cd6d2bd4a26
I don't like it, but I don't find better for now. 🤔

ncoden: > With regards to the first, I've heard arguments for and against BEM - generally it feels unnecessarily verbose to me, though I'd be open to arguments why that verbosity is helpful.

BEM was not only created to help the developer knowing what he uses, but to prevent unwanted behaviours when using CSS classes : properties overwritten by specificity, naming collisions across components or frameworks... *I remember having these problems with Foundation*

Also, there is a lot of BEM grammars, some more verbose than others (i.e. `block-foo__element-bar--modifier-baz`, `BlockFoo_ElementBar-ModifierBaz`, etc...).

kball: Very interesting.  Curious what our more SCSS-focused yetinauts think @andycochran @brettsmason @JeremyEnglert @zurbrandon

brettsmason:
I'm all for a documented naming convention and sticking to it. I personally have no experience in using BEM, and I feel it might be a bit too strict on this occasion, but I understand the reasoning behind it.

I prefer SMACSS myself, and its more flexible in its approach.
A simply example would be buttons, currently we have:

```
.button
.primary
.secondary
.large
```
You could go with something like:
```
.button
.button-primary
.button-secondary
.button-large
```
I'm happy to go with whatever the general consensus is though. Not sure how to apply this to functions/mixins...

kball:
Ok, so I propose we move forward with this as is, and move the naming conversation into a separate location.

Issue #7373, opened 2015-12-03
[Forms] Input Group button sizing variation 6.0.5


bwydev:
The changes in 6.0.5 recommend adding a wrapper for the input-group-button class. However the button ends up larger than the input field with this method. 

Gives taller button than input (6.05 method):

```
<div class="input-group">
    <input class="input-group-field" type="text" name="keywords" placeholder="Search">
    <div class="input-group-button">
        <input type="submit" class="button" value="Search">
    </div>
</div>
```

Gives equal sized button (legacy method: works):

```
<div class="input-group">
    <input class="input-group-field" type="text" name="keywords" placeholder="Search">
    <a href="#" class="input-group-button button">Search</a>
</div>
```

Docs: http://foundation.zurb.com/sites/docs/forms.html
Releases: https://github.com/zurb/foundation-sites/releases/tag/v6.0.5

joshmedeski:
I changed the $button-sizes: default value and it caused the same issue. I'm going to fix it now and submit a PR :smile: 

If you want to make sure it lines up correctly, make sure you keep the $button-sizes default value at 0.9rem. Hope it helps.

joshmedeski:
So how do you guys want to do about solving this problem?

**The Problem**
Buttons/links that are in an input group do not align unless the settings are at their default values for `$button-group:` default and `$input-prefix-padding` and `$input-font-size`. If you change any of these values, the buttons/links and inputs do not align vertically.

**The (Possible) Solution**
_Match Button with Input Styles:_ We can match the button with the input's styles so that the button's height, padding, and font-size match the input's height, padding and font-size.

**Further Discussion**
When someone is using the input group, they want the button and input to align vertically. Right now, it is really easy to change something in the settings that will break this vertical pixel perfection.

I'm willing to take some time to fix this problem, please let me know if I should take the time to invest in fixing it or if it needs further discussion.

andycochran:
@joshmedeski Seems to me that if you group a button and input, you want the button height to match the input height, regardless of the default button size. So `.input-group-button > .button { … }` should override default button styles.

stianpr:
I'm not sure this is the correct issue but having a `<button>` inside an `.input-group-button` with a large text will cause it to wrap, making the button much higher than the input field. I fixed this by adding `white-space:no-wrap` on the `<button>`. I think it's how the text for `<input>` buttons differ from a text node in ´<button>` in the case of wrapping due to width of the container. Hope this made some sense.

[...]

PR #8695, opened 2016-04-28
prevent input group buttons wrapping to multiple lines

andycochran:
Buttons in an Input Group that have multiple words were wrapping to multiple lines unless they were an `<input>` tag (which default to `white-space:pre`). So I've added `white-space:nowrap` to prevent both `<button>` and `<a>` tags from wrapping to two lines if they are multi-word. 

Bug mentioned here: https://github.com/zurb/foundation-sites/issues/7373#issuecomment-215392389.

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -46,6 +46,7 @@ $input-prefix-padding: 1rem !default;
 
   %input-group-child {
     margin: 0;
+    white-space: nowrap;
 
     @if not $global-flexbox {
       display: table-cell;


kball:
Are there any situations where we might not want this?  @brettsmason @JeremyEnglert does this seem like a reasonable default to you?

brettsmason:
I've just played around with this and I think its a good solution. If you are setting up an input group I'm confident you would want this as the default behaviour.

Good to merge!

andycochran:
@kball FWIW, this PR makes `.button` elements consistent no matter their tag. They shouldn't function differently. So `<input class="button">` should be changed to also wrap multiple lines if this isn't a desired default. 

But it is. 😉

kball:
OK sounds good

Issue #8690, opened 2016-04-27
Equalizer removes event listeners from element


flyingL123:
I'm not sure what I'm missing here, but how am I supposed to successfully bind to the `postequalized.zf.equalizer` equalizer event? The equalizer source code contains a `_pauseEvents()` method that contains the line:

    this.$element.off('.zf.equalizer resizeme.zf.trigger');

This removes any previously bound events. In other words, if I bind to the event, then initialize foundation, the event gets removed:

```
$('.productHeader').on('postequalized.zf.equalizer', function() {
  console.log('heights updated');
});

$(document).foundation();
```

In this case, the even never fires since it gets removed when equalizer is initialized. It will also be removed if the screen size changes to below the `equalize-on` value.

Shouldn't there be some sort of logic where the pausing and enabling events methods account for any other listeners that have been bound to the element by the user?

designerno1:
Same issue
http://codepen.io/marvinhuebner/pen/mPzrjN?editors=1111

kball:
I believe this is fixed in #8690 which was merged, and will be in 6.2.2.  We'll be putting out a 6.2.2 release candidate later today so if you want to test with that and verify, that would be great.

dioogbento:
Sorry to post on a closed topic. I've had a similar issue and arrived here while searching for a possible solution. I'm no expert in javascript and would like to understand why the equalizer event has to go before the foundation initialization. Thanks!

donaldG:
With @dgobnto here. I'm having no luck binding to this event or more likely I'm not understanding how to do it correctly. Following the example in the first comment but I get nothing in my console regardless of what element (document, window, div) I use & whether I do it before or after Foundation initialization.

PR #8696, opened 2016-04-28
possible fix for #8690

designerno1:
possible fix for #8690
assign a this var with a function.bind(this) to off only the function on _pauseEvents();
@kball 
don't know if this way is best practice JS, would you please take a look

### DIFF
--- a/js/foundation.equalizer.js
+++ b/js/foundation.equalizer.js
@@ -38,7 +38,10 @@ class Equalizer {
     this.hasNested = this.$element.find('[data-equalizer]').length > 0;
     this.isNested = this.$element.parentsUntil(document.body, '[data-equalizer]').length > 0;
     this.isOn = false;
-    this._bindHandler = this._bindHandlers.bind(this);
+    this._bindHandler = {
+      onResizeMeBound: this._onResizeMe.bind(this),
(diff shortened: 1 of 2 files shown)


kball:
@designerno1 thanks for pushing this forward, and for adding the visual tests.  I'd prefer we break these handlers out into individual handlers per events, but this approach of binding them is good practice.

kball:
@designerno1 do you have time to take this PR and break the handlers into individual events so we can get this in and fix #8690?

designerno1:
@kball 
really split the handlers? thats 1 more `this.var` for the bind and so for every on event to `off` we need a function
i would like to do this in vars so the _bindHandler and _bindfunctions won't be accessible from outside and don't stored on the this but i can't scope a var that's accessible from the functions, hmmmm

kball:
I'm open to other opinions @colin-marshall @Owlbertz but from my perspective if the logic is different for each type of event we should have a handler per event, not a joint handler with a condition.  I'm more concerned with readability and maintainability than one more var on this, though if you want you could make this.handlers an object and have one per event.

Owlbertz:
I think the second commit with the split makes it a bit more readable, and since it's now already there I guess we should go this way.

PR #8698, opened 2016-04-28
V6.3 drilldown scrolltop

designerno1:
Drilldown feature scroll top #8633 
adds feature ScrollTop to Drilldown Menu scrollTop scrollTopElement scrollTopOffset animationDuration animationEasing
#8617

### DIFF
--- a/docs/pages/drilldown-menu.md
+++ b/docs/pages/drilldown-menu.md
@@ -75,7 +75,7 @@ Any `<a>` without a submenu will function like a normal link.
 
 ## ScrollTop Drilldown
 
-<div class="callout">Scroll to the top on change the menu - Option scrollTop</div>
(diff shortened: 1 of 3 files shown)


kball:
@brettsmason I know you were interested in this in the issue, can you check this PR out and merge if it looks good?  Thanks

brettsmason:
@designerno1 This looks good thanks. Can i suggest a few changes to the description before I merge this so its clearer what it does? I think this will be a popular feature, so want to make sure people understand.

`Scroll to the top on change the menu - Option scrollTop`

Change to something like: `Scroll to the top of the menu when selecting a submenu/navigating back using the back button. Can be useful with a longer menu to provide a better user experience.`

Then under the plugin options on the docs we have:
`Scroll the menu after open,hide,closed` - change to something like: `Scroll to the top of the menu after opening a submenu or navigating back using the back button`

Does that make sense? Feel free to change to what you think would work best though.

brettsmason:
Awesome thanks @designerno1 this looks good to merge.

PR #8699, opened 2016-04-28
V6.3 drilldown autoheight

designerno1:
Drilldown feature autoheight #8607 

this time i used the existing heightcalc, but i think this heightcalc needs a overhaul, because not every child in the menu could have the same height (breaking into 2 lines etc.)

ads features to drilldown
data-auto-adjust-height and data-animate-height
#7821

### DIFF
--- a/docs/pages/drilldown-menu.md
+++ b/docs/pages/drilldown-menu.md
@@ -73,13 +73,14 @@ Any `<a>` without a submenu will function like a normal link.
   <li><a href="#Item-4"> Item 4</a></li>
 </ul>
 
-## autoAdjustHeight
+## autoHeight
 
 <div class="secondary callout">
-  <p>If you like to set the height to auto you can also set the autoAdjustHeight and animateHeight params</p>
+  <p>If you like to set the height to auto you can also set the autoHeight and animateHeight params</p>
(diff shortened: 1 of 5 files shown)


kball:
@colin-marshall can you take a look at this and merge if it looks good?

brettsmason:
From a non-js users point of view this is great. I've just pulled this down and it works brilliantly.
I'll leave it to @colin-marshall to merge though if he is happy with it too!

emri99:
Confirmed ! It works as expected ;) Thanks @designerno1 

agreed, heightcalc need overhaul.
Another problematic use case I experiment currently is using menuitem without text, just an icon. Even in drilldown, I don't want them to be stacked and use float property to align them left or right in one line. heightCalc add a margin-bottom. This is a very specific use case, agreed too ;)

Also wondering the real need of ```.each``` block in ```_getMaxDims``` method.
Why do you force height when drilldown is closed as the children will define height properly natively ?
I tried to completely comment it without any problems. This even fix my above use case.
Did I miss anything ?

colin-marshall:
@designerno1 functionality is great! Nice work!

A few observations:

1. Can we simplify the name of the option to `data-auto-height` or `data-variable-height` instead of `data-auto-adjust-height`?
2. Defaults are missing for the 2 new options
3. Docs page should have examples of auto height being on and off.

Other than that I think this is good to go.

designerno1:
@colin-marshall 
1. renamed it
2. defaults i have missed now added
3. i added a button with inline onclick JS to the docs to switch autoHeight

later the day i take a look @emri99 patch, and fix the conflicts

designerno1:
@emri99 
i remember why i put the heights on autoHeight not to auto with auto the menu got cut of after the initial height on change because of the position style of the element, so i had to set the calced heights to prevent it from get cut of, maybe i can verify that later with a visual test, but thats the reason i remeber

kball:
This is great!  Merged.

Issue #8685, opened 2016-04-27
[select] Arrow overlapping text when container is floated


brettsmason:
**How can we reproduce this bug?**

1. Create a select input with decent length options
2. Float the container
3. The arrow will be covering the end of the option

[CodePen link](http://codepen.io/brettsmason/pen/EKdYrx)

I have worked out how to solve the issue. At the moment the arrow is in the main area and not the padded area. I suggest changing the select padding on the right (or left if RTL?) to 1.5 x the form padding, and then setting a minus right for the background SVG.

See attached before and after screenshots.

![select-before](https://cloud.githubusercontent.com/assets/73229/14858415/f49d848a-0c97-11e6-9473-5bb9f52fd7b7.jpg)

![select-after](https://cloud.githubusercontent.com/assets/73229/14858420/f88cb516-0c97-11e6-8004-93973c6fdaba.jpg)

Does this approach to fixing it sound OK? If so I'll add a PR.

PR #8704, opened 2016-04-29
Fix overlapping select arrow

brettsmason:
Fixes the overlapping select arrow by adding additional padding to `$global-right` and changing the arrows background position accordingly.

Closes [#8685](https://github.com/zurb/foundation-sites/issues/8685)


andycochran:
Awesome. Looks good to me. Merging…

PR #8706, opened 2016-04-29
fix radius and hover on Accordion

### LINKED ISSUE #8323: No $global-radius applied to accordion-title on hover. Here's my solution
Note: Working with foundation sass.
1. Apply a noticeable global radius.
2. Place accordion with title in template.
3. Hover over title.

Hover background color to have global radius applied.

**What happened instead?**
Hover background color has no global radius applied.
**Test case:**

To solve the issue, I added a few lines of code to 
`components/_accordion.scss `
at bottom of accordion-title mixin like so:

```scss
@mixin accordion-title {
  display: block;
  padding: $accordion-item-padding;
  line-height: 1;
  font-size: rem-calc(12);
  color: $accordion-item-color;
  position: relative;
  border-bottom: $accordion-content-border;

// MY CODE BELOW
  &[aria-expanded='false'] {
    border-radius: $global-radius;
  }
  &[aria-expanded='true'] {
    border-radius: $global-radius $global-radius 0 0;
  }
// MY CODE ABOVE
  &:hover,
  &:focus {
    background-color: $accordion-item-background-hover;
  }
  @if $accordion-plusminus {
    &::before {
      content: "+";
      position: absolute;
      #{$global-right}: 1rem;
      top: 50%;
      margin-top: -0.5rem;
    }
    .is-active > &::before {
      content: "–";
    }
  }
}

```
You may find have a better way to implement this, but 'ere ya go.

andycochran:
Closes #8323. Radius was applied to the `accordion-container`. So titles overflowed the corners. I also moved the `$accordion-content-border` to the `accordion-title` and `accordion-content` (instead of `accordion-container`) so that they can work with the radius.

### DIFF
--- a/scss/components/_accordion.scss
+++ b/scss/components/_accordion.scss
@@ -46,10 +46,17 @@ $accordion-content-padding: 1rem !default;
 @mixin accordion-container {
   list-style-type: none;
   background: $accordion-background;
-  border: $accordion-content-border;
-  border-bottom: 0;
-  border-radius: $global-radius;
   margin-#{$global-left}: 0;
+}
+
+/// Adds styles for the accordion item. Apply this to the list item within an accordion ul.
+@mixin accordion-item {
+  &:first-child > :first-child {
+    border-radius: $global-radius $global-radius 0 0;
+  }
+  &:last-child > :last-child {
+    border-radius: 0 0 $global-radius $global-radius;
+  }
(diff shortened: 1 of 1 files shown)


brettsmason:
Works perfecty, thanks @andycochran :+1: 
Goot to merge.

PR #8713, opened 2016-04-30
[Bugfix] Sticky element losing its height. Fixes #8435

### LINKED ISSUE #8435: [Sticky] Sticky element loses height when anchored

gehasia:
Anchored element is losing its original height after being stuck.
Fixes https://github.com/zurb/foundation-sites/issues/8435

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -239,12 +239,11 @@ class Sticky {
 
     css[mrgn] = 0;
 
-    if ((isTop && !stickToTop) || (stickToTop && !isTop)) {
-      css[stickTo] = anchorPt;
-      css[notStuckTo] = 0;
+    css['bottom'] = 'auto';
+    if(isTop) {
+      css['top'] = 0;
     } else {
-      css[stickTo] = 0;
-      css[notStuckTo] = anchorPt;
+      css['top'] = anchorPt;
     }
 
     css['left'] = '';


colin-marshall:
This one looks good to me. @zurb/yetinauts please have a look when you get a minute, the sticky visual test for callouts really demonstrates this problem.

colin-marshall:
Discovered something worth noting for you guys to take into consideration when checking this out.

If I make the div with the Sticky class as the parent of the callout div the height issue goes away (before merging the fix from this PR). The same thing applies to menus. This syntax works as is:

```html
<div class="container" data-sticky-container>
  <div class="sticky" data-sticky data-margin-top="0" data-anchor="foo">
    <div class="callout" >
      <a href="http://foundation.zurb.com/sites/docs/stick.html" class="small button">Go to Sticky Docs</a>
    </div>
  </div>
</div>
```

This syntax has the issue:

```html
<div class="container" data-sticky-container>
  <div class="callout sticky" data-sticky data-margin-top="0" data-top-anchor="foo6" data-btm-anchor="foo8:bottom">
    <a href="http://foundation.zurb.com/sites/docs/stick.html" class="small button">Go to Sticky Docs</a>
  </div>
</div>	
```

colin-marshall:
I have a little more to on add on to my previous post. If you use the following syntax...
```html
<div class="container" data-sticky-container>
  <div class="sticky" data-sticky data-margin-top="0" data-anchor="foo">
    <div class="callout" >
      <a href="http://foundation.zurb.com/sites/docs/stick.html" class="small button">Go to Sticky Docs</a>
    </div>
  </div>
</div>
```
...the callout has a bottom margin, so the sticky container includes that in its height and the anchoring is off by that much. I changed all of the Sticky element's to use that syntax. See here:

screen shot 2016-05-10 at 11 54 55 am

My questions for the @zurb/yetinauts, @gehasia, and anybody else that wants to give feedback:

1. Should we direct users in the docs to put the `.sticky` div as a parent of the element they want to sticky? This would essentially make you use 2 container divs for Sticky, one for `[data-sticky-container]` and one for `[data-sticky]` / `.sticky`.

^^^ If we did it this way then:
2. How should we then handle margins like in the example I gave above with the callout bottom margin?
3. it would prevent potential collision between having 2 Foundation plugins initialized on the same element. I'm not sure if advising against that (2 plugins on same element) is clearly stated in the docs, and it probably should be, right?
4. Do we still implement the fix in this PR?

Sorry for the barrage of info/questions, I need some good feedback to make sure that Sticky is headed in the right direction. Thanks!

kball:
re point 3, we pretty definitely state `A single element can only have one Foundation plugin on it at a time. However, most plugins can be nested inside other ones.`  Perhaps could be louder/repeat more often, but that is clearly in the docs

kball:
Re point 1, my inclination would be only if absolutely necessary.  The more we can avoid requiring extremely bulky HTML to accomplish goals, the better.  The amount of containerization/bolierplate to handle off canvas still irks me. :P

PR #8714, opened 2016-04-30
[Bugfix] Sticky element badly anchored when it's container is in anot…

gehasia:
Fixes bad position of a sticky element when it's container is in another div than his top anchor.
See test #4 after merging https://github.com/zurb/foundation-sites/pull/8674/files


colin-marshall:
@gehasia with this fix test #4 sits in `#foo7` until you reach the anchor `#foo6`. Don't we want it to sit in `#foo6` instead of `#foo7` when it's anchored at the top? The poor placement of the callout as a child of `#foo7` was intentional to account for different use cases and developer error.

gehasia:
Hi, sorry for the delay i was on holidays !

I don't really know what would be the 'ideal' behavior as it strange to put the sticky element in a container and his anchors in others. 
But as i see it in this case :
-Parent element of the sticky element is #foo7, i don't see any reason to change that, it should sit there during the "normal flow". We don't have to reposition this element if developer put it here.
-Anchor has been set by the developer on #foo6 and #foo8, he probably has a good reason to do so, if it's not a mistake. Anchors are just triggers, so when they are reached, we trigger the 'stick to'. And when it sticks, in this case, it jumps to window top.
I don't see how it could be done differently, forcing it to sits in #foo7 would be changing the developer (un)intentional behavior, it's going against him for me. But that's just my opinion !

I know it was intentional, it was a good test case ;) !

colin-marshall:
@gehasia hope your holidays went well!

For test 4 I agree with you on those points. What about for situations like test 8 though? It's similar to test 4, but I wouldn't say the placement is bad. It's just that its top anchor point is below the top of the container it's in.

My thought was that we could put in a check during initialization to see if the Sticky element is a descendant of its top anchor div. If it's not a descendant it could get the positions of the top anchor div and set the `top` CSS value of the sticky element to place it in line with `$('#top-anchor').offset()`.

kball:
@colin-marshall @gehasia What do you consider the state of this PR?  It looks like the conversation might have still been ongoing?

gehasia:
Hi !

I'm away from the project involving foundation since june because of other priorities. I'll be back on it in two weeks if everything goes well, i'll try to finish things i left opened then  !

kball:
@gehasia do you have time to look at this now? If it is ready I'd like to get it merged ASAP - we're going to close off the 6.2.4 release soon and ship it.

kball:
Unfortunately looks like this is going to slip past 6.2.4.  @gehasia and @colin-marshall would love to get this re-invigorated for the next release.

gehasia:
I'm sorry for this, i can't move my work schedule, i'm going to be on another project until the 15th october, i hope the deadline won't be moved again. I can't look at it before i reintegrate the 'foundation' project, i'm really (really) too busy to work on it on spare time !

kball:
@gehasia don't suppose you've got time to look at this again? ;)

kball:
Pushing this out to 6.3.1

kball:
@gehasia and @colin-marshall - what is the status of this PR?

colin-marshall:
@kball maybe put this on hold since we are going to redo Sticky?

kball:
Seems reasonable.

PR #8715, opened 2016-04-30
[Bugfix] Sticky : Anchored element's not following the flow when resi…

gehasia:
Fixes anchored element not following the flow when resizing the window and going over/behind other elements.
See https://github.com/zurb/foundation-sites/issues/8587
and video 
http://sendvid.com/52a7xxmg


kball:
Looks good

PR #8718, opened 2016-04-30
[WIP] Allow to always define components in classes

ncoden:
### Purpose

Clearly separate **reset** properties from **component structural and graphics** properties. Currently, this separations is very unclear for all `foundation-forms` components

Allow to choose if Foundation must define its components into the global namespace (HTML standard tag names) or in its own classes. Currently, lot of component define their properties (no for reset) on tags, which can lead to **unwanted properties** for someone using its own components.
### Current changes

_Updated 30 April 2015, 8:00 AM_
- Add `$global-namespace {Boolean} [false]` option which when enabled, make all components be defined in classes, instead of tags.
- Use `$global-namespace` for `label` and `select`
### Why [WIP] ?

I given the direction (how components should be defined) and added the basics (`$global-namespace` setting). But my knowledge in Foundation is too limited to differentiate _reset_ from _component_ properties in `foundation-forms`.

So I hope you can bring me feedback, ideas or explanations to finish that work :)

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -94,6 +94,10 @@ $global-text-direction: ltr !default;
 /// @type Boolean
 $global-flexbox: false !default;
 
+/// Always define components with classes instead of tags (besides the reset).
+/// @type Boolean
+$global-namespace: false !default;
+
 @if not map-has-key($foundation-palette, primary) {
   @error 'In $foundation-palette, you must have a color named "primary".';
 }
--- a/scss/forms/_label.scss
+++ b/scss/forms/_label.scss
@@ -38,7 +38,9 @@ $form-label-line-height: 1.8 !default;
 }
 
 @mixin foundation-form-label {
-  label {
+  $name: if($global-namespace, '.label', 'label');
+
+  #{$name} {
     @include form-label;
 
     &.middle {
--- a/scss/forms/_select.scss
+++ b/scss/forms/_select.scss
@@ -60,7 +60,9 @@ $select-radius: $global-radius !default;
 }
 
 @mixin foundation-form-select {
-  select {
+  $name: if($global-namespace, '.select', 'select');
+
+  #{$name} {
     @include form-select;
   }
 }
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -71,6 +71,7 @@ $global-weight-bold: bold;
 $global-radius: 0;
 $global-text-direction: ltr;
 $global-flexbox: false;
+$global-namespace: false;
 $print-transparent-backgrounds: true;
 
 @include add-foundation-colors;


ncoden:
I invoke @kball ;)

DaSchTour:
I guess this might be a change considered for a 6.5 or something like this.
The idea itself isn't that bad and might lead to a more consisten behaviour.
For example, why do I need to add .button to buttons? While a table works with no class?

ncoden:
@DaSchTour Because you can want to do something like this :
```HTML
<div class="my-search-field">
    <input type="text"  class="my-search-field__input"/>
    <button class="my-search-field__button">
        <i class="fa fa-search"></i>
    </button>
</div>
```

For this particular component, you don't want graphic properties on `<input>` or `<button>`. You'll define your own graphic properties.

That's why I purpose to make it configurable, disabled by default. People who don't care and always use Foundation components will don't see any change.
But people who define their own components and don't want to overwrite all Foundation graphic properties will love the feature :)

DaSchTour:
Yes. I'm aware of this. Just wanted to point on the inconsistency ;)

kball:
interesting.  Yes, I think this is a good direction to move in... we have some other related inconsistencies in the JavaScript...

I think we need to figure out the desired state and then we can move gradually there.  As a first pass, I'm thinking something like...

1 - tags are targeted by reset, typography, and global colors
2 - all foundation-related component styling is targeted by classes (with all definitions actually in mixins so one can semantically define their own classes)
3 - All javascript behavior is manipulated by data-attributes... right now some components use classes to target things, but those should all be data-attributes.  The only way javascript should be dealing with classes is applying state classes
4 - ids can be used to target JS

Do these make sense?  Anything I'm missing?

cc @zurb/yetinauts

ncoden:
@kball It make sense 🙂. But what do you mean by "global colors" ?

Also, the best at term would be to purpose in order :
- config
- simple reset (standardize sizes and dispositions **on tags**)
- *or* hard reset (remove all graphic properties, *a future feature ?*)
- tools, base, utils
- components (**on tags or classes, on choice**)

There is an OOCSS approach specifically oriented around the rule definition order, which is called **ITCSS** :
ITCSS description

And the best of the best would be to allow **separate _structural_ and _graphical_ properties**, for the devs which doesn't like the default Foundation theme (like me, sorry), and doesn't want to manually overwrite each property.

*It could also lead to make Foundation themes more simple to create, or Foundation more usable in design companies, by purposing a set of fully functional and easily customizable components*.

kball:
@ncoden wrt global colors I meant things like setting core colors for links, typography (if you're using e.g. #333 for black), etc.

kball:
@ncoden this is currently labeled with the 6.3 milestone... given the WIP status of this, all of the other items you're working on, and the fact that the first RC for 6.3 is a week and a half from now, my inclination is to push this out to 6.4.  What do you think?

ncoden:
@kball I agree. Pushed out to `v6.4`

ncoden:
Closing this to reopen a clean and up-to-date PR.

PR #8719, opened 2016-05-01
[Bugfix] Sticky : Fixes #8393. anchor object undefined when destroyin…

### LINKED ISSUE #8393: [6.2 Sticky] Destroy errors out when "anchor" is not specified
When using something like this:
```
<nav class="sticky-nav" data-sticky-container id="container">
	<div data-stick-to="bottom" data-top-anchor="page" data-btm-anchor="container:bottom">
		Sticky Stuff
	</div>
</nav>
```
```
$('.sticky-nav').foundation('destroy');
```

Get an error:

> Uncaught TypeError: Cannot read property 'off' of undefined

It originates from this line: `this.$anchor.off('change.zf.sticky');`

And can be fixed by checking if $anchor is populated (as is done elsewhere in the code) before removing the event listener 
```
if (this.$anchor && this.$anchor.length) {
    this.$anchor.off('change.zf.sticky');
}
```

gehasia:
Fixed object undefined error when destroying sticky element with no anchors set.
Fixes https://github.com/zurb/foundation-sites/issues/8393

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -355,8 +355,9 @@ class Sticky {
                    'max-width': ''
                  })
                  .off('resizeme.zf.trigger');
-
-    this.$anchor.off('change.zf.sticky');
+    if (this.$anchor && this.$anchor.length) {
+      this.$anchor.off('change.zf.sticky');
+    }
     $(window).off(this.scrollListener);
 
     if (this.wasWrapped) {


colin-marshall:
@gehasia thank you for your hard work on improving Sticky. I'm going to go through these tonight and add some more visual tests before we merge.

gehasia:
Ok, no prob !
Don't forget to pull my PR for doc an test first then !
https://github.com/zurb/foundation-sites/pull/8674

PR #8722, opened 2016-05-01
[New Feature] New component cards

DaSchTour:
While starting a new project I started with Bootstrap 4 because of the card component, that I hoped is similar to what Semantic UI offers. But just after I started I discovered that bootstrap cards and the grid system they use does not suite my usecase so I thought I could create cards for Foundation, which I know think is the bester framework and just missing some components.

I'm waiting for some feedback to improve this as fast as possible, so that this can be released with 6.3 I hope.

### DIFF
--- a/docs/pages/cards.md
+++ b/docs/pages/cards.md
@@ -1,0 +1,150 @@
+---
+title: Cards
+description: Extending columns to cards with header, footer and image support.
+sass: scss/components/_cards.scss
+---
+
+## Basics
+
+A card is an extended column with the `.card` class applied.
+
+```html_example
+<div class="row">
+    <div class="column card">
+        <div class="card-block">
+            <h4 class="card-title">Card title</h4>
+            <p class="card-text">This is the card content block.</p>
+        </div>
+    </div>
+</div>
+```
+
+## Image
+
+A card can contain an image as first element or in between.
+
+```html_example
+<div class="row">
+    <div class="column card">
+        <img class="card-img-top" src="assets/img/generic/rectangle-1.jpg" alt="Card image cap">
+        <div class="card-block">
+            <h4 class="card-title">Card title</h4>
+            <p class="card-text">Lorem ipsum sapientem ne neque dolor erat,eros solet invidunt duo Quisque aliquid leo. Pretium patrioque sociis eu nihil Cum enim ad, ipsum alii vidisse justo id. Option porttitor diam voluptua. Cu Eam augue dolor dolores quis, Nam aliquando elitr Etiam consetetur. Fringilla lucilius mel adipiscing rebum. Sit nulla Integer ad volumus, dicta scriptorem viderer lobortis est Utinam, enim commune corrumpit Aenean erat tellus. Metus sed amet dolore justo, gubergren sed. </p>
+        </div>
+    </div>
+</div>
+```
+
+## Header
+
+The card header is emphasized.
+
+```html_example
+<div class="row">
+    <div class="column card">
+        <div class="card-header"><h4>Card header</h4></div>
+        <div class="card-block">
+            <p class="card-text">Lorem ipsum sapientem ne neque dolor erat,eros solet invidunt duo Quisque aliquid leo. Pretium patrioque sociis eu nihil Cum enim ad, ipsum alii vidisse justo id. Option porttitor diam voluptua. Cu Eam augue dolor dolores quis, Nam aliquando elitr Etiam consetetur. Fringilla lucilius mel adipiscing rebum. Sit nulla Integer ad volumus, dicta scriptorem viderer lobortis est Utinam, enim commune corrumpit Aenean erat tellus. Metus sed amet dolore justo, gubergren sed. </p>
+        </div>
+    </div>
+</div>
+```
+
+## Footer
+
+The card footer is align at the bottom of the card. All footers are align equaly.
+
+```html_example
+<div class="row small-up-2 medium-up-3 large-up-4">
+    <div class="column card">
+        <div class="card-block">
+            <h4 class="card-title">Card title</h4>
+            <p class="card-text">Lorem ipsum sapientem ne neque dolor erat,eros solet invidunt duo Quisque aliquid leo. Pretium patrioque sociis eu nihil Cum enim ad, ipsum alii vidisse justo id. Option porttitor diam voluptua. Cu Eam augue dolor dolores quis, Nam aliquando elitr Etiam consetetur. Fringilla lucilius mel adipiscing rebum. Sit nulla Integer ad volumus, dicta scriptorem viderer lobortis est Utinam, enim commune corrumpit Aenean erat tellus. Metus sed amet dolore justo, gubergren sed. </p>
+        </div>
+        <div class="card-footer">
+            <small>This is the first footer</small>
+        </div>
+    </div>
+    <div class="column card">
+        <div class="card-block">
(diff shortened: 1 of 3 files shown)


andycochran:
I think [Bootstrap Cards](http://v4-alpha.getbootstrap.com/components/card/) are something that should go in a building block. It'd be easy to do this with Callouts, Equalizer, and a custom header/footer class. 

Foundation Sites 6 dropped components that were super easy to build yourself. Like [Pricing Tables](http://foundation.zurb.com/sites/docs/v/5.5.3/components/pricing_tables.html), for which I build this: http://codepen.io/andycochran/pen/MKxMQK

Thoughts @zurb/yetinauts?

DaSchTour:
Why use JavaScript Equalizer if this works without JavaScript. BTW: this is at the moment a Flexbox only component.

brettsmason:
Interestingly, cards were a feature back in very early alpha. It took a bit of searching [but you can see here](https://github.com/zurb/foundation-sites/commit/44d2388d1d26c8e99f9eaa9478752f519c65ae91).

I'm torn on this as it is a very common pattern, and I cant really see an issue with including it (I like choice). It would be interesting to know why it was removed and compare this PR to the original.

DaSchTour:
Should I maybe add some of the features of the original cards to my branch? I really would love to make this work.

andycochran:
Does a card need to be different from a callout? Can't it just be a callout with elements that have negative margin equal to the callout padding? I'd rather extend a current component that create a new one. 

How 'bout a `.callout-header` class with negative margin on the top/sides and a `.callout-footer` class with negative margins on the bottom/sides? Use `img.callout-[header, footer]` and you have a card.

colin-marshall:
@andycochran I like the idea of extending callouts to have more features like cards. Vertical menus could be combined with callouts to make the list groups seen in the first and second examples [here](http://v4-alpha.getbootstrap.com/components/card/#content-types).

DaSchTour:
I guess maybe callouts should be renamed, because this was never a component I had in mind in this context. From my point of view callouts are equivalent to [Alerts](http://v4-alpha.getbootstrap.com/components/alerts/).
That's why I always was wondering why the hell foundation doesn't have any component like [panels](http://getbootstrap.com/components/#panels) or [listgroups](http://getbootstrap.com/components/#list-group).

brettsmason:
I'm not sure on the idea of extending callouts on this occasion. I get the reasoning behind it, but for people who dont want cards but do want callouts or vice versa aren't we just adding extra they dont need? Its more flexible as a seperate component you can just comment out to exclude (if using Sass anyway).

There's definately room to use things like menus etc to work within cards too, and I expect they would pretty much work as is anyway.

I just think extending callouts on this occasion would create confusion with people not knowing/finding how to do cards. Just my opinion anyway!

andycochran:
@brettsmason You've changed my mind. I agree.

brettsmason:
@andycochran Awesome! I think we probably need to review the base requirements for cards and implement a nice and simple version (Bootstraps version has lots going on). I haven't looked through this PR properly yet and compared to the Zurb original and see what we could do to improve it for the better.

andycochran:
Then again, maybe a cards are (stacked) media objects. I dunno; cards are so easy to make. I'm not convinced Foundation needs 'em.

brettsmason:
I guess we could say that about quite a few of the components though. Top bar, media object, callouts and labels among others are all pretty simple, its just nice to have a solid, tested foundation to work off. I agree though if I needed cards I could easily add them myself, but same applies for the above too so where do we draw the line? :smile: Probably needs some input from others.

Silvester20:
Actually I never used those cards back in days, but it should be just a few lines to add such card into the core, right? So if the cards component doesn't increase the minified CSS filesize too much, then I would like to see this in the core.

rafibomb:
The original thinking was that cards are stacked callouts with margin-bottom on the last child only. I agree with @andycochran a building block could solve for this better than adding another component to the framework. The goal of Foundation is to give people a foundation to build their visual brand upon. We're also trying to keep it lightweight without a bunch of opinionated styles.

We're leaning towards this being a building block, or multiple building blocks to extend onto your project. Though we're reading everyones input and still open a best solution. 

What do you all think of adding these as Building Blocks?

Ultimately what would be best is if code like this can be added as an npm package. 
- You'd run a command to install it like `npm install foundation-card ...`
- it would be set up in the ZURB Template package.json file
- It would create the folders needed in SCSS and/or JS
- and build the settings in

We want to know what you think!

DaSchTour:
I think, that the way Foundation is build, that it uses mixins for every component makes it possible to make a streamlined version if needed. But as this is slowly evolving I started a project for additional Foundation components. We'll see how much I'll but in there.

brettsmason:
@rafibomb I like the idea of a separate package to install, however that could be problematic to cover all bases. For example I don't use the Zurb template and pretty much only use my own WordPress theme.

Personally I still think there is value in including cards. I know there is a small overlap with callouts, but I think they are different enough, and different use cases. Also wouldnt it be better to be able to control the styling without being tied to callout styling? I may also not want one or the other in a project, and because of the brilliant way v6 is setup, this is just a case of commenting out a line.

If it helps I've spent some time working on a `_cards.scss` file based on the original implementation, but have tidied it up and (hopefully) improved it. This is ready to go so I'd be happy to share.

marcusmoore:
Just to throw in my two cents: I really like the idea of having cards readily available and as a separate component.

rafibomb:
@brettsmason We'd be into seeing what you got and exploring ways to add it in. How can we get on that?

brettsmason:
@rafibomb @kball @DaSchTour I've submitted an alternative card version #9215 finally and would welcome feedback. The goal was to make it simpler and more of a foundation.

I'd welcome thoughts on making it more compatible with flexbox though if needed as its not something I use.

kball:
@DaSchTour and @brettsmason I'd like to find a unification between your two versions... I am inclined towards the simpler approach in #9215, but want to make sure we meet everyone's use cases.  @DaSchTour is your sense that #9215 would meet your needs?

DaSchTour:
@kball the main idea was to port the basics of http://semantic-ui.com/views/card.html to Foundation. From my point of view the most important feature is having equal height cards by using flexbox column feature.
While working on this feature I saw, that for example when using rounded borders and images there are some tweaks to add. So while looking on the content of my PR it's already as foundation as possible.
I also added some additional features to my own components Repo: https://github.com/DaSchTour/foundation-components/blob/master/scss/_cards.scss
As I didn't know if there will be any progress on this feature here.
At the moment for me it's to vague what needs to be improved. While some people seam concerned about the size of the framework I think it would be a good idea to spilt the framework into different modules (which in fact is already done in a great way and the main reason I like to use Foundation) so that there could be some optional components. As the users and usecases may be very different it would be great if foundation could also support additional more advanced or complex components as optionals. This optional components would have an own CSS in dist and would be only added when explicitly using scss include of the components mixin.

brettsmason:
@kball @DaSchTour I have made a small update to my card PR to include options for border radius and box shadow, along with `flex-direction: column` to the actual `.card` container.

Would you mind seeing if the flexbox requirements are met by the PR? I'm not sure where and what needs added to which part.
Also if there's anything else that could be improved?

I like your idea of more complex optional components - definitely worth discussing further.

kball:
Closing in favor of #9215 - thanks to everyone who engaged in the conversation on this and especially @DaSchTour for getting this ball rolling.  Cards will be in 6.3.

Issue #8725, opened 2016-05-03
Magellan does not account for barOffset when setting active menu item


flyingL123:
If Magellan is applied using the `barOffset` option, the `_updateActive()` method does not account for this value when setting the active menu item. This is particularly noticeable when using a sticky menu because if you click on a menu item, you are scrolled to the appropriate section, but the menu item before the selected one will get the `active` class instead of the menu item for the section you just scrolled to. This results in an odd user experience.

I believe it's a one-line fix, and it's working for me, so unless there are some other consequences, it should be an easy update.

Line 136 needs to be changed from:

```
return isDown ? p <= winPos : p - _this.options.threshold <= winPos;
```

to:

```
return isDown ? p - _this.options.barOffset <= winPos : p - _this.options.barOffset - _this.options.threshold <= winPos;
```

JeremyEnglert:
Looks like a PR was merged to fix this. Thanks @flyingL123!

PR #8726, opened 2016-05-03
Account for barOffset

flyingL123:
Creating a pull request for this  #8725


flyingL123:
As a demo, you can look at the Magellan page in the documentation. If you click one of the links in the Magellan menu at the top of the page, the content scrolls to the correct location, but the link you clicked does not actually get the active class.

kball:
👍

flyingL123:
Thanks for this. This was actually my first PR so I don't quite understand how it all works. At what point will this updated code be available in the version of Foundation I get if I install `npm install foundation-site`? I currently don't see it so I have been continuing to use my branch.

Issue #8697, opened 2016-04-28
Multiple clicks on tooltip cause funky behavior.


kball:
**How can we reproduce this bug?**

1. Go to tooltip page in docs
2. Click on tooltip once.  Tooltip will open.
3. Click on same tooltip again without clicking anywhere else.  Tooltip will close.
4. Click on same tooltip again without clicking anywhere else.  Tooltip will not open. 

**What did you expect to happen?**
3rd click should reopen tooltip

**What happened instead?**
Tooltip stayed closed

**Test case:**

- [Docs tooltip page](http://foundation.zurb.com/sites/docs//tooltip.html)

nachtwandler:
Hi,
in the foundation.tooltip.js is an if like this (line 267):
`if ((_this.options.disableHover || !_this.$element.attr('tabindex')) && !_this.isActive) {
     _this.show();
}`

I'm not shure, why theres a check for hover and tabindex inside the 'mousedown.zf.tooltip' event ?!
If disableHover is false, the tooltip will never show up on click.

coreysyms:
Im not sure the intended behavior is correct anyhow. The click should not close the tooltip if "hover" is the intended action, especially if the click is on the target tooltip span. If you hover and click the tip should stay open until you click the outside body, why should we expect to click click click, open and close, it's "hover" to activate, "click" to extend.

nachtwandler:
I don't know, which behavior is the right here. I would say "hover" is only an addition to the "click" open and close. There isn't more content to show and if the click does nothing, it "feels" broken.

coreysyms:
Right, I'm saying we fix it to be open on hover, stay open on click, if you click the body (outside the span) it closes, if you click the span again the tip should just stay open, not close like it does now.

nachtwandler:
Ok, got it. That sounds reasonable.

coreysyms:
@zurb/yetinauts thoughts? Marking as feedback wanted

Owlbertz:
Agree with you, that should be the desired behavior.

colin-marshall:
@coreysyms I agree with you, and fwiw, that appears to be the default behavior used in Bootstrap.

coreysyms:
Ill take this and submit a PR

coreysyms:
Closing this in reference to PR https://github.com/zurb/foundation-sites/pull/8730 which should now correct the issue.

PR #8527, opened 2016-04-05
[tooltip.js] New option allowHtml

web2client:
Ref pr #8075 Add new option allowHtml (default false) to allow HTML in tooltip.
@gakimball @kball comments pls.

PR #8730, opened 2016-05-03
change default behavior of tooltips

coreysyms:
In reference to https://github.com/zurb/foundation-sites/issues/8697

This will change the default behavior of the tooltips to:
- mouseenter = open
- mouseleave = close
- mouseclick = stay open / no toggle

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -40,7 +40,7 @@ class Tooltip {
     this.template = this.options.template ? $(this.options.template) : this._buildTemplate(elemId);
 
     this.template.appendTo(document.body)
-        .html(this.options.tipText)
+        .text(this.options.tipText)
         .hide();
 
     this.$element.attr({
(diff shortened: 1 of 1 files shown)


coreysyms:
@kball, since you were the originator of https://github.com/zurb/foundation-sites/issues/8697 can you verify this PR?

designerno1:
@coreysyms 
with data-tip-hoverable set to true mouse over link opens tip hovering in span stays open as expected.
hovering back to the link the tip closes, shouldn't it stay open?

designerno1:
@coreysyms 
as #8527 allowHtml is an option in v6.3

coreysyms:
@designerno1 I wrestled with this, and went back and forth on would a user navigate the mouse back and forth between the span and the tip, ultimately I felt that the cycle had to end somewhere so I left the tip close if you leave the tip, IE giving the power to the tip instead of the span, and in a sense, 'resetting' the cycle of what gets hovered.

As for the ```allowHTML``` option in 6.3, was using 'develop' as the master, and I didn't see that and can adhere to 6.3's behavior.

coreysyms:
hang loose on this, going to break these things in to 2 different PR's, one for the fix, one for the new feature

coreysyms:
ok, changed this PR to be just a bug fix to change the default behavior of the tooltip

kball:
OK fix looks good.  Thanks @coreysyms

Issue #8739, opened 2016-05-05
[Media Object] stack-for-small doesn't work with flex grid


brettsmason:
When checking on issue [#8738](https://github.com/zurb/foundation-sites/issues/8738) I noticed the `stack-for-small` class is not working. This is because `flex-wrap: wrap;` is being applied to all breakpoints, not just small. 

See [this codepen](https://github.com/zurb/foundation-sites/issues/8738) and add the `stack-for-small` class.

I'll sort out a PR to fix both these issues now.

kball:
Fixed by #8740

PR #8740, opened 2016-05-05
Media object flexbox fixes

### LINKED ISSUE #8738: Media object causes long horizontal scroll bar with `$global-flexbox: true;`
Just adding a media object with `$global-flexbox: true;` causes a long horizontal scroll.

Seems like an easy fix, just needs 

```
.media-object-section {
    -webkit-flex: 0 0 auto;  
    -ms-flex: 0 0 auto;  
    flex: 0 0 auto;  
}
```

to change to 

```
.media-object-section {  
    -webkit-flex: 0 1 auto;  
    -ms-flex: 0 1 auto;  
    flex: 0 1 auto;   
}
``` 

Here's a codepen that demonstrates the bug, with the suggested fix that I'm using at the moment.

http://codepen.io/tbredin/pen/VaqrZY

brettsmason:
Update `media-object-section` flex property, and wrap `stack-for-small` in zero breakpoint so it doesn't get applied to all breakpoints.

Closes #8738 and #8739

### DIFF
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -32,7 +32,7 @@ $mediaobject-image-width-stacked: 100% !default;
 /// @param {Number} $padding [$mediaobject-section-padding] - Padding between sections.
 @mixin media-object-section($padding: $mediaobject-section-padding) {
   @if $global-flexbox {
-    flex: 0 0 auto;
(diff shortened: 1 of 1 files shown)


kball:
Looks good

PR #8743, opened 2016-05-05
Added showAll using down() for all elements.

Owlbertz:
Like in #8460 purposed by @crbranch the hideAll method should use the up() method.
Also showAll() could be useful.

Basically the same as #8493, but against v6.3.


kball:
Awesome

Issue #8628, opened 2016-04-19
Reveal modal with 'full' class allows page scrolling on iOS


AndresUnity:
**How can we reproduce this bug?**

1. Create a test with page that has more content than fits in an iPhone viewport. Something that requires scrolling to read.
2. Open a Reveal modal that has the 'full' class on iOS:  `class="reveal full"`
3. Scroll
´
**What did you expect to happen?**

I expect the html and body elements to not scroll. The content beneath the modal should not be visible.

**What happened instead?**

I saw the rest of the content beneath the modal. The modal just scrolled with everything else.

**Fix**

At the moment I've done to the HTML element the same thing Foundation already does for the BODY element: add the class `.is-reveal-open`

```
html.is-reveal-open {
    overflow: hidden;
}

$(document).on('open.zf.reveal', function() {
    $('html').addClass('is-reveal-open');
});
    
$(document).on('closed.zf.reveal', function() {
    $('html').removeClass('is-reveal-open');
});
```

Owlbertz:
Thanks for reporting this. Could you check on how this interferes with #7548 (and the issues that link to this one)? Reveal and scrolling on mobile, especially iOS, was always a bummer. Your solution seems pretty simple, so if you can confirm this works and fixes your issue as well as the one(s) mentioned above, we're happy to see a PR!

PR #8744, opened 2016-05-05
Fix scrolling issue with reveal on ios

kball:
Fixes https://github.com/zurb/foundation-sites/issues/8628 and https://github.com/zurb/foundation-sites/issues/7548

### DIFF
--- a/scss/components/_reveal.scss
+++ b/scss/components/_reveal.scss
@@ -113,6 +113,13 @@ $reveal-overlay-background: rgba($black, 0.45) !default;
   body.is-reveal-open {
     overflow: hidden;
   }
+  // html gets this class only in iOS
+  html.is-reveal-open,
+  html.is-reveal-open body {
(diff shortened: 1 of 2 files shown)


kball:
@zurb/yetinauts anyone else with an ios device to test that can verify this and merge?

designerno1:
LGTM
Tested on my DEV Ipad 1

Owlbertz:
I checked with an iPhone 5 with the following results:

The close button's position is somewhat odd when there is a large heading:
image1

Scrolling inside the modal works much better now, but there is a small issue; the modal itself can be scrolled down by about 50px:

img_1432

kball:
@Owlbertz I could not find a way to disable the slight 'bounce' amount of scroll in the browser, only the general scroll... based on your feedback and @designerno1's I'm going to call this good enough because it moves the ball forward, though it would be nice to do further refinement.

PR #8746, opened 2016-05-05
prevent cb() from fire multiple as move is called until animation fin…

### LINKED ISSUE #8724: [Accordion Menu] down and up fz.accordionmenu fire multiple times
I provided this codepen with the latest foundation css and js included.

http://codepen.io/anon/pen/RaqyVY

Console.log outputs the message 16 times. It changes when I add data-slide-speed = "0" to 2 times. I'm expecting it to run once either way. Is this the intended behavior or is it a bug?

designerno1:
fixes #8724 
@kball 
would you please take a look at the JS

``` js
Foundation.Move(this.options.slideSpeed, $target, function(){
      $target.slideUp(_this.options.slideSpeed, function () {
```

in the old version of Foundation.Move the `fn.apply(elem);` gets called until the anim finishes, but in the fn. `slideDown` `slideUp` is called so the slidedown and slideup gets called in a 250 duration about 11 times, so the callback specified in the slideDown/slideUp also fires 11 times

so i wrapped the fn.apply in the !start if, to only call it once the animationframe is requested.
As i don't know what `requestAnimationFrame` is about and exactly how it works i don't know if thats the right way as maybe if the animationFrame is queued the slideDown ever fires.

### DIFF
--- a/js/foundation.util.motion.js
+++ b/js/foundation.util.motion.js
@@ -25,10 +25,12 @@ function Move(duration, elem, fn){
   // console.log('called');
 
   function move(ts){
-    if(!start) start = window.performance.now();
+    if(!start) {
+      start = window.performance.now();
+      fn.apply(elem);
+    }
     // console.log(start, ts);
     prog = ts - start;
-    fn.apply(elem);
 
     if(prog < duration){ anim = window.requestAnimationFrame(move, elem); }
     else{


kball:
I think this approach breaks Move for other cases... one thing I wonder is if we need Move at all?  I think we only need move if we need to apply per-frame changes.  Can we just remove the call to Move and only use slideDown and slideUp within the accordion menu?

designerno1:
should be possible

PR #8752, opened 2016-05-07
[Bugfix] Windows visual tests

### LINKED ISSUE #8749: Running the tests under Windows works but only after physically copying the `assets` directory

Just run `npm run test:visual` under Windows. in case you have this possibility.

I expected `.html`-files to have the correct asset paths

**What happened instead?**

The assets were not loaded, because the asset paths were either wrong or the assets were missing in the paths. 

The asset path seems to be symlinked, which is not working under Windows. 

**The fix for Windows is rather simple:**

For `npm run test:visual` the directory `_build\assets`has to be copied into `test\visual` and for 
`npm run test:javascript` the directory `_build\assets`has to be copied into `test`.

A final remark for the [README.md](https://github.com/zurb/foundation-sites/blob/develop/README.md). Running the Foundation tests requires the documentation to be generated first with `npm start`, which is not explicitely stated. Generating the documentation also requires `npm install bower -g` to be installed (which is probably already installed on most systems, but I mention it anyway).

colin-marshall:
Fixes #8749.

@brettsmason can you test this to make sure it works on Windows? Thanks!

### DIFF
--- a/gulp/test.js
+++ b/gulp/test.js
@@ -19,7 +19,8 @@ gulp.task('test', ['sass:foundation', 'test:transpile-js', 'watch'], function() 
       baseDir: 'test/visual',
       directory: true,
       routes: {
-        "/assets": "_build/assets"
+        "/assets": "_build/assets",
+        "/motion-ui": "node_modules/motion-ui"
       }
     }
   });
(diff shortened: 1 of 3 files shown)


brettsmason:
@colin-marshall You beat me to it! :smile: 
Works great thanks :+1: I just noticed there is a symlink to [Motion UI](https://github.com/zurb/foundation-sites/blob/visual-tests-windows/test/visual/motion-ui) too. Probably worth adding that too at the same time if you dont mind?

colin-marshall:
Thanks for reminding me! Motion UI symlink is removed.

designerno1:
Looks good to me

PR #8755, opened 2016-05-09
foundation.orbit.js [Added Options for Callback]

simongcc:
...

``` javascript
, {
      key: '_prepareForOrbit',
      value: function _prepareForOrbit() {
        var _this = this;
        this._setWrapperHeight(function (max) {
          _this._setSlideHeight(max);

          // added callback options, user can add callback function such as 
          // to remove a class for hiding before loading is completed to avoid FOUC to enhance UX
          _this.options.callback = _this.options.callback ? _this.options.callback : function () {};
          _this.options.callback();
          // this.options.slideClass
        });
      }
```

...


rafibomb:
@simongcc Can help me understand what's this PR helps with so I can work on the docs for it? Also, since it's a feature, we're targeting those to the v6.3 branch to follow programming standards. Thanks!

simongcc:
@rafibomb Thanks for adding the features. Let me explain my first reason for this feature so that you will have a better understanding. The current design of Orbit, is loading a photos after certain time when loading. During loading process, all photos is shown to the screen, which will provide bad visual effect. To cope with this, I add a class `"visually-hidden"` to `Orbit-Container` with `visibility: none` set to the class. Then, I ask the callback to remove this class to reveal the orbit. This improve the FOUC problem.
With this improvement, I also added `height: auto !important;` to `orbit-container` to get best result.
The `important` flag is to override the inline calculation of the plugin to `orbit` element.
This is one of the usage.

Another usage is, you may interactively work with other elements, such as trigger other animation when loading is completed.

At this moment, my consideration is mainly for the visual aids for the project I am working with, I think there are many possible usages. I have used in 2 project so far.
Many thanks.

kball:
Not sure what happened but this PR doesn't appear to be containing the code mentioned in the comments above anymore... going to close, @simongcc please resubmit.  Thanks.

simongcc:
@kball Ok, I will resubmit and added by branch for merge, thanks.

PR #8762, opened 2016-05-10
Fix for bug introduced with #8511

DaSchTour:
This was introduced with changes in #8511 
Here is the bad behaviour this causes http://codepen.io/DaSch/pen/YqgPBY

### DIFF
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -99,7 +99,7 @@
   // fixes recent Chrome version not limiting child width
   // https://stackoverflow.com/questions/34934586/white-space-nowrap-and-flexbox-did-not-work-in-chrome
   @if $columns == null {
-    min-width: 0;
+    min-width: intial;
   }
   // max-width fixes IE 10/11 not respecting the flex-basis property
   @if $columns != null and $columns != shrink {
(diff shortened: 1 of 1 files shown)


scss/grid/_flex-grid.scss
```diff
@@ -99,7 +99,7 @@
   // fixes recent Chrome version not limiting child width
   // https://stackoverflow.com/questions/34934586/white-space-nowrap-and-flexbox-did-not-work-in-chrome
   @if $columns == null {
-    min-width: 0;
+    min-width: intial;
```
kball: looks like a typo - should this be initial?

kball:
@ncoden based on your comments on @DaSchTour 's feedback on the other PR, I assume this seems like an appropriate fix to you?  So when the typo here is fixed we'll merge this... would either of you like to submit some visual tests around flexbox to help prevent this type of regression in the future?

ncoden:
@kball It isn't a fix, it's just a revert. `initial` value is equel to removing the property.
I'll search for a fix that resolve my problem, without side effects.

DaSchTour:
I hope we can find a fix very soon, so we can have this fixed in 6.2.2

DaSchTour:
Is it this bug? https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored

kball:
Merging this to include the reversion in the 6.2.2 rc, but it would be great if we get a fix for the original issue as well.

DaSchTour:
@ncoden does min-width: min-content; maybe fix the problem? Autoprefixer adds -moz-min-content and -webkit-min-content

PR #8763, opened 2016-05-10
add scss var for input font weight

### LINKED ISSUE #8283: $form-label-font-weight also controls weight of <input> contents :(
I just tried bolding my labels and came across this little treat.  Reset it back to normal and inputs changed as well.  Simple enough to override, but somthing that shouldn't happen IMO.

andycochran:
Closes #8283. 

This PR sets an explicit font weight for inputs to prevent cascade if an `input` is inside a `label` and `$form-label-font-weight` is bold.


rafibomb:
This is a good variable to add because changing in CSS is not straight forward.

PR #8764, opened 2016-05-10
Add ifequal helper documentation to readme.md

NeonArray:
The helper function is present in the source code, but not documented. This commit adds the function and a single use case to the helpers section of the readme.


rafibomb:
This is a great start and this needs to be documented! I'll merge this in, can you add some specific examples like layout or page variables?

```
{{#ifequal layout 'default'}}
  stuff
{{/ifequal}}
```

PR #8767, opened 2016-05-10
Added and optional background parameter to the reveal-overlay() mixin

grgcnnr:
As per the documentation, `@mixin reveal-overlay` should accept an optional `background-color` value to override `$reveal-overlay-background`


rafibomb:
Is the purpose of this to be able to use different overlay colors in different places?

grgcnnr:
@rafibomb yep, edge case I know. But also just so it matches the documentation.

andycochran:
👍 This looks good to merge.

Issue #8039, opened 2016-01-29
F6 Abide hidden input fields preventing proper error message display


wgstjf:
Hi guys,

It's taken me a while to figure out what is going on here....the problem is that any form I have with hidden input fields is preventing the .form-error fields from showing when the form is submitted and invalid.  I recognise that I can use the data-abide-ignore feature but as I am working on a Wordpress template I cannot manipulate the output of the nonce or the _wp_http_referer fields to add the attribute.  

My first thought was to detect these via jQuery and to retrospectively add the data-abide-ignore attribute to these fields but it hasn't made any difference to the outcome.  Is it that Abide needs to be re-initialised?  If so how does one do this?  If not...please help :-)

Cheers,

Will

zurbchris:
`$('[data-abide]').foundation('_reflow');`

wgstjf:
Outstanding!  Thanks for the quick response. :-)

scudco:
I'm experiencing this exact issue when Rails adds hidden inputs. I know the workaround works, but I think this issue ought to be re-evaluated. 

The problem ultimately occurs in `Foundation.Abide.prototype.findFormError` because it simply traverses up to the hidden field's parent to find _all_ its `.form-error` elements. This parent element may contain any number of invalid fields with visible form-errors that will then be hidden because the hidden field is valid. Because this is the case, I think a much cleaner hack is to simply override this method to ignore hidden fields:

```js
Foundation.Abide.prototype.findFormError = function($el){
  if ( $el.is('[type=hidden]') ) { return; }

  var $error = $el.siblings(this.options.formErrorSelector);
  if(!$error.length){
    $error = $el.parent().find(this.options.formErrorSelector);
  }
  return $error;
};
```



A few questions

* Why would a hidden input be validated (by default) at all? 
* Would it be reasonable to add a `data` attribute-level option for ignoring fields by jQuery selector or simply one for ignoring hidden fields?
* Why are inputs not evaluated for `[data-abide-ignore]` at the time of validation so that a reflow wouldn't be necessary?
* Could/should `form-error`s be tightly bound to inputs to prevent this kind cross-contamination of valid/invalid states?

sdhull:
> this issue ought to be re-evaluated

This issue ought to be re-opened. If hidden inputs break abide then I'd call that a major, show-stopping bug.

[...]

PR #8770, opened 2016-05-11
[bugfix] Abide: dont validate hidden fields

sdhull:
This addresses #8039 where hidden fields can cause error messages to be hidden, making it unclear why a form was not submitted.

It also moves the skipping of `[data-abide-ignore]` fields to validation-time such that if that property is added or changed with javascript, it will still be respected by Abide.

I welcome feedback and suggestions. Sorry I don't have time to really backfill tests for Abide.

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -29,7 +29,7 @@ class Abide {
    * @private
    */
   _init() {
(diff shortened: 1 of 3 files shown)


kball:
Great, thanks @sdhull.  Nice work.

PR #8773, opened 2016-05-11
Add Browserify directive to transform single source files with Babelify

RemiCollin:
When requiring single Foundation js components and compiling them with browserify, source files aren't transform with the Babelify transformer (see : https://github.com/babel/babelify#why-arent-files-in-node_modules-being-transformed). 

Adding this directive to package.json fix this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,5 +111,8 @@
         ]
       }
     }
+  },
+  "browserify": {
+    "transform": [["babelify", { "presets": ["es2015"] }]]
   }
 }


rafibomb:
@colin-marshall Are you able to look at this one for merge?

kball:
This makes sense to me.

timwis:
Hey guys, FYI this fails when I try to build it because `Cannot find module 'bablieyf' from /node_modules/foundation-sites/`. Would it make sense to add `babelify` to foundation-sites' `devDependencies`?

Issue #8096, opened 2016-02-05
[Forms] Number Input Spinner Bug


foodgy:
These [sass lines](https://github.com/zurb/foundation-sites/blob/c66d458b24d72f98a830f517878c0909df46016f/scss/forms/_text.scss#L142) should be fixed:

```
  // Number input styles
  [type='number'] {
    @if not $input-number-spinners {
      -moz-appearance: textfield;

      &::-webkit-inner-spin-button,
      $::-webkit-outer-spin-button {
        -webkit-appearance: none;
        margin: 0;
      }
    }
  }
```

Removed nesting ```[type='number']```.

colin-marshall:
@MandiSaeteun I'm going through old issues and noticed you were assigned this one. Was there a reason for that or should we pass it off to somebody else? Thanks!

rafibomb:
Closing for PR - thanks!

PR #8776, opened 2016-05-12
Fix bad nesting that breaks input-number-spinners styles

pdeffendol:
The generated styles for input number spinners are wrong - there's an addition level of nesting.  This was mentioned in #8096.


rafibomb:
Thanks!

Issue #8427, opened 2016-03-18
Sticky plugin makes div larger


iammichiel:
Hi, 

I am having a strange issue when using the sticky plugin. 
I simply create two divs `large-9` and `large-3`. The `large-3` is a sticky div for `large-9`.

``` html
<div class="row">
    <div id="leftbar" class="large-9 columns">
    </div>

    <div id="rightbar" class="columns large-3" data-sticky-container>
        <div class="sticky" data-sticky data-anchor="leftbar">

        </div>
    </div>
</div>
```

Without the sticky part, the `#rightbar` is 369 pixels wide. When adding the sticky class, it gets 12 pixels larger.

gehasia:
Please provide **full** html+css code to reproduce the bug.
You tell rightbar is 369px wide, but the width will depend of your browser windows width, do you mean you fixed the column width ???

andycochran:
If you can, please reproduce this in a [CodePen](http://codepen.io/rafibomb/pen/ZQEGog).

iammichiel:
Yep, sure. Here is a link to a [Codepen](http://codepen.io/iammichiel/pen/WwmqmZ) illustrating the problem.

I can now go a little further in my explanation, when adding the .sticky behaviour, the right gutter is no longer displayed.

andycochran:
@iammichiel, thanks for clarifying. 

I actually just noticed this bug in the docs today. The sticky space images spill into the right gutter. This isn't visually evident because there's some bad markup. There are columns nested in a column without a row: 

```html
<!-- ```html_example -->
<div class="row">
  <div class="columns small-12">
    <div class="columns small-6" id="example1" data-something>...</div>
    <div class="columns small-6 right" data-sticky-container>...
```
The problem is very evident if you remove that superfluous `<div class="columns small-12">`.

andycochran:
@gehasia, you're good with .js and familiar with Sticky, right? Would you like to take a stab at fixing this? I'd love to see a PR.

gehasia:
Sure, i'll try to have a look on it next week probably (as i'll be busy for few days) !

Jflem1290:
@gehasia Hey! I was wondering if you had time to take a look at this issue. I would love to collaborate and get this issue resolved for the community.

gehasia:
@Jflem1290 
Hi ! Unfortunately i don't have any time right now, if you want to look at it, feel free to do it !

Jflem1290:
@gehasia Sound great!

[...]

PR #8777, opened 2016-05-12
[Sticky] fix bad grid markup in docs

andycochran:
There was a superfluous `.column` in the markup. It was breaking the grid and causing a Sitcky bug not to be seen. This PR makes #8427 evident.


colin-marshall:
@andycochran there's another bad grid in the Stick To Bottom section, can you please take care of that one too while we're at it? Thanks!

andycochran:
@colin-marshall Got it. There was a `<strongr />` in there too. 😲

Issue #8409, opened 2016-03-15
[6.2.0] top-bar-right doesn't work with foundation-flex


hylke94:
The alignment of the top-bar-right class doesn't work correctly when including the [foundation-flex.min.css](https://cdnjs.cloudflare.com/ajax/libs/foundation/6.2.0/foundation-flex.min.css) file.

**How can we reproduce this bug?**
1. Create a div with class "top-bar".
2. Inside top-bar create a div with class "top-bar-title" (or "top-bar-left").
3. Inside top-bar create a div with class "top-bar-right".

**What did you expect to happen?**
The div with class top-bar-right aligns to the right.

**What happened instead?**
The div with class top-bar-right align to the center of the top-bar.

**Test case:**
[JSFiddle](https://jsfiddle.net/1wxvjq4o/)

hylke94:
The same issue is on a "row align-justify" div. The second div doesn't align to the right, but somewhere in the middle ([JSFiddle](https://jsfiddle.net/r9su0wh1/)).
"align-spaced" also doesn't work properly.

lucianosemerini:
It happens to me too, any workaround?

sculler:
You can fix like this:

`.top-bar .top-bar-left, .top-bar .top-bar-right {
    flex: none;
}`

hylke94:
Is this issue fixed in the 6.2.1 release of yesterday?

hylke94:
Nope. The top-bar-right is now full-width (beside the top-bar-left/title) and aligns to the left. And the "row align-justify" acts like "row align-spaced".

DaSchTour:
Time to release 6.2.2 :stuck_out_tongue_winking_eye:

hylke94:
As shown in this [JSFiddle](https://jsfiddle.net/1wxvjq4o/2/), above 'fix' doesn't fix the issue completely. The alignment is somewhat better, but now they also align towards the middle...

The same for the second issue, alignment in a ['align-justify' row](https://jsfiddle.net/r9su0wh1/1/)...

cabelodomato:
Check out the solution given by sculler.

It worked for me!

hylke94:
Not in my JSFiddle tests... 

How is your code, @cabelodomato ?

Lesha-spr:
@hylke94 ,
Seems :before/:after pseudo-elements are the breakers. You should remove non-flex Foundation from Resources.

hylke94:
@Lesha-spr 
[...]

PR #8778, opened 2016-05-12
Fix for #8409

natewiebe13:
Fixes topbar in flex mode.


kball:
This fix doesn't work in the case where you have all three of top-bar-title, top-bar-left, and top-bar-right.

I created a visual test case and submitted a PR with a more complete fix here: https://github.com/zurb/foundation-sites/pull/9177

PR #8779, opened 2016-05-12
Partially fixes #8325

### LINKED ISSUE #8325: Responsive top bar has issues with $global-flexbox: true

Note: Using Foundation 6.2
1. Use flex grid in sass as [described here](http://foundation.zurb.com/sites/docs/flex-grid.html#importing)
2. `$global-flexbox: true;`
3. Create a top bar like this:

```
<div class="top-bar" id="main-menu">
   <div class="top-bar-left">
      <ul class="menu vertical medium-horizontal expanded" data-responsive-menu="drilldown medium-dropdown">
        <li>
          <a href="#">Item 1</a>
          <ul class="menu">
            <li><a href="#">Item 1B</a></li>
            <li><a href="#">Item 1C</a></li>
          </ul>
        </li>
        <li>
          <a href="#">Item 2</a>
          <ul class="menu">
            <li><a href="#">Item 2A</a></li>
            <li><a href="#">Item 2B</a></li>
            <li><a href="#">Item 2C</a></li>
          </ul>
        </li>
      </ul>
   </div>
</div>
```

- Menus are expanded across width of screen
- Menus go vertical when the screen size is shrunk

**What happened instead?**
- Menus are aligned to the left instead of expanded
- When the screen is made smaller, the menus do not go vertical, they stay horizontal, although it looks like the drilldown functionality is still mostly working.

Changing `$global-flexbox: false;` fixes these 2 issues.

Thanks.

natewiebe13:
`.expanded` was overwriting the vertical styles. Not sure if the other part of #8325 can be fixed without deeper changes.

### DIFF
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -228,6 +228,9 @@ $menu-icon-spacing: 0.25rem !default;
     @include menu-base;
     @include menu-icons;
 
+    // Orientation
+    @include menu-direction(horizontal);
+
     // Even-width
     &.expanded {
       @include menu-expand;
@@ -236,9 +239,6 @@ $menu-icon-spacing: 0.25rem !default;
         width: 100%;
       }
     }
-
-    // Orientation
-    @include menu-direction(horizontal);
 
     &.vertical {
       @include menu-direction(vertical);
@@ -249,6 +249,15 @@ $menu-icon-spacing: 0.25rem !default;
         @include breakpoint($size) {
           &.#{$size}-horizontal {
             @include menu-direction(horizontal);
+
+            // Even-width
+            &.expanded {
+              @include menu-expand;
+
+              > li:first-child:last-child {
+                width: 100%;
+              }
+            }
           }
 
           &.#{$size}-vertical {
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -83,6 +83,10 @@ $menu-icon-spacing: 0.25rem !default;
     display: table;
     table-layout: fixed;
   }
+
+  > li:first-child:last-child {
+    width: 100%;
+  }
 }
 
 /// Sets the direction of a Menu.
@@ -234,10 +238,6 @@ $menu-icon-spacing: 0.25rem !default;
     // Even-width
     &.expanded {
       @include menu-expand;
-
-      > li:first-child:last-child {
-        width: 100%;
-      }
     }
 
     &.vertical {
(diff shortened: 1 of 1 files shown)


kball:
@andycochran @brettsmason can you guys take a look at this and see if it seems good to go?

scss/components/_menu.scss
```diff
@@ -231,6 +231,15 @@ $menu-icon-spacing: 0.25rem !default;
     // Orientation
     @include menu-direction(horizontal);
 
+    // Even-width
+    &.expanded {
+      @include menu-expand;
+
+      > li:first-child:last-child {
+        width: 100%;
+      }
```
ncoden: Why isn't it in `@mixin menu-expand` instead of being duplicated ?

ncoden (commented):
Hi @natewiebe13, thank you for your PR !

According to me (i'm not familiar with flexbox), `menu-expand` should provide different flex values on `> li` if the menu is vertical or horizontal. Maybe it's the deeper changes you're talking about.

ncoden:
Also, PR commit names must be in *imperative* :wink:

scss/components/_menu.scss
natewiebe13: Moved to the mixin. The idea was to keep the code the same, just move it around and fix the specificity issues

natewiebe13:
@ncoden, this only fixes the expanded class from being overridden by fixing the specificity issues, so adding different flex values to list items were the deeper changes I was referring to. Also, since we have to support flex and non-flex styles, it may be more work than expected. I just haven't had time to look at it.

kball:
@natewiebe13 do you by chance have time to look at this again for the updates mentioned above?

natewiebe13:
@kball I haven't had a chance yet. Not sure when I'll be able to tackle this.

kball:
@natewiebe13 and @ncoden given we may not have time to address those deeper changes, what is your sense of this PR as it stands to make incremental progress?

ncoden:
This fix duplicate `@mixin menu-expand` across all breakpoints, so with a high specificity. I don't think it would be a good idea to merge it as it stands. I'll look at it later to find a proper solution, if @natewiebe13 doesn't have the time to.

natewiebe13:
@ncoden: Go ahead, I won't have time for the next few weeks.

kball:
Going to push this out to 6.3.1

ncoden:
@natewiebe13 I created https://github.com/zurb/foundation-sites/pull/9442. You was right. When we work with media queries, we have to duplicate everything 😭 .

kball:
Closing in favor of #9442

PR #8783, opened 2016-05-13
v6.2.2 fixes abide radio check if only one radio is required

designerno1:
fixes bug when only one radio has required and adds a visual test for that

according to docs from jQuery http://api.jquery.com/attr/
Get the value of an attribute for the **first element in the set of matched elements** or set one or more attributes for every matched element.

``` js
// .attr() returns undefined if no elements in $group have the attribute "required"
if ($group.attr('required') === undefined) {
  valid = true;
}
```


kball:
👍

PR #8785, opened 2016-05-13
[JS-dropdown] no-reggression unit test for getPosition()

iamdey:
Hi, I've spotted a bug on dropdown that has already been fixed.

Here is the unit test. Feel free to merge or not.

Cheers


rafibomb:
This is great! One focus of v6.3 is unit testing. 

We just need to solidify where these tests go. @kball Have any insight on this?

kball:
This is right - unit tests go in in test/javascript/components/

PR #8786, opened 2016-05-13
Clarified language: "instead" to "inside"

wesf90:
Changed to better describe the appropriate action that needs to be taken by the user. You don't place links "instead" of list items, but rather, "inside" of them.


colin-marshall:
👍

Issue #8792, opened 2016-05-16
[Compatibility] Webkit prefixes missing


slavanga:
The [docs](http://foundation.zurb.com/sites/docs/compatibility.html) state that foundation is compatible with Android Browser 2.3+ but prefixes for properties like ```border-box``` are missing.

DaSchTour:
I created a PR #8794
Android was added in a wrong way to the autoprefixer

Geczy:
Autoprefixer?

rafibomb:
autoprefixer is there but it could be solved by @DaSchTour 's PR. Thanks for the PR @DaSchTour !

PR #8794, opened 2016-05-16
Fix for #8792

DaSchTour:
I guess this is a misunderstanding about android browser versions
To target Android Webview (Android 2.3) another syntax is needed than for Android Chrome which it was until now.

### DIFF
--- a/gulp/sass.js
+++ b/gulp/sass.js
@@ -24,7 +24,7 @@ var LINT_PATHS = [
 var COMPATIBILITY = [
   'last 2 versions',
   'ie >= 9',
-  'and_chr >= 2.3'
+  'Android >= 2.3'
 ];
 
 // Compiles Sass files into CSS


slavanga:
This change would add prefixes for ```border-box``` but it would also output the old flexbox syntax (which does not support wrapping):
```display: -webkit-box;``` instead of ```display: -webkit-flex;```

I wonder were the correct webkit prefixes for flexbox in [foundation-flex.css](https://github.com/zurb/foundation-sites/blob/develop/dist/foundation-flex.css) come from...

DaSchTour:
That's right. But Android 2.3 only supports old flexbox syntax and normaly id should add both in a way that the newest supported syntax will be used.

slavanga:
This config doesn't output all required prefixes. Android had the ```-webkit-box``` syntax and then jumped to the prefixless variant in version 4.4.

We should add an old Safari version (6 or older) to the Autoprefixer config to get ```-webkit-flex```.

PR #8805, opened 2016-05-18
Clean up slider logic and fix bug with top end

kball:
Fixes issue discovered in QA with top end of slider.  Also cleans up slider logic to be more readable.

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -275,11 +275,14 @@ class Slider {
           eventOffset = vertical ? e.pageY : e.pageX,
           halfOfHandle = this.$handle[0].getBoundingClientRect()[param] / 2,
           barDim = this.$element[0].getBoundingClientRect()[param],
-          // touch events emulated by the touch util give position relative to screen, add window.scroll to event coordinates...
           windowScroll = vertical ? $(window).scrollTop() : $(window).scrollLeft();
 
 
-      var eventFromBar = eventOffset - this.$element.offset()[direction];
+      var elemOffset = this.$element.offset()[direction];
+
+      // touch events emulated by the touch util give position relative to screen, add window.scroll to event coordinates...
+      if (eventOffset < elemOffset) { eventOffset = eventOffset + windowScroll; }
+      var eventFromBar = eventOffset - elemOffset;
       var barXY;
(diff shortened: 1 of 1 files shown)


brettsmason:
@kball I can confirm this fixes the issue in the QA I reported :+1:

designerno1:
@kball 
this makes it better for drag but unusable for touch android

kball:
@designerno1 nice catch, I'll dig into it

kball:
@designerno1 updated and I believe it resolves on android.  Touch slider on android feels a little clunky to me but going back to the initial RC it feels clunky there too, and in 6.2.1 it was broken entirely, so I think this is good enough.  Can you try it out again?

designerno1:
@kball 
now it works on android with touch
but jumps back on drag from top to bottom
keyboard go to top with up arrow works also fine

kball:
Ah the jump makes sense... ok still tinkering.  The discrepancy between the real events and faked event stuff is a pain.

kball:
@designerno1 Got another pass on this.  This feels kinda clunky, but after mucking around a bit I couldn't figure out how to do the better solution (fixing the touch util faked events), and this seems to be working now both on my android test device and the non-android browsers I've tested

designerno1:
now works by drag
but still buggy on android swipe and touch move

kball:
@designerno1 Was it working in those cases before?  If not, I'd like to merge this and open those as a separate issue to tackle

designerno1:
@kball 
yeah in the v6.2.2 branch touch on vertical slider works great

kball:
ok I'll keep pounding on it then

kball:
@designerno1 one more fix in for an issue on android 6, but I'm not sure exactly what behavior you're seeing still on this branch that isn't there on v6.2.2.  Can you explain more what you mean by swipe as distinct from drag for the vertical slider?  Thanks

designerno1:
@kball 
with drag i mean click and hold mouse and move
with swipe i mean touch hold and pull up
with touch i mean tap on slider to move

designerno1:
@kball 
now it looks goog on chrome,edge,firefox, android 6
awesome

PR #8807, opened 2016-05-18
Fix background scrolling for Android and improve on iOS

kball:
Apply the reveal iOS background scrolling fix to Android as well to prevent background scrolling, and improve the recovery of position on mobile reveal close.  Fixes issue turned up by QA.


designerno1:
don't know how but it works

PR #8808, opened 2016-05-18
Fix issue with explicit close logic

kball:
Fixes issue turned up in QA where explicit close link was not closing reveal.


designerno1:
looks good to me and works

PR #8809, opened 2016-05-18
Add missing styles for collision changed dropdown menu in RTL

kball:
Fixes issue noticed in QA with RTL dropdown menu on small screens.


designerno1:
looks better to me, but if screen is too small dropdown goes out left window just the same

kball:
@designerno1 updated to fix that tiny case.  @abdullahsalem can you take a look at this one too?

abdullahsalem:
Great! the issue has been fixed!
Thanks a lot @kball !

PR #8829, opened 2016-05-22
Clarify documentation for off-canvas menu

whatcould:
I ran into two confusing issues transitioning my off-canvas menu to Foundation 6.

The first is just that the `data-position` attribute was used in some examples but not documented (first commit) — see http://foundation.zurb.com/forum/posts/39294-right-off-canvas-not-working for an example of someone else who ran into the same doc issue.

The second issue is that it was confusing that the _page content_ is contained in a class of `.off-canvas-content`. The first time I was adding my classes, I thought the content I wanted to display off-canvas would go in `.off-canvas-content`. Right? It makes sense if you start thinking of `.off-canvas` as a namespacing prefix, so you have [prefix]-content and [prefix]-wrap. Tried to clarify that for future users in the second commit, but you may have suggestions for better copy?

Thanks for all your work on Foundation 6!

### DIFF
--- a/docs/pages/off-canvas.md
+++ b/docs/pages/off-canvas.md
@@ -26,7 +26,7 @@ To start, create two wrappers to house the page. These are necessary to prevent 
 </body>
 ```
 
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks!

PR #8831, opened 2016-05-23
Added documentation for orbit image mixin.

abarnwell:
Small addition to orbit rotator documentation. Added the `orbit-image` mixin to the docs as it wasn't included but fairly important if you're using images in your rotator.


rafibomb:
Thanks for this!

PR #8838, opened 2016-05-24
Remove CSS-class dot from html

Ichag:
Remove CSS-class dot from html


Ichag:
Also requested by #8830

PR #8842, opened 2016-05-24
Update sass.md

paxperscientiam:

Happy coding! :)

Suggestion to use DiffMerge for visualizing file differences. Instrumental for adapting _settings.scss changes.

### DIFF
--- a/docs/pages/sass.md
+++ b/docs/pages/sass.md
@@ -78,7 +78,7 @@ You're also going to want a settings file for your project, which will allow you
 @import 'foundation';
 ```
 
-Note: To relieve the tedium of comparing a newly released _settings.scss file with your project's current version, there is [DiffMerge](http://twobitlabs.com/2011/08/install-diffmerge-git-mac-os-x/). This handy application can either be integrated into yourgit workflow or used separately. Its GUI makes visualizing the line-by-line differences between two files a snap.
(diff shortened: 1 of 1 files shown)


andycochran:
I'm not sure we should recommend specific third-party tools for Git workflows. 
@zurb/yetinauts, thoughts?

rafibomb:
So I like this kind of information, and I think there is a better place than the documentation for it. Perhaps we start a page in the Wiki for workflow optimizations or a page in the docs specifically for this. 

@paxperscientiam Would love to hear some ideas where it would allow workflow optimizations to be shared in the community.

rafibomb:
We also have a [foundation code syleguide](https://github.com/zurb/foundation-code-standards) which could be linked to from a page. Maybe it's a hub from Foundation to other resources?

colin-marshall:
I think if the specific tool recommendation was changed to the more broad recommendation of simply using a "diff" tool this would be good to include in the docs.

paxperscientiam:
@rafibomb I'm obsessed with efficiency -- to a fault. I would love to share my thoughts on workflow. Is the consensus that any related write-ups should submitted via PR to zurb/foundation-code-standards?

@colin-marshall I second that.

kball:
Given the discussion above, going to close this.

PR #8844, opened 2016-05-24
Change no scroll default - Foundation v5

joeworkman:
I like that this feature was added. However, the default value should be the same as it was prior to this change. You don't want to break people's sites by simply updating the framework.


rafibomb:
Thanks Joe - agreed!

rafibomb:
I think the JS need updating here: http://foundation.zurb.com/sites/docs/v/5.5.3/components/reveal.html  Can you help on that?

PR #8850, opened 2016-05-25
Account for hOffset

flyingL123:
Currently if you set `data-h-offset` on a dropdown menu, the hoffset value is not used.

### DIFF
--- a/js/foundation.util.box.js
+++ b/js/foundation.util.box.js
@@ -186,7 +186,7 @@ function GetOffsets(element, anchor, position, vOffset, hOffset, isOverflow) {
       break;
     default:
       return {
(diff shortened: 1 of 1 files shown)


Owlbertz:
Hi @flyingL123,
The Dropdown Menu component does not have a `data-h-offset` parameter. Maybe you are talking about the Dropdown component? However, unfortunately your PR does only fix this for LTR oriented sites. Can you adjust this so it will also work for RTL sites?

kball:
I'm going to go ahead and merge this for now - it looks like there are a number of positioning situations that don't take offsets properly into account right now, but we're planning on tackling positioning as a part of 6.2.5.

PR #8852, opened 2016-05-25
Fixed id typos

StephenOrJames:
`aria-describedby` contains `emailLabel1` and `emailLabel2`, but `emailLabel` and `emailLabe2` were used as the `id`s of the `span` elements.


rafibomb:
Thanks!

PR #8853, opened 2016-05-26
Fix: undo margins for flex-grid nested rows with collapse

saadshahd:
Hi, this undoes the margin declaration for `.row.collapse > .column > .row`

Adds:

``` CSS
.row.collapse > .column > .row {
  margin-left: 0;
  margin-right: 0;
}
```

### DIFF
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -165,7 +165,13 @@
     }
 
     &.collapse {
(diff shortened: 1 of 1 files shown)


saadshahd:
Please check the descriptive Codepen [here](http://codepen.io/anon/pen/ezYEeg).

rafibomb:
Thanks for the demo! Are you able to add this to the visual tests so we can QA it for the next release?

PR #8871, opened 2016-05-31
Add $gutter parameter to the grid-layout mixin.

### LINKED ISSUE #8857: Block Grid Gutter
In Foundation 6 you use the 'grid-layout' mixin for a block grid, but what controls the gutter? The child elements have no padding left or right?

This is my code:

```
.block {
  &__list {
    @include grid-layout($n:4, $selector:li);
    list-style: none;
    margin-left: 0;
  }
}

<div class="block block--front">
   <ul class="block__list">
        <li><img src="http://placehold.it/640x480"></li>
       <li><img src="http://placehold.it/640x480"></li>
       <li><img src="http://placehold.it/640x480"></li>
       <li><img src="http://placehold.it/640x480"></li>
   </ul>
</div>
```

brettsmason:
Add a 3rd `$gutter` parameter to the `grid-layout` mixin. This allows you to use the mixin to create a block grid that works like the block grid from v5 (including a gutter). $grid-column-gutter can be passed as the option to enable responsive gutters.

This closes #8857

### DIFF
--- a/scss/grid/_layout.scss
+++ b/scss/grid/_layout.scss
@@ -10,13 +10,38 @@
 ///
 /// @param {Number} $n - Number of elements to display per row.
 /// @param {String} $selector ['.column'] - Selector(s) to use for child elements.
+/// @param {Number|List} $gutter
+///   The gutter to apply to child elements. Accepts multiple values:
+///   - $grid-column-gutter will use the values in the $grid-column-gutter map, including breakpoint sizes.
+///   - A fixed numeric value will apply this gutter to all breakpoints.
 @mixin grid-layout(
   $n,
-  $selector: '.column'
+  $selector: '.column',
+  $gutter: null
 ) {
   & > #{$selector} {
     width: percentage(1/$n);
     float: $global-left;
+
+    // If a $gutter value is passed
+    @if($gutter) {
+      // Gutters
+      @if type-of($gutter) == 'map' {
+        @each $breakpoint, $value in $gutter {
+          $padding: rem-calc($value) / 2;
+
+          @include breakpoint($breakpoint) {
+            padding-left: $padding;
+            padding-right: $padding;
+          }
+        }
+      }
+      @else if type-of($gutter) == 'number' and strip-unit($gutter) > 0 {
+        $padding: rem-calc($gutter) / 2;
+        padding-left: $padding;
+        padding-right: $padding;
+      }
+    }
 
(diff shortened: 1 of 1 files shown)


andycochran:
This is rad! Testing…

andycochran:
@brettsmason Should this third parameter be added to Block Grid's `@include` in `grid/_classes.scss`? 
```scss
@include grid-layout($i, ".#{$column}", null);
```
And is any documentation needed?

rafibomb:
@brettsmason Anything we can do to get this across the line?

brettsmason:
@andycochran @rafibomb I'm not sure about including it in the Block Grids include, as the aim in v6 for the block grid seems to be reliant on using a `.column` as the child element. Changing it would potentially cause confusion? Though I agree it probably should of worked like this to begin with to mimic v5 more closely.

Re the documentation - do you think the description in the code is enough or not? THis would then be displayed on the docs anyway wouldnt it?

Let me know and I'll make any needed changes.

kball:
@brettsmason I think this is good, the description in the code does show up automatically in the docs... I think it might be helpful to include an example, though. 

As someone coming in without having followed the whole thread, especially without having experience in F5, I'm a little confused. It seems like the block grid in the docs already has gutters coming from the columns class... I *think* I'm following the conversation to indicate the need for this is when defining your own block grid using the mixin, but an example would definitely help.

brettsmason:
@kball Any ideas what you would like to see as an example? I'll get it added to the PR.

The purpose of this like you say was if you are creating your own block grid, for example I could do this:

```
.gallery {
    @include grid-layout(3, '.gallery-item', $grid-column-gutter);
}
```

That would output the CSS for `.gallery-item` with padding values in the `$grid-column-gutter` map, including the breakpoint sizes. You can also pass a number value, and that will then be static padding (no seperate breakpoints). It cuts down on the amount of Sass you have to write.

Sorry I've been out of the loop for a while and its been months since I actually did the PR so details are a bit fuzzy!

andycochran:
@brettsmason Maybe add that `.gallery` example to a new example in `docs/assets/scss/` and on `docs/pages/grid.md`?

kball:
@brettsmason @andycochran I'm going to move forward with merging this as it is backwards compatible without any updates, but would still love some more examples of use.

brettsmason:
@kball sorry struggling for time at the moment. I'll see if I can put something together in the next few days.

PR #8878, opened 2016-06-01
Added double-quotes to href data targets to prevent jQuery errors

rickcurran:
Added double-quotes to href data targets on Foundation tabs to prevent errors with jQuery 1.12.3. WordPress 4.5.2 includes jQuery 1.12.3 by default so this was causing a JavaScript error when using deeplinking: true on Foundation tabs. 

Adding double-quotes around these three locations in foundation.tab.js fixes this issue.

### DIFF
--- a/js/foundation/foundation.tab.js
+++ b/js/foundation/foundation.tab.js
@@ -102,19 +102,19 @@
             var hash_element = S(hash);
             if (hash_element.hasClass('content') && hash_element.parent().hasClass('tabs-content')) {
               // Tab content div
-              self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href=\\' + hash + ']').parent());
+              self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href="' + hash + '"]').parent());
             } else {
               // Not the tab content div. If inside the tab content, find the
               // containing tab and toggle it as active.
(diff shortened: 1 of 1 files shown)


andycochran:
This would disallow the use of single quotes. 😞 

> By default, SGML requires that all attribute values be delimited using either double quotation marks (ASCII decimal 34) or single quotation marks (ASCII decimal 39). Single quote marks can be included within the attribute value when the value is delimited by double quote marks, and vice versa. Authors may also use numeric character references to represent double quotes (`&#34;`) and single quotes (`&#39;`). For double quotes authors can also use the character entity reference `&quot;`.
–[w3.org](https://www.w3.org/TR/html4/intro/sgmltut.html#h-3.2.2)

andycochran:
@datenimperator, any thoughts on this?

andycochran:
This would close #8956.

datenimperator:
Since IDs must not contain neither single nor double quotes, I think it's appropriate.

rickcurran:
Hi, thanks for taking a look at this pull request. Note that this is an issue for Foundation 6's codebase as well so this fix would need applied there too.

datenimperator:
May I ask why this was closed without merging? Also, 5.5.4 hasn't been released afaik. We still have sites built on Foundation 5 so an official fix would be greatly appreciated.

PR #8882, opened 2016-06-02
Error on Collapse/Uncollapse Rows example

OndrejVasicek:
The description doesn't correspond to the example code. To me, the description seems to be fine, so I fixed the code.


andycochran:
In the spirit of mobile first, the `small-collapse` class can just be `collapse`.

OndrejVasicek:
Well it could I guess, but that's rather a design decision. I was trying to show you the inconsistency between the docs and the example.

rafibomb:
@andycochran Maybe we should change the class name but `.small-collapse`  or `.collapse` works. I decided to leave it but you're right and we should be consistent.

PR #8884, opened 2016-06-02
[V5] throttle scroll handler for performance gain

dragthor:
While profiling our Zurb 5 implementation, we noticed a slight bottle neck using a "sticky" top bar.  The user would sometimes experience a "stutter" or "glitch" while scrolling.  Users with lower end devices would experience the issue more often.

The problem: `update_sticky_positioning` is called a lot.  Too much.

The solution: `throttle` the `scroll` handler when calling `update_sticky_positioning`

I introduced a new `setting` named `scroll_throttle` with a default value of 300 milliseconds.  Throughout our test cycles the value works well.

Profiler before:

before

Profiler after:

after

### DIFF
--- a/js/foundation/foundation.topbar.js
+++ b/js/foundation/foundation.topbar.js
@@ -16,7 +16,8 @@
       is_hover : true,
       scrolltop : true, // jump to top when sticky nav menu toggle is clicked
       sticky_on : 'all',
-      dropdown_autoclose: true
+      dropdown_autoclose: true,
+      scroll_throttle: 300
     },
 
     init : function (section, method, options) {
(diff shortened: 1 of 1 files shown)


dragthor:
This change has been working well for us the past 9 months.  Hope to see it merged soon.

DanielRuf:
Not sure if we should use `300`. What about using some request animation frame based throttling (at most once per animation frame on a browser or per 1000/60 ms on Node)?

DanielRuf:
Or is this still too much?

JeremyEnglert:
Unfortunately support for V5 has been dropped.

PR #8885, opened 2016-06-02
[V5] IE focus dropdown jumping issue

dragthor:
IE likes to "jump" the user into position when `.focus()` is called.  We don't like this.  Notice in the video when the dropdown is at the bottom of the screen, the dropdown "jumps" up -

https://www.youtube.com/watch?v=OmuSAKOMjcQ

This seems to only be an issue with IE.  Here is a video after the fix (and behaving similar to Firefox and Chrome).  Simply set your `data-options="no_focus:true"` and then no more jumping.

https://www.youtube.com/watch?v=D5BbHLIG4cE

### DIFF
--- a/js/foundation/foundation.dropdown.js
+++ b/js/foundation/foundation.dropdown.js
@@ -14,6 +14,7 @@
       is_hover : false,
       hover_timeout : 150,
       no_pip : false,
+      no_focus : false,
       opened : function () {},
       closed : function () {}
     },
@@ -165,7 +166,13 @@
       dropdown.data('target', target.get(0)).trigger('opened.fndtn.dropdown', [dropdown, target]);
       dropdown.attr('aria-hidden', 'false');
       target.attr('aria-expanded', 'true');
-      dropdown.focus();
+      
+      var settings = target.data(this.attr_name(true) + '-init') || this.settings;
+      
+      if (settings.no_focus === false) {
+        dropdown.focus();
+      }
+      
       dropdown.addClass('f-open-' + this.attr_name(true));
     },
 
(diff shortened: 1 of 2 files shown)


andycochran:
This is not just IE. You can see this in the [first example in the docs](http://foundation.zurb.com/sites/docs/dropdown.html) in any browser. But I'm not sure this is a bug. In your example, `data-auto-focus` is doing what it should be doing, bringing the focused item into view.

dragthor:
Zurb 5, not 6 - http://foundation.zurb.com/sites/docs/v/5.5.3/components/dropdown.html

Swagetha:
I have this same issue in my app. Just in IE, not the other browsers.

If this is not a bug, then how should do you prevent the window from jumping when a dropdown is clicked?

andycochran:
@dragthor @Swagetha Can y'all try removing the `tabindex="..."` on your `data-dropdown-content`? 

(In the docs, is seems that in the examples with `tabindex`, the content causes the window to scroll. But in the examples without, the content flows out of the viewport.)

Swagetha:
@andersonaguiar , I have removed that attribute from the element because it also caused issues in other browsers, but it does not fix the issue in internet explorer.

rafibomb:
@dragthor Is this solution still working for you? I'd like to merge it in if it's a definite improvement for F5.

dragthor:
Rafi,

We've been live with it for over a month now.  No more IE jumping.

Note: its developer "opt in" meaning there is no change to default behavior.

dragthor:
Hope to see this merged soon.

DanielRuf:
Is `no_focus` set / updated somewhere? Can not find a direct reference.

JeremyEnglert:
Unfortunately support for V5 has been dropped.

PR #8888, opened 2016-06-03
Error in documentation

mhindery:
There was a slight error in the docs regarding the native progress bar usage with sass. Simply replaced @import by @include

Native element sass usage: @import replaced by @include, which it should be


andycochran:
👍 , but documentation fixes should target `master`.

Issue #8304, opened 2016-03-01
flex grid nesting on fluid layout bug


takl95:
I just made a new  foundation 6.2 project and enabled the flex-grid
I also changed the width to 100%

With these settings if I created a .row inside a .column div it wont stretch out to the full width. (Nesting)
The moment I changed $global-width to rem-calc(1200) everything works again.

gakimball:
Could you be more specific about what you did to cause the bug?

When you say "changed the width", did you change a Sass variable, or modify the width in CSS?

And could you post the HTML used to create the grid? Thanks :)

mmjaeger:
There's certainly something going on when nesting  - I added flex: 1 0 100% to the nested row and that seems to work?!

tinyfly:
I'm running into this as well. I want to have a 100% fluid-width container.

If you set the Sass Variable `$global-width` to "100% then have a nested flex grid like this:

```html
<div class="row">
  <div class="column">
    <h1>Some content</h1>

    <div class="row">
      <div class="column">...content...</div>
      <div class="column">...content...</div>
    </div>
  </div>
</div>
```

![nested-column-bug](https://cloud.githubusercontent.com/assets/91300/13988152/abcd54cc-f0d8-11e5-9e59-b79b0b3ebb5c.png)

The nested row will not be 100% of the parent column. It'll be too narrow and it'll have some sort of space along the right margin. See the blue bar in the screenshot. It should stretch to the edge of the red parent.

As the original poster mentioned if you set `$global-width` back to a fixed value like `rem-calc(1200)` then the nested row stretches edge-to-edge of its parent column as expected.

tinyfly:
Some more experimenting looks like the Sass variable `$global-width` directly sets the max-width on the `.row` which is true for every row so if I set `$global-width: none` I get a fluid width container and and nested rows are 100% as expected.

So my solution is to add a css class in my site styles of:

```css
.row .row {
  max-width: none;
}
```

This way you can use a percentage value for `$global-width` that doesn't mess with the row width of nested rows.

gakimball:
[...]

PR #8889, opened 2016-06-03
Contain nested rows to grid

### LINKED ISSUE #8886: Expanded row children no longer contained to grid in Foundation 6.2.1
#8304 introduces a pretty significant break to rows nested under a parent row with the expanded class; previously, they were contained to the grid - with the option to add the expanded class to nested rows as needed - but now all nested rows under an expanded row are also expanded and there is no fix once the left and right margin's are set to zero.

See [codepen example](http://codepen.io/nicwinn/pen/VjLZvE) - (change the foundation css version from 6.2.1 to 6.2.0)

I propose rolling back this commit until another solution is found for the original issue.

edit: update with correct issue number

gpspake:
This commit removes a single line that was recently added that results in the following style:

``` css
.row .row {
  max-width: none
}
```

The line was added in response to #8304. However, the appropriate solution would have been to add the expanded class to nested rows. Setting max-width to none introduces a break. If a nested row needs to be full width, the `expanded` class can be applied to that row; however, if nested row max-width is set to none, there's no way to contain a row to the grid inside of an expanded row.

If a row is nested inside of an expanded row, it should be contained to the grid unless the expanded class is explicitly added to the nested row. 

This fixes #8886

### DIFF
--- a/scss/grid/_row.scss
+++ b/scss/grid/_row.scss
@@ -82,7 +82,6 @@
   @if type-of($gutter) == 'number' {
     $gutter: ($-zf-zero-breakpoint: $gutter);
   }
-  max-width: none;
 
   @each $breakpoint, $value in $gutter {
     $margin: rem-calc($value) / 2 * -1;


gpspake:
For further reference, here's how this has worked up until 6.2.1. The `expanded` class already sets `max-width:none` so applying it to all nested rows is too broad.
```html
<div class="row expanded"> <!--Parent - Full width-->
    
    <div class="row"> <!--Nested - Contained to grid-->

    </div>

    <div class="row expanded"> <!--Nested - Full width-->
   
    </div>
    
</div>```

**Comment by andycochran** [comment]
:+1: I agree; [this commit](https://github.com/zurb/foundation-sites/commit/3822efecf799fe56d6a8bcfe1a9eb96312b6050e) was a mistake.

**Comment by andycochran** [comment]
@gakimball, are you okay with reverting this and requiring `.expanded` to fix nested rows? Other ideas?

**Comment by rafibomb** [comment]
I'll take this one! I'll test it and see if it's the way to go.

**Comment by jrock17** [comment]
@rafibomb I believe this fix also needs to be made for the flex grid in _flex-grid.scss around line 165

**Comment by jrock17** [comment]
@rafibomb I think there is a related issue on the flex-grid. It's missing the margin-left and margin-right for nested rows inside an expanded row.

In _flex-grid.scss around line 165 I think you need to add the margin stuff for the nested row:

```scss
    // Expanded row
    &.expanded {
      max-width: none;

        .row {
            margin-left: auto;
            margin-right: auto;
        }
    }
```

gpspake:
Small little anecdote that doesn't really change anything here...
We added a temporary fix for this which explicitly adds a max-width back to `.row .row`
which we just realized overrides the nested expanded rules which prevents us from doing nested expanded rows :(
If we get rid of our temporary rule, though, we're back to square one.
best bet is to undo that pesky [one-liner](https://github.com/zurb/foundation-sites/commit/3822efecf799fe56d6a8bcfe1a9eb96312b6050e)

PR #8892, opened 2016-06-04

simshaun:
Hi,

I'm creating a tabbed interface using only the minimal syntax required for Tabs to work (foregoing the default Tabs styling). However, I need to be able to use different _active_ class names than the Tabs javascript expects.

This PR makes the _active_ class name configurable for both the tab links and tab panels.


simshaun:
Note: Travis failed because it couldn't download PhantomJS.

rafibomb:
This is a valuable PR!

rafibomb:
Anyway you can point this to v6.3? Trying to follow a standard of point releases and features. Thanks!

simshaun:
Opened a new PR pointed at v6.3.  https://github.com/zurb/foundation-sites/pull/9018

PR #8904, opened 2016-06-06
Enable the script to load in DoneJS SSR

marshallswain:
Adding this check for `script && script.parentNode` allows foundation-sites to work with DoneJS's Server Side Rendering.

This fixes an issue where the limited DOM used by done-ssr would prevent the script from running and would fail with the following error:

```
Error evaluating file:/Users/marshall/Sites/gh-test/public/node_modules/foundation-sites/dist/foundation.js
Cannot read property 'parentNode' of undefined
    at file:/Users/marshall/Sites/gh-test/public/node_modules/foundation-sites/dist/foundation.js:855:13
```

I don't think I can make a test for this.

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -142,7 +142,7 @@ window.matchMedia || (window.matchMedia = function() {
     style.type  = 'text/css';
     style.id    = 'matchmediajs-test';
 
-    script.parentNode.insertBefore(style, script);
+    script && script.parentNode && script.parentNode.insertBefore(style, script);
 
     // 'style.currentStyle' is used by IE <= 8 and 'window.getComputedStyle' for all other browsers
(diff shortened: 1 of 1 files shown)


marshallswain:
I watched the tests pass (why wouldn't they, it's just a variable check?) and then the phantom install failed.

AdamBarnwell:
@marshallswain I've had the same problem with my tests. Travis CI has a hiccup now and again.

andycochran:
@kball You wanna weigh in on this PR and nudge Travis to rebuild if you think it's a good change?

kball:
This seems fine; I don't know if it will generally help other server-side rendering frameworks, or even if it is sufficient for DoneJS (there may be other assumptions in other parts of the chain) but it's not going to break anything in the client-side environment, so I see no harm in merging.

PR #8905, opened 2016-06-07
Fix the slider to set var on offset percent

jasonraimondi:
When I ran npm test locally, both before and after the change and it was successful. After the change, I am able to compile and successfully run foundation.slider.js through both gulp (laravel elixir) and webpack with gulp.

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -292,7 +292,7 @@ class Slider {
       } else {
         barXY = eventFromBar;
       }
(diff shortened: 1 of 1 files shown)


jasonraimondi:
The CI integration seems to have failed on an `npm install` which my change should be unrelated.

aldenfbrown:
This fix worked for me as well. @jasonraimondi thanks!

PR #8906, opened 2016-06-07
Fix display for .input-group-button

OndrejVasicek:
I believe that in flexbox is no need to use display: table-cell. I tried that and it works happily without it.


rafibomb:
Good solution!

Issue #8851, opened 2016-05-25
[Input Group] Flexbox input group overflows the width of parent containers


johnbacon:
The flexbox input group overflows the width of any parent container that is smaller than the input group thinks it should be.

Desired behavior (`max-width: 120px;`):
![](http://i.imgur.com/r6MiI96.png)
Actual behavior (`max-width: 120px;`):
![](http://i.imgur.com/GbeQZlu.png)
Here is a Codepen that demonstrates the issue and includes a fix: http://codepen.io/johnbacon/pen/xOKdNG

The root cause is the `width: auto` that was applied to the flexbox-based `.input-group-field` in [this commit](https://github.com/zurb/foundation-sites/commit/4ff0759a6c3abc36c48e0610088b6c0af86b99e0#diff-36a3c48d0968fa90c112e137ecd866ffR91). I'm not sure sure why it was there (if it was done for a reason), so I don't feel comfortable submitting a pull request without further testing. However, removing the `width: auto;` appears to to fix the issue.

OverFlow636:
I stumbled on this issue as well. However its not just max-width that it overflows, it will overflow any parent that is smaller than it thinks it needs to be.

http://codepen.io/OverFlow636/pen/zBajyo

johnbacon:
Thanks for the update on this, @OverFlow636. I've modified the title and content of the issue to match.

kball:
Fixed via #8907, will be in 6.2.4

PR #8907, opened 2016-06-07
Fix proportions of children in input-group

OndrejVasicek:
If I understand flex model correctly, there is no need to force the ‘.input-group-field‘ item to have certain width. Actually it breaks the design in some cases.

The point is, the flexbox container calculates the width of its items itself. If we set flex-grow to 1 for ‘.input-group-field‘ (which we did), the flexbox gives that item “width priority” over the rest, which is set 0 (default).

Before commit:
input-group-error

After commit:
input-group-fix

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -91,7 +91,6 @@ $input-prefix-padding: 1rem !default;
     // scss-lint:disable ZeroUnit
     @if $global-flexbox {
       flex: 1 1 0px;
-      width: auto;
       height: auto;
     }
     @else {


rafibomb:
So odd that it would make a difference - good PR! We'll pull it down to test!

OndrejVasicek:
Yeah, I'm playing with the flexbox and usually it's wonderful, but sometimes it behaves like magic :) A lot of is hidden in the black box called browser.

But someone with an ability to see a bigger picture should check it. It may have some secret meaning.

OverFlow636:
this pr will fix issue #8851

kball:
This makes sense - the combo of flex-grow 1, flex-shrink 1, and flex-basis 0 should result in these all being sized automatically to fit, if there's room, and extra space distributed evenly across them.  Width auto shouldn't be needed, and somehow I'm not surprised it is breaking.

dobromir-hristov:
I had to add the input into a `div.input-group-field` for it to shrink.

PR #8911, opened 2016-06-08
Make access to $foundation-palette a bit easier

DaSchTour:
Accessing colors from $foundation-palette through map-get needs a lot of code. This way it would be a lot easier to use colors from palette.


kball:
@andycochran @brettsmason what do you guys think of this?

andycochran:
:+1: This is a fine idea. I can think of many uses for this function.

PR #8912, opened 2016-06-08
Change radius of input-groups to use $input-radius

DaSchTour:
I think this is a way better solution. I don't think that there are many users that would like to have different radius on input than on input-groups.


rafibomb:
That's a good point - like it!

rafibomb:
Restarting travis for merge

rafibomb:
Since this is a "breaking change" - can you point the PR to branch `v6.3`? We're keeping that for features or things that need migration.

DaSchTour:
New PR #9116

PR #8916, opened 2016-06-08
Fixed invalid class in example.

rubiii:
Just removed a dot really.


andycochran:
Oops! Nice catch. Docs PRs should target the `master` branch. Could you redo this for `master`?

rubiii:
Just found out this was already fixed in `master` by #8830.

PR #8919, opened 2016-06-09
Update grid.md

error500:
Just a typo


error500:
ok that was not a mistake

PR #8922, opened 2016-06-10
Correct vertical tab 5/6 links

tuarrep:
Replace links anchors for vertical tabs 5 and 6 in example. They point to panel 3 and 4


andycochran:
👍 Good to merge.

Issue #8915, opened 2016-06-08
Issue with nested rows in collapsed parent


saadshahd:
Hi,

There's an issue with these rules:
```CSS
.row .row {
  margin-right: $negative-gutter;
  margin-left: $negative-gutter;
}
```
Which undo the gutters of the nested row's parent column. But in this scenario:
```CSS
.row.collapse
  .columns
    .row
```
It causes an issue, because the parent column here doesn't have gutters. So I tried to fix it here #8853.
By adding this new rules:
```CSS
.row.collapse > .column > .row {
  margin-left: 0;
  margin-right: 0;
}
```
But the issue remains for scenarios like here http://codepen.io/SaadShahd/pen/zBvXXr?editors=1100

So I added to my codebase a modifier class and it determines when the row is a child of a collapsed row.

saadshahd:
@abdullahsalem No, my PR fixes the first scenario but this [one](http://codepen.io/SaadShahd/pen/zBvXXr?editors=1100) differs and doesn't get fixed.

abdullahsalem:
@saadshahd You are welcome to send another PR ;) would you?

saadshahd:
@abdullahsalem sure

saadshahd:
@abdullahsalem Updated!

rafibomb:
Thanks for the PR! Closing in favor of PR

abdullahsalem:
Thanks @saadshahd ! and sorry for this delay in reply.

saadshahd:
@rafibomb @abdullahsalem Thanks Guys!

PR #8925, opened 2016-06-10
Add is-collapse-child class

saadshahd:
#8915


rafibomb:
This is great! Can you add the codepen to a visual test to help can prevent regressions.

PR #8937, opened 2016-06-16
Abide checkboxes group validation

SassNinja:
Handle checkboxes similar to radio:
if one checkbox within the group is required, the whole group requires at least one checkbox to be checked.
The default value for the min amount of required checkboxes is 1 but can be changed by adding data-min-required to the checkbox.

Example can be found in test/visual/abide/abide-checkboxes.html


andycochran:
Woah. There are 137 commits and 91 changed files in this PR. It's hard to tell what's going on here. Are you maybe working on the `develop` branch and trying to pull that into `v6.3`?

rafibomb:
I made sure that v6.3 is in sync with develop. I think we'll need a clean PR with only the intended files. @SassNinja (cool name BTW) can you resubmit?

kball:
Closing this because the PR is unclean.  Please resubmit against v6.3 with just the commits of the change.  Thanks!

Issue #8470, opened 2016-03-26
6.2 Orbit changes slide on arrow, even if accessible = false


nviitala:
If you create Orbit with parameters like:
data-orbit data-bullets="false" data-nav-buttons="false" data-auto-play="false" data-swipe="false" data-accessible="false" data-infinite-wrap="false"

And you have input fields inside an Orbit slide, when user presses right or left key inside the input, the orbit will slide.

Outside input fields data-accessible="false" works as expected.

On foundation.orbit.js there is "this.$wrapper.add(this.$bullets).on('keydown.zf.orbit', function(e) {" on line 219. It should be inside an if:
```
if (this.options.accessible) {
this.$wrapper.add(this.$bullets).on('keydown.zf.orbit', function(e) {
...
});
}
```

Owlbertz:
Fixed via #8942.

PR #8942, opened 2016-06-17
Change slides with arrow keys only if the data-accessible option is true

dhargitai:
The related bug was reported under https://github.com/zurb/foundation-sites/issues/8470


Owlbertz:
Seems to work out, thanks!

PR #8943, opened 2016-06-17
Components settings

andreamk:
Added the variable map "$components_setting " in global , with which you can enable the foundation components.


rafibomb:
Wow - intersting approach! How does this improve on the current method of controlling includes in app.scss?

andreamk:
Hello
I'm developing a wordpress plugin that allows to change the settings in the backend .

Add the components in an array improves two things .
1 ) You no longer need to modify the inclusions and it changes only the file settings.scss
2 ) you can create external components that depend on the active modules .

for example, in an external scss (wordpress theme scss) you can do this 

` @if (not foundation_component_enabled("forms")) {
  	input {
            box-sizing : border-box;
            width: 100%;
        } 
  }`

andreamk:
It also solves a problem of inconsistency .

include "flex-grid" and set "$global-flexbox" false I think is a possible error.
With the $components-settings you can replace 
`@if $global-flexbox {`
with
`@if (foundation_flex_grid_enabled()) {`
removing any possibility of error

andycochran:
Ooh. This *is* interesting. What do the @zurb/yetinauts think? 
(It's a major change tho, so it'd need to target the `v6.3` branch.)

rafibomb:
Yeah - this is a big change! @andreamk Can you point this at branch `v6.3`? In the meantime I'll get some more eyes on it see how we can fit it in!

colin-marshall:
Haven't had a chance to test this out, but it seems logical to me to consolidate this into `_settings.scss`. I'm all for it!

andreamk:
hello
I don't understand how to point the pull request at branch 6.3

brettsmason:
@andreamk unfortunately the only way as far as I know is to close this PR and open a new one, targetting the 6.3 branch.

Owlbertz:
Yes, but please perform a merge of v6.3 into your feature branch locally before opening the PR.

andreamk:
Ok

Issue #8900, opened 2016-06-06
Slider : offsetPct variable undefined


guoni:
foundation-sites/js/foundation.slider.js

Since 6.2.2, the variable offsetPct is not defined

```
  _handleEvent(e, $handle, val) {
    var value, hasVal;
    if (!val) {//click or drag events
      e.preventDefault();
      var _this = this,
          vertical = this.options.vertical,
          param = vertical ? 'height' : 'width',
          direction = vertical ? 'top' : 'left',
          eventOffset = vertical ? e.pageY : e.pageX,
          halfOfHandle = this.$handle[0].getBoundingClientRect()[param] / 2,
          barDim = this.$element[0].getBoundingClientRect()[param],
          windowScroll = vertical ? $(window).scrollTop() : $(window).scrollLeft();
         **missing variable offsetPct**

      var elemOffset = this.$element.offset()[direction];

      // touch events emulated by the touch util give position relative to screen, add window.scroll to event coordinates...
      // best way to guess this is simulated is if clientY == pageY
      if (e.clientY === e.pageY) { eventOffset = eventOffset + windowScroll; }
      var eventFromBar = eventOffset - elemOffset;
      var barXY;
      if (eventFromBar < 0) {
        barXY = 0;
      } else if (eventFromBar > barDim) {
        barXY = barDim;
      } else {
        barXY = eventFromBar;
      }
      **offsetPct** = percent(barXY, barDim);
```

Best regards,
Guillaume

abdullahsalem:
@guoni Could you please use this [issue template](https://github.com/zurb/foundation-sites/blob/develop/.github/ISSUE_TEMPLATE.md)

jasonraimondi:
I am seeing the same issue. This is happening both when I am running through Gulp/babel (laravel elixir babel to be exact), and webpack.

**How can we reproduce this bug?**

1. Webpack build with babel preset to ['es2015']
2. Place slider on an html page
3. The slider renders fine, but does not slide at all, trying to slide the slider causes console errors
4. ![screen shot 2016-06-06 at 7 17 45 pm](https://cloud.githubusercontent.com/assets/5787967/15844001/787ebc22-2c1b-11e6-9d3a-aebb481a86d5.png)

**What did you expect to happen?**
The slider to slide

**What happened instead?**
Slider errors and nothing happens, tick doesnt move, console logs
[...]

PR #8948, opened 2016-06-19
fix unset var on foundation slider

jasonraimondi:
Foundation slider has an unset variable where the result is the slider renders properly, but when slid the slider throws a console error stating the variable 'offsetPct' is undefined.

Please see issue #8900

### DIFF
--- a/dist/foundation.js
+++ b/dist/foundation.js
@@ -7272,7 +7272,7 @@ function _classCallCheck(instance, Constructor) { if (!(instance instanceof Cons
           } else {
             barXY = eventFromBar;
           }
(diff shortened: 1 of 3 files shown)


rafibomb:
It looks like this affects the `dist` folder. Can we change it to src then should be good to merge

Owlbertz:
@jasonraimondi are you able to fix this PR changing the dist files anytime soon? There are already two duplicates of this PR and I really want to go ahead and merge it to prevent future duplicates.

kball:
The dist stuff will get overwritten in release, but this does update the src as well, so it should be fine.

Issue #8956, opened 2016-06-22
Tab deep linking fails w/ jQuery unrecognized expression error


datenimperator:
**How can we reproduce this bug?**

Our sites use Foundation 5.5.3 with jQuery 1.12.1
1. Open a page containing tabs with deep linking enabled, include the selected tab id in the url

**What did you expect to happen?**
1. Page opens and displays selected tab initially

**What happened instead?**

There's a jQuery error on the console:

`Error: Syntax error, unrecognized expression: [data-tab] > * > a[href=#contacts]`

Here, `#contacts` is the id of our tab.

**Test case:**

Please see https://jsfiddle.net/5x5eqap7/

**Fix:**

In file `foundation/foundation.tabs.js` line 105, 111, 117 it needs double quotes around the hash. Eg line 111:

``` js
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href="#' + hash_tab_container_id + '"]').parent(), hash);
```

rafibomb:
Good catch! Is that something you can submit a PR for? We'll merge it in.

datenimperator:
Done, not sure if I got the branches right. Please see the PR linked above.

andycochran:
@rafibomb #8878 has a fix for this. I'd appreciate your feedback on it disallowing single quotes.

earksiinni:
@datenimperator I had this exact same problem. Changing the `data-tab` attribute to `data-tabs` fixed it.

earksiinni:
This appears to be a bug in jQuery. Newer versions don't have this behavior.

earksiinni:
Specifically this: https://github.com/jquery/jquery/issues/2824

colin-marshall:
Very sorry for the late response to your issue. We are moving on to Foundation 7, and there will be no more development or support for Foundation 5. I am closing this one out.

PR #8968, opened 2016-06-24
Fix jquery expression error w/ deep linked tabs

datenimperator:
Not sure where to base that branch onto, since it's targeting Foundation 5.

Fixes zurb/foundation-sites#8956


andycochran:
This is a duplicate of #8878. As I mentioned there, it would disallow the use of single quotes. But if you have any feedback on that, comments would be appreciated on the open PR.

PR #8970, opened 2016-06-26
Added max-width 100 to shrunken column

abarnwell:
Changes based on feedback on http://foundation.zurb.com/forum/posts/43843-flex-column-shrink-max-width-issue

PR #8973, opened 2016-06-27
Moved $gutter calcs to dedicated function and added max-width 100% to shrunken column

abarnwell:
I've moved the calculations for $gutter padding to their own function as I would like to use the returned value for purposes of vertical rythm and various areas where I may not be using the row and column layout but I want spacing to be the same.

I've also added max-width 100% to shrunken columns based on feedback on http://foundation.zurb.com/forum/posts/43843-flex-column-shrink-max-width-issue

### DIFF
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -51,25 +51,23 @@
 
 /// Calculates the gutter sizes of a column based on the $gutter values and breakpoints.
 ///
-/// @returns {Number} The gutter width.
-@function grid-column-gutters {
+/// @param {String} the attribute to use.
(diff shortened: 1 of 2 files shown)


AdamBarnwell:
I stupidly made my last pull request changes on the same branch so that pull request (https://github.com/zurb/foundation-sites/pull/8970) will need to be merged prior to this, sorry. :-(

kball:
Hey @AdamBarnwell this appears to break the sass build due to missing variables - I think it needs a little more work.  Going to close, please feel free to fix and reopen.

PR #8923, opened 2016-06-10
Remove references to deprecated jQuery.fn.load()

silverelizard:
The following changes are to eliminate deprecated references to jQuery.fn.load() and convert them to jQuery.fn.on() with load as parameter.

PR #8979, opened 2016-06-27
Bind 'load' event instead of using the ajax function

spyesx:
Everything is in the title :)
We don't need to load `window` but to know when it is loaded.


spyesx:
Ok the error doesn't come from my fix but from `phantomjs` :) I guess you can merge.

Owlbertz:
As I can see this issue is already addressed in #8923.
I guess we will go ahead with the other PR since it also fixes the deprecation in other components.

Owlbertz:
Closed with #8923.

spyesx:
Yes sure! Thx

PR #8987, opened 2016-06-30
nix unnecessary negative right margin in docs' flex grid scss

### LINKED ISSUE #8893: [Flex Grid] Bugs in docs
1. Visit the http://foundation.zurb.com/sites/docs/flex-grid.html page
2. Watch any example with row/columns

Nice columns using the flexbox.

**What happened instead?**
* Last columns has “float: right”, which I believe it’s not necessary with flexbox.
* There is a rule “margin-right: -1px“ on columns, which makes either missing pixel in column (white 1px column without background) or missing right border (column covers it).
  
  foundation2

    foundation1

* The basics example shows “12/6/4 columns“, but there isn’t any “small-12” class. It’s ok in regular grid, where the full width is the default, but not in flex. Here we apply the expand behaviour, as you write further.

  foundation3

andycochran:
Fixes #8893. 

This PR fixes a `margin-right` bug in the Flex Grid docs. 

There was a line with a comment in `docs/assets/scss/_grid.scss`: 

``` scss
margin-right: -1px; // Not sure why this has to be here
```

I found no reason for it to remain. So I nixed it. Everything works fine.

### DIFF
--- a/docs/assets/scss/examples/_grid.scss
+++ b/docs/assets/scss/examples/_grid.scss
@@ -65,7 +65,6 @@
   .columns {
     float: none;
     width: auto;
-    margin-right: -1px; // Not sure why this has to be here
 
     &:nth-child(odd) {
       background: #eee;


rafibomb:
Wow - that was weird. Thanks for removing it!

PR #8988, opened 2016-06-30
add settings variable for accordion title font size

### LINKED ISSUE #8935: Font-size accordion item title should not be set to arbitrary font-size

in the @mixin accordion-title

`font-size: rem-calc(12);`

The font-size should reference the global style and should be customizable

**What happened instead?**

the font-size is set to an arbitrary size: `font-size: rem-calc(12);`

andycochran:
Closes #8935.

### DIFF
--- a/scss/components/_accordion.scss
+++ b/scss/components/_accordion.scss
@@ -14,6 +14,10 @@ $accordion-background: $white !default;
 /// @type Boolean
 $accordion-plusminus: true !default;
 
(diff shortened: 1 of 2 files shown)


rafibomb:
I'm all for adding variable for things that are hard to change - this one is one line of CSS: http://codepen.io/rafibomb/pen/pbdzyG

rafibomb:
It's a good PR though - I think people will use it! Well done!

Issue #8899, opened 2016-06-05
grid-column-position(center) override issue?


carpusmedia:
Hello,

We are having issues with the `grid-column-position` mixin in _Foundation 6 (ZURB Template)_. We have one column that is set with

`grid-column(10)`

and we are trying to center this with

`grid-column-position(center)`

however, the column in question does not center.

When we check the source

```
element:last-child:not(:first-child) {
  float: right;
}
```

seems to override the code of grid-column-position, which seems to be

```
element {
  float: none;
  margin-left: auto;
  margin-right: auto;
  padding-left: 0.625rem;   
  padding-right: 0.625rem;
  width: 83.3333%;
}
```

Anything we are missing, or is this just a bug?

Thanks

ar-rm:
I think I'm seeing exactly the same as well.

ar-rm:
I figured it out (at least for me).

The clue was "last" and "not first" in the selector:
`element:last-child:not(:first-child)`

This was my markup:
```
<div class="container">
    <h4>Header</h4>
    <div class="centerplease">Center this</div>
</div>
```

This was my sass:
```
.container {
    @include grid-column(12);
}
h4 {
    visibility: hidden;
}
.centerplease {
    @include grid-column(10);
    @include grid-column-position(center);
}
```

The problem was the `h4` in the **markup**.
`grid-column-position(center)` will only work if it's the only element in that row. (And by "row" I don't necessarily mean `grid-row()`)

I didn't think about the `h4` being related since it was visually *hidden* in the css. But the first-child, etc. pseudo selectors only care about the *actual* markup, not the how it's styled.

IOW:
- Not a bug (unless you count my brain as a bug)
- Check your **markup** to see if there are any other elements in same containing row as the column you want to center.

I hope this helps.

andycochran:
I'll fix this so that any number of elements can be centered…

ar-rm:
@andycochran Ohhh! That would be ideal. Thanks.

rafibomb:
Closing for PR #8990

PR #8990, opened 2016-07-01
add specificity to centered columns to override :last-child

andycochran:
This PR addresses #8899. The specificity of `:last-child:not(:first-child)` caused columns to float right even if they should be centered. Now the `grid-column-position` mixin will work for the last grid column.

### DIFF
--- a/scss/grid/_position.scss
+++ b/scss/grid/_position.scss
@@ -17,7 +17,9 @@
     #{$global-left}: $offset;
   }
   @else if $position == center {
-    float: none;
+    &, &:last-child:not(:first-child) {
+      float: none;
+    }
     margin-left: auto;
     margin-right: auto;
   }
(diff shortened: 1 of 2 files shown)


rafibomb:
This looks looks good! I'd like us to have a good test in the test folder to make sure there are no regressions.

Would this work? http://codepen.io/rafibomb/pen/RRVqww

andycochran:
@rafibomb Cool. I'll add a test. Maybe something like this so the columns' edges are more obvious. http://codepen.io/andycochran/pen/xOdmKG

andycochran:
@rafibomb Test is ready to go.

rafibomb:
Sweet! I thought that's what this was for! Great PR!

rafibomb:
So we can sum this up as: now you can have multiple centered columns stacked in one row.

andycochran:
Yup! They can even be various widths. Which just made me realize a bug: if a centered column follows uncentered columns that don't fill the row, this can happen: 

image

Eek! I think the solution here is `clear:both` on centered rows?

PR #8991, opened 2016-07-02
fix centered columns flowing around uncentered columns

andycochran:
@rafibomb as soon as ya merged #8990, I realized there was a bug ([explained here](https://github.com/zurb/foundation-sites/pull/8990#issuecomment-230072603)). This PR should fix it.


rafibomb:
Awesome good catch! Test look great!

Issue #8978, opened 2016-06-27
Buggy keyboard behavior on drilldown menu.


sassanh:
First of all thanks for your efforts for providing this great library. I like many others appreciate it. Here's the description of a bug I found:
1. Go to http://foundation.zurb.com/sites/docs/drilldown-menu.html in Chrome 51 or Safari 9.1 or Firefox 46.0.1 or Firefox 47 (actually I guess browser version or kind doesn't matter)
2. Click on "Item1" in the sapmle
3. Click on "Item 1A"
4. Click on "Back"
5. Press down arrow key on your keyboard
6. Whoops, suddenly you're back into "Item 1A" (Bug 1)
7. Press Up
8. press Enter
9. Whoops, it animates like it's going back to the parent menu, but it doesn't (Bug 2)
10. Try to go back with clicking on Back item.
11. Whoops you can't. (Bug 3)
12. Keep Enter pressed for 5 sec on Back item.
13. OK enough, seems like keyboard navigation is not production ready on drilldown menu at all.

**What did you expect to happen?**
It to work.

**What happened instead?**
It didn't work.

**Test case:**

Give us a link to a CodePen or JSFiddle that recreates the issue.

No need for CodePen or JSFiddle, http://foundation.zurb.com/sites/docs/drilldown-menu.html is enough.

OK, actually I don't need this keyboard navigation on my drilldown menu at all. All I need is to be able to disable it (it being keyboard navigation on drilldown menu) so that I can deploy my code on the server without worrying about users who try to navigate it with keyboard.
It'd be great if I could remove the outlines it renders on items (trying to show which item is selected by keyboard, but just trying). It'd be great if I could disable fancy features and have a simple working drilldown menu.

Owlbertz:
Hi,
you're right, the keyboard navigation for Drilldown seems to be broken (again). This seems to be one of the most broken feature during releases... It seems like this happened from [6.2.1](http://codepen.io/anon/pen/dXRNYL) to [6.2.2](http://codepen.io/anon/pen/YWQNxz).

If you don't care about modifying the Foundation files, you can remove the event listener in the module.

sassanh:
@Owlbertz thanks for info, I do care about modifying Foundation files, is it alright in 6.2.1? I can downgrade.

Owlbertz:
[...]

PR #9002, opened 2016-07-06
Fix drilldown keyboard

Owlbertz:
Quick fix for #8978 by adding tabindex to sub-menu parents.

Second commit removes the tabindex on destroy, messed up the message.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -149,8 +149,6 @@ class Drilldown {
         }
       });
 
-      console.log('Next:', $nextElement, 'Prev', $prevElement);
-
       Foundation.Keyboard.handleKey(e, 'Drilldown', {
         next: function() {
           if ($element.is(_this.$submenuAnchors)) {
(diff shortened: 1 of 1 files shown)


andycochran:
Nice! Why do you consider it a quick fix? Is there a better (more involved) way to handle this?

Owlbertz:
Well the issue here is that the existing code removes the `href` attribute of all the sub-menu parents on initialization, therefore they are not focusable anymore (links without href are not considered interactive). That's why adding `tabindex` fixes the issue by simply making them focusable again.
Therefore I consider it quick, because it makes it work again without changing anything of the existing code. 

Maybe there is a need to streamline the way menus are initialized in general, as (afaik) other menu types do not remove `href` attributes - but that would involve some tinkering and that's why I decided to put out this quick thing.

Owlbertz:
I also pushed another commit that I had in my pipeline that will allow the actual activation of links with the keyboard.

rafibomb:
I think it's a good move forward! We should figure out why the `href`'s are being removed and see if there is a better way to handle that.

Owlbertz:
I think removing the `href` is done because the sub-menu parents don't actually function as links but only open the sub-menu. Maybe that behavior can be changed in the future so that the developer can decide whether he wants the sub-menu parents as a trigger only (e.g. `href="#"`, in which case `preventDefault()` should be called to prevent the URL/history from changing) or as actual links.

andycochran:
Perhaps sub-menu parents should be `<button>` elements not `<a>`.

kball:
LGTM.

PR #9003, opened 2016-07-06
Fixed keyboard usage for Accordion Menu.

### LINKED ISSUE #8680: AccordionMenu not accessible via VoiceOver on iOS

1. Create an Accordion Menu

Menu is accessible via screen reader

**What happened instead?**
VoiceOver on iOS only reads "menu item" for each element, and not the text of the link within the list item.  The link is not accessible or clickable.

By using Foundation.Nest, `role="menuitem"` is added to all of the `li` elements within the accordion menu. `menuitem` requires that it is contained within an element with role `group`, `menu`, or `menubar` ([reference](https://www.w3.org/TR/wai-aria/roles#menuitem)) but AccordionMenu applies `tablist`.

Owlbertz:
Fixes #8680.

No need to handle tab and shift tab events since the normal behavior seems to work out just fine.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -34,9 +34,7 @@ class AccordionMenu {
       'ARROW_UP': 'up',
       'ARROW_DOWN': 'down',
       'ARROW_LEFT': 'close',
-      'ESCAPE': 'closeAll',
-      'TAB': 'down',
-      'SHIFT_TAB': 'up'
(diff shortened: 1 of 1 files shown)


Owlbertz:
Also removed unnecessary tabindex Settings.

andycochran:
Which browsers have been tested?

Owlbertz:
Chromium 48 and Firefox 44. Now that you asked also IE 11.

Issue #8864, opened 2016-05-30
Equalizer - No option for specific media queries


Ne-Ne:
Hi All,

Looking for a foundation solution, can we include support for 'only' breakpoints?

E.g.

```html
<div class="row" data-equalizer data-equalize-on="medium" id="test-eq">
...
</div>
```

This solution with give users the power to be specific range, in my case the page design needs to to be equal height for medium only. Large and small are completely different. The only solution is to use `!important` as an override which terrible to have to use..

Ne-Ne:
Any ideas @gakimball @zurbchris?

kball:
Addressed in #9008, will be in 6.3

PR #9008, opened 2016-07-07
Feature - Foundation.MediaQuery.is() Method

auginator:
I was thinking this would be a good way for plugins to target a specific breakpoint, rather than being limited to a minimum breakpoint, as they are now with atLeast().  Accepts either ‘[breakpoint] only’ or ‘[breakpoint]’ - in the former case, it will see if the breakpoint matches exactly, and in the latter, it falls back to atLeast(). 

For instance: #8864 

@brettsmason @rafibomb  - Any thoughts on this? I would be happy to take a stab at integrating this into the plugins if the community thinks it is worthwhile. 

PS: It's my first attempt at contributing to Foundation - I hope I did not donk things up too badly!

### DIFF
--- a/js/foundation.equalizer.js
+++ b/js/foundation.equalizer.js
@@ -87,7 +87,7 @@ class Equalizer {
    * @private
    */
   _checkMQ() {
-    var tooSmall = !Foundation.MediaQuery.atLeast(this.options.equalizeOn);
+    var tooSmall = !Foundation.MediaQuery.is(this.options.equalizeOn);
     if(tooSmall){
       if(this.isOn){
         this._pauseEvents();
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -267,7 +267,7 @@ class Sticky {
    * @private
    */
   _setSizes(cb) {
-    this.canStick = Foundation.MediaQuery.atLeast(this.options.stickyOn);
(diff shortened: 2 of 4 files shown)


rafibomb:
This is a cool feature - so if I read it right this will allow `data-equalize-on="medium only"`

rafibomb:
We'll pull it down and test it - thanks for pointing to 6.3!

auginator:
Nice – I would still have to change methods on the plugins that use Foundation.MediaQuery to use this function instead of the atLeast() method, but I am more than willing to give it a shot and update the pull request. 

Happy I got the version number for feature requests right!

auginator:
I updated the branch with a few changes :) Now it works as planned with Equalizer, Tooltip, and Sticky. Cheers! 

**Equalizer:** 
`<div class="row" data-equalizer data-equalize-on="medium only" id="the-equal">...</div>`

**Sticky:** 
`<h3 class="callout" data-sticky data-stick-to="bottom" data-sticky-on="medium only">I stick on medium only!</h3>`

**Tooltip:** *(I must admit this one seems a little weird to me, not sure if it would practically be that useful.)*
`<span data-tooltip data-show-on="small only" …>…</span>`

auginator:
@rafibomb I forgot to @ mention you when I updated the branch one week ago.

kball:
I like this.  Nice work.

PR #9009, opened 2016-07-08
Allow definition of global-font-size in rem

DaSchTour:
Browser scaling through changing the base font-size in browser settings doesn't work correctly if the base is defined in percent. To make this work we define the base in rem. But in this case rem-calc stops working.

### DIFF
--- a/scss/util/_unit.scss
+++ b/scss/util/_unit.scss
@@ -40,7 +40,7 @@ $global-font-size: 100% !default;
     $base: ($base / 100%) * 16px;
   }
   
-  // Using rem as base allwos correct scaling
+  // Using rem as base allows correct scaling
   @if unit($base) == 'rem' {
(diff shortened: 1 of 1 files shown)


scss/util/_unit.scss
```diff
@@ -39,6 +39,11 @@ $global-font-size: 100% !default;
   @if unit($base) == '%' {
     $base: ($base / 100%) * 16px;
   }
+  
+  // Using rem as base allwos correct scaling
```
natewiebe13: There's a typo. `allwos` should be `allows`

rafibomb:
@andycochran What are your thoughts on this? Seems like it would be more universal.

rafibomb:
@DaSchTour Were you able to see why the test failed?

DaSchTour:
@rafibomb I don't see any failed test 😕

Issue #8536, opened 2016-04-06
off-canvas sidebar is aria-hidden by default, and can't change this!


Geczy:
The off canvas side bar has aria-hidden set to true on by default. I have the side-bar visible for large desktops, but that means accessibility software can't see the sidebar at all

Geczy:
@brettsmason  ?

PR #9011, opened 2016-07-08
Enhanced keyboard navigation for Offcanvas.

Owlbertz:
On open not the first element within the side panel, but the whole panel should receive focus.
If the first interactive element is focussed the user might miss some content above it, e.g. the panel heading.
Also added usage of handleKey() to the component to close the panel.
Moved the panel focussing outside of the transitionEnd listener because this was never called and it also works outside...

I would really appreciate feedback of @sethkane and @Geczy here.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -27,7 +27,11 @@ class OffCanvas {
     this._init();
     this._events();
 
-    Foundation.registerPlugin(this, 'OffCanvas');
+    Foundation.registerPlugin(this, 'OffCanvas')
+    Foundation.Keyboard.register('OffCanvas', {
+      'ESCAPE': 'close'
+    });
+
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Geczy:
Took a look and that definitely makes the off canvas menus behave better under accessible software. But my issue still remains #8536 

I work around it by adding `$('#my-info').attr('aria-hidden', 'false');` on dom ready

Owlbertz:
Huh, didn't saw that issue yet. Will try to get this done soon as well in the same branch.

Owlbertz:
@Geczy Do you work with standard visibility classes? I feel like a simple `this.$element.is(':visible')` check before applying the `aria-hidden` (and also the `aria-expanded` stuff I guess) would do the trick during the initialization.

rafibomb:
@Owlbertz That solution should work!

feeekkk:
since this added `keyboard` as a dependency to off canvas, it would have been nice to have noted that in the documentation!

PR #9014, opened 2016-07-11
Keyboard enhancements v6.3

Owlbertz:
Some general enhancements of the Keyboard util I wanted to make.
Fixes some odd behavior by tweaking `parseKey`.
Reduces the code base and offers functions to trap and release keyboard focus with `trapFocus` and `releaseFocus`.
Also got rid of a debug log.

I feel like this PR is not yet ready to be merged as there may be other stuff to add, but I'm interested in some feedback for the current state.


kball:
I like the concept here.  Would love some feedback from other accessibility focused folks like @sethkane and @gerardkcohen

kball:
@Owlbertz can you rebase this to resolve conflicts?  Also pinging again @sethkane and @gerardkcohen for accessibility-focused feedback.

kball:
Still looking for feedback on this from @sethkane and @gerardkcohen if you have time, given the accessibility implications. @Owlbertz from your perspective, how are you feeling about this PR? In the initial open you said that you felt it still needed work... it looks like some has gone in... is this ready?  Or needs more work and feedback?

Owlbertz:
Well I just fixed some issues with `parseKey` (should be fully backwards compatible) and added `trapFocus` and `releaseFocus` that can be used to trap focus (really!) so that there is less redundant code in the components (Reveal, Offcanvas and Dropdown for now). Also of course this means that sites can now call this function in other, non-Foundation-component contexts as well.

kball:
Merged!

PR #9016, opened 2016-07-12
Example didn`t work

tinacolada:
Added large-3 classes, as they were missing.


andycochran:
@tinacolada Could you clarify what wasn't working? Is it a specific browser/OS? I'm not seeing a bug. The columns shouldn't need the `large-3` class.

kball:
Closing in lieu of a response to @andycochran's question.

PR #9018, opened 2016-07-12

simshaun:
Hi,

I'm creating a tabbed interface using only the minimal syntax required for Tabs to work (foregoing the default Tabs styling). However, I need to be able to use different active class names than the Tabs javascript expects.

This PR makes the active class name configurable for both the tab links and tab panels.

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -48,7 +48,7 @@ class Tabs {
     this.$tabTitles.each(function(){
       var $elem = $(this),
           $link = $elem.find('a'),
-          isActive = $elem.hasClass('is-active'),
+          isActive = $elem.hasClass(`${_this.options.linkActiveClass}`),
           hash = $link[0].hash.slice(1),
           linkId = $link[0].id ? $link[0].id : `${hash}-label`,
           $tabContent = $(`#${hash}`);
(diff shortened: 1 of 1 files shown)


Owlbertz:
Hi @simshaun,
I pulled this down and had a look - seems to work out.
But I noticed that when you set the `panelActiveClass` option, the tabs will not be visible when active - because the `display: block` is linked to the `.tabs-panel.is-active` CSS-rule.

However, his is not the case for the tab links, because they are highlighted with  `.tabs-title > a:focus, .tabs-title > a[aria-selected='true']` - so it's based on the ARIA attribute. I think in order to make your changes not cause more work for the user (they would have to handle `display: block` for the active panel on their own), I think you should also move the `display: block` for the panels to a rule based on the ARIA attribute (i.e. `.tabs-panel[aria-hidden="false"]`).

Thoughts from any @zurb/yetinauts ?

simshaun:
Hm. That's not something I thought about since we were using completely custom styling.

I think changing the active class names is more of an advanced usage. It's not something most people are going to do, but it's nice to have the option and accept that the default styling may not work.

However, if it can be made to work based on the aria-hidden attribute, that'd be cool.

kball:
Ooh interesting idea hooking it into the aria, in this case I think that would work. @simshaun do you want to take a pass on it?

Broadly speaking we should make sure that all of the javascript class stuff handles cases where folks are using the mixins and not using our default classes.

simshaun:
Sorry for the late reply. I'll take a look and see if I can figure it out.

simshaun:
Alright, I've opened a new PR (#9369) that merges these changes with all the new collapsible stuff from the develop branch & also separates the tab panels' visibility from the `is-active` CSS class.

PR #9019, opened 2016-07-13
[JS-dropdown] fix overriden positionClass option

iamdey:
the `positionClass` is always overriden by element properties

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -51,10 +51,6 @@ class Dropdown {
 
     });
 
-    if (this.options.positionClass.length === 0) {
-      this.options.positionClass = this.getPositionClass();
-    }
-
     this.counter = 4;
     this.usedPositions = [];
     this.$element.attr({
@@ -77,7 +73,12 @@ class Dropdown {
     var horizontalPosition = /float-(\S+)\s/.exec(this.$anchor[0].className);
         horizontalPosition = horizontalPosition ? horizontalPosition[1] : '';
     var position = horizontalPosition ? horizontalPosition + ' ' + verticalPosition : verticalPosition;
-    return position;
+
+    if (position.length > 0) {
+      return position;
+    }
+
+    return this.options.positionClass;
   }
 
   /**
@@ -148,6 +149,8 @@ class Dropdown {
     }
 
     this.$element.offset(Foundation.Box.GetOffsets(this.$element, this.$anchor, position, this.options.vOffset, this.options.hOffset));
+
+    this.initialPositionClass = this.getPositionClass();
 
     while(!Foundation.Box.ImNotTouchingYou(this.$element, false, true) && this.counter){
       this._reposition(position);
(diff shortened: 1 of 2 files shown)


iamdey:
npm failed to install on travis :(

Owlbertz:
I restarted the Travis check, it passed now.
Could you give a small example code on what was not working correctly?

test/javascript/components/dropdown.js
```diff
@@ -23,6 +23,27 @@ describe('Dropdown', function() {
       plugin.$element.should.be.an('object');
       plugin.options.should.be.an('object');
     });
+
+    it('uses user\'s defined options', function() {
```
iamdey: > Could you give a small example code on what was not working correctly?

easy; this test will fail with current code @Owlbertz.

Owlbertz:
Can you provide the HTML/JS you use to set up the Dropdown?

iamdey:
this is my real world example : https://github.com/CaliOpen/caliopen.web-client-ng/blob/master/src/js/component/layout/header/user-menu.js#L8-L37

this may not be a simple example since we use foundation as a library not a framework (we use foundation' scss mixins in a BEM approach and foundation JS components only when it's required in  an ES6 angular project). In my use case it is a non-sense to use foundation specific stylesheet classes (`bottom`, `float-right`).

You probably prefer a not running example, I will create a plunkr asap (probably monday).

iamdey:
@Owlbertz I've added a working example showing that defining manually `positionClass` has no effects. The only way is by using foundation classes, but it's not in the way I'm using foundation.

http://codepen.io/esion/pen/mEXPxR

Owlbertz:
Thanks for that stunning example. As I see the same applies for the configuration via the `data-position-class` attribute.

Owlbertz:
Alright, after checking this out I feel like this has to be placed right in the `getPositionClass()` method.
As far as I can see there also needs to be a change in `open()` because otherwise it will not work (at least not when using the `data-position-class`). Putting your change right on top of the `getPositionClass()` should get this running for all cases.
What do you thing?

iamdey:
Do you mean, `getPositionClass` is responsible to give the correct depending on option or data-* or stylesheet classes ? so yes :+1: 

1. it prevents variable mutations
2. it improves separation of concerns

Owlbertz:
Do you mind adjusting the PR accordingly?

js/foundation.dropdown.js
```diff
@@ -312,7 +318,8 @@ class Dropdown {
       if(curPositionClass){
         this.$element.removeClass(curPositionClass);
       }
-      this.$element.addClass(this.options.positionClass)
+
+      this.$element.addClass(this.initialPositionClass)
```
iamdey: the code in `close()` and `_reposition()` is too much legacy and my fix not a good solution at all; please reject the PR.

iamdey:
In my mind the positioning thing needs to have a whole refactor, adding and removing classes should not happen in that way.

kball:
Closed based on the above.  Refactoring positioning is on the agenda for 6.3 or 6.3.1

PR #9020, opened 2016-07-13
[JS-dropdown] fix float-right is ommited when last in classes

iamdey:
the regex requires a empty character after `float-something` 

I didn't get why `\s` has been added (the unit test is ok without it)


Owlbertz:
I agree, it makes no sense to expect a space after the float class.

iamdey:
done @Owlbertz

Owlbertz:
Ok, I pulled this down and had a look. Seems like your change is unnecessary because the `hover` class will always be applied to the anchor before `getPositionClass()` is called. Therefore, `float-*` will never be the last class.
However, to make this more future proof, I'm happy you found this.

Issue #9021, opened 2016-07-14
[bug scss] Hover disabled buttons reset color


hylke94:
Every button with the class of button and disabled is being reset to the base color when hovering over the button.
For example: a button with classes "button alert disabled" is red with an opacity to show the button is disabled. When you hover over this button the color is turning black to blue.

**What did you expect to happen?**
The button stays red when hovered.

**What happened instead?**
The button turns back to blue when hovered.

**Test case:**
[JSFiddle](https://jsfiddle.net/Hylke94/pyh76hyd/)

kball:
Fixed in #9022.

PR #9022, opened 2016-07-14
Fix for resetting background color on hover and focus state of a disabled flavored button.

denisahac:
Fix for #9021

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -136,12 +136,12 @@ $button-opacity-disabled: 0.25 !default;
 }
 
 /// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
-@mixin button-disabled {
+@mixin button-disabled($color: $primary-color) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
 
   &:hover, &:focus {
-    background-color: $button-background;
+    background-color: $color;
     color: $button-color;
   }
 }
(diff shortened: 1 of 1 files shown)


Silvester20:
Works great for a normal download, thanks for that! But unfortunately this PR doesn't work for the custom build download. Or actually I probably should say this PR does work but somehow it is not included in the "foundation.min.css" from the [custom build download](http://foundation.zurb.com/sites/download.html/).

Can you take a look at it @kball ?

kball:
@Silvester20 I just tried doing a download of the full from there and it appeared to work fine... can you give me more details about the options you used for the download?

Silvester20:
@kball Yes, I didn't made any change to the custom download. Just go to the [download page](http://foundation.zurb.com/sites/download.html/) and scroll to bottom and press the big orange button "Download Custom Build". Then I took the minified version of the CSS file.

kball:
Thanks @Silvester20, that helped me find it.  There was a caching problem in the customizer and in some cases it was still serving out 6.2.3.  has been fixed.

scss/components/_button.scss
```diff
@@ -136,12 +136,12 @@ $button-opacity-disabled: 0.25 !default;
...
 /// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
-@mixin button-disabled {
+@mixin button-disabled($color: $primary-color) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
 
   &:hover, &:focus {
-    background-color: $button-background;
+    background-color: $color;
     color: $button-color;
```
stianpr: Shouldn't this use `foreground($color, $button-color-alt, $button-color)`? if the button background color is very light? The secondary color is usually gray so we don't want the font color to be white then...

PR #9026, opened 2016-07-15
Components settings

andreamk:
https://github.com/zurb/foundation-sites/pull/8943
target on 6.3

### DIFF
--- a/assets/foundation-flex.scss
+++ b/assets/foundation-flex.scss
@@ -1,3 +1,40 @@
+$foundation-components: (
+	grid: true,
+	flex-grid: true,
+	typography: true,
+	forms: true,
+	button: true,
+	accordion: true,
+	accordion-menu: true,
+	badge: true,
+	breadcrumbs: true,
+	button-group: true,
+	menu: true,
+	menu-icon: true,
+	drilldown-menu: true,
+	dropdown: true,
+	dropdown-menu: true,
+	flex-video: true,
+	label: true,
+	media-object: true,
+	off-canvas: true,
+	orbit: true,
+	pagination: true,
+	progress-bar: true,
+	slider: true,
+	sticky: true,
+	reveal: true,
+	switch: true,
+	table: true,
+	tabs: true,
+	thumbnail: true,
+	title-bar: true,
+	tooltip: true,
+	top-bar: true,
+	visibility-classes: true,
+	float-classes: true
+);
+
 @import '../scss/foundation';
 
-@include foundation-everything($flex: true);
+@include foundation-components-include;
--- a/scss/util/_components.scss
+++ b/scss/util/_components.scss
@@ -1,0 +1,173 @@
+@function foundation_component_enabled($key: '') {
+	@if (not map-has-key($foundation-components, $key)) {
+		@return false;
+	} @else {
+		@return map-get($foundation-components, $key) ;
+	}
+}
+
+@function foundation_grid_enabled() {
+	@if foundation_component_enabled("flex-grid") {
+		@return false;
+	} @else {
+		@return foundation_component_enabled("grid");
+	}
+}
+
+@function foundation_flex_grid_enabled() {
+	@return foundation_component_enabled('flex-grid');
+}
+
+
+@mixin foundation-components-include() {
+  @include foundation-global-styles;
+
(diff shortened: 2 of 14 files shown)


kball:
Curious what @brettsmason @andycochran @ncoden and @zurbrandon think of this

assets/foundation-flex.scss
```diff
@@ -1,3 +1,40 @@
+$foundation-components: (
```
ncoden: Is it usefull to duplicate component enabling settings in `assets/*` and `scss/foundation.scss` ? We could provide a boolean setting like `$foundation-components-enable` which enable or disable all components by default.

scss/util/_components.scss
```diff
@@ -0,0 +1,164 @@
...
+		@return false;
+	}
+	
+	@return map-get($foundation-components, $key);
+}
+
+@mixin foundation-components-include() {
+  @include foundation-global-styles;
+
+  @if (foundation_component_enabled("grid")) {
```
ncoden: In France, we call this an "If forest".
Because we can't `@include` dynamically, the only way I see to avoid it is to declare the components by using a `@mixin` checking if the component is enabled/disabled, and directly generating the CSS.
Well, almost like here but split into the component files.

Also, it is not possible anymore to easily customize the component generation order, or add its own components/code between. I think it is an important possibility that we shouldn't drop.

ncoden (commented):
@andreamk @kball Great idea :+1:. I have only tree remarks

brettsmason:
Hmm so I'm not sure I fully understand the purpose of this.
My main concern is does this still allow you to use the mixins to include components wherever you like in your code? Or does this only effect the mixin that includes everything? I feel like this might add a layer of complexity thats generally not needed, but I may be completely missing the point, which is quite possible 😄

ncoden:
At https://github.com/zurb/foundation-sites/pull/8943, @andreamk gives good arguments for the controle of components in settings. However, we can't drop the possibility for the user to generate himself the components (so change the order, adding code between, generate some or all of them in a scope...).

So I think the win-win situation would be to:
- Make the `@mixin foundation` generating the components enabled/disabled in settings (see my other comments above)
- Make it clear that the components `@mixin foundation-*` can be directly called.

andreamk:
Hello 
You can see the problem generated by the current system with the variable $global-flexbox.
Having no information on active modules was necessary to introduce a variable to manage dependencies between modules.

With my system you can extend foundation with functions that depend on the active modules without introducing variables that can generate inconsistency errors.

ncoden:
@andreamk True. Edited ;)

DaSchTour:
I think this change introduces a very bad pitfall, as it is possible to include normal grid and flexbox grid at the same time. I also would also suggest to keep the old `$global-flexbox` setting and use it for the new settings, so that backward compatibility is kept for 6.3 at least.

kball:
So my sense based on the comment threads here is that this is an interesting direction but needs more work.  @andreamk do you want to take another pass/iteration on this?

andreamk:
Hello
sorry for the wait.

I restored $global-flexbox
`
$global-flexbox: foundation_component_enabled("flex-grid") !default;
`
It is true if flex-grid is enabled but you can assign a custom value

ncoden:
@andreamk Hi, sorry for not replying to you sooner.

Your branch has conflicts with `zurb:6.3`. Could you merge `zurb:6.3` in it, then squash ?
Also, `v6.3` has been released, so maybe you should open a new PR to merge into `v6.4`. (@kball ?)

kball:
Definitely target should now be 6.4.  I think this should probably be a part of the larger conversation about how we handle settings etc as in https://github.com/zurb/foundation-sites/issues/9492

ncoden:
@andreamk Do you still have time for this PR ?

DanielRuf:
Is this PR still relevant for 6.4 or 6.5 as it introduces some changes?

ncoden:
Hi @andreamk. As we started working on F7, we stopped working on any new complex features on F6, or others deep changes.

I like the objectives of this PR and the direction it gives to Foundation, but I have close it, we do not want to introduce deep architecture changes in v6.5. We'll try to consider this for F7. ;)

PR #9031, opened 2016-07-18
Fixes button groups radius #8447

imabdulsal:
@rafibomb 
This PR fixes the issue #8447, and replaces the PR #8448, however it contains additional code that fixes the unstacked scenario.

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -44,13 +44,21 @@ $buttongroup-expand-max: 6 !default;
     margin-#{$global-right}: $buttongroup-spacing;
     margin-bottom: $buttongroup-spacing;
     font-size: map-get($button-sizes, default);
+    border-radius: 0;
 
     @if $global-flexbox {
       flex: 0 0 auto;
     }
 
+    &:first-child {
+      border-top-#{$global-left}-radius: $global-radius;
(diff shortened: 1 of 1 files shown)


kball:
@andycochran @brettsmason @ncoden thoughts on this?

ncoden (approved):
:+1: Nothing to say

andycochran:
Still not convinced this is the right move. All my thoughts are in #8447. But I'm not saying this should be rolled back. I think it solves an issue for everyone that wants rounded corners on the whole button group. Just remember, that's a design decision which makes assumptions. If I want a set of buttons, each with its corners rounded, I'll just have to override this with app-specific styles.

DaSchTour:
I totaly agreey with @andycochran 
At the moment I use button groups with border-radius on every button.
So you can bet after I'll get the bug report from my project you'll see my PR changing this again 😆

kball:
Its in 6.3, so we still have time to adjust... @DaSchTour and @andycochran do you want to take another pass on this?  Can we get something that straightforwardly supports both usecases?

andycochran:
Could be as simple as: 
```sass
$buttongroup-radius-on-group: true !default;
```

andycochran:
@kball On it. 😉

PR #9032, opened 2016-07-18
Abstracting font-size calculation to its own mixin

Minimilk93:
This should help for re-use.

### DIFF
--- a/scss/typography/_base.scss
+++ b/scss/typography/_base.scss
@@ -303,15 +303,7 @@ $abbr-underline: 1px dotted $black !default;
   }
 
   // Heading sizes
-  @each $size, $headers in $header-sizes {
-    @include breakpoint($size) {
-      @each $header, $font-size in $headers {
-        #{$header} {
-          font-size: rem-calc($font-size);
-        }
-      }
-    }
-  }
+  @include typography-header-font-size;
 
   // Links
   a {
@@ -438,6 +430,29 @@ $abbr-underline: 1px dotted $black !default;
 
     @if has-value($keystroke-radius) {
       border-radius: $keystroke-radius;
+    }
+  }
+}
+
+/// Calculates the header font-size based on breakpoints.
+///
+/// @param {String} the size of the header to output.
+@mixin typography-header-font-size (
+  $requested-size
+) {
+  // Heading sizes
+  @each $size, $headers in $header-sizes {
+    @include breakpoint($size) {
+      @each $header, $font-size in $headers {
+        @if $requested-size == $size {
+          font-size: rem-calc($font-size);
+        }
+        @else {
+          #{$header} {
+            font-size: rem-calc($font-size);
+          }
+        }
+      }
     }
   }
 }


scss/typography/_base.scss
```diff
@@ -441,3 +433,26 @@ $abbr-underline: 1px dotted $black !default;
     }
   }
 }
+
+/// Calculates the header font-size based on breakpoints.
+///
+/// @param {String} the size of the header to output.
+@mixin typography-header-font-size (
+  $requested-size
+) {
```
kball: Not sure I'm fully understanding the intended behavior from requested-size... looks like we still iterate through all sizes and breakpoints and output them, so what exactly is the intent?

scss/typography/_base.scss
```diff
@@ -441,3 +433,26 @@ $abbr-underline: 1px dotted $black !default;
...
+        }
+        @else {
+          #{$header} {
+            font-size: rem-calc($font-size);
+          }
+        }
+      }
+    }
+  }
+}
```
ncoden: I don't like the way it put a mixin at the end of the file. If we use a mixin to make it reusable (which is a good practice), it is an *util* and should be in an *util typography file*. Please create `/scss/util/_typography.scss` and move the mixin to it.

scss/typography/_base.scss
```diff
@@ -441,3 +433,26 @@ $abbr-underline: 1px dotted $black !default;
...
+
+/// Calculates the header font-size based on breakpoints.
+///
+/// @param {String} the size of the header to output.
+@mixin typography-header-font-size (
+  $requested-size
+) {
+  // Heading sizes
+  @each $size, $headers in $header-sizes {
+    @include breakpoint($size) {
```
ncoden: This mixin does two things:
**1** - generate headers (or any other selectors) from a font map. It could probably be used by the users.
**2** - generate this set of headers (or any other selectors) for each breakpoint. It is used by Foundation, and maybe some users.

So I think this mixin should be divided in two mixin : `typography-font-sizes` (**1**) and `typography-bp-font-sizes` (**2**).

@kball What do you think ? Do you have better names ?

scss/typography/_base.scss
ncoden: @kball I think `$requested-size` is intended to generated the default font-size, like in `/scss/_global.scss`. I don't see other purposes. <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_twitter_64.png);background-position:80% 10%;background-size:4100%" title="thinking_face"></span></span>

@Shebs72 Can you give us an exemple where `requested-size` would be useful ? Otherwise, please remove it.

ncoden (changes requested):
Hi @Shebs72, thank you for the PR.

scss/typography/_base.scss
ncoden: @kball I think `$requested-size` is intended to generated the default font-size, like in `/scss/_global.scss`. I don't see other purposes. <span class="emoji-outer emoji-sizer">

@Shebs72 Can you give us an exemple where `requested-size` would be useful ? Otherwise, please remove it.

scss/typography/_base.scss
kball: I agree that this would serve better as two separate mixins. Especially for things that are not included in the final output css, I far prefer explicit to brief.

kball:
@Shebs72 are you going to be able to address the changes requested by @ncoden?

andycochran:
I'm going to this PR, as headers are handled by `$header-styles` as of v6.3. @Shebs72, take a look at #9419 and see if you have any suggestions on how font sizes are calculated there.

PR #9033, opened 2016-07-18
Feature/issue 8680 into v6.3

sethkane:
Various HTML, SCSS and JS updates to make the accordion menu works better with accessibility.

Key points:

Removed not needed events
Changed html to being more semantic
Added conditionals Foundation.Nest for proper roles
Added focus ring
Tested in:

IE 11 and JAWS
Firefox and NVDA
Mac and VoiceOver
iPhone and VoiceOver


Geczy:
Yess

Owlbertz:
Seems to be a bit messed up. Can you get rid of the conflicts?

sethkane:
@Owlbertz weird, I thought I did that already. I will try it again.

sethkane:
@Owlbertz  Updated and passed!

Geczy:
Can't wait for this merge!

Owlbertz:
@Geczy I already had a first glance at the PR, but I currently have not that much time to spend on that topic.
Are you able to review the changes and give me your feedback on code-level?

Geczy:
I took a quick look too, I"m not as familiar with the accessibility standards to comment on that aspect. 

I will note that this will be a breaking change to existing accordion menus. 

Also, there seems to be an unrelated change to the menus in this PR
https://github.com/zurb/foundation-sites/pull/9033/files#diff-77edfaa71bc712336c47b29b8cddc927R13

sethkane:
@Geczy The reason it is in v6.3 is due to all the accordion changes.  We know it will effect older version but that just has to happen to make it more accessible.

I also don't know why in my PR there were other changes maybe I branched/merged from the wrong branch.  Let me know if I need to redo my work for a cleaner PR.

Owlbertz:
@sethkane this can happen because the develop and the v6.3 branch are not always synced.
These issues can be avoided by branching your feature branch directly from v6.3, but since you already started developing on a branch branched from the develop branch there are now some more commits from other people. I think we can deal with that.

Regarding the changes: maybe you can summarize the changes you made if you got the time so we can use that as a little migration guide to v6.3?

@Geczy I think every thing helps, so if you can pull this down and just play a round with the result a bit and check for inconsistencies that would be great.

Geczy:
I will do that and report back.

sethkane:
I will be happy to write some upgrade notes.  I don't think there will be many because the HTML is the only thing that really changes and it isn't much.

Owlbertz:
Great!

@sethkane Some people say that Foundation is changing too rapidly, and I would really like to encounter this by having a solid guide to perform the upgrade. Also I hope that this will make it easier and thus encourage more people to take care of the accessibility of their site.

Geczy:
Gave this a spin

First issue I noticed is tabbing isn't quire working. It'll highlight the first block. But it won't go past this block either to the next parent block, or deeper into a child block. When I hit tab again it goes to the next screen (outside the demo menu). 

Expected behavior is to be able to tab each menu line item. 

image

Geczy:
I do have good news though. I realized if I use the keyboard to navigate, I'm able to go to each menu line item. 

NVDA is behaving as expected with the keyboard arrow keys. We just need to make each item tabbable now, I think.

Geczy:
https://www.w3.org/TR/wai-aria-practices/#accordion

Tab - When focus is on an accordion header, pressing the Tab key moves focus in the following manner:
If interactive glyphs or menus are present in the accordion header, focus moves to each in order.
When the corresponding panel is expanded (its aria-expanded state is 'true'), then focus moves to the first focusable element in the panel.
If the panel is collapsed (its aria-expanded state is 'false' or missing), OR, when the last interactive element of a panel is reached, the next Tab key press moves focus as follows:
If a subsequent accordion panel is already expanded, focus moves to the first focusable element in this subsequent panel.
If no subsequent accordion panel is expanded, focus moves to the first focusable element outside the accordion component.

sethkane:
@Geczy I like the idea of following the WC3 specification to the letter however not even the example that is presented by them follows the specifications to that degree.  http://www.oaa-accessibility.org/examplep/accordian1/

So I am happy to spend more time on this but i am afraid that it might be more work that what is actually needed.

Thoughts?

Geczy:
@sethkane Tab navigation is absolutely necessary, it is indeed odd that those examples of accordions don't follow the WC3 tab specs...

I'm coming from an accessibility standpoint at a major university, we get in a lot of trouble when things like this aren't followed to spec.

Though, what you've done now is leaps and bounds better than what we have,

Let me know if I can help you in any way

sethkane:
@Geczy Again not sure I totally agree 100% Do you have an example of one that works exactly how you want it to work?

Every good example out there leaves the navigation when inside of the widget to the arrow keys. Here are some other examples that follow what I have done vs. what the exact spec is.

https://dequeuniversity.com/library/aria/tabpanels-accordions/tabbed-accordion
http://jqueryui.com/accordion/ (via https://www.paciellogroup.com)
http://a11y.nicolas-hoffmann.net/accordion/

I am good with doing it to the exact spec I just won't have time this week.

sethkane:
@Geczy  One thing I would like to add before you disagree with me ;-)
If you run the current code in JAWS the verbal commands for the role="tree" are to use the arrows to navigate.  While I understand that NVDA doesn't have the same comments JAWS is the most widely used screen reader out there.  Just my $0.02

kball:
@sethkane given that this is changing behavior from 6.2 to 6.3, can you write a quick migration guide for migrating a 6.2 accordion to a 6.3 accordion? Lets put this directly in the accordion-menu docs page, in a section marked by a data-migration attribute.  E.g.:
```
<div data-migration="6.3">
  ...Migration guide
</div>
```

sethkane:
@kball I will be happy to do that sometime this week. I will keep you all posted.

sethkane:
Merged Develop in and updated merge conflicts.  Hopefully this is good.

sethkane:
@Owlbertz any idea why it still says I have a conflict?  I am pretty sure I resolved them in my local.

Owlbertz:
@sethkane I will check.

kball:
Hey @Owlbertz and @sethkane not sure what happened here, but I just cleaned it up as a new PR here: https://github.com/zurb/foundation-sites/pull/9283

Going to close this PR in favor of that one.

PR #9036, opened 2016-07-19
Commenting out console.log in Reveal

okofish:
This appears to be left over from someone's debugging, but it's present in the current distribution of 6.2.3 and logs "focus" to console whenever Motion UI is used to animate-in a Reveal.


Owlbertz:
Better remove that!

PR #9006, opened 2016-07-07
Fix assignment to undeclared variable

bsa01:
Before submitting a pull request, make sure it's targeting the right branch:
- For documentation fixes, use `master`.
- For bug fixes, use `develop`.
- For new features, use the branch for the next minor version, which will be formatted `v6.x`.

If you're fixing a JavaScript issue, it would help to create a new test case under the folder `test/visual/` that recreates the issue and show's that it's been fixed. Run `npm test` to compile the testing folder.

Happy coding! :)

Fix ReferenceError: assignment to undeclared variable in strict mode

PR #9037, opened 2016-07-19
[FIX] offsetPct undefined

a2nt:
offsetPct variable was not defined


Owlbertz:
Already addressed in #8948 and #9006.

kball:
Fixed in #8948

PR #9039, opened 2016-07-21
Added .top-bar-title strong as shown in docs

### LINKED ISSUE #9029: v6.2.3 - .top-bar-title strong is missing styling
I'm not sure if this was intended or not but I have spotted a slight styling issue when it comes to `.top-bar-title strong`. I copied and pasted the code from the documentation (Top Bar, Advanced Layout) into a codepen so you can see the issue with the title.

[http://codepen.io/JLT93/pen/ZOvdpG](http://codepen.io/JLT93/pen/ZOvdpG)

Shouldn't there already be appropriate styling for the whole title bar?

mattferderer:
This is a request/suggestion to fix #9029 & enhance #7907 by adding styling to the strong element to match the suggested usage in the documentation for the top-bar-title class in [Top Bar - Advanced](http://foundation.zurb.com/sites/docs/top-bar.html#advanced-layout)


colin-marshall:
@mattferderer your pull request uses the class selector `.strong` but I think you meant to use the element `strong`?

mattferderer:
@colin-marshall you're correct. The PR has been fixed. Hopefully my quick fix using GitHub's editor worked.

kball:
@colin-marshall @andycochran @brettsmason @JeremyEnglert This seem reasonable to you?

brettsmason:
@kball I'm not entirely sure the need for this if I'm honest - should we really be forcing the use of `<strong>` rather than letting it be whatever element you like?

Also I don't think its a good idea having the `$menu-item-padding` used directly - I remember Geoff mentioning not to have dependancies on other components [in this conversation](https://github.com/zurb/foundation-sites/pull/7944) so I think we would need a need a new variable if this went ahead.

mattferderer:
@brettsmason  I believe the original request or need for this was so that someone could include Foundation in their project, copy & paste code directly from the Foundation documentation & expect it to work the same as it does in the documentation. I feel that either the styles or the documentation need to change. 

Geoff makes an excellent point in the conversation mentioned. I can make a dedicated variable if requested & if this specific pull request seems like the best method to go forward.

brettsmason:
@mattferderer Thanks for coming back to me and confirming the use case. I understand a bit more now about the purpose.

I'm wondering however if it would be better to create another class, eg `top-bar-title-text` and style this appropriately. Maybe add a few other variables like font size etc. That way users aren't restricted to a particular tag (for example I use a <h1> on home and <p> on sub pages on my sites).

What do you think? If you and others think its overkill though I'm happy to go with it 😄

mattferderer:
I really like the idea of using a class named `top-bar-title-text` more than using the strong element. Maybe I'm a purist but I don't like adding a bunch of padding or margin styling to an element like strong in cases such as this one. I also think that class name does a good job of describing the element & purpose of it. 

I'm not sure if this creates an issue breaking old code or not. I would assume not, since you had to add additional styling outside of Foundation to make this work like the documentation anyways.

kball:
Mmm... I like the direction of having the styling based on class, not element.  This ties into the conversation @ncoden and I were having on https://github.com/zurb/foundation-sites/pull/8718#issuecomment-251820827

ncoden:
Using tags is never a good idea. It look simple at first, but lead often to *totally unsolvable properly* problems like overwrite by an arbitrary hight specificity.

So yes, if you're adding elements, please only use classes. It's a first step. We should later remove all no-reset properties on tags.

mattferderer:
I was messing around with the editor in Chrome. It looks like it might currently be possible to adjust the `top-bar-title` class margin-right property to be a margin property. The scss/settings/_settings.scss file would also need to have the setting changed in there though. 

If this is an acceptable solution, I can look into it closer & make another pull request.

kball:
@mattferderer that approach seems reasonable to me.

kball:
@ncoden @andycochran @brettsmason what do you think of where this PR ended up?

andycochran:
👍 This makes a lot of sense to me. 

(I often mess with the margins here when I add logos and custom fonts. It'll be nice to be able to do so in the settings.)

brettsmason:
I agree with @andycochran - makes sense now.

mattferderer:
I committed a bit too soon it appears. I still should remove the unnecessary sass file in the docs & test the flex part of this code.

ncoden:
Seems great. Ready to merge ?

mattferderer:
Let me test this a bit closer & make one more commit.

mattferderer:
I think I botched that merge conflict... Suggestions?

kball:
@mattferderer if you can list out the commit names you want included, I can pull this down, cherry-pick them out, and create a new branch/PR for us to look at

mattferderer:
32e1956b6dc8a8765d3408cf52a7ad7d76f9492d  (this was the merge conflict with the docs.scss file)
eefd44670dc96bc6e51215c502dec00cb9bd131a
5359b779ebf49ad065c500113b41baf8f99666e8
c256c46e5a7e94b99c368d16f4bd99fcf27906a9
167767a5f3ef86b88a76ee54e1e199d686daa8e0
9373dc01b232cd434d09cb926e8fb8a4f37d1dd8

If you want, I can create a new branch too. Still trying to get the hang of pull requests with Github.

kball:
@mattferderer probably easiest would be if you can create a new branch, but if you run into trouble I'll take a pass on it

mattferderer:
@kball  do you want me to create a new PR as well then & reference this one?

kball:
@mattferderer yes please!  Closing this one in favor of that.

PR #9041, opened 2016-07-22
Fix dropdown alternative anchor selection

jeromelebleu:
According to the dropdown JavaScript source code, the anchor could be defined either with `data-toggle` or `data-open`. The later will actually never be selected due to `$(`[data-toggle="${$id}"]`)` which will always be true - even if no object is found. Note that I took this syntax from the reveal component - see [here](https://github.com/zurb/foundation-sites/blob/master/js/foundation.reveal.js#L47).


kball:
👍

PR #9048, opened 2016-07-25
Add settings for the hover color for thead and tfoot

natewiebe13:
This fixes a bug where hovering the header or footer with a modified background color, would change the background color to be the same as the body.

### DIFF
--- a/scss/components/_table.scss
+++ b/scss/components/_table.scss
@@ -48,9 +48,17 @@ $table-stripe: even !default;
 /// @type Color
 $table-head-background: smart-scale($table-background, $table-color-scale / 2) !default;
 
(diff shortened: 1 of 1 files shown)


natewiebe13:
Reopened to rebuild.

andycochran:
Should this maybe use the `smart-scale` function instead of `darken`? In some cases, `lighten` might be preferable.

natewiebe13:
I used `darken` to be consistent with the existing functionality. If requested, I can change it to use `smart-scale`.

andycochran:
Ooh, my mistake; `$table-row-hover` does default to darken. Apologies. This looks good to me. 👍

Issue #8866, opened 2016-05-30
[Foundation for Sites] Error in grid-context mixin


ikemarrus:
**How can we reproduce this bug?**
In version 6.2.1
1.set a grid context of 10 (or another number other than the default) with nested scss that calls grid-column to set up columns - the column width is correctly calculated based on 10 columns
2. Outside that nested scss, call grid-column to set up columns. look at how widths for columns are being calculated - they are still being calculated with a grid context of 10 instead of the default grid context (e.g. 12)

I looked at the code for grid-context:

 $old-grid-column-count: $grid-column-count;
    $grid-column-count: $columns !global;
    @if $root {
        @at-root {
            @content;
        }
    }
    @else {
        @content;
    }
    // Restore the old column count
    $grid-column-count: $old-grid-column-count; 

Note that $grid-column-count is specified as a global variable when it's being set, but as a local variable when it's being reset. When I changed the reset to also be global, the grid-context was restored correctly

andycochran:
Is this still an issue in v6.2.3?

Baedda:
@andycochran  yes it is. I dont know why, but there was merged a pull request in december 2015: 
https://github.com/zurb/foundation-sites/commit/56d819af094a41e20c21726224a16cbd005ce668

In my opionion this pull request is responsible for the _false_ behaviour. Take a look to what the description of the pr says:
https://github.com/zurb/foundation-sites/pull/7538
The old behaviour was correct. Using the grid-row-mixin should **not** effect the **default** $grid-column-count.  I think this change has do be reverted. Otherwise the default-variable doesnt make any sense. If you have a default grid with 10 columns and switch only one row to have e.g. 15 columns then every following row will be calculated with 15 instead of 10 columns. 

You can fix this on your side by using the grid-contetxt-mixin at each use but this is really annoying :/
Hope there will be a fix soon.

andycochran:
@Baedda Do we just need to add `!global` back into that variable?

Baedda:
@andycochran in my opinion: yes.
Take a look ... i did a small test case.

``` scss
$grid-column-count: 10;

.row-a {
    @include grid-row;
    .half-width-column {
        @include grid-column(5);
        @debug $grid-column-count;
    }
}

.row-b {
    @include grid-row(12) {
        .half-width-column {
            @include grid-column(5);
            @debug $grid-column-count;
        }
    }
}

.row-c {
    @include grid-row;
    .half-width-column {
        @include grid-column(5);
[...]

PR #9055, opened 2016-07-27
Fix $grid-column-count beeing overwritten by grid-context-mixin

Baedda:
Commit fixes issue described in issue: https://github.com/zurb/foundation-sites/issues/8866


andycochran:
@kball You okay with this PR? @brettsmason and I both looked over #8866 and think this it the right fix.

kball:
Sounds good, lets do it.

PR #8860, opened 2016-05-27
Modify cite element to use a variable for the pseudo content

nessthehero:
Adds an override variable to specify the cite element's pseudo ::before content.

This pull request resolves #8845

PR #9058, opened 2016-07-28
Modify cite element to use a variable for the pseudo content

### LINKED ISSUE #8845: Add setting variable for cite ::before content
Can the content for the ::before for the cite element in typography/_base.scss be specified with a variable, so that we can set an override?

Yes, I can just change it easily in my CSS, but it'd be nice to just be able to set a variable and have the default style be what I want from the start.

nessthehero:
Adds an override variable to specify the cite element's pseudo ::before content.

This pull request resolves #8845

There is also an older pull request of this change in #8860


andycochran:
I like this idea. 👍

andycochran:
@kball can you kick Travis? No reason this PR should cause a failure.

kball:
@andycochran kicked into gear

Issue #9060, opened 2016-07-29
[Tabs] Missing key ARIA role="tablist"


gerardkcohen:
Tab container is missing key parent `role="tablist"` which may produce compatibility issues with various assistive technology/ browser combinations.

As specified in the official WAI-ARIA Authoring Practices for [Tab Panel](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel):

> A tablist is the container role for a set of elements with the role attribute set to tab.

This can be seen on current [Tabs documentation](http://foundation.zurb.com/sites/docs/tabs.html) by inspecting any Tab `<ul>` container.

abdullahsalem:
@kball Should this be closed?

kball:
yes. Fix will be in 6.3

PR #9061, opened 2016-07-29
[Tabs] Missing key ARIA role="tablist"  #9060

gerardkcohen:
This PR addresses issue #9060 by adding `role="tablist"` to Tab `<ul>` containers, as documented in the WAI-ARIA Authoring Practice for [Tab Panels](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)

cc: @rafibomb @gakimball

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -42,6 +42,7 @@ class Tabs {
   _init() {
     var _this = this;
 
+    this.$element.attr({'role': 'tablist'});
     this.$tabTitles = this.$element.find(`.${this.options.linkClass}`);
     this.$tabContent = $(`[data-tabs-content="${this.$element[0].id}"]`);
 
@@ -94,10 +95,10 @@ class Tabs {
     this._addKeyHandler();
     this._addClickHandler();
     this._setHeightMqHandler = null;
-    
+
     if (this.options.matchHeight) {
       this._setHeightMqHandler = this._setHeight.bind(this);
-      
+
       $(window).on('changed.zf.mediaquery', this._setHeightMqHandler);
     }
   }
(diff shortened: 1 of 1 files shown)


Owlbertz:
@sethkane I think this makes sense?

sethkane:
@Owlbertz yes you should have role="tablist" on the <ul> and you should have role="tab" on the <li>
http://oaa-accessibility.org/example/34/

gerardkcohen:
It's ok to keep the role of `tab` on the anchor as long as the role of `presentation` stays on the `<li>`. The requirement is that `tablist` contain only `tab` elements, and `presentation` will remove semantics from the `<li>`

Owlbertz:
Alright, I guess this can be merged then. I think this can go into ` develop` as it's only changes the ARIA-attributes that are added automatically, although it might of course break the design if someone made up some CSS rules that target those attributes - but imo this should be a rare case.

gerardkcohen:
@Owlbertz Should not have any impact as this is only adding an attribute that looks like was never there.

Owlbertz:
@kball, maybe you can check and give feedback?

Silvester20:
Thanks for this PR! Is this PR already included v6.2.4? (Because the W3 validator says the tabrole is missing on my site)

Owlbertz:
@Silvester20 it should be part of 6.2.4. Can you provide a link to your site?

Silvester20:
@Owlbertz  Thanks for the quick reply. I don't have a domain so I made a [fiddle](https://jsfiddle.net/zs2z7s38/4/).

Owlbertz:
@Silvester20 you are right, the change is not part of 6.2.4.
[This line](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.tabs.js#L45) is missing in [the release version](https://github.com/zurb/foundation-sites/blob/v6.2.4/js/foundation.tabs.js#L44).

This is strange, because the PR got merged on October 17th and the [release tag](https://github.com/zurb/foundation-sites/releases) has been created on October 21st.

Maybe @kball knows what's up?

kball:
Sorry guys, this didn't quite make it in for 6.2.4... it was merged after we'd branched off the first RC.  It will be in 6.3, the first RC will be in 2 weeks.

PR #9066, opened 2016-07-30
Fix magellan `_updateActive()` behavior. Fixes #8960

### LINKED ISSUE #8960: Magellan doesn't work as expected when the links are not in the same order as the targets
### Steps to reproduce:

1. Make a page with several sections, each with `id` and `data-magellan-target` specified.
2. Make a nav element with `data-magellan` containing `<a>` links to the `id`s, but don't put them in the same order as the sections
3. Scroll

### Expectation:

The correct link should get the `active` class.

### Reality:

The link corresponding to the order of the sections got the `active` class instead.

mark-tan:
When the links are not in the same order as the targets, or a link exists without a corresponding target, or several links with the same `href` are in a magellan container, the wrong link may get the .active class.

A visual test has been added for this case.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -139,7 +139,7 @@ class Magellan {
     }
 
     this.$active.removeClass(this.options.activeClass);
-    this.$active = this.$links.eq(curIdx).addClass(this.options.activeClass);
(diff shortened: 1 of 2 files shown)


Owlbertz:
Thanks, this seems to work out.

PR #9070, opened 2016-07-30
Fixed Magellan issue when target does not exist.

Owlbertz:
Prevent any scrolling action when the target of the clicked link does not exist.
Previously the `.getOffset().top` created an JS error because `Cannot read property 'top' of undefined`.
Also added a basic test for scrolling and a test for the graceful failing when the target does not exist.

Noticed this issue when checking #9066.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -103,6 +103,8 @@ class Magellan {
    * @function
    */
   scrollToLoc(loc) {
+    // Do nothing if target does not exist to prevent errors
+    if (!$(loc).length) {return false;}
     var scrollPos = Math.round($(loc).offset().top - this.options.threshold / 2 - this.options.barOffset);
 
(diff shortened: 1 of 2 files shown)


Owlbertz:
Any feedback from @zurb/yetinauts ?

coreysyms:
This looks good, would we want a console log stating there is "no target, please check existence or spelling". Just a thought. Otherwise, merge it!

Owlbertz:
@coreysyms Not sure if there is any Foundation-wide guideline on console logs. Maybe @kball can help?

kball:
This looks good; I've fought with exactly this bug before.

On console.logs... historically we don't include them in the codebase.  I think @coreysyms has a good idea... I wonder if we should have a set of debug only logs that get compiled out in the production builds... for now, lets not worry about it.  Going to merge both this and #9066.

Issue #9072, opened 2016-08-01
[DropdownMenu] Invalid ARIA usage


gerardkcohen:
DropdownMenu widget, following the [Menu](https://www.w3.org/TR/wai-aria-practices-1.1/#menu) ARIA pattern, uses an incorrect property of `aria-expanded` on menu items. [aria-expanded](https://www.w3.org/TR/wai-aria-1.1/#aria-expanded) is not supported on the role of [menuitem](https://www.w3.org/TR/wai-aria-1.1/#menuitem).

There is also an unnecessary `aria-hidden` on inactive menuitem's. The `aria-hidden` is not needed since `display:none` is being used to hide menu items, which enough to hide from assistive technologies.

Lastly, `tabindex` of 0 is being added to the first menu item but is unnecessary because the anchor tag used is already focusable.

brettsmason:
This has been fixed in PR 9073 which was merged.

PR #9073, opened 2016-08-01
[DropdownMenu] Invalid ARIA usage #9072

gerardkcohen:
This PR addresses issue #9072 with the following:
1. Removed `aria-expanded` on menuitems's in dropdownMenu.js and nest.js. This state is unsupported with `role="menuitem"` and may cause incompatibility issues with assistive technologies. Changes to nest.js will affect DropdownMenu, DrilldownMenu, and AccordionMenu.
2. Removed unnecessary/ redundant `aria-hidden` on menuitems in dropdownMenu.js and nest.js. `display:none` is enough to hide from assistive technologies. This is just a cleanup change, and reduces surface area for future defects.
3. nest.js adds `tabindex="0"` to the first menu item, but this is not necessary since the `<a>` tag is already natively focusable. This is just a cleanup change, and reduces surface area for future defects.
4. Added a missing `href` in the DropdownMenu docs
## Recommendation

Examples show using an anchor (`<a>`) tag for each `menuitem` but the expected child of `menuitem` should just be a text node. Ideally, the `<a>` would not be required or at least marked with `role="presentation` except the intent/ usage of this widget is unclear to me. Based on the docs, it looks this like _could_ be used as an actual anchor (Item 3 and Item 4). This can also be seen with the main doc navigation at the upper right, which I believe is coming from the nest.js util. These are, in fact, navigational items but a `menubar` widget is a little overkill here, as simple navigation `<ul>` would suffice. I attempted to address this but without any context I did not want to break anything.

cc: @rafibomb @gakimball

### DIFF
--- a/docs/pages/dropdown-menu.md
+++ b/docs/pages/dropdown-menu.md
@@ -41,7 +41,7 @@ To create dropdown menus, nest a new `<ul>` inside an `<li>`. You can nest furth
 
 <ul class="dropdown menu" data-dropdown-menu>
   <li>
-    <a>Item 1</a>
+    <a href="#Item-1">Item 1</a>
     <ul class="menu">
       <li><a href="#Item-1A">Item 1A</a></li>
       <li>
@@ -97,7 +97,7 @@ Add the `.vertical` class to the top-level menu to make it vertical. Sub-menus a
 
 <ul class="vertical dropdown menu" data-dropdown-menu style="max-width: 300px;">
   <li>
-    <a>Item 1</a>
+    <a href="#Item-1">Item 1</a>
     <ul class="menu">
       <li><a href="#Item-1A">Item 1A</a></li>
       <li>
@@ -138,13 +138,13 @@ Add the `.vertical` class to the top-level menu to make it vertical. Sub-menus a
 
 ## Sticky Navigation
 
-See the documentation for the [Sticky](sticky.html#sticky-navigation) plugin to see how to easily make a sticky nav bar. 
+See the documentation for the [Sticky](sticky.html#sticky-navigation) plugin to see how to easily make a sticky nav bar.
 
 ---
 
 ### Preventing FOUC
 
-Before the JavaScript on your page loads, the dropdown menus will not have arrows. However, once the JavaScript file has loaded, the arrows will appear causing a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content). You can prevent this by adding the `.is-dropdown-submenu-parent` class manually. 
+Before the JavaScript on your page loads, the dropdown menus will not have arrows. However, once the JavaScript file has loaded, the arrows will appear causing a [flash of unstyled content](https://en.wikipedia.org/wiki/Flash_of_unstyled_content). You can prevent this by adding the `.is-dropdown-submenu-parent` class manually.
 
 ```html
 <ul class="dropdown menu" data-dropdown-menu>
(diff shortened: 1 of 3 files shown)


Owlbertz:
Hi @gerardkcohen, thanks for one more a11y improving PR!
Is this fully backwards compatible?

gerardkcohen:
@Owlbertz In all honesty, I am not sure. I don't have enough experience using Foundation to know. Maybe we can work through that together? I can tell you, I am not sure about the last [commit](https://github.com/zurb/foundation-sites/pull/9073/commits/60538ddcc07e6a79dfae681955d80029082a1989) as I think leaving the `href` might be intentional. It just did not feel right when I tested with VoiceOver/ Safari.

Owlbertz:
Well I guess the important thing regarding backwards compatibility here is if your change requires any HTML changes.
Regarding your `href`, taking into account what @sethkane did [in this line of code](https://github.com/zurb/foundation-sites/pull/9033/commits/ebfa2b7841ee5915cc25e05eb853e908040fa8b8#diff-efbe0bc06492f62fff77f7569ad2971dR40) in #9033, he might have some feedback if using an `a` as a toggle would be suitable at all?

gerardkcohen:
@Owlbertz In that case, then the important changes do not require any changes to markup. I am ok with leaving out the change to the DropdownMenu docs.

sethkane:
@Owlbertz @gerardkcohen it is my belief to use as much semantic markup as possible. I removed all the `<a>` because they no longer act or behave like anchors in the Accordion Menus.  By switching out the tag you don't need to use any additional markup like a role or aria tag to make augment it's behavior and speech announcements with screen readers. There are a thousand examples out there that with the combination of `<a>`  and the right aria tags work well.  I personally like this one:  http://oaa-accessibility.org/example/25/ which doesn't.

I am happy to pull this branch down and test it when time permits.

gerardkcohen:
@sethkane Agreed, but I don't have any context as to the intent of the markup, .e.g progressive enhancement, usage/ options, history, etc so I understand the need to keep breaking changes to a minimum, hence my "recommendation" above. My only intent was to bring consistency with the other items in the same example. It didn't make sense to recommend, via example code, that the first item be a readonly anchor and then force `tabindex="0"` via the JS to make it focusable, with some of the anchors being toggle triggers and others looking like actual navigational links. For sure, lines are blurry with the intent of these examples.

kball:
Hi @gerardkcohen @sethkane @Owlbertz - what is the state of this PR?

gerardkcohen:
@kball IMO this is complete as is.

kball:
Testing this out (having merged in our various updates), this causes one of the Drilldown unit tests that @Owlbertz wrote to fail.  The failing test is here: https://github.com/zurb/foundation-sites/blob/develop/test/javascript/components/drilldown.js#L67

My understanding is that the 'hidden' was taken out of nest because for dropdowns display:none was sufficient, but for drilldowns they are hidden by translation outside of a container that has overflow hidden, so I would guess that they still need to have aria-hidden.

I think our options are to restore the hidden/unhidden attributes on submenus in nest or add some logic based on menu type... given this was nominally a cleanup change, I lean towards the former with a comment added about why it is needed.

kball:
@gerardkcohen Can you take a look at this issue turned up above and advise on course of action?

gerardkcohen:
@kball No problem. So to understand, there is difference in implementation with nest and the way menu's are hidden? If that is the case, I too would agree some logic based on type might be the best route. I assume that is something that I will be able to do based on CSS classes, e.g. the drilldown class will require the `aria-hidden`?

kball:
@gerardkcohen I think the Feather and Burn methods get the type passed into them, so that should be enough to add the logic.

kball:
I want this in for 6.3 so going to go ahead and do updates mentioned above in comments above myself

kball:
Updated and merged!

PR #9076, opened 2016-08-02
[Reveal] Fix iOS bug where opening a modal would cause the page to jump to top.

jonathanmelville:
When opening a Reveal modal on iOS Safari, the page jumps the top of the screen while the modal is open, and jumps back to it's original position when the modal is closed. This is corrected by changing `height` to `min-height` for the iOS-only class `html.is-reveal-open, html.is-reveal-open body`.


kball:
.is-reveal-open is applied regardless of browser. That said, with a little additional JS work I believe this is an improvement, and doesn't appear to break behavior on other browsers.  Going to merge it and we'll QA thoroughly for the release.

PR #9077, opened 2016-08-02
[Bugfix] Fix Centering Mixins on Safari 8

jessedoyle:
Hello!

We noticed that the `absolute-center` SASS mixin wasn't working properly on Safari 8.

After looking into the problem, it looks like Safari 8 only supports the `transform` CSS3 attribute through the `-webkit` prefix (see [here](http://caniuse.com/#feat=transforms2d)).

This PR adds the appropriate vendored-transform properties to the `vertical-center`, `horizontal-center` and `absolute-center` mixins.

Here's some before and after screenshots from browserstack on Safari 8 (using `absolute-center`):
###### Before

safari-8
###### After

safari-8-fixed

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -196,6 +196,7 @@
   position: absolute;
   top: 50%;
   transform: translateY(-50%);
+  -webkit-transform: translateY(-50%);
 }
 
 /// Horizontally centers the element inside of its first non-static parent,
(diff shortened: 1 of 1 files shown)


natewiebe13:
I would use something like [https://github.com/postcss/autoprefixer](url) instead of prefixing in the source.

andycochran:
Hmm… We're already using [Autoprefixer](http://foundation.zurb.com/sites/docs/sass.html#autoprefixer-required). Shouldn't need browser prefixes in the SASS.

kball:
We've seen some issues where the current zurb stack autoprefixer config isn't catching ios8.  I think we should add an explicit callout for `ios > 8` to the gulp files for both the templates.

andycochran:
That seems like the right solution. I'll go ahead and close this. @kball is that autoprefixer config bug documented in an issue?

PR #9084, opened 2016-08-04
[BugFix] Sticky plugin max-width calculation is insufficient

JerBear90:
The calculation between `newElemWidth` and `pdng` seems to be insufficient.
The max-width should be calculated by `newElemWidth - (pdng * 2)`

This PR addresses the following Issue:
Sticky plugin makes div larger #8427
https://github.com/zurb/foundation-sites/issues/8427

I have provided a test at test/visual/sticky/adjust-max-width


kball:
Lets try 2 things to make sure this is a general fix:

1. Remove the implementation of sticky within the test page; it should get your new version automatically by including the dist/foundation.js

2. Build a testcase with padding-left and padding-right being different to make sure the doubling works generally, or if we should instead be explicitly pulling out padding-left and padding-right and adding them.

ncoden:
Hi @Jflem1290. Can you rebase your work ?

poke @kball

kball:
There's a lot of conflicts in here and I'm no longer sure what was being fixed in this.  Based on the title though, I suspect this is addressed in https://github.com/zurb/foundation-sites/pull/9298

Going to close for now.  @Jflem1290 please feel free to reopen with a rebased update if you think there is something being lost not addressed in #9298

PR #9087, opened 2016-08-06
focus state and input-transition

pine3ree:
maybe it's simpler to just import (@extend) the global form-element

### DIFF
--- a/scss/forms/_select.scss
+++ b/scss/forms/_select.scss
@@ -43,6 +43,22 @@ $select-radius: $global-radius !default;
     padding-#{$global-right}: ($form-spacing * 1.5);
   }
 
+  @if has-value($input-transition) {
+    transition: $input-transition;
+  }
+
+  // Focus state
+  &:focus {
+    border: $input-border-focus;
+    background-color: $input-background-focus;
+    outline: none;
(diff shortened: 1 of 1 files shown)


ncoden:
Hi @pine3ree. Thanks for your PR.

*The next time, please provide a better PR description.*

`form-element` can't be extended for this component, due to unwanted properties (like `display: block;`. But because a select input *is* a form component and currently has a lot of properties in common, we should envisage a refactor of *at least* the form components.

Poke @kball, Foundation uses almost always mixins, and very few times extends. It leads to a lot of 	
unnecessary code duplication. *ITCSS* in a powerfull sass approach (https://github.com/zurb/foundation-sites/pull/8718#issuecomment-251820827) should help to define core abstract components (objects) and easily extend them without CSS order problems.

Nicolas

ncoden (approved):

kball:
Seems reasonable to me - is there a filed issue associated with this?

pine3ree:
@ncoden , @kball ,

yes sorry guys. It seemed obvious to me at the time but just because i was working on it.
The $input-transition was not applied to `select` elements, only to "text" elements (see _text.scss).
This pr meant to extend input transition it to selects.
kind regards

PR #9090, opened 2016-08-07
vertically center left-right dropdown arrows

pine3ree:
height is 10px (2*5px = 2xtriangle-border). To center vertically top: 50%, margin-top: height/2 => triangle border size 5px

what about parameterizing the triangle size (now 5px hardcoded ) as well to allow custom arrow sizes via configuration?

### DIFF
--- a/scss/components/_dropdown-menu.scss
+++ b/scss/components/_dropdown-menu.scss
@@ -34,7 +34,7 @@ $dropdownmenu-border-width: nth($dropdownmenu-border, 1);
 @mixin left-right-arrows {
   > a::after {
     #{$global-right}: 14px;
-    margin-top: -3px;
+    margin-top: -5px;
   }
 
   &.opens-left > a::after {


rafibomb:
I agree it should be vertically centered with math if the `<li>'s are going to change height. If that's the use case you are p[proposing, that would be the best method or using the vertical-center mixin.

pine3ree:
@rafibomb 
Yes, that's the purpose. Even in the vanilly foundation the left-right carets are positioned a little lower of the middle of their &lt;a&gt; (the a is position: relative; as well). You can verify the correct positioning  by changing the triangle-height via _settings.

something similar could perhaps be done for up-down carets too.

kind regards

pine3ree:
@rafibomb 
for up-down carets the height is === border width, so we need to move up by half the border to get correct positioning (we are positioning a box that is twice the border width high, but the triangle only use the upper half of the box, so we need to move it down by half the border size again which means just move up by half the border width).

kball:
Going to merge as is, but could well be worth parameterizing.  @pine3ree do you want to put together a PR replacing hard-coded sizes?

pine3ree:
@kball 
I will take a look in the next days (it's past midnight here now).
I remember i already parametrized the ddm-arrow size in my projects including the correction for the vertical indicator as explained in a previous comment and I used the same arrow size for all menus drilldown, dropdown, accordion)
kind regards

pine3ree:
@kball 
i forgot....we could also use 
```css
transform: translateY(-50%);
```
instead of a negative margin-top

kball:
Yeah I think setting position to 50% and translate back -50% is going to be more portable

PR #9094, opened 2016-08-09
appendTo feature for reveal

phifa:


js/foundation.reveal.js
```diff
@@ -85,7 +85,7 @@ class Reveal {
   _makeOverlay(id) {
     var $overlay = $('<div></div>')
                     .addClass('reveal-overlay')
-                    .appendTo('body');
+                    .appendTo('this.options.appendTo');
```
Owlbertz: In order to have the code aligned I would like to see ` .appendTo($(this.options.appendTo));

Owlbertz (approved):
Seems like a useful feature to me.
Just one small think I commented on a line.
I would like to have feedback from another @zurb/yetinauts on that since it changes the public API.

js/foundation.reveal.js
kball: I agree, putting it in the quotes won't work, because jquery will try to find that string, rather than the variable.

kball: In fact, I think I need to change this to a review because without this change this PR is not correct.

kball (changes requested):
We need to change the overlay append to properly reference the appendTo variable rather than a string.

kball:
I'm going to just do the update for the bug in place and get this merged

PR #9111, opened 2016-08-15
Set trapFocus default for OffCanvas to be true. Matches current docum…

### LINKED ISSUE #8957: [Off-canvas] `data-trap-focus` is not true by default
In the [docs for Off-canvas](http://foundation.zurb.com/sites/docs/off-canvas.html#js-options) it says that `data-trap-focus` is true by default, but [it's not](https://github.com/zurb/foundation-sites/blob/v6.2.3/js/foundation.offcanvas.js#L392).

rspraymond:
Fixes #8957.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -389,7 +389,7 @@ OffCanvas.defaults = {
    * @option
    * @example true
    */
-  trapFocus: false
+  trapFocus: true
 }
 
 // Window exports


rspraymond:
closing and reopening to skip travis build failure

rspraymond:
Trying build again...

kball:
Unless there's a good argument why this should be the default, I'd prefer we update the documentation to match the behavior rather than the other way around, to avoid breaking existing projects on update.

PR #9117, opened 2016-08-16
custom button and hollow button hover lightness settings

pine3ree:
I just didn't like the default -50% lightness for hollow buttons, maybe alowing further customization via settings could be a better solution for users


kball:
This seems reasonable to me.

PR #9118, opened 2016-08-16
drilldown height calculation

pine3ree:
Calculate min-height by iterating through all menus and submenus and building each menu/submenu height with the sum of its li outerHeight, then take the max height of each menu.

multiplying max li number x 1st li height does not take into account variable/multirows li height.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -299,13 +299,16 @@ class Drilldown {
    * @private
    */
   _getMaxDims() {
-    var max = 0, result = {};
-    this.$submenus.add(this.$element).each(function(){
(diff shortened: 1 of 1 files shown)


pine3ree:
It would be nice to have auto-height for every menu level as in f5, though.... :-)

pine3ree:
I've just found this another way to solve the same issue: https://github.com/zurb/foundation-sites/pull/8320.

my idea was to check for every list element in all (nested) submenus, take the sum of them to get each submenu height and take the highest value.

kball:
Closing this as #8320 was identified as an alternative, and it has been merged.

PR #9119, opened 2016-08-17
Add data-option-closeable to accordion menu. If set to "false", parent links of open submenus will be followed by clicking on.

relthyg:


kball:
I"m not sure what happened but this PR appears to have a bunch of extraneous stuff in it. Please resubmit a clean PR against v6.3.

PR #9120, opened 2016-08-17
Dropdown Menu keyboard accessibility

dsturley:
This fixes two accessibility issues with dropdown menu:
1.  When escape is pressed, focus returns to the last menuitem in the menubar.  It should go to the top level menuitem which contains the opened menu.
2.  Makes down arrow wrap to first element.  This matches the behavior of up arrow.  For accessibility, the menus should either wrap both ways (e.g. down on last item goes to first; up on first goes to last) or not at all.

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -163,7 +163,7 @@ class DropdownMenu {
         open: openSub,
         close: function() {
           _this._hide(_this.$element);
-          _this.$menuItems.find('a:first').focus(); // focus to first element
+          _this.$tabs.has(e.target).children('a').focus(); // focus to first element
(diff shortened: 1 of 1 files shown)


Owlbertz:
Hi @dsturley, I agree with you you on your second point. However, I feel that for the first point there is room for interpretation. Can you reference any resource that documents the desired behavior?

kball:
ping @dsturley - can you respond to @Owlbertz 's question?  Would love to get this confirmed and merged for 6.3.

kball:
@Owlbertz @dsturley what is the status of this PR?

Owlbertz:
For me the reason for @dsturley 's first point is still not clear...

PR #9125, opened 2016-08-18
configurable pagination current item on mobile

pine3ree:
Let the current page indicator be visible on small screens. via configuartion

### DIFF
--- a/scss/components/_pagination.scss
+++ b/scss/components/_pagination.scss
@@ -50,9 +50,14 @@ $pagination-item-color-disabled: $medium-gray !default;
 /// @type Color
 $pagination-ellipsis-color: $black !default;
 
-/// If `false`, don't display page number links on mobile, only next/previous links.
+/// If `false`, don't display page number links on mobile, only next/previous links
+/// and optionally current page number.
 /// @type Boolean
 $pagination-mobile-items: false !default;
+
+/// If `false`, don't display current page number on mobile.
+/// @type Boolean
+$pagination-mobile-current-item: false !default;
 
 /// If `true`, arrows are added to the next and previous links of pagination.
 /// @type Boolean
@@ -79,6 +84,12 @@ $pagination-arrows: true !default;
       &:last-child,
       &:first-child {
         display: inline-block;
+      }
+
+      @if $pagination-mobile-current-item {
+        &.current {
+          display: inline-block;
+        }
       }
 
       @include breakpoint(medium) {
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -436,6 +436,7 @@ $pagination-item-color-current: foreground($pagination-item-background-current);
 $pagination-item-color-disabled: $medium-gray;
 $pagination-ellipsis-color: $black;
 $pagination-mobile-items: false;
+$pagination-mobile-current-item: false;
 $pagination-arrows: true;
 
 // 27. Progress Bar


kball:
Interesting... so the use case for this is that you want to be able to show current page even if there's no way for them to page through?  Presumably because you're letting them swipe or something similar?

pine3ree:
Hello @kball, 
the main purpose is to send a bit of information to the visitor (which page you are on).
This could be useful when the number of the page has meaning, like product ordering by price.
But....you can discard this pr at once if you think it's not worth it...it's just what i needed to do for a customer and so i decided to share the idea.

kind regards

kball:
@andycochran @brettsmason curious what you guys think on this

brettsmason:
So if I'm understanding this correctly, this allows the current page number to show on mobiles/small screens?

Eg: Previous  Current Page No.  Next

If so that makes sense to me and is probably useful for the user in most cases.

pine3ree:
Hello @brettsmason ,
Yes, just like that. kind regards.

kball:
Oh got it - somehow I was thinking the prev/next were also covered by $pagination-mobile-items and it wasn't making sense to me to have just the current number.  Now I get it.  Makes sense, thanks!

andycochran:
❤️ Smart PR.

scss/components/_pagination.scss
```diff
@@ -50,10 +50,15 @@ $pagination-item-color-disabled: $medium-gray !default;
...
 
-/// If `false`, don't display page number links on mobile, only next/previous links.
+/// If `false`, don't display page number links on mobile, only next/previous links
+/// and optionally current page number.
 /// @type Boolean
 $pagination-mobile-items: false !default;
 
+/// If `false`, don't display current page number on mobile.
+/// @type Boolean
+$pagination-mobile-current-item: false !default;
```
DaSchTour: Shouldn't this be defaulted to true? This way it's an unnecessery breaking change 😟

pine3ree:
Hello @DaSchTour,
the feature was not available in previous versions, so i decided to use a default value of false to keep the current behaviour by default.
kind regards

DaSchTour:
Well yes, I added this feature to display the numbers in mobile. But I added it the way, that users that already use pagination don't have to change anything after updating, because the old behavior is used.
With your addition of `$pagination-mobile-current-item` everybody with `$pagination-mobile-items` set to `true` will have to also set `$pagination-mobile-current-item` to `true` to gain the behavior of previous version. This means, that after upgrading and without changing the configuration the behavior changes. So everybody is forced to update the configuration to have the same with as with the version before.

DaSchTour:
Okay, well now I got it. The description and naming is very confusing. It guess it should be made clear, that the options are disjunctive. Saying "If `false`, don't display current page number on mobile." looks like. Set it to false and the current page number isn't displayed independent of other settings. So my idea was, that setting `$pagination-mobile-current-item: false` and `$pagination-mobile-items: true` will display all page numbers but not the current one. It's really confusing.

pine3ree:
When i re-read the code and parameter comment i felt the same, so I'll issue a PR to make the comment clearer.
Thank You,
maks

DaSchTour:
@pine3ree thanks. Good to see that I'm not the only one getting confused

pine3ree:
@DaSchTour , I added the new parameter and wrote the comment myself....I totally agree with you that it can be confusing... :-)

Issue #8684, opened 2016-04-27
[Equalizer] wrong height


pavelblazek:
**How can we reproduce this bug?**
1. I'm using Drupal CMS
2. have two divs - first with text, second with image with data-interchange
3. seems it takes longer to load the image so the height is set by first (smaller) div with text
4. the height of div with img is wrong - img overflows
5. if I use just img without the data-interchange the height is set correctly
6. the only difference is if I use the data-interchange on img

**What did you expect to happen?**
Same height - I guess load the image first and than set the height. Might be caused by the Drupal
**What happened instead?**
The height is not set right - img overflows
**Test case:**
tested on CodePen - works fine

Give us a link to a CodePen or JSFiddle that recreates the issue.
- [CodePen with Foundation 6 and MotionUI](http://codepen.io/rafibomb/pen/xVVGOB)
- [CodePen with Foundation 6, Flexbox grid and MotionUI](http://codepen.io/rafibomb/pen/jqqPra)

denisahac:
Can you resubmit the correct CodePen/JSFiddle links.

pavelblazek:
http://codepen.io/pavelblazek/pen/YqObeW

It sets the height correctly when you resize the window, but not when it's loaded for the first time

designerno1:
check it without interchange = works
Issue related to interchange img, equalizer check for loaded img, interchange (img without src) `img.complete = true`
maybe adding a case to the foundation onImagesLoaded function?

jsilver951:
This is how I addressed this on a recent project. 

`	
$('*[data-equalizer] img').on('load', function() {
    new Foundation.Equalizer($("body"));
    // console.log("Equalizer after image loads");
});
`

Jflem1290:
@jsilver951 - I found that the similar solution worked for me when using  `$(window).on("load", function(){})`. This way everything on the page gets initialized / loaded. 

Ref: Initializing
  http://foundation.zurb.com/sites/docs/javascript.html
   Although in the docs() under JavaScript it states that "$(document).foundation(); will kick off every Foundation plugin at once. " be sure to include a new instance of [Equalizer](http://foundation.zurb.com/sites/docs/equalizer.html).
` var elem = new Foundation.Equalizer(element, options);`

Reference from jQuery [document.ready](https://learn.jquery.com/using-jquery-core/document-ready/) - 
    Initializing the code below will run once the entire page is ready and has been loaded. I.e our images and text. 

    $(window).on("load", function() {
[...]

Issue #8556, opened 2016-04-08
Hidden .tabs-panel with equalizer


JTallis:
Equalizer sets all heights to 0 when it's in a `.tabs-panel` that is hidden.

I'm certain that this doesn't happen in F5 as it was working fine until upgrading to F6 and changing the HTML so that it matches that of F6.

Here is an: [example](http://codepen.io/anon/pen/groKXO)

Click the Equalizer tab and you will see that all heights will be 0 until you resize the browser and only then will it re-equalize and be the right heights.

Is this a bug? Does anyone have any suggestions on how to fix this?

designerno1:
quick and dirty fix you can try something like this
```js
$('[data-tabs]').on('change.zf.tabs', function(e, $target) {
  var $tabsEq = $('[data-tabs-content=' + $target.parent('[data-tabs]').attr('id') + '] > .tabs-panel:eq(' + $target.index() + ') > [data-equalizer]');
  if ($tabsEq.length) {
    $tabsEq.data('zfPlugin')._reflow();
  };
});
```

andycochran:
Just works if you add this: 

```scss 
.tabs-panel {
  display: block;
  height: 0;
  overflow: hidden;
  padding: 0;
}

.tabs-panel.is-active {
  height: auto;
  padding: 1rem;
}
```
Maybe `display:none` is the wrong way to go for tabs?

madhanan:
Can someone solve this ??

JTallis:
I had added a quick fix to the one and only page where I use both components together. designerno1's fix is likely to be more efficient as it only targets the selected content element's children to equalize.

This was my fix:

```javascript
$("[data-tabs]").on("change.zf.tabs", function() {
    Foundation.reInit('equalizer');
});
```

Someone should definitely come up with a permanent fix for this, if one hasn't already been added. I haven't upgraded my Foundation to the latest yet. I'm still on 6.2.1.

andycochran:
@zurb/yetinauts Does anyone have concerns with me switching Tabs to a method other than `display:none`? A fix like I mentioned above would solve the height issues with nested components.

t22james:
@andycochran I specifically came to git now to suggest this, so glad you are already considering it. 

[...]

Issue #8955, opened 2016-06-22
Foundation 6 Orbit carousel shrinking in Chrome after page refresh


rita-patel:
Hi

I am using Foundation 6 starter project to create carousel using Orbit. I have realised that on page refresh in chrome the carousel squeezes. I checked if the same is happening on the example given on Foundation Zurb (http://foundation.zurb.com/sites/docs/orbit.html) and the same happens on there too. 

So if you go to http://foundation.zurb.com/sites/docs/orbit.html where the Orbit example is
Once page is loaded in Chrome refresh page using F5 and the carousel shrinks most of the time.
I was told this is happening on Android however it is happening on Desktop too after refreshing page.

Is this a known bug and is there a fix for this please?

sharkyl:
I'm having the same issue. Anyone have a fix?

timdj:
Same problem here. Seems to have to do with cache. When images are loaded from cache it sometimes miscalculates the height. If I have dev-tools open it works fine.

timdj:
#https://github.com/zurb/foundation-sites/pull/8478

kball:
Should be fixed by #9272

PR #9126, opened 2016-08-18
updating triggers for mutation listeners, and equalizer to use them

coreysyms:
Adding this as a "bug fix / feature" to develop because if it works will fix a lot of issues.
EX: https://github.com/zurb/foundation-sites/issues/8684 or https://github.com/zurb/foundation-sites/issues/8556

This is the implementation of mutation in addition to scroll and resize
triggers for universal plugin usage. Foundation can now fire events
when elements that are listening for mutation events or their children
change in the DOM either by element addition, element removal, or
attribute manipulation.

This can be very useful for ajax caused changes, display none on load
issues, element manipulation, and any other DOM mutation events.

To see what I mean, I added in the mutation trigger for Equalizer, see how it interacts with interchange, and a "display:none" div. And added in a fire event for Tabs as well.

This is what Foundation currently does with interchange and equalizer, and Tabs with JS modified content. (not so nice)
http://tangerineindustries.com/Foundation/feature-mutation-trigger/without_mut.html

This is with the new mutation trigger (everything lays out proper)
http://tangerineindustries.com/Foundation/feature-mutation-trigger/with_mut.html

### DIFF
--- a/js/foundation.equalizer.js
+++ b/js/foundation.equalizer.js
@@ -36,6 +36,7 @@ class Equalizer {
 
     this.$watched = $watched.length ? $watched : this.$element.find('[data-equalizer-watch]');
     this.$element.attr('data-resize', (eqId || Foundation.GetYoDigits(6, 'eq')));
+	this.$element.attr('data-mutate', (eqId || Foundation.GetYoDigits(6, 'eq')));
 
     this.hasNested = this.$element.find('[data-equalizer]').length > 0;
     this.isNested = this.$element.parentsUntil(document.body, '[data-equalizer]').length > 0;
@@ -71,6 +72,7 @@ class Equalizer {
     this.$element.off({
       '.zf.equalizer': this._bindHandler.onPostEqualizedBound,
       'resizeme.zf.trigger': this._bindHandler.onResizeMeBound
+	  'mutateme.zf.trigger': this._bindHandler.onResizeMeBound
     });
   }
 
@@ -101,6 +103,7 @@ class Equalizer {
       this.$element.on('postequalized.zf.equalizer', this._bindHandler.onPostEqualizedBound);
     }else{
       this.$element.on('resizeme.zf.trigger', this._bindHandler.onResizeMeBound);
+	  this.$element.on('mutateme.zf.trigger', this._bindHandler.onResizeMeBound);
     }
     this.isOn = true;
   }
--- a/js/foundation.equalizer.js
+++ b/js/foundation.equalizer.js
@@ -71,7 +71,7 @@ class Equalizer {
     this.isOn = false;
     this.$element.off({
       '.zf.equalizer': this._bindHandler.onPostEqualizedBound,
-      'resizeme.zf.trigger': this._bindHandler.onResizeMeBound
+      'resizeme.zf.trigger': this._bindHandler.onResizeMeBound,
 	  'mutateme.zf.trigger': this._bindHandler.onResizeMeBound
     });
   }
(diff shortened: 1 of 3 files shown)


coreysyms:
@zurb/yetinauts Bump, I could really use others testing to see if this breaks anything.

kball:
@coreysyms this is super cool.  It took me a while and some conversation with you to get what its doing, but I think this approach will enable some amazing 'automatic', and performant responsiveness to data changes.  We can build upon this for all sorts of automatic ajax content fetching, etc.

@Owlbertz and @colin-marshall you should take a look and try this out.

We need to add some testing & documentation, but I think this will be amazing.

Lets target this for 6.2.5 or 6.3.

kball:
Would this be a natural way to address https://github.com/zurb/foundation-sites/issues/8955 as well?

coreysyms:
> Would this be a natural way to address #8955 as well?
No. The issue seems to be with Orbit calculating images height on the **initial page load**, whether it be a primed cache or not Orbit can't seem to consistently calculate the image height. This listener is method is intended for elements that were added, removed, or manipulated **after load**. 

Now if the user utilized Interchange for the images in the Orbit carousel, Interchange would fire a mutate event because it removes and or adds images to the DOM after initial page load, so if Orbit had a mutate listener, it could fire a re-calc.

Also for what it's worth, I can't replicate that issue, tried windows 7, 10 mac, nothing.

kball:
@coreysyms @Owlbertz I'm super excited about the potential for this... what do we need to get it past the line?

Owlbertz:
Looks good to me on the first glance.

kball:
@coreysyms testing this out... THIS IS AWESOME! :)   One thing I notice is that the mutate observers don't appear to be polyfilled back for IE9/IE10 in the same way scroll & resize were... does it make sense to do so?  Or is the performance impact too large?

coreysyms:
@kball 

> mutate observers don't appear to be polyfilled back for IE9/IE10 in the same way scroll & resize were...

The mutation observer for scroll and resize is effectively replacing the for loop, it is meant as a more efficient way to trigger the scroll or resize event. (we are cleverly using a mutation observer)

The mutation observer for mutation events is, well, what it is, and no fallback is needed... which brings us to the fork in the road of IE9/10. There really isn't anything that we can do in this instance for those browsers. However, since it is a trigger event, it is non-destructive, and IE9/10 can continue to exist in its terrible environment. 

_I do realize the confusion of using a mutation observer to observer mutation among other things, I have to re-read my code and what I'm typing as to not confuse myself let alone others._

kball:
@coreysyms I think we can merge this then... I tested it, and other than the aforementioned ie9/10, it all is looking good. I think reveal is another potential target, and there's a ton of possible applications.

Do you have any reasons this is not ready to merge?

coreysyms:
I agree, Reveal would be greatly complemented by this. I feel good about this PR. I'm good for merge!

Corey Snyder

coreysyms:
Perhaps we should roadmap the plugins to update for usage? I've touched tabs. Interchange and equalizer in this PR. 

The good news is. It shouldn't take long as the plugin just needs to "register" and hit whatever function it already has to reflow itself. 

Accordion. Would be another plugin that "display none" effects. 

So once this is merged. We can start leveraging it. 

Corey Snyder

coreysyms:
@kball just made a quick change to wrap the mutate listener in an `if` so IE9/10 does not throw an error in the console window. I'm ready to let it run wild.

coreysyms:
odd, not sure what the problem is with Tabs and Travis

kball:
Found it.  Was a legit bug (missing var), but I fixed it.  Merged!

PR #9127, opened 2016-08-19
fix flex-video docs w3c validation

pine3ree:
- frameborder is obsolete and not valid in htnl5
- webkitAllowFullScreen and mozAllowFullScreen are not valid attributes


kball:
Looks good.  This appears to be fine for all supported browsers, so merging.

PR #9132, opened 2016-08-23
Reload focusableElments at every keydown to fix issue #9130

### LINKED ISSUE #9130: [Reveal] 'Tab' key does not work for forms loaded via ajax in reveal modals
1. Setup a reveal modal to load a form via ajax on 'open.zf.reveal'
2. Once open, click the first input in the form and then hit 'Tab'

Tab should have taken you to the next field in the form.

**What happened instead?**
Tab key does not work

**Test case:**
http://codepen.io/sujayjaju/pen/akAYzP

samuelmc:
Fixes #9130 
Reload focusableElments at every keydown for the tab key to work properly when the content inside the reveal changes after opening.

Also avoid dropdown closing upon click of an element not in the DOM like in pr #9146

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -249,7 +249,7 @@ class Dropdown {
             if(_this.$anchor.is(e.target) || _this.$anchor.find(e.target).length) {
               return;
             }
-            if(_this.$element.find(e.target).length) {
+            if(_this.$element.find(e.target).length || !$(document).find(e.target).length) {
               return;
             }
             _this.close();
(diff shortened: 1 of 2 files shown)


Owlbertz:
Hi @samuelmc, please make sure that this PR only addresses a single issue so we can better decide what PR to merge.

Also on your fix, while it may be necessary that the focusable elements are calculated again for modifications with AJAX, I feel like calculating them on every keydown is bad for the performance. Maybe there is another way to fix the issue?

samuelmc:
Thanks for the comment @Owlbertz, I'll create new branches for these fixes.

PR #9135, opened 2016-08-24
Anchor link fix for example links

TravisHi:
Changed &lt;a href="#"  &gt; to &lt;a href="#0" &gt; throughout page, to prevent infuriating snap to top behaviour of &lt;a href="#"    &gt;. (Solution thanks to Chris Coyier)


kball:
Cool, I hadn't seen that trick before!  If you want to apply this to other docs pages as well that would be great.

PR #9139, opened 2016-08-29
check if cb is a function

### LINKED ISSUE #8801: [Sticky] "cb is not a function" on _calc run
Hi, I have found bug of init _calc on sticky plugin for recalc sticky position.

example of invoke function:
```
$('.sticky').foundation('_calc', true);
```
then in console appear an error:
"Unhandled promise rejection TypeError: cb is not a function"

problem is in sticky plugin code of "_calc" function:
https://github.com/zurb/foundation-sites/blob/develop/js/foundation.sticky.js#L162
function _setSizes is init without callback function and _setSizes function try run missing callback founction which is "Undefined".

Please help, to correct function an argument in _calc or add check for callback is it a function in setSizes.

pine3ree:
and fixes https://github.com/zurb/foundation-sites/issues/8801

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -267,7 +267,9 @@ class Sticky {
    */
   _setSizes(cb) {
     this.canStick = Foundation.MediaQuery.atLeast(this.options.stickyOn);
-    if (!this.canStick) { cb(); }
+    if (!this.canStick) {
+      if (cb && typeof cb === 'function') { cb(); }
+    }
     var _this = this,
         newElemWidth = this.$container[0].getBoundingClientRect().width,
         comp = window.getComputedStyle(this.$container[0]),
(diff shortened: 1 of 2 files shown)


Owlbertz (approved):
Seems reasonable.

kball:
It doesn't make sense to me that this would fix #8801... `_calc` expects an argument that is a boolean, and then calls `_setSizes` without a callback.  `_setSizes` checks for existence of callback, it doesn't in any place get called with a boolean.

Also, as far as I can see Timer should never be called without a callback... so not sure what updating that to have this check helps.

As a note, I'm also currently unable to replicate #8801, so not sure what that bug is targeting.

I'm going to close this as unnecessary; if you have a counterexample of a case requiring this please reopen with that example.

pine3ree:
Hello @kball,

here's a codepen

http://codepen.io/anon/pen/xEAEdp

steps to reproduce:

1. start in full width (desktop) browser
2. scroll down a few time to activate sticky
3. resize the browser window to a fondation small width (menu is drilldown)
4. scrollup and click one of the drilldown links, you will get the console error 
`foundation.js:7889 Uncaught TypeError: cb is not a function`

using latest chrome and opera

PR #9141, opened 2016-08-30
orbit: add/trigger event beforeslidechange

pine3ree:
I use this event to smoothly resize the orbit container for carousels having images with different heights.
the callback will receive 2 arguments: the current slide and the next slide.

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -268,6 +268,8 @@ class Orbit {
     }
 
     if ($newSlide.length) {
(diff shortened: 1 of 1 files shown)


kball:
Looks good.  This is an event that makes a ton of sense for Orbit, regardless of the resize issue.

We're talking about doing a lifecycle/event standardization and audit across all of our plugins; @coreysyms when you get a first pass on that lets loop in @pine3ree

PR #9143, opened 2016-08-30
topbar-title-padding

pine3ree:
Add the same vertical padding and line-height as the ones used for topbar menus in order to achieve a centered vertical alignment of the top-bar-title element in the default foundation setup.

### DIFF
--- a/scss/components/_top-bar.scss
+++ b/scss/components/_top-bar.scss
@@ -22,6 +22,10 @@ $topbar-submenu-background: $topbar-background !default;
 /// @type Number
 $topbar-title-spacing: 1rem !default;
 
+/// Padding for the top bar title.
+/// @type Number
+$topbar-title-padding: 0.7rem 0 !default;
+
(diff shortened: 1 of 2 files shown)


pine3ree:
closing and reopening pr to trigger rebuild after travis `npm install phantomjs `failure

pine3ree:
I've just found this PR
https://github.com/zurb/foundation-sites/pull/9039
which fixes the same style, but i believe we'd better keep menu and topbar-title padding independent but equal by default. Furthermore by default topbar-title does not need an horizontal padding.

kball:
Since this is changing default behavior on existing pieces, it should target 6.3 instead of develop... #9039 is a bit different in my mind because its adding a behavior on a subpiece.

All that said, would be interested in @andycochran @brettsmason and @JeremyEnglert's thoughts on the desirability of this behavior.

kball:
#9039 turned into #9316, but I think it resolves this same issue.  Given #9316 was merged, can we close this @pine3ree

PR #8001, opened 2016-01-25
Added Motion-UI support to ResponsiveToggle #7940

DominikPieper:
Add support for Motion-UI to the ResponsiveToggle component, using data-animate=".." to configure the animations.

PR #9145, opened 2016-08-31
responsiveToggle slideToggle animation

pine3ree:
new js/data-attribute options:
- data-slide-toggle="true" (default: false)
- data-slide-speed="300"  (default: 250 as in accordionMenu)

### DIFF
--- a/js/foundation.responsiveToggle.js
+++ b/js/foundation.responsiveToggle.js
@@ -84,13 +84,20 @@ class ResponsiveToggle {
    */
   toggleMenu() {   
     if (!Foundation.MediaQuery.atLeast(this.options.hideFor)) {
-      this.$targetMenu.toggle(0);
-
-      /**
-       * Fires when the element attached to the tab bar toggles.
-       * @event ResponsiveToggle#toggled
-       */
-      this.$element.trigger('toggled.zf.responsiveToggle');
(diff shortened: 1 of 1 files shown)


pine3ree:
triggering new travis-ci build for the usual npm install issue....

pine3ree:
this could maybe be merged also into 6.3 (see https://github.com/zurb/foundation-sites/issues/8001) as motion-ui slide-in-down slide-out-up does not actually slide down/up the toggled menu in the "classic" meaning, it just moves dow/up the menu in a fixed height containing top-bar.

ncoden:
Can we make these attributes more generic, as a way to control appear/disappear events for all utilities or components, instead of for this particular case ?

@kball Who are the Js pro Yetinauts ?

kball:
Great question @ncoden.  @colin-marshall @Owlbertz would love your thoughts on how we can make this more generic

DanielRuf:
Mh, can we not reuse Motion UI (classes) and use just classes which are bound to transitions / animations?

Triggering 1 event or 2 events makes much sense sense (toggle vs toggled) but the animation should be probably applied using a callback, class names or using the triggered events imo.

DanielRuf:
Closing as we did not receive further feedback and there are conflicts.

PR #9149, opened 2016-08-31
remove redundant else

pine3ree:
expressions like:

``` js
function func() {
    if (a) {
        //...
        return x;
    } else if (b) {
        //...
        return y;
    } else {
        //...
        return z;
    }
}
```

can be safely rewritten as

``` js
function func() {
    if (a) {
        //...
        return x;
    }

    if (b) {
        //...
        return y;
    }

    //...

    return z;
}
```

question 1:
why in line 215 inside the transitionend method the local variable `end` is being assigned to the value of a timer?

question 2:
inside parseValue should we just test for equality with string literals 'false', 'true'...otherwise a string like 'untrue' would be converted to the boolean true.

kind regards

### DIFF
--- a/js/foundation.core.js
+++ b/js/foundation.core.js
@@ -209,14 +209,16 @@ var Foundation = {
         end = transitions[t];
       }
     }
+    
     if(end){
       return end;
-    }else{
-      end = setTimeout(function(){
-        $elem.triggerHandler('transitionend', [$elem]);
-      }, 1);
-      return 'transitionend';
     }
+    
+    end = setTimeout(function(){
+      $elem.triggerHandler('transitionend', [$elem]);
+    }, 1);
+      
+    return 'transitionend';
   }
 };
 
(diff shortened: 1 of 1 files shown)


kball:
On question 1:  No idea, doesn't seem like we need it.

Question 2:  great question... string literals is a good idea, because otherwise anywhere someone is passing in text content we could have a bug if we have a longer sentence containing true or false.

For the expression transformation, I'd be curious to know if you have any reason for stripping the elses beyond them being unnecessary.  Stylistically I think we're inconsistent on this, but lacking additional information I'd err on the side of being explicit.

pine3ree:
Hello @kball,
it could be seen just as a trend, but i have found that early returns inside conditionals and inner loops end in clearer/cleaner code. But you can treat this just as a styling proposal and kill this PR at once ! :-)
kind regards.

kball:
Given this doesn't match the rest of the code styling, going to kill this.  However would be interested in a PR updating to string leterals

pine3ree:
@kball 
here it is https://github.com/zurb/foundation-sites/pull/9279
kind regards
the second commit encountered a travis error but it seems unrelated (i only added 3 spaces after 3 if)

> 78 passing (2s)
  1 failing
  1) Reveal "after each" hook:
     Uncaught TypeError: 'null' is not an object (evaluating '$element.find') (file:///home/travis/build/zurb/foundation-sites/_build/assets/js/foundation.js:673)

maybe I'll close and reopen to trigger a new travis check

kind regards

kball:
The error is unrelated, digging into it.

pine3ree:
@kball 
in the 1st build yes.
the 2nd (phatntom-js install) is unrelated too, but it's not uncommon in foundation-sites travis tests. I (and other devs) encountered it a few times.

kind regards

Issue #9103, opened 2016-08-13
ORBIT carousel inside toggable element and dropdown menu


pine3ree:
The orbit carousel works fine inside a dropdown pane,(dropdown plugin)  but doesn't work inside a dropdown (initially hidden) menu item (dropdown menu plugin); images disappear and/or bullets/navs clicks have no effect. If the menu item is open while images are loading the carouselt works fine until is closed and re-opened and something else is clicked in between.

I discovered this issue while building a fancy mega menu.

ps
the orbit placed inside the dropdown-pane continues to work after toggling it several times.

kind regards

pine3ree:
updates: when using the opacity property to show/hide the submen the carousel works fine. Using display (none|block) - the default framework behaviour -  breaks the carousel.

pine3ree:
update:
when the orbit or any outer element is toggled (display: none) the orbit stop working (not always at the first toggle).

pine3ree:
update:
the dropdownMenu/drilldownMenu/accordion plugins sometimes utilize jQuery.find('.is-active') and remove the class 'is-active' from all the matching elements, including orbit-slide elements inside menu list item elements. This causes styling issues and js issues (the changeSlide method expects to find a least 1 'is-active' orbit-slide element)

pine3ree:
Disabling mui makes the orbit slider work even after beiing toggled,
see: https://jsfiddle.net/akp40a46/2/

pine3ree:
the bug does not depend on timing function:

see [jsfiddle example no-autoplay](https://jsfiddle.net/akp40a46/5/)

the next-slide button works as well as the other controls until the next button is clicked while the carousel is hidden (via the toggle button).

In my tests I added debugging output and found out that when changeSlide is called while the orbit is hidden current/next slides are left with mui related classes on.

pine3ree:
getting closer:
it appears that after calling changeSlide when the orbit is hidden the foundation.motion finish() method is not called. The transitionend event is triggered, though....

pine3ree:
got it:
i've found that the mui finish() method is never called for elements for which jQuery is(':hidden') is true.
The quicker fix for the orbit carousel would be early return within changeSlide() for those cases.

https://github.com/zurb/foundation-sites/pull/9152

pine3ree:
update: motion-ui related issue when orbit is `:hidden` is now fixed by https://github.com/zurb/foundation-sites/issues/9311

kball:
[...]

PR #9152, opened 2016-09-01
stop changeSlide if carousel is hidden

pine3ree:
fixes the 2nd part of https://github.com/zurb/foundation-sites/issues/9103

motion-ui finish method is not called if the element is hidden (or inside hidden elements).
Therefore if changeSlide() is called for hidden carousels the "is-active" class attribute will be set for the next/chosen slide but mui-transition classes will be not removed from current/next slide. Further changeSlide() calls will therefore exit early as mui-classes are found in the current  lide (= the previous transition's next/chosen slide).

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -248,7 +248,9 @@ class Orbit {
   */
   changeSlide(isLTR, chosenSlide, idx) {
     var $curSlide = this.$slides.filter('.is-active').eq(0);
-
+    
+    if (this.$element.is(':hidden')) { return false; } // do not changeSlide if the element is not visible because mui fisnish event will not be called leaving mui classes on in current and next slides
+    
     if (/mui/g.test($curSlide[0].className)) { return false; } //if the slide is currently animating, kick out of the function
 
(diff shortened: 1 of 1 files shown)


kball:
Hmm... so I think I understand the problem this is trying to fix, but I'm not sure this is the right fix.

My understanding is that running the full transition breaks if the carousel is hidden, leaving the carousel in a broken state, so this prevents that by preventing a transition if the carousel is hidden.

However, it seems plausible to me that someone would want to want to change the slide to set it to something while it was hidden before showing it again.  Can we instead address this by in the hidden case changing the active slide without going through the transitions?

pine3ree:
That would be event better. My fix was the quickest idea that came to my mind to prevent the broken motion-ui transition while being hidden, but I believe that a definitve fix should address the motion-ui lib.
Your idea of just stopping the transition, leaving the ability of setting the active slide is in my opinion a step further, so, i agree, it would be better than my quick fix.
kind regards

pine3ree:
closing in favor of THE appropriate fix https://github.com/zurb/foundation-sites/issues/9311

PR #9159, opened 2016-09-06
allow easier mega-menu like items

pine3ree:
I often use `.menu.[size]-[num]-up li.column` to build mega menus. This small change will allow more complex mega menus for dropdown list items instead of equally divided boxes only, without redefining all grid column classes ad hoc.

### DIFF
--- a/scss/components/_dropdown-menu.scss
+++ b/scss/components/_dropdown-menu.scss
@@ -207,7 +207,7 @@ $dropdownmenu-border-width: nth($dropdownmenu-border, 1);
       }
     }
 
-    > li {
+    > li:not('.column') {
       width: 100%;
     }
 


kball:
This makes sense to me.  @andycochran @brettsmason @JeremyEnglert as our SCSS experts, any objections?

scss/components/_dropdown-menu.scss
```diff
@@ -207,7 +207,7 @@ $dropdownmenu-border-width: nth($dropdownmenu-border, 1);
       }
     }
 
-    > li {
+    > li:not('.column') {
       width: 100%;
```
ncoden: - Why do we automatically apply this property on all `<li>` in the first place, instead of using a class ?
- Why do we declare again every menu-related properties inside a `dropdown-menu` component, instead of simplifying `dropdown` and `menu` ?

Also, like [here](https://github.com/zurb/foundation-sites/pull/9098#pullrequestreview-3417021), we can't assume the behavior the user expect by the classes he uses. Filtering `.column` fix this particular case and nothing more, can make weird behavior, and lead to an unmaintainable spaghetti code.

pine3ree:
@ncoden 
Hello,
I am fine with adding an extra class for this but, since i am not adding an feature, just preventing the reset of the width to 100%, we would need to use it in the same way with a `not` selector, otherwise the menu selector wins.

I used the `.column` class because when i use it for building mega menus it's its own properties being reset, but i recognize that someone else could need it for other purposes. 

kind regards

andycochran:
Yeah, this won't work for semantic classes. 😞

kball:
Based on @ncoden and @andycochran 's feedback, I think this is probably not the right approach and we should close this.  It sounds like the right approach is to rework the approach to dropdown menu to not make assumptions based on selector.

pine3ree:
@kball 
I agree...it was a shurtcut I use to avoid adding kbytes of css for this simple goal. The approach you are suggesting is very welcome.
kind regards

PR #9165, opened 2016-09-08
This fix #9122 by changing z-index for the tooltip.

### LINKED ISSUE #9122: Tooltip doesn't appear inside Reveal

1. Create a Reveal modal window
2. Create a tooltip inside the modal window
3. hover on the tooltip

I expect the tooltip to appear when hovering inside the modal window.

What happened instead was the tooltip didn't appear.

Link to a CodePen:

- [Foundation 6 Reveal Tooltip bug](https://codepen.io/ableslayer/pen/yJwkWq)

isaak-ordonez:
I changed the z-index to fix the issue with the tooltip inside the reveal. 

Could you review if the fix is good? @abdullahsalem

### DIFF
--- a/scss/components/_tooltip.scss
+++ b/scss/components/_tooltip.scss
@@ -56,7 +56,7 @@ $tooltip-radius: $global-radius !default;
   font-size: $tooltip-font-size;
   padding: $tooltip-padding;
   position: absolute;
-  z-index: 10;
+  z-index: 1006;
   top: calc(100% + #{$tooltip-pip-height});
   max-width: 10rem !important;
   border-radius: $tooltip-radius;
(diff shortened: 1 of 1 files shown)


abdullahsalem:
Thanks @isaak-ordonez for this PR.
Setting up the `z-index` of the **Tooltip** to `1006`  will work fine. However, in your PR, both **Tooltip** and **Reveal** elements have the same `z-index` order -`1006`-, so from **[Stacking Context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context)**  perspective, the **Tooltip** will overlap because it comes after the **Reveal** element in the HTML hierarchy. _Therefore_, I would suggest that **Tooltip** takes a greater number than **Reveal** to avoid any unexpected scenario in the **Stacking Context** process.

isaak-ordonez:
@abdullahsalem  Thanks for the feedback, I made a change to the z-index from 1006 to 1200.

abdullahsalem:
Thanks, @isaak-ordonez !

PR #9166, opened 2016-09-09
Avoid duplicate reveal-overlays

DaSchTour:
the _makeOverlay function is not save to create a new overlay although one already exists.
As id is already passed to the function using it to make it unique.

BTW: I couldn't run tests because _estraverse-fb_ is broken. May it be, that there is something missing in package.json in 6.3 branch?

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -83,10 +83,10 @@ class Reveal {
    * @private
    */
   _makeOverlay(id) {
-    var $overlay = $('body').find(`.reveal-overlay-${id}`);
+    var $overlay = $('body').find(`.reveal-overlay.overlay-${id}`);
     if ($overlay.length < 1) {
       $overlay = $('<div></div>')
-        .addClass(`reveal-overlay-${id}`)
+        .addClass(`reveal-overlay overlay-${id}`)
         .appendTo('body');
     }
     return $overlay;
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -83,9 +83,12 @@ class Reveal {
    * @private
    */
   _makeOverlay(id) {
-    var $overlay = $('<div></div>')
-                    .addClass('reveal-overlay')
-                    .appendTo('body');
+    var $overlay = $('body').find(`.reveal-overlay-${id}`);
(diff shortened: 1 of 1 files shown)


kball:
Is there a reason you need this, or a case you're running into where the overlay is being duplicated?  The plugin already checks to see if it has an overlay before it calls _makeOverlay...

DaSchTour:
Well I had the case in which two overlays where created, my ajax injection added content to the second overlay and calling open opened the first empty overlay. May be because of the usage of new Foundation.Reveal() to instantiate and the $().reveal("open") it's not Instance-Safe?

Fact is that I ran into this problem and could only fix it by adding this patch.

kball:
That's a reasonable argument... @Owlbertz did you run into this at all as you were putting together unit tests for Reveal?  Could you write a test to repro the problem?

Owlbertz:
Nope, I didn't run into this problem. Seems to be somewhat use-case specific based on how you create new objects using the constructor.

I can create test case; maybe @DaSchTour could come up with some of his code causing the issue to help me with that?

DaSchTour:
Well it's really hard so extract this from my code. I have a wrapper around all jquery plugins to provide an unified interface to initialization. My code create an HTML element, that is attached to the DOM and than adds the Reveal functionality. I have different groups of triggers that use the same modal that was previously create through javascript and change the content of the modal through AJAX.

DaSchTour:
I dived into this a little bit and found, that this is an issue with `new Foundation.Reveal()` on the same jQuery layer Element. So this was also an issue with using Foundation. But I think it would be great if there was some handling to avoid this behavior.
I created a repro here: http://s.codepen.io/DaSch/debug/LRvGEX

kball:
Interesting... I think there is a whole slew of potentially undefined behavior associated with initializing multiple JS components on a single element.

I think instead, we should update the _init functionality to verify that there isn't already an id in the data-reveal attribute... and more broadly, make components throw an error (or warning) upon attempting to do this, and possibly even just return the original.

DaSchTour:
Well than also the id as parameter for _makeOverlay should be removed, as it isn't used.

PR #9172, opened 2016-09-09
Add license to bower.json

rbayliss:
Notes the MIT license in bower.json.  This is good practice, and stops VersionEye from complaining :)


rbayliss:
I screwed this up... I'll fix and do it again.

PR #9174, opened 2016-09-09
Fix mouseenter issue with touch devices

kball:
Fixes an odd issue encountered in a client project.  Hoverable dropdowns fall back to be clickable on touch devices, but due to some touch devices triggering a mouseenter event on click, you can end up in a situation where the hover show is getting triggered on click, with the result being that a dropdown can be opened on mobile but never closed.  This fixes it.

I'm open about if we want to use a javascript API to whatinput instead of the dom reference... anyone have an opinion on that?

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -169,12 +169,14 @@ class Dropdown {
 
     if(this.options.hover){
       this.$anchor.off('mouseenter.zf.dropdown mouseleave.zf.dropdown')
-          .on('mouseenter.zf.dropdown', function(){
-            clearTimeout(_this.timeout);
-            _this.timeout = setTimeout(function(){
-              _this.open();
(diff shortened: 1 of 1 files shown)


bwydev:
This fix now seems to require what-input to be included, or hover dropdown seems to fail. Is this expected, as I thought using what-input was optional?

kball:
@bwydev good catch, I can fix that case, but that won't necessarily handle the original bug, where hoverable dropdowns don't close correctly on mobile... do you know of a better way to attack this?

rafibomb:
I think this is still an open issue. My simple solution is to add my own hover function to the dropdown: https://codepen.io/rafibomb/pen/oGEZRV

vcarneiro:
This is definitely still an open issue, Rafi your suggestion is good but a bit glitchy. In some scenarios I hover out of the dropdown remains open.

PR #9176, opened 2016-09-13
Bump to version 6.2.2 [Orbit] Internet Explorer 11 Image Load Behaviour

simongcc:
/\* 
    problem: if image has no source or problematic source, the following function will not work properly in Internet Explorer 11(believably predecessors IE are also having the same issue )
    as the logic in `onImagesLoaded()` is based on `this.complete`, when image has no source or source cannot be loaded, 'complete' will never be happened, so `callback` is never called and `unloaded` will not be counted. Modern Chrome and Firefox does not have this problem, to resolve, I have added another testing below the 3rd condition.
So, when image have problem in loading, `singleImageLoaded()` will still be called to deduct count
*/

```
function onImagesLoaded(images, callback) 
// ...
{
 if (typeof this.naturalWidth === 'undefined' || this.naturalWidth === 0) {
  singleImageLoaded();
 }
// ...
}
```


kball:
Not sure what happened but I'm not seeing any actual changes submitted... please take a look and resubmit your PR

PR #9178, opened 2016-09-14
Don't attach accordion elements that are not marked as accordion-item

DaSchTour:
I think this is a relict from an older version, where data-accordion-item was not used.
I'm aware of the fact, that this is a breaking change, but as the documentation tell's the developer to use data-accordion-item I think it shouldn't fallback to use li. This way it is also possible to add list items to an accordion that have no content. Till now having to content section throws an ugly hard to debug javascript error.

### DIFF
--- a/js/foundation.accordion.js
+++ b/js/foundation.accordion.js
@@ -38,7 +38,7 @@ class Accordion {
    */
   _init() {
     this.$element.attr('role', 'tablist');
-    this.$tabs = this.$element.children('li, [data-accordion-item]');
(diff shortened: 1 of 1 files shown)


kball:
I think this is the right direction to go, but we'll need to make sure we highlight this as a breaking change requiring a possible migration.

kball:
Migration Note:  Accordions used to instantiate items for any child `li`, but they should only instantiate an item for a child marked with `data-accordion-item`.  If you depended on the former behavior, you will need to add the `data-accordion-item` attribute.

PR #9187, opened 2016-09-18
Reload focusableElments at every tab keydown to fix issue #9130

### LINKED ISSUE #9130: [Reveal] 'Tab' key does not work for forms loaded via ajax in reveal modals

samuelmc:
Fixes #9130 
Reload focusableElments at every tab keydown for the tab key to work properly when the content inside the reveal changes after opening.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -317,6 +317,7 @@ class Reveal {
       // handle keyboard event with keyboard util
       Foundation.Keyboard.handleKey(e, 'Reveal', {
         tab_forward: function() {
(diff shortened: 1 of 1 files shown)


Owlbertz:
This looks better to me, having it in the `TAB` and `SHIFT_TAB` handlers only will probably only have a small influence on the performance. Thanks for the update.

PR #9201, opened 2016-09-22
fix width body after open reveal

Andrey9300:
The width of the body is changed when opening/closing the reveal.

### DIFF
--- a/dist/foundation-flex.css
+++ b/dist/foundation-flex.css
@@ -3190,7 +3190,7 @@ body {
     bottom: 0; }
 
 body.is-reveal-open {
-  overflow: hidden; }
+  overflow: auto; }
 
 html.is-reveal-open,
 html.is-reveal-open body {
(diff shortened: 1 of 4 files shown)


Owlbertz:
@Andrey9300 what about those travis test commits?

Andrey9300:
@Owlbertz  I'm sorry, this is my first request, so I could not get from the travis test. I began to test on your own repository and not removed do not need commits.

Owlbertz:
The Travis CI test sometimes fails because it cannot download all resources via NPM/Bower/..., you should be able to manually re-start the test by following the "Details" link next to the failed build.

However, could you please describe what your PR does and why you made it in a more detailed way?

Andrey9300:
When we open the pop-up, for example directly on the website documentation http://foundation.zurb.com/sites/docs/reveal.html, the width of the <body> and changing the background like jumps at 17px. For clarity, we made pictures.
https://goo.gl/photos/rxGT4eRjuEXULax58

Owlbertz:
This behavior is intended, because otherwise you could scroll the page under the Reval backdrop while the modal is opened.

Andrey9300:
Okay, I thought it would be my first change, but not all the first time)

Owlbertz:
Don't worry, contributing is always a good thing. Making this change in your project might be required, but the general use case is that the modal is in the focus and the scrolling should be disabled.

PR #9204, opened 2016-09-24
Added a "small" selector for close-button

ghost:
The close button looks great on a multiline-content callout but a little big when the content is only one line and the callout is sized to `small`.

I think it might be convenient to add a `small` selector to close-button, just like the `small` and `large` selectors in callout.

Code example:

```
<div class="alert callout small" data-closable>
    <p>This is a system alert.</p>
    <button class="close-button small" aria-label="Dismiss alert" type="button" data-close>
    </button>
</div>
```

### DIFF
--- a/scss/components/_close-button.scss
+++ b/scss/components/_close-button.scss
@@ -57,5 +57,10 @@ $closebutton-color-hover: $black !default;
 @mixin foundation-close-button {
   .close-button {
     @include close-button;
+    
+    &.small {
+      top: 0.25rem;
+      font-size: 1.5em;
+    }
   }
 }
--- a/scss/components/_close-button.scss
+++ b/scss/components/_close-button.scss
@@ -10,17 +10,26 @@
 /// @type List
 $closebutton-position: right top !default;
 
-/// Right (or left) offset for a close button.
-/// @type Number
-$closebutton-offset-horizontal: 1rem !default;
+/// Right (or left) offset(s) for a close button.
+/// @type Number|Map
+$closebutton-offset-horizontal: (
+  small: .66rem,
+  medium: 1rem,
+) !default;
 
-/// Top (or bottom) offset for a close button.
-/// @type Number
-$closebutton-offset-vertical: 0.5rem !default;
+/// Top (or bottom) offset(s) for a close button.
+/// @type Number|Map
+$closebutton-offset-vertical: (
+  small: .33em,
+  medium: .5rem,
+) !default;
 
-/// Default font size of the close button.
-/// @type Number
-$closebutton-size: 2em !default;
+/// Default font size(s) of the close button.
+/// @type Number|Map
+$closebutton-size: (
+  small: 1.5em,
+  medium: 2em,
+) !default;
 
 /// The line-height of the close button. It affects the spacing of the element.
 /// @type Number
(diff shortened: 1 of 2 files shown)


DaSchTour (changes requested):
Size and position look a bit to hardcoded for me. The size and position should depend on configuration of the close-button. It should work with other configurations than the basic one.

ncoden:
Like @DaSchTour said, it is arbitrary values. Also, there is no reasons for this feature to not be available for custom close-buttons generated with `@mixin close-button`.

So, we should rather modify `scss/components/close-button.scss`:
```scss
...

/// Right (or left) offset(s) for a close button.
/// @type Number|Map
$closebutton-offset-horizontal: (
  small: .66rem,
  medium: 1rem,
) !default;

/// Top (or bottom) offset(s) for a close button.
/// @type Number|Map
$closebutton-offset-vertical: (
  small: .33em,
  medium: .5rem,
) !default;

/// Default font size(s) of the close button.
/// @type Number|Map
$closebutton-size: (
  small: 1.5em,
  medium: 2em,
) !default;

...
```

These values can be easily used with an alternative version of `@function -zf-get-bp-val` (for sizes indexs and not breakpoints, working like `@function grid-column-gutter`. I'll call it `-zf-get-size-val`.

And we replace the actual `@mixin close-button` with:

`scss/components/close-button.scss` :
```scss
...

@mixin close-button-size($size) {
  $x: nth($closebutton-position, 1);
  $y: nth($closebutton-position, 2);

  #{$x}: -zf-get-size-val($closebutton-offset-horizontal, $size);
  #{$y}: -zf-get-size-val($closebutton-offset-vertical, $size);
  font-size: -zf-get-size-val($closebutton-size, $size);
  line-height: -zf-get-size-val($closebutton-lineheight, $size);
}

@mixin close-button {
  ... // (other properties)

  &.small { @include close-button-size(small) }
  &, &.medium { ... }
  // &.large { ... }
  // (etc...)
}

...
```

What do you think ?
(poke @kball)

*__Note__: The best would be to standardize the whole Foundation sizes gesture, which for now differs from a component to an other. I think it can be done by making sizes like we currently make breakpoints.*

john-wei:
@DaSchTour  You are right. It is a better way to configure the size and position within `@mixin close-button` rather than hardcode them.

john-wei:
@ncoden This is much better than my hardcoding way :)

john-wei:
@ncoden This is much better than my hardcoding way :)

Please consider creating a new pull request.

john-wei:
I closed this pull request a day ago simply because I thought this pull request is no longer needed.

I reopened this because @ncoden gave a very detailed solution and I think keep this open is convenient for further conversations.

ncoden:
@john-wei Can you apply the changes I purposed to your PR ?

john-wei:
@ncoden Yes. I'll do that. Thanks :)

john-wei:
@ncoden Please help check the changes.

Should I place `@function -zf-get-size-val` in `scss/components/close-button.scss`? If we standardize the whole Foundation sizes gesture, like you said, then I think we should make `@function -zf-get-size-val` a general function and place it outside `close-button.scss`.

john-wei:
Another thought about sizes:

Is it a good idea to make the classes consistent, like "tiny - small - default - large"?

Currently, in `$button-sizes`, we use:
```
$button-sizes: (   
  tiny: 0.6rem,   
  small: 0.75rem,   
  default: 0.9rem,   
  large: 1.25rem, 
);
```
in `$dropdown-sizes`, we use:
```
$dropdown-sizes: (   
  tiny: 100px,   
  small: 200px,  
  large: 400px, 
);
```
and in `$closebutton-sizes`, we use:
```
$closebutton-sizes: (
   small: 1.5em, 
  medium: 2em, 
);
```
How about we classify all the sizes with the same classes? Like:
```
$everyobject-sizes: (
  tiny: ()rem,   
  small: ()rem,   
  default: ()rem,   
  large: ()rem, 
);
```
I think, in this way, it might be easier for users to remember the classes, but it can also make things more complex than it currently is.

ncoden:
Thanks @john-wei. Standardize the whole Founation size gesture is a big and breaking change. I think we can keep `@funciton -zf-get-size-val` in `/scss/components/close-button.scss` for now, and factorize it later, in another PR.

I think we can merge. @kball, what do you think ? and `:develop` ?

kball:
I agree that full standardization is a bigger change, but seems like the right thing to do.  We can evolve that direction as we go.  This looks good to me - We've branched off for 6.2.4, so develop will become 6.3, this seems reasonable to merge.

PR #9146, opened 2016-08-31
Avoid reveal modal closing upon click of an element not in the DOM

ardeshireshghi:
Before submitting a pull request, make sure it's targeting the right branch:
- For documentation fixes, use `master`.
- For bug fixes, use `develop`.
- For new features, use the branch for the next minor version, which will be formatted `v6.x`.

If you're fixing a JavaScript issue, it would help to create a new test case under the folder `test/visual/` that recreates the issue and show's that it's been fixed. Run `npm test` to compile the testing folder.

Happy coding! :)

Added check to make sure e.target is still in the DOM. I came across a bug when I created a google map canvas inside reveal modal, whereby upon closing the info popup on a map marker, the entire reveal modal gets closed. This is because when the overlay click handler is invoked, Google code has already removed the e.target  element and the overlay handler only gets a ghost element and can not find it in the modal element

PR #9207, opened 2016-09-26
Add visual test for PR #9146.

Owlbertz:
Visual test I made while checking out #9146, thought it may be useful in the code base.


kball:
great!

PR #9209, opened 2016-09-28
Update grid.md with grid-layout options

brettsmason:
Added a custom block grid writeup to the grid.md docs.

This is a followup based on @andycochran suggesting an example in #8871.
If you could take a look and see if the example reads OK and I'll amend as needed.

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -564,3 +564,52 @@ Refer to the Sass documentation below to learn how each mixin works.
   }
 }
 ```
+### Custom Block Grid
+
+Use the `grid-layout()` mixin to create your own block grid.
+By default the mixin takes 3 parameters:
+- Number of columns
+- The child element selector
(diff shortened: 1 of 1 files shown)


kball:
I think this makes sense... @andycochran and @rafibomb what do you think?

docs/pages/grid.md
```diff
@@ -564,3 +564,52 @@ Refer to the Sass documentation below to learn how each mixin works.
...
 ```
+### Custom Block Grid
+
+Use the `grid-layout()` mixin to create your own block grid.
+By default the mixin takes 3 parameters:
+- Number of columns
+- The child element selector
+  - An optional padding value
+
+The padding value can be set to `$grid-column-gutter` to use the values from that map.
```
andycochran: 😮 Can I use a custom map?

andycochran:
This is great.

PR #9210, opened 2016-09-28
check if cb is a function

pine3ree:
Hello @kball,

https://github.com/zurb/foundation-sites/pull/9139

here's the codepen

http://codepen.io/anon/pen/xEAEdp

steps to reproduce:
1. start in full width (desktop) browser
2. scroll down a few time to activate sticky
3. resize the browser window to a foundation small width (menu is drilldown)
4. scrollup and click one of the drilldown links, you will get the console error foundation.js:7889 Uncaught TypeError: cb is not a function

using latest chrome and opera


kball:
Perfect, thank you.

PR #9212, opened 2016-09-28
Accordion tests

Owlbertz:
Added unit tests for Accordion.
Changed the way the toggling works to fix some issue detected during testing.

### DIFF
--- a/js/foundation.accordion.js
+++ b/js/foundation.accordion.js
@@ -76,16 +76,8 @@ class Accordion {
       if ($tabContent.length) {
         $elem.children('a').off('click.zf.accordion keydown.zf.accordion')
                .on('click.zf.accordion', function(e) {
-        // $(this).children('a').on('click.zf.accordion', function(e) {
           e.preventDefault();
-          if ($elem.hasClass('is-active')) {
-            if(_this.options.allowAllClosed || $elem.siblings().hasClass('is-active')){
-              _this.up($tabContent);
-            }
-          }
-          else {
-            _this.down($tabContent);
-          }
+          _this.toggle($tabContent);
         }).on('keydown.zf.accordion', function(e){
           Foundation.Keyboard.handleKey(e, 'Accordion', {
             toggle: function() {
@@ -115,14 +107,12 @@ class Accordion {
 
   /**
    * Toggles the selected content pane's open/close state.
-   * @param {jQuery} $target - jQuery object of the pane to toggle.
+   * @param {jQuery} $target - jQuery object of the pane to toggle (`.accordion-content`).
    * @function
    */
   toggle($target) {
     if($target.parent().hasClass('is-active')) {
-      if(this.options.allowAllClosed || $target.parent().siblings().hasClass('is-active')){
-        this.up($target);
-      } else { return; }
+      this.up($target);
     } else {
       this.down($target);
     }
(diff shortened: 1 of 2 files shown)


js/foundation.accordion.js
```diff
@@ -165,16 +155,16 @@ class Accordion {
...
    * @function
    */
   up($target) {
     var $aunts = $target.parent().siblings(),
         _this = this;
-    var canClose = this.options.multiExpand ? $aunts.hasClass('is-active') : $target.parent().hasClass('is-active');
+    var canClose = $aunts.hasClass('is-active');
 
-    if(!this.options.allowAllClosed && !canClose) {
+    if(!this.options.allowAllClosed && !$aunts.hasClass('is-active')) {
```
kball: I'm a little confused by this change... do we not need to check if the target is open?  And if not, why do we need the 'canClose' variable at all?

test/javascript/components/accordion.js
```diff
@@ -1,20 +1,142 @@
...
+      });
+    });
+
+    it('toggles attributes of title of the targeted container', function() {
+      $html = $(template).appendTo('body');
+      plugin = new Foundation.Accordion($html, {});
+
+      plugin.up($html.find('.accordion-content').eq(1));
+      $html.find('.accordion-title').eq(1).should.have.attr('aria-expanded', 'false');
+      $html.find('.accordion-title').eq(1).should.have.attr('aria-selected', 'false');
```
kball: Do we know that these were true prior to the up?  It's not clear to me that we're actually testing the toggle.

test/javascript/components/accordion.js
```diff
@@ -1,20 +1,142 @@
...
+    });
+
+    it('not closes the open container if allowAllClosed is false', function() {
+      $html = $(template).appendTo('body');
+      plugin = new Foundation.Accordion($html, {allowAllClosed: false});
+
+      plugin.up($html.find('.accordion-content').eq(0));
+      $html.find('.accordion-content').eq(0).should.be.visible;
```
kball: In the prior examples we waited for 'up.zf.accordion' before checking on visibility, presumably because of the animation.  Shouldn't we wait here as well?

test/javascript/components/accordion.js
```diff
@@ -1,20 +1,142 @@
...
+        done();
+      });
+    });
+
+    it('not closes open container if multiExpand is true', function() {
+      $html = $(template).appendTo('body');
+      plugin = new Foundation.Accordion($html, {multiExpand: true});
+
+      plugin.down($html.find('.accordion-content').eq(1));
+      $html.find('.accordion-content').eq(0).should.be.visible;
```
kball: Another place where I think we have to wait for the event before we make the assertion...

test/javascript/components/accordion.js
```diff
@@ -1,20 +1,142 @@
...
+        done();
+      });
+    });
+
+    it('not closes the open container if allowAllClosed is false', function() {
+      $html = $(template).appendTo('body');
+      plugin = new Foundation.Accordion($html, {allowAllClosed: false});
+
+      plugin.toggle($html.find('.accordion-content').eq(1));
+      $html.find('.accordion-content').eq(1).should.be.visible;
```
kball: Same point on waiting for the event.

kball (changes requested):

js/foundation.accordion.js
Owlbertz: As far as I saw checking is not explicitly needed, but I guess checking none the less would be a good idea.
Regarding the variable - yes, I made it obsolete and not removed it.

I will fix both issues.

test/javascript/components/accordion.js
Owlbertz: The usage of `toggle()` is a bit limited as it only checks if the target is active and closes it if so, or opens it otherwise. Basically `toggle()` just delegates to `up()` and `down()`.

test/javascript/components/accordion.js
Owlbertz: When checking if it is visible it is not necessarily needed to wait as it is already visible as soon as the animation starts. When hiding waiting is mandatory as it will only be hidden once the animation completed.

test/javascript/components/accordion.js
Owlbertz: Same as above.

test/javascript/components/accordion.js
kball: Got it, makes sense

test/javascript/components/accordion.js
kball: I guess my point is... we we know that prior to triggering 'up', that these had values of true?  In particular, given that I believe we default the content in slot 0 to be open, and here we're checking against 1,  I suspect these started out with values false

Owlbertz: You're right, the test should of course toggle the 0th element.

Owlbertz:
Included the feedback.

js/foundation.accordion.js
kball: This looks good

test/javascript/components/accordion.js
kball: ok once this test update is in I think this is good to merge.

kball:
Update looks good, some of the tests were updated to target the correct elem, I think one more needs update, and then we're good to merge.

kball (approved):

PR #9214, opened 2016-09-29
Add option for non-closeable accordion menu. Parent links will be followed.

relthyg:
Hi, 
according to [this](https://github.com/zurb/foundation-sites/issues/7770) discussion, I wrote a patch for the accordion menu that offers an option to follow the parent links of open submenus. You can see it in action [here](https://relthyg.github.io/foundation-sites/).

Basically, I check for a "data-option-closeable" attribute of the `<ul>`-Element. If this is set to "false", parent links will be followed instead of closing the submenu.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -185,13 +185,19 @@ class AccordionMenu {
 
   /**
    * Toggles the open/close state of a submenu.
+   * If data option "closeable" is set to "false", submenu will not close
+   * but parent link will be followed then.
    * @function
    * @param {jQuery} $target - the submenu to toggle
    */
   toggle($target){
     if(!$target.is(':animated')) {
       if (!$target.is(':hidden')) {
-        this.up($target);
+        if (this.options.closeable === false) {
+          window.location.href = $target.parent().children('a').attr('href');
+        } else {
+          this.up($target);    
+        }
       }
       else {
         this.down($target);


kball:
Hi @relthyg, this seems like a reasonable thing to enable, but before we merge we need to think about the keyboard behavior for this.  Right now it is kind of confusing because you can't close the subnav with the mouse but keyboard navigation will still close it.

relthyg:
Hi @kball, as far as I can see, keyboard users can as well choose to "open" and "close" submenus, while mouse users can only "toggle" them. I don't see the need to remove any features here, but maybe we should consider renaming the "data-option-closeable" to something else, like "data-option-follow-parent-links" or so, to avoid this confusion. What do you think?

kball:
@brettsmason would love your thoughts on this, as it is related to your issue on accordion menus

brettsmason:
@kball for my personal use case it doesn't quite solve it unfortunately.

I think the main issue is naturally you would click the parent menu item to open the menu because of the arrow indicator. Then however you would go to collapse it but instead get directed to the parent link.

I think personally this needs to be a completely seperate submenu toggle like I'm working on to avoid user confusion, eg: http://samikeijonen.github.io/responsive-nav.js/demos/multiple-levels/index.html

However thats just my opinion!

relthyg:
@kball @brettsmason I agree with you. Brett's approach is far better.

kball:
Ok, closing this in favor of @brettsmason's approach, which we're planning to have wrapped up for 6.3

PR #9218, opened 2016-09-30
Tooltip tests

Owlbertz:
Added unit tests for Tooltips.

Fixed a wrong reference to the triggerClass option for Tooltips.


kball:
Nice work!

PR #9222, opened 2016-10-01
Added unit tests for Tabs.

Owlbertz:
Please note the `TODO:` comment.
@kball

### DIFF
--- a/test/javascript/components/tabs.js
+++ b/test/javascript/components/tabs.js
@@ -1,20 +1,150 @@
 describe('Tabs', function() {
-	var plugin;
-	var $html;
+  var plugin;
+  var $html;
+  var template = `
+    <div>
+      <ul class="tabs" data-tabs id="example-tabs">
+        <li class="tabs-title is-active"><a href="#panel1" aria-selected="true">Tab 1</a></li>
+        <li class="tabs-title"><a href="#panel2">Tab 2</a></li>
+        <li class="tabs-title"><a href="#panel3">Tab 3</a></li>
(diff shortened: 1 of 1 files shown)


test/javascript/components/tabs.js
```diff
@@ -1,20 +1,150 @@
...
+      $html.find('[href="#panel1"]').should.have.attr('role', 'tab');
+      $html.find('[href="#panel1"]').should.have.attr('aria-controls', $html.find('panel1').attr('id'));
+      $html.find('[href="#panel1"]').should.have.attr('aria-selected', 'true');
+      $html.find('[href="#panel2"]').should.have.attr('aria-selected', 'false');
+
+      // Tab list items
+      $html.find('[href="#panel1"]').parent().should.have.attr('role', 'presentation');
+    });
+
+    // TODO: Check if classes should be applied by the Plugin or not.
```
kball: I don't think the plugin should apply classes; rather it looks like it takes some arguments that tell it what classes to look for.  This is relatively inline with our general policy that data-attributes should be communicating to the JS and classes should be style related.

Part of me thinks instead we should be labeling links & panels with data attributes, thus leaving the classes to be completely style related and not hunted for by the JS, similar to how off canvas works, but regardless I think the only classes that JS should be adding are state related ('is-active'), not the base classes.

Owlbertz: I agree. But this would mean some work in various plugins.
I will adjust the tests as the first step to match the fact classes are not applied.

PR #9229, opened 2016-10-05
Add "alert" as required map value to $foundation-palette

sprrw:

Happy coding! :)

Undefined "alert" color in $foundation-palette breaks SCSS compilation. Documentation incorrectly identifies "primary" as the only required value in the user redefined map.

### DIFF
--- a/docs/pages/global.md
+++ b/docs/pages/global.md
@@ -83,7 +83,7 @@ Many components can also be colored with four other colors: secondary, alert, su
   </div>
 </div>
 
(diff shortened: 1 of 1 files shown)


ncoden:
Thanks. Think to delete the default PR text next time ;)

@kball *Should* `alert` color be required ? I think we should instead use `primary` color when `alert` color is not given.

kball:
Mm... a good point.  I think this PR is good due to current situation, but having alert fall back to primary if it is unset would be even better.

PR #9233, opened 2016-10-05
9092 fix

### LINKED ISSUE #9092: Block grid last column should be float to the right on RTL
On a RTL document, The last .column div on a block grid should be floated to the right not to the left.

brettsmason:
Sets the last child to float to global left. Fixes #9092


brettsmason:
Closing this - will do a new PR as I messed this one up some how...

PR #9240, opened 2016-10-06
Fixes nesting default/flex grids within a container

### LINKED ISSUE #9225: [Grid] Compiled CSS for child `.row`s not as expected when wrapped in ID/Class
Hi all.

I have a situation in which I'm trying to encapsulate the Foundation Grid styles behind an ID - we have a product that can be injected into other websites and I'm trying to ensure that any styles we need don't bleed out and interfere with the parent site.

However, I've got small issue when I wrap the Foundation Sites grid in a class or ID in that the output CSS for the child `.row` classes is not as expected. The wrapper ID/Class is placed in some unexpected places.

Consider the following setup in `my-app.scss`:
```scss
@import 'foundation-sites/scss/util/util';
@import 'foundation-sites/scss/global';
@import 'foundation-sites/scss/grid/grid';

#WRAP {
    @include foundation-flex-grid();
}
```

I expected the Foundation Sites grid styles to have the additional specificity of my ID/Class.
**E.g.:** `#WRAP .row .row`
```css
#WRAP .row {
  max-width: 75rem;
  margin-left: auto;
  margin-right: auto;
  display: flex;
  flex-flow: row wrap;
}
#WRAP .row .row {
  max-width: none;
  margin-left: -0.625rem;
  margin-right: -0.625rem;
}
@media screen and (min-width: 40em) {
  #WRAP .row .row {
    margin-left: -0.9375rem;
    margin-right: -0.9375rem;
  }
}

[...]

#WRAP .column.row.row, #WRAP .row.row.columns {
  float: none;
  display: block;
}
#WRAP .row .column.row.row, #WRAP .row .row.row.columns {
  padding-left: 0;
  padding-right: 0;
  margin-left: 0;
  margin-right: 0;
}

#WRAP .small-1 {
  flex: 0 0 8.33333%;
  max-width: 8.33333%;
}

[...]
```

**What happened instead?**
The parent ID/Class is added in some strange positions in relation to `.row`.
**E.g.:** `#WRAP .row #WRAP .row`
```css
#WRAP .row {
  max-width: 75rem;
  margin-left: auto;
  margin-right: auto;
  display: flex;
  flex-flow: row wrap;
}
#WRAP .row #WRAP .row {
  max-width: none;
  margin-left: -0.625rem;
  margin-right: -0.625rem;
}
@media screen and (min-width: 40em) {
  #WRAP .row #WRAP .row {
    margin-left: -0.9375rem;
    margin-right: -0.9375rem;
  }
}

[...]

#WRAP .column.row.row, #WRAP .row.row.columns {
  float: none;
  display: block;
}
.row #WRAP .column.row.row, .row #WRAP .row.row.columns {
  padding-left: 0;
  padding-right: 0;
  margin-left: 0;
  margin-right: 0;
}
#WRAP .small-1 {
  flex: 0 0 8.33333%;
  max-width: 8.33333%;
}

[...]
```

brettsmason:
In the flex grid there were 2 issues:
- The use of `& &` on line 158
- `.row &` on line 194

The issues on the standard grid were almost the same

Fixes #9225 
##


ncoden:
Thanks, I never thought to it when I was working on grids.
You just forgot an empty line at the end of `scss/grid/_flex-grid.scss`

@kball It can be merged now in develop. But the nesting often causes problems (arbitrary higher specificity), so I think it should be removed it later (it's a breaking change).

brettsmason:
@ncoden Thanks for pointing that out, fixed 😄

ncoden (approved):
:+1:

scss/grid/_classes.scss
```diff
@@ -69,14 +69,14 @@
...
 
-    // To properly nest a column row, padding and margin is removed
-    .#{$row} & {
-      padding-left: 0;
-      padding-right: 0;
-      margin-left: 0;
-      margin-right: 0;
-    }
+  // To properly nest a column row, padding and margin is removed
+  .#{$row} .#{$column}.#{$row}.#{$row} {
```
andycochran: `.row.row`?

andycochran: Oh, never mind. This is for specificity. 😬

Issue #7308, opened 2015-11-30
[Magellan]  On page load (without any anchor) first menu item is always rendered active, even if anchor is not yet active. 


phifa:
**How can we reproduce this bug?**
1. Go to the sample page: http://stage.dasachtegebot.de/magellan/lorem.html 
2. Click on the menu items in the Top Bar. 
3. At the destination, the headlines are covered with the Top Bar. 

**What did you expect to happen?**
I expect the scrolling to end where the section starts, including the Fixed Top Bar! So the height of the Magellan Container needs to be included in the scroll position, or there needs to be a scroll offset option. When I add **threshold: 290**, the scrolling is correct (for one media query), but the magellan active stops working properly. I have to use **threshold: -1** though for it to work at all... But I guess that is another issue. Here is my app.js: http://stage.dasachtegebot.de/magellan/assets/js/app.js

**What happened instead?**
Magellan scrolls too far. 

**Test case**
http://stage.dasachtegebot.de/magellan/lorem.html

zurbchris:
I see the problem. If you come up with a solution, I'd love to see a PR. Otherwise this fix will take a bit to get to. Thanks for pointing it out.

phifa:
Too bad, I have no idea how to do this. But, I can report another error :smile: 
As you can see on the example page, once the first nav item is marked "active" and you scroll back to the top, it always stays active. 
http://stage.dasachtegebot.de/magellan/lorem.html

TvrtkoM:
Hi all,
@phifa it works for me. The links are changing 'active' class when i scroll up. What browser are you using?

Regarding a feature maybe add an offset option which acts on on *Foundation.MediaQuery.current* state. For example options like offset_small, offset_medium ... and if its just *offset* it's value will be used for all screens.

Another solution comes to mind is referencing another element and calculate offset on event time, but this seems weird.

phifa:
Well, they render active but the first item stays active, even if I scroll way before it. See the screenshot.
![blue](https://cloud.githubusercontent.com/assets/3061129/11554297/177de5cc-9997-11e5-9219-80ee9d447263.JPG)

zurbchris:
That's the default, to add the active class to the first list item. I'll put it on the list of minor bug fixes unless you want to submit a pr to solve it :smile:

phifa:
Not sure if I can do that, but ok. It definitely needs to be done. I recently used BS for a website and there it works out of the box: http://www.durr.com/challenge/

[...]

PR #8949, opened 2016-06-19
Magellan issue #7308 bug fix

rawiki:
Modified function _updateActive() to stop first magellan menu item being rendered active before it reaches anchor point. Scrolling above the anchor point will also remove any active classes. If someone can guide me in writing a test I'll be happy to do that.

PR #9243, opened 2016-10-08
Magellan issue #7308 - deep linking support

rawiki:
Bug fix for Magellan issue #7308 now with support for deep linking.


ncoden:
poke @zurbchris @andycochran

@rawiki Can https://github.com/zurb/foundation-sites/pull/8949 be closed ?

PR #9251, opened 2016-10-12
Slider tests

Owlbertz:
Added unit tests for Slider.

Properly clear timeout used to fire `changed.zf.slider` on `destroy()`. Otherwise timeout ma trigger the event after the element has been set to null.

Make `Move()` fire callback directly when duration is 0. Useful for testing and I guess in general. Maybe @kball can give some feedback.


kball:
Looks good.  I've seen a lot of places Move is used incorrectly, but this one appears to actually be using it appropriately, and this update makes sense.

PR #9255, opened 2016-10-14
Fixes aria roles in accordion menu

MichellanneLi:
This pr fixes the aria roles in the accordion menu by defining the <ul> as "menu." 

The original aria roles in the accordion menu defined it as a tablist, which was flagged as causing accessibility problems when I ran the following command line auditor: https://github.com/addyosmani/a11y

To replicate the issue, install and run "$ a11y http://localhost:3000/accordion-menu.html." Among other accessibility issues, you will see this one: 

"  ✖ Elements with ARIA roles must be in the correct scope

  #lsha5t-acc-menu > .is-submenu-item.is-accordion-submenu-item
  #lsha5t-acc-menu > LI:nth-of-type(2)
  #lsha5t-acc-menu > LI:nth-of-type(3)
  #c6n5i7-acc-menu > LI:nth-of-type(2)
  #c6n5i7-acc-menu > LI:nth-of-type(3)
  #pik1va-acc-menu > .is-submenu-item.is-accordion-submenu-item
  #pik1va-acc-menu > LI:nth-of-type(2)
  #docs > .vertical.menu > LI:nth-of-type(3)"

To validate that this pr corrects the problem, please pull this branch into your local repo and run "$ a11y http://localhost:3000/accordion-menu.html" again. You will see that this error no longer appears.

You can also use your inspector to view the aria roles and confirm that they are correct.

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -47,7 +47,7 @@ class AccordionMenu {
   _init() {
     this.$element.find('[data-submenu]').not('.is-active').slideUp(0);//.find('a').css('padding-left', '1rem');
     this.$element.attr({
-      'role': 'tablist',
+      'role': 'menu',
       'aria-multiselectable': this.options.multiOpen
     });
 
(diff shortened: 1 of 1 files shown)


Owlbertz:
Hi @MichellanneLi, we already have a pending PR for v6.3 that addresses this (#9033, yes it is quite messed up because v6.3 and develop branches are not synced). However, this PR suggests using the Tree sytax (see https://github.com/zurb/foundation-sites/pull/9033/files#diff-ff345c8516b8e29d396e298367201c2c).

kball:
I think this is a step forward, even though Tree may end up being the final approach, the current approach is broken and this addresses that in the short term, and is minimally invasive.

PR #9261, opened 2016-10-15
Offcanvas tests

Owlbertz:
Added unit tests for Off Canvas.

Remove unused variable.
Fix condition when close() should do nothing.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -154,8 +154,7 @@ class OffCanvas {
    */
   open(event, trigger) {
     if (this.$element.hasClass('is-open') || this.isRevealed) { return; }
-    var _this = this,
-        $body = $(document.body);
+    var _this = this
 
     if (this.options.forceTop) {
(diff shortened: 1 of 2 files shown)


js/foundation.offcanvas.js
```diff
@@ -154,8 +154,7 @@ class OffCanvas {
    */
   open(event, trigger) {
     if (this.$element.hasClass('is-open') || this.isRevealed) { return; }
-    var _this = this,
-        $body = $(document.body);
+    var _this = this
```
kball: Do we no longer need $body?  In which case we should update with semicolon after this. Otherwise we should put this dropped line back in.

Owlbertz:
@kball the variable was not used in this block, but you were right, I missed the semicolon. I pushed a new commit that adds it again.

kball:
Ended up reverting this... thought the travis error was transient but turned out it was real, reproducing it locally.  @Owlbertz can you take a look and resubmit with errors fixed?  Thanks!

PR #9272, opened 2016-10-18
Check readyState that works with different browsers.

milose:
This fixes #8478 and #8955 and probably others with regards to image loading from cache.

I removed usage of `this.naturalWidth` as this produces unexpected behaviour in some browsers. Apparently, it only works reliably the first time.


kball:
Awesome, thanks @milose. Looks good.

Issue #9138, opened 2016-08-27
password validation in abide example flaky


akostadinov:
This might or might not be related to #6740. I'm using Firefox 47.0.1. The issue is that password validation doesn't work properly sometimes. I am just starting to explore foundation and went to:
http://foundation.zurb.com/sites/docs/kitchen-sink.html#abide

It is somehow tricky to reproduce, sometimes warning for matching password re-typed does not disappear and sometimes it does not appear when it should. See below one of the sequences I could reproduce with:
1. type password field "asd"
2. type re-enter password field "asd"
3. change password field to "dsa" (re-enter field does not become red while it should)
4. type in re-enter password field to "asd" (field still not red while it should)

IMO if re-enter password field is non-empty, validation should happen on password field update.
Also when re-enter password field is typed in (even if string is left the same as it was before), validation should be performed when field loses focus.

Owlbertz:
I made a PR here #9273. Maybe you can check if this solves your issue @akostadinov?

Owlbertz:
New PR here: #9275.

kball:
I believe this is fixed by #9275

PR #9273, opened 2016-10-18
Abide password validation fix

Owlbertz:
Re-validate fields depending on the changed field.

Adresses #9138.


Owlbertz:
Closed because of messed up history.

PR #9278, opened 2016-10-20
[Orbit] Detach events on reInit

### LINKED ISSUE #9245: Foundation.reInit('orbit') works incorrectly
See example http://codepen.io/anon/pen/JRLxyX

The issue is in foundation.orbit.js _events() method.
Navigation buttons do not unregister events so when doing re-init the events will be attached again.

For now I am detaching events manually before calling reInit(): 

```
$orbit.off('.zf.orbit').find('*').off('.zf.orbit')
Foundation.reInit($orbit)
```

milose:
Fixes #9245.

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -43,7 +43,8 @@ class Orbit {
   * @private
   */
   _init() {
-    this._detachEvents();
+    // @TODO: consider discussion on PR #9278 about DOM pollution by changeSlide
+    this._reset();
 
     this.$wrapper = this.$element.find(`.${this.options.containerClass}`);
     this.$slides = this.$element.find(`.${this.options.slideClass}`);
@@ -241,9 +242,9 @@ class Orbit {
   }
 
   /**
-   * Detaches event listeners
+   * Resets Orbit so it can be reinitialized
    */
-  _detachEvents() {
+  _reset() {
     // Don't do anything if there are no slides (first run)
     if (typeof this.$slides == 'undefined') {
       return;
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -245,7 +245,7 @@ class Orbit {
    */
   _detachEvents() {
     // Don't do anything if there are no slides (first run)
-    if (typeof this.$slides == "undefined") {
+    if (typeof this.$slides == 'undefined') {
       return;
     }
 
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -43,6 +43,8 @@ class Orbit {
   * @private
   */
   _init() {
+    this._detachEvents();
+
     this.$wrapper = this.$element.find(`.${this.options.containerClass}`);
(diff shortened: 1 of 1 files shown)


js/foundation.orbit.js
```diff
@@ -239,6 +241,44 @@ class Orbit {
...
+      // Reset all sliddes
+      this.$slides.each(function(el) {
+        $(el).removeClass('is-active is-active is-in')
+          .removeAttr('aria-live')
+          .hide();
+      });
+
+      // Show the first slide
+      this.$slides.first().addClass('is-active').show();
+
```
kball: A little confused why this is in _detachEvents

js/foundation.orbit.js
```diff
@@ -239,6 +241,44 @@ class Orbit {
...
+      // Triggers when the slide has finished animating
+      this.$element.trigger('slidechange.zf.orbit', [this.$slides.first()]);
```
kball: This also seems like it might be in the wrong place

js/foundation.orbit.js
```diff
@@ -239,6 +241,44 @@ class Orbit {
...
+
+      // Select first bullet if bullets are present
+      if (this.options.bullets) {
+        this._updateBullets(0);
```
kball: and this as well.  Should these be in _init?  Or

kball (changes requested):
Broadly speaking, love the direction of this.  I'm a bit confused why some stuff is in the _detachEvents method, as it doesn't appear to be actually related to detaching events.  Would like those either pealed off into a different method or explain why they are a part of detach.

js/foundation.orbit.js
milose: This is needed to reset the state of the component because of poorly implemented `changeSlide` method.

js/foundation.orbit.js
milose: I am not sure this is necessary, but users might be listening to this event and by switching to the first slide, this must be triggered.

milose:
Perhaps we should rename `_detachEvents` to `_resetOrbit` or something similar, because it does more then detaches events.

milose:
Because of the poor implementation of `changeSlide` method, I was almost inclined to copy initial DOM of the Orbit in local state, then pull it back in if `_init` was called by `Foundation.reInit($orbit)`. That would also fix the issue but it is not an elegant solution. This proves that `changeSlide` method pollutes the DOM of the Orbit and needs to be changed in order for `_init` to do it's job without this pull request.

On the other hand, shy of `reInit`, the `changeSlide` method works as expected, and is battle-tested.

kball:
@milose I think if we change the function name to more semantically match all of its behavior (`_resetOrbit` works for me), I'm good with this PR, and it moves the ball forward.  Afterwards, if you want to work on another one that cleans up `changeSlide` that would be awesome! :)

kball (approved):

PR #9279, opened 2016-10-21
parseValue() full match with string literals

pine3ree:
the original regex matching would have made values like "untrue" to be interpreted as `true` and values like "not-false" as `false`
see discussion here: https://github.com/zurb/foundation-sites/pull/9149

PS
i kept the `else`s this time even if not necessary as we return inside every conditions... :-)

### DIFF
--- a/js/foundation.core.js
+++ b/js/foundation.core.js
@@ -364,8 +364,8 @@ function functionName(fn) {
   }
 }
 function parseValue(str){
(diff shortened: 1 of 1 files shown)


pine3ree:
cloing/reopening to trigger travis build

pine3ree:
now a travis phantom-js install error .... will retry later

PR #9283, opened 2016-10-21
Cleaned history version of #9033 for #8680

kball:
From @sethkane's original PR:

various HTML, SCSS and JS updates to make the accordion menu works better with accessibility.

Key points:

Removed not needed events
Changed html to being more semantic
Added conditionals Foundation.Nest for proper roles
Added focus ring
Tested in:

IE 11 and JAWS
Firefox and NVDA
Mac and VoiceOver
iPhone and VoiceOver


kball:
Testing this a little bit, keyboard functionality looks good, but the mouse interactions aren't working like I'd expect.

kball:
@sethkane @Owlbertz This is a clean version of #9033... looks like it still needs a little work to get past the line.  What are our next steps?

sethkane:
@kball Can you be a little more specific on what you are expecting that you aren't seeing.   The only thing I notice is I should remove the tabindex="-1" if the items shouldn't be clickable. e.g. An item with no child objects. I will take care of this today or tomorrow.

kball:
@sethkane a few things I'm noticing:

1)  In the visual test, the menu items don't respond to click (don't expand)
2)  Within the docs pages, the focus styling applies on click, not just on keyboard interaction
3)  Also within the docs page, the focus styling seems off

sethkane:
@kball Ha... I am stupid I didn't even modify the visual test or documents. I will have to add that.  The only thing I was working from was the  Accordion Menu page. http://localhost:3000/accordion-menu.html  I will update the other stuff in the next day or two.  Thank you for pointing that out.

sethkane:
@kball Can you check it again, I just made some more updates.

kball:
@sethkane pulled into the PR... looking better!  Awesome!  One thing I'm noticing is that this overrides the current focus behavior in 2 ways.

1. Previously we appear to depend on browser/user styles for focus (e.g. on chrome is a blue outline).
2. Previously we hide that focus style using whatinput for mouse interactions.

I think it is very reasonable to consider number 1... looking at what the behavior is without the override, its ugly and distracting (at least on chrome).  However, if we're doing this, I think we should wrap up the focus styles in some variables so people have control over them, and then apply that consistently across all of our components.  @Owlbertz @brettsmason @andycochran what do you think?

For number 2, that seems to me like it is probably a bug, and we should re-exert the behavior of not showing the focus ring for mouse interactions.

sethkane:
@kball I added a custom focus indicator because at the time it was part of the conversation, I am happy to remove it and just allow the default browser manage that.

Let me make that update now.

sethkane:
@kball Just updated the branch

kball:
Awesome, pulled in.  This is looking really good to me now.  @Owlbertz what's your sense, is this good to go?

Owlbertz:
I will have a look at the updated branch again.

kball:
I just tested again... I'm super stoked about this.... @Owlbertz just waiting on your signoff ;)

Owlbertz:
@kball went through it and it looks good!

kball:
@sethkane @Owlbertz a couple of issues that came up with this... first, it breaks the responsive menus feature, I think because it changes these menus to behave differently from our other menus... second the class `.tree` is not especially descriptive and mostly duplicates `.menu`.

I'm sorry I didn't catch this earlier - I only really started digging into responsive menus recently, and realizing how awesome it is to have consistent markup and easy responsiveness across all of these menu plugins.

I *think* that making the internal pieces of the menu that don't link off into spans makes sense, but it seems like if we're going to do that we should change it universally across the menus.  @brettsmason @Owlbertz @andycochran what do you think?

On the second point, was there a reason for changing the class name and having duplicate styles across `.menu` and `.tree` all over the place?

kball:
Unfortunately based on the above breakage (and how close we are to 6.3) I think I'm going to need to revert this for now... we're talking about doing a large menu refactor for 6.4, which could include this and other related changes across all of the menus.

Geczy:
@kball Can we revisit this now that 6.3 is already released?

Issue #9291, opened 2016-10-25
is-reveal-open class deleted if you call second reveal from other one with MotionUI


artems-hyuna:
1. Open any Reveal subwindow (.is-reveal-open added to `<body>`).
   `Login` link in example below.
2. Click link or button in this Reveal window to open other Reveal window.
   `Forgot your password?` link in example below.
3. MotionUI features must be added:
   `Foundation.Reveal.defaults.animationIn = "slide-in-left";
   Foundation.Reveal.defaults.animationOut = "slide-out-left";`

**What did you expect to happen?**
`<body class="is-reveal-open">`

**What happened instead?**
`<body class="">`
Result: vertical scroll appear.

[CodePen](http://codepen.io/anon/pen/LRaLor)

kball:
Thanks for the bug report and repro case!  Will take a look.

artems-hyuna:
@kball Tried your solution. It doesn't work.

colin-marshall:
Resolved in #9295

PR #9295, opened 2016-10-25
Fix a race condition when reveal is using animations

kball:
Fixes issue #9291


kball:
@Owlbertz @colin-marshall can one of you review this and merge if it looks good?  Thanks

colin-marshall (approved):
LGTM

Issue #9290, opened 2016-10-24
Flex Block Grid doesn't work in Safari 8


Visiluna:
**How can we reproduce this bug?**
1. Step one

Assuming page already includes all the necessary bits to support flex grids, write this code.

```
<div class="row small-up-1 medium-up-2 large-up-3">
    <div class="column">Item 1</div>
    <div class="column">Item 2</div>
    <div class="column">Item 3</div>
    <div class="column">Item 4</div>
    <div class="column">Item 5</div>
    <div class="column">Item 6</div>
</div>
```
1. Step two

View above code in Safari 8.0.x.

**What did you expect to happen?**

At the small breakpoint, I should see six rows of one item each. At the medium breakpoint, I should see three rows of two items. At the large breakpoint, I should see two rows of three items.

**What happened instead?**

At all breakpoints, I see one row containing all six items with no wrapping.

**Test case:**

Here's a CodePen of the above, for viewing in Safari 8.0.x
- [CodePen with Foundation 6, Flexbox grid and MotionUI](http://codepen.io/visiluna/pen/kkqXbL)

kball:
I'm not seeing this on that codepen using browserstack with Safari 8.0.8... can you confirm more about your environment?  Also can you double check that the Foundation CSS loaded for you ok?

Visiluna:
Sorry. I've updated the CodePen to `@import` the `app.css` file generated by SASS (libsass and autoprefixer) from the Foundation 6 Basic Template. I have changed a view settings like colors, fonts, etc., but it does not contain any custom CSS other than that generated by Foundation Basic Template SASS. I have omitted Foundation components I'm not using (Orbit, Slider, etc.).

My site does use custom CSS, but I have omitted that to eliminate it as a possible cause for this problem.

Interestingly enough, if I copy the contents of `app.css` directly into CodePen's CSS box (instead of `@import`ing it), it seems to work correctly. However, using `@import` in CodePen produces the same result I'm seeing on the live site. (I don't use `@import` on the live site. I use a `<link>` tag.)

If you want to see the live page in context, you can view it [here](http://cpeak.dev.visiluna.com/members).

kball:
Ok, able to reproduce it.  Found a couple other holes in flexbox for safari 8 as well.  If you're building from SCSS you can fix it by adding 'ios >= 7' to your autoprefixer configuration.  Will submit an update, built version should be updated for 6.3.

PR #9296, opened 2016-10-25
Add ios >= 7 to autoprefixer

kball:
Fixes https://github.com/zurb/foundation-sites/issues/9290 and brings the build config for the css distribution up to par with what we're now including in the template.


dragthor:
I think this change should also be included with the customizer.  We got bit assuming it `ios > -7` was in the customizer.  Want me to add and create a pull request?

PR #9298, opened 2016-10-26
Bug fixes for sticky

karland:
I have added a test-case for sticky, that is showing a simple menu. The test case reveals two issues:
1. Refreshing the page or pressing the links shows an artefact (only visible in blinks) which is fixed with changing the `position` from `absolute` to `relative`
2. The menu, when stuck jumps to the right, since the `max-width` calculation does not take account of the padding-left from the parent.

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -273,7 +273,8 @@ class Sticky {
     var _this = this,
         newElemWidth = this.$container[0].getBoundingClientRect().width,
         comp = window.getComputedStyle(this.$container[0]),
-        pdng = parseInt(comp['padding-right'], 10);
+        pdngl = parseInt(comp['padding-right'], 10),
+        pdngr = parseInt(comp['padding-right'], 10);
 
     if (this.$anchor && this.$anchor.length) {
       this.anchorHeight = this.$anchor[0].getBoundingClientRect().height;
@@ -282,7 +283,7 @@ class Sticky {
     }
 
     this.$element.css({
-      'max-width': `${newElemWidth - pdng}px`
+      'max-width': `${newElemWidth - pdngl - pdngr}px`
     });
 
     var newContainerHeight = this.$element[0].getBoundingClientRect().height || this.containerHeight;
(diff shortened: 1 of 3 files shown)


kball:
This looks good to me.  @colin-marshall you looked at sticky a lot - do you see any issues with this?

mlang38:
Changing the `position` from `absolute` to `relative` introduces a bug. The menu is now positioned to its **own** initial position, not relatively to its container position, and therefore doesn't follow its container anymore.
I have a header which has its height set in `vw` unit, so it shrinks in height when the viewport gets thinner. The sticky container follows this movement, but the menu doesn't, so the content below it starts to slip under it. I think the position should be `absolute`. (Tested on both FF and Chrome)

ncoden:
I agree with @mlang38. `position: relative` should only be used to be a reference of its absolute childrens, not to position an element.

ncoden:
@karland I cannot reproduce the first bug, and it seems the fix you made create a new bug. Can you give more informations about this bug, with maybe a GIF ?

karland:
I understand the argument of @ncoden. At the time, I tried to fix an artefact not a bug, i.e. a part of the Menu moved in from the right, which was looking quite odd. However, the moving part ended up at the right position. The artefact seems to be visible only on slow browsers and I created a test case for it. Maybe the artefact has in the meantime disappeared due to changes in sticky?

Anyway, if removing this artefact introduces a bug, `position: absolute` should be reintroduced.

kball:
The change from absolute to relative fixed a real FOUC issue for sticky... prior to the JS being run, having the element absolutely positioned can cause the container to have no height.

From what I can see, .sticky once it is initialized will always have either `.anchored` or `.is-stuck`. `.anchored` is now also position relative.  What if `.sticky` gave no position attribute at all, but we change `.anchored` to be absolute? Will that resolve both issues?

mlang38:
Indeed I have the FOUC issue. Perhaps could we make `position: absolute` by JS after the container height is first computed and applied ?

Issue #7620, opened 2015-12-18
Move from Ruby scss-lint to Node sass-lint


gakimball:
There's now a Node-powered Sass linting library, appropriately called [sass-lint](https://github.com/sasstools/sass-lint). Right now we're using gulp-scss-lint, which is a wrapper around the Ruby-based [scss-lint](https://github.com/brigade/scss-lint).

Let's migrate our linting config file to sass-lint and drop that Ruby dependency.

colin-marshall:
I'm on it.

colin-marshall:
Using this: http://sasstools.github.io/make-sass-lint-config/

It came back that these are not supported yet in sass-lint:
```yml
# The following scss-lint Linters are not yet supported by sass-lint:
# ElsePlacement, PropertyCount, SelectorDepth, UnnecessaryParentReference
# VendorPrefixes, Compass::*
# 
# The following settings/values are unsupported by sass-lint:
# Linter Indentation, option "allow_non_nested_indentation"
# Linter Indentation, option "character"
# Linter PropertySortOrder, option "separate_groups"
# Linter SpaceBeforeBrace, option "allow_single_line_padding"
```

Should I still proceed or should we wait until these are added to sass-lint?

gakimball:
Hmm, I like SelectorDepth, but maybe if a big open source project adopts sass-lint we can press them to add more linters ;)

Go ahead and keep working on the implementation. I'll do some linting with the existing file today just so everything is up-to-date.

colin-marshall:
I know they plan on adding PropertyCount: https://github.com/sasstools/sass-lint/issues/406

I will open an issue over there and see what's up.

colin-marshall:
@gakimball: We got a great response from @DanPurdy: https://github.com/sasstools/sass-lint/issues/457#issuecomment-165935624

I can open up some new issues over there for specific Linter requests. Besides SelectorDepth are there any other important ones you would like before the switch can happen?

gakimball:
We can do the switch right now and I'll keep tabs on the library as it develops.

gakimball:
Also it looks like `nesting-depth` covers what `SelectorDepth` is doing.

colin-marshall:
There is a problem with making the switch right now. sass-lint throws a parse error when there is a string interpolated on a CSS property so it doesn't complete the linting process. I opened a new issue on the gonzales-pe repo to get this fixed: https://github.com/tonyganch/gonzales-pe/issues/140

[...]

PR #9301, opened 2016-10-27
Migrate from scss-lint to sass-lint (#7620)

colin-marshall:
This one has been in the works for nearly a year as we waited for sass-lint to mature but it's finally ready thanks to @DanPurdy, @bgriffith, @tonyganch and the @sasstools team! (See #7620 for history)

We drop a ruby dependency with this migration, which might be the last one remaining in Foundation? @zurb/yetinauts please take a look when you get a chance. Not all of the linting rules were the same from scss-lint to sass-lint so feel free to suggest any changes. Check the sass-lint [Rules Documentation](https://github.com/sasstools/sass-lint/tree/master/docs/rules) for further info.

This PR depends on:
- [x] #9335


gulp/lint.js
```diff
@@ -13,7 +14,12 @@ gulp.task('lint', ['lint:sass', 'lint:javascript']);
 
 gulp.task('lint:sass', function() {
   return gulp.src(PATHS)
-    .pipe(scssLint());
+    .pipe(plumber())
```
DaSchTour: Just out of curiosity. What is this line doing?

ncoden: `plumber()` make the script ignoring `sass-lint` error and continue the compilation

abdullahsalem:
I've not looked at the code yet, but I encourage this orientation.

DanPurdy:
Just a mention that the tool to migrate your config looks to be a bit out of date, there's plenty of new rules that should now provide direct equivalents.

Also 1.10 looks to be dropping in the next week or so which adds a few new rules, a whole host of fixes and the ability to 'FINALLY' disable linters via comments.

We're glad it's finally able to happen for you all!

kball:
Awesome!  Lets get this to happen for 6.3.  Lets see if we can incorporate @DanPurdy 's feedback and get the config file a little more directly equivalent, but 100% this should be in.

colin-marshall:
@DanPurdy you're right the config file is old. I made it back when I first started this branch last year. I will have another go at it. Thanks again for your help.

@kball sounds good! Do you want me to target the 6.3 branch instead then?

kball:
@colin-marshall develop is fine - 6.3 will be our next release.  Excited to see you back on github! :)

gulp/lint.js
ncoden: I think it is. It is not "compilation errors", but only "formating warnings". It shouldn't interrupt the compilation.

colin-marshall:
@DanPurdy the only rule I could find an equivalent for was `VendorPrefix` -> `no-vendor-prefixes`. Here's what I'm left with:

```yml
# The following scss-lint Linters are not yet supported by sass-lint:
# ElsePlacement, PropertyCount, SelectorDepth, UnnecessaryParentReference
# 
# The following settings/values are unsupported by sass-lint:
# Linter Indentation, option "allow_non_nested_indentation"
# Linter Indentation, option "character"
# Linter PropertySortOrder, option "separate_groups"
# Linter SpaceBeforeBrace, option "allow_single_line_padding"
```

What are some of the other equivalent rules for the above? I see you're still working on PropertyCount. Thanks!

ncoden:
@colin-marshall I work on a fork of your branch `colin-marshall:sass-lint` to reorder properties, following https://gist.github.com/ncoden/d42f55df7c7970f548a02cd3468f9c86.

DanPurdy:
@colin-marshall There was a bug with the AST that was causing fairly annoying problems with the property count rule in .sass format. It's been on hold for quite a while because of that. The good news is v4.0 of the AST will be landing very very soon which fixes all the remaining issues that blocked it, I'll finish it then. 

Our 1.10 update should be landing within the next week too with even more rules and the long awaited ability to disable lint warnings etc via comments (finally!).

colin-marshall:
@ncoden awesome! Updated the PR with your sort order. Just out of curiosity, is this a common/recommended order that other developers use or just what you prefer personally?

@DanPurdy no worries, just wanted to make sure I wasn't missing anything else. Looking forward to 1.10!

.sass-lint.yml
```diff
@@ -0,0 +1,347 @@
...
+  placeholder-name-format:
+    - 1
+    - convention: '([a-z0-9]+-?)+'
+  property-sort-order:
+    - 1
+    - ignore-custom-properties: false
+    - 
+      order :
+        # Position
+        - position
```
ncoden: Shouldn't we add `'` (single quotes) on properties names ?

.sass-lint.yml
```diff
@@ -0,0 +1,347 @@
...
+  no-trailing-zero: 1
+  no-url-protocols: 1
+  no-vendor-prefixes: 1
+  placeholder-in-extend: 1
+  placeholder-name-format:
+    - 1
+    - convention: '([a-z0-9]+-?)+'
+  property-sort-order:
+    - 1
+    - ignore-custom-properties: false
```
ncoden: Please set `ignore-custom-properties` to `true`. We can't handle all CSS properties.

.sass-lint.yml
colin-marshall: I went by this example, which uses no quotes.
https://github.com/sasstools/sass-lint/blob/master/docs/rules/property-sort-order.md

Can add them if you think it's necessary.

ncoden:
@colin-marshall This is my mix between [the tree standards I found](https://github.com/sasstools/sass-lint/tree/develop/lib/config/property-sort-orders) : `smacss` (global organization), `recess` (picked missing properties) and `concentric` (from disposition/ouside to appearance/inside).

So yes, this is sort of "my" sort order. I have too much critics to do to others standards.

.sass-lint.yml
colin-marshall: @ncoden can you elaborate on why this is this please? If this was linting being done in the template I see where setting this to true would be necessary, since the end user can throw any property in there, but this linting will have no effect on what the end user does.

.sass-lint.yml
ncoden: Their internal config files use quotes : https://github.com/sasstools/sass-lint/tree/develop/lib/config/property-sort-orders

I think it is better because of properties are following *external* naming rules (CSS), and not YML ones. Even if in practice it don't change anything.

ncoden:
@colin-marshall Actually, you didn't have to make these changes for `property-sort-order`. This rules wasn't used with `scss-lint` before. You can only migrate to `sass-lint` without too much rules changes, I am adding `property-sort-order` and all the required changes in SCSS files in my branch.

colin-marshall:
@ncoden even so I think `property-sort-order` would be useful for us to have here as an easy way to "enforce" consistency in how Foundation's SCSS is written.

ncoden:
I'm not sure to understand you. Like I said, I am adding property-sort-order and all the required changes in SCSS files in my branch. So i'm for a consistency too, but changing the lint rules is not the subject of *this* PR.

colin-marshall:
@ncoden ohhhh ok. I was confused when you said remove it and thought you meant you were adding the sort orders for your own personal use in your projects and we didn't need to add them here. The fact it came from a gist threw me off. I now see you just meant you were working on the rules in your own branch for a separate PR. My bad! I will revert that commit.

ncoden:
@colin-marshall After `position, disposition, margin, padding, global appearance...`, do you prefer:
- `... border, background, shadow ...` (smacss)
- `... shadow, border, background ...` (concentric)
- `... background, border, shadow ...` (recess / classic)

?

(poke @kball)

ncoden:
@colin-marshall Don't forget to remove `gem install scss_lint` from `.travis.yml`
https://github.com/zurb/foundation-sites/blob/6c7b0b4af5502ca5e2495f9c50caedb398bd2d12/.travis.yml#L9

.travis.yml
```diff
@@ -1,9 +1,8 @@
 language: node_js
 node_js:
-  - "0.12"
+  - "4.0"
 before_install:
   - rvm install 2.2.2
```
colin-marshall: @zurb/yetinauts does anybody know if we need this `rvm install` script for anything besides scss-lint? If that's all it's being used for I think I will just remove it in this PR.

ncoden: There is no other possible uses of rvm than for our own gems, isn't it ?

colin-marshall: @ncoden none that I'm aware of but I am not 100% sure that scss-lint is the last thing Ruby related in the repo so I wanted to verify that before removing rvm in the travis scripts.

.sass-lint.yml
```diff
@@ -0,0 +1,128 @@
...
+  trailing-semicolon: 1
+  url-quotes: 1
+  variable-for-property:
+    - 0
+    - properties: []
+  variable-name-format:
+    - 1
+    - allow-leading-underscore: true
+      convention: hyphenatedlowercase
+  zero-unit: 1
```
ncoden: Line end

colin-marshall: Fixed. Thanks!

ncoden (approved):

ncoden:
:tada:

Issue #9250, opened 2016-10-12
Print styles


anselmdk:
I find myself again and again jumping hoops to get Foundation 6 based sites to print properly.
Am I missing something here, or could we do some improvements in that area?

See attached for a very simple example, and also https://github.com/zurb/foundation-sites/issues/8992

I think for the very least showing the urls should be configurable in the `_setting.scss` and by default the large grid should show when printing.

# Browser:

![](https://cloud.githubusercontent.com/assets/1316533/19299933/88915a70-9057-11e6-8e68-b3b08977ba59.png)

# Print Preview:

![](https://cloud.githubusercontent.com/assets/1316533/19299931/88765ca2-9057-11e6-8ca9-96f253a051a4.png)

kball:
This is interesting... there's a couple of things going on. 

Looking at how all of the breakpoints etc are defined, they all include a 'screen only' as a part of their media query... so by default things are going to lay out essentially ordered as they would on a small screen, but only using the default styles.

It also looks like the ONLY thing that is configurable right now in the print typography styles is whether background should be transparent or not.

This definitely feels like a set of oversights... it kind of feels like you should be able to specify if you want things to lay out via the grid or not, and probably have a set of additional common configurations (like link styles) available via settings.

@ncoden @brettsmason @andycochran interested in your thoughts

ncoden:
Currently, at the end of http://foundation.zurb.com/sites/docs/ :

![Zurb Foundation screeshot](https://cloud.githubusercontent.com/assets/9939075/19370762/ffcac138-91ad-11e6-9c6f-e2c114fdf4a5.png)

andycochran:
I'm finding it difficult to determine the best approach for print styles without making app-specific assumptions. 

It'd be grueling to have to comb through markup and apply classes for print (e.g. `hide-for-print`) like we do for screen readers (e.g. `show-for-sr`). I often keep a list of components that I don't want to print and apply a `display:none` to the group in a print stylesheet. I wonder if that could be automated or abstracted somehow. 

Maybe the grid could be handled with print-specific classes: 
```html
<div class="row">
  <div class="columns medium-3 large-4 print-4">…</div>
  <div class="columns medium-9 large-8 print-8">…</div>
</div>
```
…or maybe a settings variable would be better: 
```sass
$print-breakpoint: large;
```

ncoden:
[...]

PR #9304, opened 2016-10-27
Print Styles

andycochran:
Adresses #9250. 
- Adds `print,` to the media types in media queries for named breakpoints up to and including the one matching `$print-breakpoint`
- Adds a settings variable to disable printing href values after link text

### DIFF
--- a/scss/typography/_print.scss
+++ b/scss/typography/_print.scss
@@ -6,6 +6,7 @@
 /// @type Boolean
 /// @group global
 $print-transparent-backgrounds: true !default;
+$print-hrefs: true !default;
 
 // scss-lint:disable all
 
@@ -36,7 +37,9 @@ $print-transparent-backgrounds: true !default;
     // Display the URL of a link after the text
     a,
     a:visited { text-decoration: underline;}
-    a[href]:after { content: ' (' attr(href) ')'; }
+    @if $print-hrefs {
+      a[href]:after { content: ' (' attr(href) ')'; }
+    }
 
     // Don't display the URL for images or JavaScript/internal links
     .ir a:after,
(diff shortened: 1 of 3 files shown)


kball:
This is great!  Nice work @andycochran.  @anselmdk can you try this out for your use case?

anselmdk:
Nice guys, I'll try it out shortly.

Owlbertz:
Because 6.3 will be the next release and the v6.3 branch already got merged into develop, shouldn't this PR (and all other new ones) target develop rather than v6.3?

andycochran:
Ack! I did not know that 6.3 was in develop now. @kball let me know if I need to retarget this.

andycochran:
@brettsmason I'd be interested to hear what you think about the `@function map-to-list` that I added. The list `$-zf-breakpoints-keys` basically gives us a way to compare the indexes of the `$print-breakpoint` and whichever breakpoint passed to the breakpoint mixin. Can you think of a more efficient way to compare breakpoint indexes? I suppose the `map-to-list` function could be simplified, but it's a nice function that could be reused (I don't think it's too bloaty).

brettsmason:
No problem I'll take a look this evening when I'm home

brettsmason:
@andycochran The PR looks great to me, nice job!
That function looks especially useful, and might be of use for the whole vertical rhythm problem I'm working on. I would rather keep the function as you have it, rather than stripping it back.

Thumbs up from me 👍

kball:
I'm keeping 6.3 and develop in sync for right now... by our previous policy develop should be for next release, and 6.3 is our next release, however we have some outstanding PRs against 6.3, so for now either is fine.

kball:
Love the way this turned out.  Thanks for putting it together @andycochran!

anselmdk:
Just added a screenshot to https://github.com/zurb/foundation-sites/issues/9250 of how my initial use case looks with the new code - great work guys!

Issue #9307, opened 2016-10-28
input-group sizing wrong when using the file picking styling suggested


jellisii:
**How can we reproduce this bug?**
1. Use input-group with a file input type demonstrated https://foundation.zurb.com/sites/docs/forms.html#file-upload-button

**What did you expect to happen?**

I expect the button, label, and textbox to all be the same size.

**What happened instead?**

The button and label are not the same size as the textbox. 

**Test case:**

http://codepen.io/jellisii/pen/dpEayx

This may be related to #8818

andycochran:
This is fixed in #9308.

PR #9308, opened 2016-10-29
Force input-group buttons to match input size

### LINKED ISSUE #7373: [Forms] Input Group button sizing variation 6.0.5

andycochran:
This PR closes #7373 and #9307, by making sure that buttons in input groups match the size of inputs, regardless of button settings vars.

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -114,7 +114,8 @@ $input-prefix-padding: 1rem !default;
 
     a,
     input,
-    button {
+    button,
+    label {
       @extend %input-group-child;
       font-size: $input-font-size;
       height: $height;
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -115,7 +115,11 @@ $input-prefix-padding: 1rem !default;
     a,
     input,
     button {
-      margin: 0;
+      @extend %input-group-child;
+      font-size: $input-font-size;
+      height: $height;
+      padding-top: 0;
+      padding-bottom: 0;
     }
   }
 


kball:
So I think given our current approach, this is good.  I don't know the backstory behind making .input-group-button a wrapper, but it seems a little odd to me.  

Would be interested in @ncoden's opinion on if things like this (styling tags based on container classes) this makes sense in a world where we're trying to move things as much as possible to be purely class based.

ncoden:
That's *useful*, it's true. But we **can't** assume that everyone will expect this behavior to all `<button>` tags. That's why we should use classes and nothing else as selectors. It's the only way we can ensure that nobody is blocked with unwanted properties (which are always very hard to cancel).

About *styling something based on container classes*: if you want something maintainable, you can style the dom element:
- as an BEM *element* par of a *block*
- as a child of a [scope](http://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/) form.

kball:
Ah interesting... so based on that writeup, the approach taken in this PR (and also in this class prior to this PR) is to treat `.input-group-button` essentially as creating a scope.

Given this is extending an existing paradigm (and only applies within that `.input-group-button` pseudo-scope), I think this is still mergeable... but I'm seeing more and more clearly the impetus to redesign our css architecture.

ncoden:
Scope is a particular thing that is relevant in only a few situations, where only `<tags>` are expected (no classes, no custom styles...) or advanced nesting. For example an article (`<h1>...<h6>`, `<p>`, `<b>`...) or a standard form (`<input>`, `<select>`...).

A *scope* is not a component. It *contain* components that all have the same semantic context, its only purpose is to simplify their creation (no classes needed). It does not make sense to consider `.input-group-button` as a scope. I was talking generally about *styling something based on container classes*.

If you want to apply properties (that are not resets) on buttons or one if its sub-elements, apply it with classes and try to keep an 1-level specificity.

andycochran:
I think it can be rightly assumed that `button` elements within an `.input-group-button` should be treated a certain way.

kball:
Yeah, I think within the paradigm we have set up with v6, this is a go. This is one of many areas we'll want to revisit for the v7 rearchitecture.

PR #9315, opened 2016-11-01
Adds pow and log transforms to the Slider component

murb:
Introduces two settings, `nonLinearBase` and `positionValueFunction` which change the way in which the position of the slider is translated to an actual value. 

Besides the default value of ‘linear' for `positionValueFunction`, 'pow' and a 'log' can be used.

*Example application*

In example below we want to focus on the most recent years, hence we use a logarithmic transformation of the slider:

          <div class="slider" data-slider data-initial-start="2010" data-initial-end="2015" data-step="1" data-start="1950" data-end="2016" data-position-value-function="log">
            <span class="slider-handle" data-slider-handle role="slider" tabindex="1" aria-controls="car_search_year_from"></span>
            <span class="slider-fill" data-slider-fill></span>
            <span class="slider-handle" data-slider-handle role="slider" tabindex="1" aria-controls="car_search_year_to"></span>
          </div>

Since I didn't get the test-framework to run I submitting this for now without tests, if it is absolutely necessary to do so, and this functionality is desired, allow me some time to fix that.

### DIFF
--- a/docs/pages/slider.md
+++ b/docs/pages/slider.md
@@ -140,7 +140,7 @@ Alternatively there is also a `pow`-type position value function available, maki
 
 ```html_example
 <div class="small-10 columns">
-<div class="slider" data-slider data-initial-start="50" data-step="1" data-position-value-function="log" data-non-linear-base="10">
+<div class="slider" data-slider data-initial-start="50" data-step="1" data-position-value-function="log" data-non-linear-base="5">
   <span class="slider-handle" data-slider-handle role="slider" tabindex="1" aria-controls="sliderOutputNonLinear"></span>
 </div>
 </div>
@@ -150,4 +150,4 @@ Alternatively there is also a `pow`-type position value function available, maki
 </div>
 ```
 
-The nonLinearBase-option is optional and defaults to 10.
+The nonLinearBase-option is optional and defaults to 5.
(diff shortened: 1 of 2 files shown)


kball:
I think this is really cool... I'd like to see at least an example on the docs page for the slider which we could also use for testing.

With the refactoring happening already, I almost wonder if we could allow a user-defined function as well as these 3 pre-defined functions... this would of course require using the JS api and not the html data-attributes api, but seems like it might well be worthwhile... @coreysyms @colin-marshall what do you think?

murb:
Just added an example to the docs :)

kball:
OK, this looks good. One thing I notice is that the nonlinear example is hard to select every value down at the low end, but I think that is the nature of this type of reaction function... would love to get one more @zurb/yetinauts opinion on this, but I think it is probably good to go

coreysyms:
Trying out the "inline comments" in the code so we will see if that works, but just in case; I made a note that the docs say default is 5 for "base" but in the code, it's 10. This is really cool, nice work, and works great as far as I can see.

murb:
Fixed the doc; and yes it is harder to select values at the either the beginning or the end, making it actually easier to select it at the other side, which is as intended (use cases: maximum mileage selector for an old car, or a period selector where you want to focus on the more recent years (after a certain point the precise value doesn't matter)). As to whether it is a good idea to add support for custom functions... not sure whether it is ok to hide it away in a javascript API (I prefer the data values myself at least :) )... maybe some extra default functions could be added in the future.

PR #9316, opened 2016-11-01
Pr top bar title docs

### LINKED ISSUE #9029: v6.2.3 - .top-bar-title strong is missing styling

mattferderer:
This is a PR to fix #9029 & enhance #7907. This is hopefully a nicer, cleaner PR to replace PR #9039 which I made a bit messy. @kball 

Let me know if there is anything else I can do on this one.

### DIFF
--- a/scss/components/_top-bar.scss
+++ b/scss/components/_top-bar.scss
@@ -20,7 +20,7 @@ $topbar-submenu-background: $topbar-background !default;
 
 /// Spacing for the top bar title.
 /// @type Number
-$topbar-title-spacing: 1rem !default;
+$topbar-title-spacing: 0.45rem 1rem 0 1rem !default;
 
 /// Maximum width of `<input>` elements inside the top bar.
 /// @type Number
@@ -144,7 +144,7 @@ $topbar-unstack-breakpoint: medium !default;
   @if $global-flexbox {
     .top-bar-title {
       flex: 0 0 auto;
-      margin-right: $topbar-title-spacing;
+      margin: $topbar-title-spacing;
     }
 
     .top-bar-left,
@@ -155,8 +155,8 @@ $topbar-unstack-breakpoint: medium !default;
   @else {
     .top-bar-title {
       float: left;
-      margin-right: $topbar-title-spacing;
-
+      margin: $topbar-title-spacing;
+      display:inline-block;
     }
 
     .top-bar-left {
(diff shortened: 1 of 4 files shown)


kball:
@ncoden @andycochran @brettsmason you reviewed #9039, does this look good to go?

scss/components/_top-bar.scss
```diff
@@ -155,8 +155,8 @@ $topbar-unstack-breakpoint: medium !default;
   @else {
     .top-bar-title {
       float: left;
-      margin-right: $topbar-title-spacing;
-
+      margin: $topbar-title-spacing;
+      display:inline-block;
```
ncoden: @kball  BTW we should apply that to all Foundation

mattferderer: @ncoden - The change has been fixed but just out of curiosity, were there any public discussions on the topic of display order for Foundation? I would be interested in reading what made this decision & what the pros/cons are to see if they are something my team might be interested in adopting in our projects.

ncoden: @mattferderer Currently, there is no standard for properties order in Foundation. I'm trying to improve the code standard of the framework and your PR make me think of *properties order*.

I prefer ordering CSS properties by *type* instead of alphabetically because it is easier to:
- Find a property. You know the property you are searching better by what it does (approximately) than its name.
- Understand the global component behaviour. You see and distinguish immediately the different "groups" of properties, and each property is next to those related to it.
- Understand the consequences of each property on others. Properties are ordered from the *nature/inside* to the *appearance/outside* of the component, the first having consequences on the seconds.

You can see also: https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

mattferderer: I happened to glance at that exact same article right before asking the above question. Thanks for the info.

ncoden: @kball What do you think ?

Issue #7551, opened 2015-12-15
[Orbit] slide and container height on resize


designerno1:
Slide and container height doesn't get refreshed on resize

designerno1:
a autoHeight function for each slide would be great

zurbrandon:
This isa known issue we're hoping to get to soon, if you have expercine with this at all a PR would be super helpful.

designerno1:
a very very simple workaround is to set .orbit-container to height: auto !important;

SimonHeimberg:
related: #7815 #7982

kball:
@designerno1 @SimonHeimberg I took a pass on fixing this in #9317.  Would appreciate if you could take a look and see if it addresses your issues.

kball:
Fixed in #9317

PR #9317, opened 2016-11-01
Fix resize behavior of orbit

### LINKED ISSUE #7815: [Orbit] heights should change on resize
max-height and height should be recalced on resize
or there should be a funktion to recalc it _prepareForOrbit() does not work for me

kball:
Fixes #7815, #7551 by adding letting slides be sized purely in css and then adding a resize handler that resizes the container.

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -48,8 +48,15 @@ class Orbit {
 
     this.$wrapper = this.$element.find(`.${this.options.containerClass}`);
     this.$slides = this.$element.find(`.${this.options.slideClass}`);
+
     var $images = this.$element.find('img'),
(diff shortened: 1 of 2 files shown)


DanFeather:
I'm curious. Is there any way to use the JS and CSS files from this branch (i.e., just the orbit-related files) in a Foundation 5 website? I'm using Orbit there and hate the FOUC. I've seen some hacks in the Foundation forum going back to Foundation 3 and 4, but I'd like to use the newer Orbit code if possible.

ncoden:
@DanFeather You can try looking at the changes in this PR and applying them to the Foundation 5 code on your own repo. But we are not maintaining F5 anymore as we began working on F7.

PR #9318, opened 2016-11-01
parametrize menu arrows and vertical alignment

pine3ree:
What this PR does:
- add a global parameter for menu arrow size `$global-arrow-size: 6px`
- add individual parameters for dropdown-menu, drilldown menu and accordion-menu (`$dropdownmenu-arrow-size`, `$drilldown-arrow-size`, `$accordionmenu-arrow-size`) . All of them inherits from global arrow size (note: the original foundation arrow size for dropdown-menu was 5px not 6px as in the other plugins). They can be also individually set to different sizes.
- use `transform: translateY(-50%)` to achieve vertical center alignment

demo: http://codepen.io/anon/pen/MjNzEx
- in the demo a span element is added inside all anchors to highlight the arrow position and a bottom margin is added to all menu list items
- the compiled default foundation.css is inserted inline in codepen

### DIFF
--- a/scss/components/_accordion-menu.scss
+++ b/scss/components/_accordion-menu.scss
@@ -10,23 +10,27 @@ $accordionmenu-arrows: true !default;
 /// @type Color
 $accordionmenu-arrow-color: $primary-color !default;
 
+/// Sets accordion menu arrow size if arrow is used.
+/// @type Length
+$accordionmenu-arrow-size: 6px !default;
+
 @mixin foundation-accordion-menu {
   @if $accordionmenu-arrows {
     .is-accordion-submenu-parent > a {
       position: relative;
 
       &::after {
-        @include css-triangle(6px, $accordionmenu-arrow-color, down);
+        @include css-triangle($accordionmenu-arrow-size, $accordionmenu-arrow-color, down);
         position: absolute;
         top: 50%;
-        margin-top: -4px;
+        transform: translateY(-50%);
         right: 1rem;
       }
     }
 
     .is-accordion-submenu-parent[aria-expanded='true'] > a::after {
       transform-origin: 50% 50%;
-      transform: scaleY(-1);
+      transform: scaleY(-1) translateY(50%);
     }
   }
 }
--- a/scss/components/_accordion-menu.scss
+++ b/scss/components/_accordion-menu.scss
@@ -23,14 +23,14 @@ $accordionmenu-arrow-size: 6px !default;
         @include css-triangle($accordionmenu-arrow-size, $accordionmenu-arrow-color, down);
         position: absolute;
         top: 50%;
-        transform: translateY(-50%);
+        margin-top: -1 * ($accordionmenu-arrow-size / 2);
         right: 1rem;
       }
     }
 
     .is-accordion-submenu-parent[aria-expanded='true'] > a::after {
       transform-origin: 50% 50%;
-      transform: scaleY(-1) translateY(50%);
+      transform: rotate(180deg);
     }
   }
 }
--- a/scss/components/_drilldown.scss
+++ b/scss/components/_drilldown.scss
@@ -18,6 +18,10 @@ $drilldown-arrows: true !default;
 /// @type Color
 $drilldown-arrow-color: $primary-color !default;
 
+/// Sets drilldown arrow size if arrow is used.
+/// @type Length
+$drilldown-arrow-size: 6px !default;
+
 /// Background color for drilldown submenus.
 /// @type Color
 $drilldown-background: $white !default;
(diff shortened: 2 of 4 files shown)


pine3ree:
Hello @kball ,
I added the parametrization of menu arrows (https://github.com/zurb/foundation-sites/pull/9090#issuecomment-257441483).

I tested locally both the float and the flex grid version with chrome, chromium, firefox, opera. 
The codepend demo is using the float-grid.
Please read the notes in the PR. I am not sure about the global arrow size, i just find it more elegant.
If we want to keep the original 5px for the dropdown-menu we can remove the global parameter.

kind regards

kball:
Seems reasonable to me. @brettsmason @ncoden @andycochran any thoughts?

ncoden:
I heard a lot of bad things about `transform`. I think it must be used only for visual modifications, not to center or move a semantical element.

To center something, see http://howtocenterincss.com/
In this case, you can apply a static or dynamic margin.

Static margin:
```scss
top: 50%;
margin-top: -($accordionmenu-arrow-size / 2);
```

Dynamic margin:
```scss
top: 0;
bottom: 0;
margin auto;
```
*Yes, that's awesome*.

This PR makes me notice that all menu arrows are duplicated across components in the generated CSS. I think we should rather `@extend` a `.arrow` component, or make it used manually.

About `$dropdownmenu-arrow-size`, `$drilldown-arrow-size`, `$accordionmenu-arrow-size` (...), I don't know. That's a great idea but we should keep it only if it is possible to avoid duplicated code.

pine3ree:
Hello @ncoden,
the original idea was to move up half the parametrize arrow size (like it was done in my previous pr https://github.com/zurb/foundation-sites/pull/9090), but i have seen using transform very often latelty.

PLS, Let me know if you want me to update the pr to the static version (which is actually what i have been actually using in my previous project)

The presence of arrows is now controlled via scss parameters (and the caret itself is generated by the css-triangle mixin using a pseudo element). Are you suggesting to trigger the caret using a class on the menu?

kind regards

ncoden:
Yes, it would be great to avoid `transform` in this case.

About the way we generate the triangle, what I want can involve breaking changes (poke @kball :smile:).

Solution with `@extend`:
```scss
@mixin icon-arrow(
    $color: $global-arrow-color,
    $size: $global-arrow-size,
    $direction
) {
        @extend %css-triangle-base;
        @include css-triangle-color($color);
        @include css-triangle-size($size);
        @include css-triangle-direction($direction);
}

...

.is-accordion-submenu-parent > a {
    position: relative;

    &::after {
        @include icon-arrow($accordionmenu-arrow-color, $accordionmenu-arrow-size, down);
    }
}
```

I also thought about only using classes, like:
```html
<ul class="vertical menu" data-accordion-menu>
  <li>
    <a href="#">
        Item 1
        <span class="icon-arrow icon-arrow-down"></span>
    </a>
  </li>
</ul>
```
But it's not customizable across components.

Is it really useful ? I think it is a very basic element that will be designed the same way across all the components.

We can also use a "mix" `@mixin` which uses a basic `%icon-arrow` (with global color and size), unless the properties given are not the default ones.

pine3ree:
@ncoden 
Another option would be adding a class `.has-arrows` to the `.menu` elements. Their nature, nesting level and attributes (as `vertical` class) could be used to determine arrows' direction.

a reason no to use transform would be having non-0 transition durations when rotating the carets on open/close events as in the accordion menu. It would not be nice see the caret sliding up/down :-D

kind regards

ps: I'll leave you thinking about the best solution and will catch-up tomorrow....it's past midnight here

ncoden:
Applying properties from a parent element or auto-determining them from the context causes a lot of problems due to a higher specificity.

It is past-midnight here too :D (Paris)

pine3ree:
i updated the online example as well: http://codepen.io/anon/pen/MjNzEx (the transition in the accordion menu example is added inline)

***notes***:
using static margins (we can do that here because we know the element height) requires a different margin size:
- for vertical carets (top level dropdown menu and accordion menu) the parametrized size represents its height so we move up half of it
- for horizontal carets (dropdown submenu and drilldown menu) the parametrized size represents half its height so we move up the actual parameter size

in the accordioon  menu I have replaced `scaleY(-1)` with a rotate transform as to me it's semantically more appropriate. I can revert it back if I missed the reason for using scale.

pine3ree:
Hello @ncoden (north-east Italy here!! :-D)

I am not sure we are targetting the very same subject, so maybe I could be a little off-topic in regards to your proposal and apologize for that. 

Here's what I was thinking: 

so far we can trigger the visibility of menu arrows via scss parameters. This is good because it also saves a few bytes in the compiled css. 

another way to trigger the presence of menu arrows is by adding a class to the top level menu but this would mean that either we always compile the arrow related css code (more bytes) or we require both setting the sccs parameters to true and using the extra top level menu class.

About the proposal of an arrow element I believe that i can be useful but that it should be an addon not a replacement of the way the menus are now working. The actual implementation allows us to choose the presense, the color (and hopefully the size) of menu arrows without extra html code and in my opinion it's a very handy feature. Having to add span elements for arrows should be given as a choice not a reqirement.

The one point I disagree is the use of `arrow-down` `arrow-left` etc etc. I do not object about having those classes (of course we need them to set the direction in a never-changing arrow), but their usage inside menus can result in non-semantic code.

For instance an arrow inside an accordion menu is initially a down-caret but it becomes an up-caret when the menu is open. A dropdown nested menu can open in the right direction as well as in the left direction, and it starts differently in a rtl context. I believe that in these cases the context dictates the behaviour of the arrow and having the framework do that means less work for the developer.

Pls let me know your opinions and if I misunderstood any/all of your points.

kind regards

ps
I've found vertical alignment via transform-translateY in the switch and tooltip components...but of course we don't have heights in those cases

PR #9319, opened 2016-11-01
Ally-compliant color palette & functions

andycochran:
This PR, which effects every component that used the `foreground()` function, addresses #7387. I've adjusted the `$foundation-palette` to meet AA contrast requirements. The `foreground()` function has been replaced with a new `pick-best-color()` function that uses luminosity instead of lightness (which fails as #7387 explains).

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -25,7 +25,7 @@ $global-lineheight: 1.5 !default;
 $foundation-palette: (
   primary: #1a7cbd,
   secondary: #767676,
-  success: #238748,
+  success: #3adb76,
   warning: #ffae00,
   alert: #cc4b37,
 ) !default;
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -23,11 +23,11 @@ $global-lineheight: 1.5 !default;
 /// Colors used for buttons, callouts, links, etc. There must always be a color called `primary`.
 /// @type Map
 $foundation-palette: (
-  primary: #2199e8,
-  secondary: #777,
-  success: #3adb76,
+  primary: #1a7cbd,
+  secondary: #767676,
+  success: #238748,
   warning: #ffae00,
-  alert: #ec5840,
+  alert: #cc4b37,
 ) !default;
 
 /// Color used for light gray UI items.
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -23,7 +23,7 @@ $global-lineheight: 1.5 !default;
 /// Colors used for buttons, callouts, links, etc. There must always be a color called `primary`.
 /// @type Map
 $foundation-palette: (
-  primary: #1a7cbd,
+  primary: #1779ba,
   secondary: #767676,
   success: #3adb76,
   warning: #ffae00,
--- a/scss/components/_accordion.scss
+++ b/scss/components/_accordion.scss
@@ -20,7 +20,7 @@ $accordion-title-font-size: rem-calc(12) !default;
 
 /// Default text color for items in a Menu.
 /// @type Color
-$accordion-item-color: foreground($accordion-background, $primary-color) !default;
+$accordion-item-color: pick-best-color($accordion-background, ($primary-color, $white)) !default;
 
 /// Default background color on hover for items in a Menu.
 /// @type Color
@@ -40,7 +40,7 @@ $accordion-content-border: 1px solid $light-gray !default;
 
 /// Default text color of tab content.
 /// @type Color
-$accordion-content-color: foreground($accordion-content-background, $body-font-color) !default;
+$accordion-content-color: pick-best-color($accordion-content-background, ($body-font-color, $white)) !default;
 
 /// Default padding for tab content.
 /// @type Number | List
--- a/scss/components/_accordion.scss
+++ b/scss/components/_accordion.scss
@@ -20,7 +20,7 @@ $accordion-title-font-size: rem-calc(12) !default;
 
 /// Default text color for items in a Menu.
 /// @type Color
-$accordion-item-color: pick-best-color($accordion-background, ($primary-color, $white)) !default;
+$accordion-item-color: $primary-color !default;
 
 /// Default background color on hover for items in a Menu.
 /// @type Color
@@ -40,16 +40,18 @@ $accordion-content-border: 1px solid $light-gray !default;
 
 /// Default text color of tab content.
 /// @type Color
-$accordion-content-color: pick-best-color($accordion-content-background, ($body-font-color, $white)) !default;
+$accordion-content-color: $body-font-color !default;
 
 /// Default padding for tab content.
 /// @type Number | List
 $accordion-content-padding: 1rem !default;
 
 /// Adds styles for an accordion container. Apply this to the same element that gets `data-accordion`.
-@mixin accordion-container {
+@mixin accordion-container (
+  $background: $accordion-background
+) {
   margin-#{$global-left}: 0;
-  background: $accordion-background;
+  background: $background;
   list-style-type: none;
 }
 
@@ -65,26 +67,32 @@ $accordion-content-padding: 1rem !default;
 }
 
 /// Adds styles for the title of an accordion item. Apply this to the link within an accordion item.
-@mixin accordion-title {
+@mixin accordion-title (
+  $padding: $accordion-item-padding,
+  $font-size: $accordion-title-font-size,
+  $color: $accordion-item-color,
+  $border: $accordion-content-border,
+  $background-hover: $accordion-item-background-hover
+) {
   position: relative;
   display: block;
-  padding: $accordion-item-padding;
+  padding: $padding;
 
-  border: $accordion-content-border;
+  border: $border;
   border-bottom: 0;
 
-  font-size: $accordion-title-font-size;
+  font-size: $font-size;
   line-height: 1;
-  color: $accordion-item-color;
+  color: $color;
 
   :last-child:not(.is-active) > & {
-    border-bottom: $accordion-content-border;
+    border-bottom: $border;
     border-radius: 0 0 $global-radius $global-radius;
   }
 
   &:hover,
   &:focus {
-    background-color: $accordion-item-background-hover;
+    background-color: $background-hover;
   }
 
   @if $accordion-plusminus {
@@ -103,18 +111,23 @@ $accordion-content-padding: 1rem !default;
 }
 
 /// Adds styles for accordion content. Apply this to the content pane below an accordion item's title.
-@mixin accordion-content {
+@mixin accordion-content (
+  $padding: $accordion-content-padding,
+  $border: $accordion-content-border,
+  $background: $accordion-content-background,
+  $color: $accordion-content-color
+) {
   display: none;
-  padding: $accordion-content-padding;
+  padding: $padding;
 
-  border: $accordion-content-border;
+  border: $border;
   border-bottom: 0;
-  background-color: $accordion-content-background;
+  background-color: $background;
 
-  color: $accordion-content-color;
+  color: $color;
 
   :last-child > &:last-child {
-    border-bottom: $accordion-content-border;
+    border-bottom: $border;
   }
 }
 
--- a/scss/components/_badge.scss
+++ b/scss/components/_badge.scss
@@ -53,7 +53,7 @@ $badge-font-size: 0.6rem !default;
       @if $name != primary {
         &.#{$name} {
           background: $color;
-          color: pick-best-color($color, ($badge-color, $badge-color-alt));
+          color: color-pick-contrast($color, ($badge-color, $badge-color-alt));
         }
       }
     }
--- a/scss/components/_badge.scss
+++ b/scss/components/_badge.scss
@@ -12,7 +12,11 @@ $badge-background: $primary-color !default;
 
 /// Default text color for badges.
 /// @type Color
-$badge-color: pick-best-color($badge-background) !default;
+$badge-color: $white !default;
+
+/// Default text color for badges.
+/// @type Color
+$badge-color-alt: $black !default;
 
 /// Default padding inside badges.
 /// @type Number
@@ -49,7 +53,7 @@ $badge-font-size: 0.6rem !default;
       @if $name != primary {
         &.#{$name} {
           background: $color;
-          color: pick-best-color($color, ($white, $black));
+          color: pick-best-color($color, ($badge-color, $badge-color-alt));
         }
       }
     }
--- a/scss/components/_badge.scss
+++ b/scss/components/_badge.scss
@@ -14,7 +14,7 @@ $badge-background: $primary-color !default;
 /// @type Color
 $badge-color: $white !default;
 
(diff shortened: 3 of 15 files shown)


andycochran:
While replacing `foreground()` with a function that uses luminosity, I noticed that Tabs is using `foreground()` for the content font color. There's really no reason for this, as the content background color is explicitly set. Now `$tab-content-color` is just simply set to `$body-font-color`. 

I also noticed that Tabs' mixins didn't allow for passing any vars when creating your own tabs. Fixed this too. 

Finally, I added settings for tab item font colors. You could previously set the tab items' background, but not their color.

andycochran:
Just like Tabs, it's the same case for Accordion & Pagination (foreground function sets the text color but the background color is explicitly set, and mixins don't pass vars). Fixing those too…

ncoden:
:+1: 
Wouldn't `color-contrast()` / `color-*()` / `pick-contrast-color()` (but I prefer a function with a namespace) be better than `pick-best-color()` ?

andycochran:
I can change that to whatever everyone agrees is most semantic. That's just what the function is called in [the gist](https://gist.github.com/voxpelli/6304812) from which I ported the luminosity functions. I've updated the function (and its description which was old/wrong) based on your feedback. Thanks, @ncoden!

kball:
Travis failure appears to be real - @andycochran can you take a look?

ncoden:
I don't know why `sass-true` crash now. But we can try with node v4.0.

kball:
v4.2.2 throws the same error:

```
Error: Undefined operation: "0.2126 times pow(0.99628, 2.4)".
    at Object.module.exports.renderSync (/Users/kball/git/foundation-sites/node_modules/n
ode-sass/lib/index.js:440:16)
    at Object.module.exports.runSass (/Users/kball/git/foundation-sites/node_modules/sass
-true/lib/main.js:34:18)
    at Object.<anonymous> (/Users/kball/git/foundation-sites/test/sass/test_sass.js:14:10
)
    at Module._compile (module.js:435:26)
    at Object.Module._extensions..js (module.js:442:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:311:12)
    at Module.require (module.js:366:17)
    at require (module.js:385:17)

```

kball:
Found the issue. The scss tests import files individually... this PR introduces a dependency for color on math.  You can fix the failure by adding an @import 'math' into `_color.scss`

kball:
I can fix the above issue as a part of the merge if you like.  Other than that, is this ready to go @andycochran @ncoden?

andycochran:
@kball fixed. I think this is good to go so long as Zurb is cool with the new ally-compliant color palette. It's not as vibrant. I also wouldn't mind a "LGTM" from @ncoden or @brettsmason.

andycochran:
Hmm… looks like Travis is still upset.

kball:
Bah, kicking travis, looks like that spurious phantom download error.
 
@zurbrandon @tdhartwick @gakimball any of you have thoughts on this new default palette?

scss/components/_accordion.scss
```diff
@@ -65,24 +67,30 @@ $accordion-content-padding: 1rem !default;
...
+  $background-hover: $accordion-item-background-hover
+) {
   display: block;
-  padding: $accordion-item-padding;
+  padding: $padding;
   line-height: 1;
-  font-size: $accordion-title-font-size;
-  color: $accordion-item-color;
+  font-size: $font-size;
+  color: $color;
```
ncoden: Why does some components have an alternative color, and others doesn't ?
It could be useful to use a common *color format* like a `($color-light, $color-dark)` list (for background too), and a mixin like `component-skin($background-list, $color-list, $pick-best-contrast)` to use it in components.

scss/components/_accordion.scss
```diff
@@ -65,24 +67,30 @@ $accordion-content-padding: 1rem !default;
...
+  $padding: $accordion-item-padding,
+  $font-size: $accordion-title-font-size,
+  $color: $accordion-item-color,
+  $border: $accordion-content-border,
+  $background-hover: $accordion-item-background-hover
+) {
   display: block;
-  padding: $accordion-item-padding;
+  padding: $padding;
   line-height: 1;
```
ncoden: `font-size` before `line-height`

scss/components/_accordion.scss
```diff
@@ -101,16 +109,21 @@ $accordion-content-padding: 1rem !default;
...
-@mixin accordion-content {
-  padding: $accordion-content-padding;
+@mixin accordion-content (
+  $padding: $accordion-content-padding,
+  $border: $accordion-content-border,
+  $background: $accordion-content-background,
+  $color: $accordion-content-color
+) {
+  padding: $padding;
   display: none;
```
ncoden: `display` before `padding`

Please rebase, then run `gulp sass:lint` ;)

scss/_global.scss
```diff
@@ -23,11 +23,11 @@ $global-lineheight: 1.5 !default;
...
 /// @type Map
 $foundation-palette: (
-  primary: #2199e8,
-  secondary: #777,
+  primary: #1779ba,
+  secondary: #767676,
   success: #3adb76,
   warning: #ffae00,
-  alert: #ec5840,
+  alert: #cc4b37,
```
ncoden: If you have a reference for these colors, can you add the link in comment ?

scss/components/_accordion.scss
andycochran: I avoided `-light` and `-dark` because the default button and label do not use the `$pick-best-contrast` function. You might make an app with a dark background and a bright default button with dark text. If necessary, the buttons created from the palette use the alt color, which might be light.

andycochran:
@ncoden Thanks for the feedback. I'll work on this tomorrow.

scss/components/_accordion.scss
andycochran: `gulp lint:sass` 😉

scss/components/_badge.scss
```diff
@@ -47,7 +51,7 @@ $badge-font-size: 0.6rem !default;
       @if $name != primary {
         &.#{$name} {
           background: $color;
```
andycochran: I'm not sure about this. The `$foundation-palette` is a list of colors, not a list of backgrounds. We're using a `$color` from the palette to set a `background`. Seems logical as-is to me.

scss/components/_accordion.scss
andycochran: Oh, and I'm only using alt colors for component that create classes with an `@each` loop for the `$foundation-palette`. Components that do not use a `$foundation-palette` loop do not need an alt color.

scss/_global.scss
ncoden: I mean : How did you find that we should use _these particular colors_, and not, for example, `#cc4b38` ?

andycochran:
FYI, here's the diff in the colors: 

image

image

scss/_global.scss
andycochran: I used Photoshop to incrementally adjust the original color's luminosity (keeping the same hue and saturation, changing only the brightness), checking each incremental change in the [WebAIM Contrast Checker](http://webaim.org/resources/contrastchecker/), until the change allowed the new color to meet AA requirements.

ncoden:
Can't we keep background colors for a white text ?

scss/_global.scss
ncoden: Ok. I think it could be useful to add the link to test the contrast, with an explanation.

andycochran:
@ncoden In order to keep white text for the Success and Warning buttons, the colors have to darken quite a lot. The gold ends up being an ugly brown. 

image

In order to change the palette as little as possible—and because I like the existing green/gold—I decided to leave the Success and Warning colors as they are. They meet AA requirements with the alt text color.

ncoden:
@andycochran I understand. But the different text colors disturb me. And I don't find the `disabled` text very visible.

Maybe we should change the hue, saturation and luminosity to have beautiful _and_ ally-compliant colors.

andycochran:
@ncoden I'd love it if we could come up with that palette. But AA contrast requirements quite limiting. I think we can only get two out of these three: 1) beautiful 2) ally-compliant 3) white text. Please feel free to propose new colors. 

I imagine most designers who don't care about AA requirements will simply set `[component]-color` and `[component]-color-alt` to the same value—e.g. white. :cry:

Maybe the alt text color is visually awkward because all the buttons are shown in a row here? In practice, buttons won't often be arranged this way.

ncoden:
@andersonaguiar 
I'll search for better colors that meet these three critera. Otherwise, I would prefer `beautiful` and `white text`. I care about a11y, but not if it break the design for everyone.

Also, In practice, I often see a primary button next to a warning button. :/

andycochran:
:frowning_face: Caring more about aesthetics than a11y is directly opposed to the spirit of #7387. 

@kball / @rafibomb, can Zurb and its designers can weigh in on this?

ncoden:
I checked with [WebAIM](http://webaim.org/resources/contrastchecker/), and didn't find a a11y-compliant (AA) color for a white text that is not ugly. :disappointed:  Also, `disabled` button have a `.25` opacity, so the background color is not relevant at all for a11y tests.

But could we, instead of changing the default colors, provide a set of tools to help websites to provide an alternative `A11y High Contrast` color palette dynamicly (like DuckDuckGo).

Also, Firefox allow the user to [overriding the page colors](https://support.mozilla.org/en-US/kb/accessibility-features-firefox-make-firefox-and-we#w_overriding-page-colors) (but not Chrome).

andycochran:
Don't worry about contrast for `.disabled`. 

> But even tough the guidelines do not require a contrast ratio for disabled UI components, I think the current style for the disabled buttons (opacity: 0.25) is a bit too light/has too low contrast ratio. Maybe the number could be tweaked at least a little bit.
– @Owlbertz's [comment in #7387](https://github.com/zurb/foundation-sites/issues/7387#issuecomment-161886371)

Again, I think this comes down to a judgement Zurb should make. Should Foundation Sites actually be "fully-accessible" as advertised, and must that include meeting AA contrast requirements out-of-the-box?

Owlbertz:
We are focusing so much on accessibility with the framework that imo we should make it be compliant out of the box. Of course this is a thing that the ZURBians have to decide, but as we are promoting to be "fully-accessible" as @andycochran said this should be the goal. Maybe we have to go even further and add some contrast checking to the SASS and/or the "custom download" page. 

The biggest problem with a11y will always be that people are not aware of it or just don't care and thus we should give them as much help as we can to create compliant websites so they'll actually do.

MandiSaeteun:
The updated compliant palette @andycochran suggested is a super solid compromise. @shannon-zurb and I tinkered with the palette this morning and arrived at a similar conclusion - white text on green and yellows is hard to make look beautiful _and_ accessible. 

Question came up in the office here - **should we make the compliant palette a variable that can be disabled (or enabled)**? We fear that defaulting the framework's color palette to be compliant maybe regressive. Sites may have intentionally used the current out-of-the-box colors because it aligns with their brand. @zurbrandon @tdhartwick

andycochran:
@MandiSaeteun I don't think an extra variable just to turn on a11y colors is necessary—especially if it's on by default. Just as easily as turning it off, you could set your alt-colors to white. If this was a patch, I might agree that we shouldn't break folks' button colors. But it's a minor version with other changes that require reviewing your `_settings.scss`. Isn't advocating for AA compliance is more important than not bothering people to review the new version's settings?

andycochran:
Looks like there are some new conflicts. I'll get these fixed… 

@kball / @MandiSaeteun, any further thoughts from Zurb on out-of-the-box compliance?

IllusionPerdu:
I think the callout have similar problem, in the setting they are a $callout-font-color-alt but never used and it sould be great if the background is dark the alt color is used.

scss/components/_orbit.scss
```diff
@@ -96,7 +96,8 @@ $orbit-control-zindex: 10 !default;
   padding: $orbit-caption-padding;
 
   background-color: $orbit-caption-background;
-  color: foreground($orbit-caption-background);
+  color: color-pick-contrast($orbit-caption-background);
+  background-color: $orbit-caption-background;
```
IllusionPerdu: "background-color: $orbit-caption-background;" is duplicated line 98 and 100

andycochran: Ack! Must've missed that when rebasing and fixing a merge conflict. I'll clean this up…

kball:
@andycochran Our consensus has been that out of the box compliance is very important, and worth sacrificing a little bit of backwards compatibility.  Lets write a migration note that addresses this and lets people know how they can revert to the previous palette.  Other than this, is this ready to merge?

andycochran:
@kball AFAIK, it's ready to merge. Didn't wanna add that label to my own PR tho.

kball:
Merged!  @andycochran can you write a quick migration note (comment in this PR would be fine) highlighting what the change is and how someone would get back to the previous palette if they wanted to?

andycochran:
# Migration notes

### New functions

- The `foreground()` function has been replaced with a new `color-pick-contrast()` function that uses luminosity instead of lightness. 

- The `color-pick-contrast()` function requires new `_math.scss` utilities for 1) finding the greatest common divisor of two integers and 2) handling decimal exponents by trying to convert them into a fraction and then use a nth-root-algorithm for parts of the calculation. 

### Revised settings 

- The `$foundation-palette` variables have been adjusted to meet WCAG 2.0 level AA contrast requirements. For backwards compatibility (ignoring AA requirements), revert `primary` to `#2199e8`, `secondary` to `#777`, and `alert` to `#ec5840`.

- All instances of the `foreground()` function have removed from `_settings.scss` in favor of setting explicit colors and alternate colors. 

- The `$tab-content-color` and `$pagination-item-color-current` variables are now set explicitly. There's no need for a function to determine these color, as their associated background colors are also set explicitly. Previously, the `foreground()` function for Pagination only allowed for black or white text. 

### New settings variables

- The Badges and Labels components now have alternate colors: `$badge-color-alt` and  `$label-color-alt`. These are used the same way as `$button-color-alt`. The `$badge-color` and `$label-color` now default to white (like `$button-color`). Previously, the badge and label colors defaulted to a function which made assumptions about the lightness of the background colors. Buttons, Badges, and Labels components now all work the same way. The default can be any foreground/background color combination, and for each background color defined in `$foundation-palette`, the foreground is automatically set to the color or alternate color (whichever has more contrast). For backwards compatibility (ignoring AA requirements so all button colors have white text), set the alt colors to white. 

- The Tabs component now allows you to set the color of tab titles with `$tab-color` and `$tab-active-color`. Previously, tab titles were hardcoded to the primary color, which was not visible on some tab background colors. 

### Mixins

- Accordion, Pagination, Tabs now allow for passing vars when creating your own components.

kball:
@andycochran I think it would be valuable to add a note to the docs for buttons, badges, and labels that explains why they have mixed text colors out of the box, and how to change it to all white (with a warning that it will reduce your contrast below AA contrast requirements)

andycochran:
@kball Will do…

PR #9329, opened 2016-11-04
Made Accordion documentation consistent

karland:
1. Made accordion.md documentation consistent to embedded code.
2. Replaced unrepresentative code from Kitchen Sink Accordion with code from accordion.md.

### DIFF
--- a/docs/pages/accordion.md
+++ b/docs/pages/accordion.md
@@ -20,9 +20,13 @@ Each pane has a **title**, an `<a>` with the class `.accordion-title`, and a **c
 ```html
 <ul class="accordion" data-accordion>
   <li class="accordion-item is-active" data-accordion-item>
+    <!-- Accordion tab title -->
     <a href="#" class="accordion-title">Accordion 1</a>
+
+    <!-- Accordion tab content: it would start in the open state due to using the `is-active` state class. -->
     <div class="accordion-content" data-tab-content>
-      I would start in the open state, due to using the `is-active` state class.
+      <p>Panel 1. Lorem ipsum dolor</p>
+      <a href="#">Nowhere to Go</a>
     </div>
   </li>
   <!-- ... -->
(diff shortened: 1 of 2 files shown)


abdullahsalem:
Thanks @karland !
- Your modifications on `kitchen-sink.md` are fine, except changing the input type to `text`, so why do think that is necessary?
- Regarding to `accordion.md` file, I believe that the examples must not be the same everywhere, so if the single example has a valid code but doesn't look like the other examples, that is fine.

@kball you are welcome to share us this conversation :)

karland:
Hi @abdullahsalem,
* In the code example in the docs I expected a date picker or something similar to open when the tab "Accordion 3" suggested to enter a date. I think this is what nowadays normally happens when you click on a date field in a form. Therefore I changed it to something that would make it clearer to the user that nothing is going to happen.
* I agree, the kitchen sink must not necessarily show the same code. However, I felt the code was misleading. I believe, the idea of Accordion is to set the IDs and references automatically, so that the user does not have to bother. In the kitchen sink, however, everything is set manually, which is kind of unrepresentative and error prone (as you see from the comment, that triggered this PR). Furthermore manual referencing is not described in the docs, so I wondered why it is done in the kitchen sink which after all is just an overview.

Also the comment is uncorrect to my understanding  in as such that the href does not need any matching ID.

```HTML
<!-- The content pane needs an ID that matches the above href, role="tabpanel", data-tab-content, and aria-labelledby. -->
```

So I thought, it is easiest to just take the code from the docs to have a representative example.

kball:
I agree the comment was incorrect, probably worth reworking... would actually be nice to point out that you can but do not have to have an id on the content... perhaps the accordion docs page should have another section highlighting that?

Other than that... @rafibomb do you have an opinion on this change?

karland:
I did not have a deep look at the Accordion code, so I am not exactly sure what one would achieve by setting `href` to an an ID. However, my feeling was that the `href` is ignored by Accordion.

ncoden:
up

kball:
@abdullahsalem as the first one who reviewed this, what are your thoughts on the response by @karland? Is this good to go?

rafibomb:
I like it! @kball What is the id on the content for?

kball:
in accordions id is probably generally irrelevant unless you're using deep linking (added in https://github.com/zurb/foundation-sites/pull/9664), but in that case you'd need it to manage the urls

IamManchanda:
Seems like that this vintage commit is been reproduced again by me #9882 
( although as bug fix and not feature request )

kball:
Merging

PR #9332, opened 2016-11-06
[docs] fixed programmatic use of interchange

joernroeder:
I was running into this issue and it took me a while to figure out that by providing a single string with multiple rules as `rules` option the [loop](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.interchange.js#L109) iterates over each character in the option string.  
Providing the rules as an array works. 

I also propose to do the [`.match`](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.interchange.js#L106) right after the condition which desides where the options are loaded from and simply do

```javascript
rules =  typeof rules === 'string' ? rules.match(/\[.*?\]/g) : rules;
```

right before entering the loop. I could also create a pull request for it if you like.

### DIFF
--- a/docs/pages/interchange.md
+++ b/docs/pages/interchange.md
@@ -85,5 +85,11 @@ When using Interchange programmatically, you need to pass in your ruleset in the
 
 ```js
 var $photoFrame = $('#some-container');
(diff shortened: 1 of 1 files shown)


kball:
@joernroeder your proposed update sounds good if you want to submit a PR for it.

joernroeder:
@kball submitted a pull request #9589. better late than never ;)

PR #9333, opened 2016-11-06
Sort CSS properties

ncoden:
Sort the CSS properties by categories.

### Why ?
Like said [here](https://github.com/zurb/foundation-sites/pull/9316#pullrequestreview-6675668), I prefer ordering CSS properties by *categories* instead of alphabetically because it is easier to:
- Find a property. You know the property you are searching better by what it does (approximately) than its name.
- Understand the global component behaviour. You see and distinguish immediately the different "groups" of properties, and each property is next to those related to it.
- Understand the consequences of each property on others. Properties are ordered from the *kind* to the *appearance* of the component, the first having consequences on the seconds.

### What ?
The full sort order is [Specific](https://gist.github.com/ncoden/d42f55df7c7970f548a02cd3468f9c86) (I'm still looking for a better name).

These categories are:
- Position (`position`, `top`, `z-index`...)
- Disposition (`display`, `flex-*`, `float`...)
- Dimension (`width`, `height`, `margin`, `padding`...)
- Global appearance (`appearance`, `opacity`, `transform`...)
- Borders (`border`, `outline`)
- Graphics (`background`, `box-shadow`, `transition`...)
- Content/Text (`content`, `color`, `font-*`, `text-*`...)

This is my mix between [the tree standards I found](https://github.com/sasstools/sass-lint/tree/develop/lib/config/property-sort-orders) : `smacss` (global organization), `recess` (picked missing properties) and `concentric` (from disposition/ouside to appearance/inside).

### How ?
- [x] Global
- [x] Components
- [x] Forms
- [x] Grids
- [x] Typography
- [x] Utils
- [x] Require pull request https://github.com/zurb/foundation-sites/pull/9301

*9333 : winrar !*


ncoden:
@colin-marshall After `position, disposition, margin, padding, global appearance...`, do you prefer:
- `... border, background, shadow ...` (smacss)
- `... shadow, border, background ...` (concentric)
- `... background, border, shadow ...` (recess / classic)

?

(poke @kball)

colin-marshall:
@ncoden I prefer SMACSS.

ncoden:
@colin-marshall Why ? ;)

colin-marshall:
Because that's the style guide I follow when I write SCSS 😄. To be honest I don't have a good reason for why we should use any of those tree standards over the other than that SMACSS is probably the most well known of them.

brettsmason:
Awesome work @ncoden. Thinking about what @colin-marshall said I guess he has a point with SMACSS. There's more chance of another user being familiar with that sort order, and so would be easier to write code as they are already familiar with it. The goal of this is to standardise the code base now and going forward, so I wonder if using a well known approach might help with adoption?

ncoden:
Thanks @colin-marshall @brettsmason. So I'll continue with this sorting order.

ncoden:
I finished to reorder a part of properties across the whole codebase. @zurb/yetinauts, I would appreciate your feedback, and maybe your advices for some problems I encountered.

## Specific

First, I used the [Specific](https://gist.github.com/ncoden/d42f55df7c7970f548a02cd3468f9c86) sort order. That's my first order (I'm still looking for a better name), a mix between [smacss](https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/smacss.yml), [recess](https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/recess.yml) and [concentric](https://github.com/sasstools/sass-lint/blob/develop/lib/config/property-sort-orders/concentric.yml).

:point_right: Do you have any remarks / critics / suggestions about this sorting convention ?

## Left and Right

Probably because of `margin: * * * *;` and `padding: * * * *;` define margin/padding for in order `top`, `right`, `bottom` and `left`, [all the standards I found](https://github.com/sasstools/sass-lint/tree/develop/lib/config/property-sort-orders) agree on the following property sort order:
```scss
margin-top: ...;
margin-right: ...;
margin-bottom: ...;
margin-left: ...;
```

And I'm totally agree with this sort order. However, when only `left` and `right` margins/paddings have to be defined (like for [grids](https://github.com/zurb/foundation-sites/blob/develop/scss/grid/_classes.scss)), it leads to the following order:
```scss
margin-right: ...;
margin-left: ...;
```

I find it *logical*, but *not intuitive at all*.

:point_right: Should we move anyway to this sort order, `right` *then* `left` ?

## Mixins and conditions

Currently, mixins and conditions (`@if ...`) are randomly declared in the top or the bottom of the declaration block, but never where it is *relevant* (i.e. where `border:...;` should be for a mixin/condition containing border properties).

:point_right: Should we, when it does not make the declaration block difficult to read, move the mixins and conditions to their *relevant* position ?

kball:
I also find right before left to be somewhat unintuitive, but given the logical coherence of it, and given that is what all of the various standards we've looking into suggest, I think it is fine to push it.  Foundation hasn't historically shied away from pushing people towards things that were uncomfortable but good practice.

On mixins and conditions... hmm.  I'm not sure... the part that feels very vague to me is 

> when it does not make the declaration block difficult to read

Additionally, what happens when a condition or mixin impacts multiple properties at once?

ncoden:
> the part that feels very vague to me is: "when it does not make the declaration block difficult to read"
For example:

```scss
@mixin block(
    $border,
    $shadow,
    $fontsize
) {
    position: relative;
    display: block;

    @if has-value($border) {
        border: $border;
    }

    background-color: #fff;

    @if has-value($shadow) {
        box-shadow: $shadow;
    }

    font-familiy: sans-serif;

    @if has-value($fontsize) {
        font-size: $font-size;
    }

    text-decoration: underline;
}
```

> Additionally, what happens when a condition or mixin impacts multiple properties at once?

If it is multiple properties of different types, so the condition or mixin doesn't have a relevant position, and will be at the top or bottom.

ncoden:
@kball 
> the part that feels very vague to me is: "when it does not make the declaration block difficult to read"

For example:

```scss
@mixin block(
    $border,
    $shadow,
    $fontsize
) {
    position: relative;
    display: block;

    @if has-value($border) {
        border: $border;
    }

    background-color: #fff;

    @if has-value($shadow) {
        box-shadow: $shadow;
    }

    font-familiy: sans-serif;

    @if has-value($fontsize) {
        font-size: $fontsize;
    }

    text-decoration: underline;
}
```

---

> Additionally, what happens when a condition or mixin impacts multiple properties at once?

If it is multiple properties of different categories, so the condition or mixin doesn't have a relevant position, and will be at the top or bottom.

kball:
The one thing that I think is outstanding here is better documenting for the future what these policies are, not just in a gist linked in the PR.  In our readme we link to this repo: https://github.com/zurb/foundation-code-standards which I believe has not been fully maintained... I've just added @zurb/yetinauts to it, perhaps we can have a related PR there that highlights this standard for property orders?

ncoden:
@kball Why don't we have all the help, installation and contribution advices, code standards (...) in the documentation or in the wiki, on this repo ?

kball:
@ncoden great question, I don't know.  Perhaps we should move in that direction... @gakimball do you know the history on why standards are a separate repo?

andycochran:
+1 to keeping the code standards in the docs or wiki (as is the Testing Guide).

andycochran:
Actually @kball, as far as why the standards are a separate repo, those standards aren't only for Foundation Sites. They're also for Foundation for Emails/Apps/CLI, Supercollider, Motion UI, Inky, Panini, Yeti Launch, Octophant… 

So, does it make sense to keep all Foundation code standards together, or are there specific Foundation Sites standards that should be separate?

colin-marshall:
Could we keep the coding standards in that separate repo and add something in the docs build process for each framework that pulls in the standards? At the least the standards repo should be linked in each set of docs.

ncoden:
If [Foundation Code Standards](https://github.com/zurb/foundation-code-standards) is used across multiple Zurb projects, I think we should keep in its own repo instead of duplicate it across projects, with only the particular cases and implementation advices added in each projects. So _Foundation Code Standard_ would be a _web standard_ like the Airbnb one.

However, `foundation-sites` is moving (or will move) to a very different standard (SMACSS/ITCSS). The question is: will the other Zurb projets follow ? If yes, we can update the Foundation Code Standards. If not, we should add it the Foundation doc its own standard.

kball:
Good question @ncoden, its unclear exactly if all of the projects will follow, and even if they do the process will likely be gradual.

I suspect the fact that the doc is separate is one reason why it has been unmaintained, so my general inclination would be to pull in the foundation-sites specific standard to this repo.

ncoden:
@kball So, what should we do with:
- [the Foundation doc](https://github.com/zurb/foundation-sites/tree/develop/docs/pages) which don't give contributing advices or styleguides.
- [CONTRIBUTING.md](https://github.com/zurb/foundation-sites/blob/develop/CONTRIBUTING.md) which is very basic and redirect to [Foundation Code Standards](https://github.com/zurb/foundation-code-standards) (and I guess that only a few folks read it).

Maybe we should split the Foundation doc into tree parts : *installation, customization and global help*, *components doc and examples*, and *contribution and styleguide*. We can begin by a simple `styleguide.md`, then gradually move `CONTRIBUTING.md` and *Foundation Code Standards* into their `.md` files and folders.

What do you think ?

kball:
@ncoden I like your proposal. to split this out and include it in the docs. I don't think `styleguide.md` is a blocker to merging this, but it should follow fairly shortly.

ncoden:
@kball Ok. So we can merge it ? :tada:

kball:
Boom!  Merged

PR #9335, opened 2016-11-07
Upgrade node version

ncoden:
Upgrade node version to v4.0 to prevent future bugs with sass-lint (see https://github.com/zurb/foundation-sites/pull/9301) and probably others new packages. I think the v0.12 begin to be outdated.

The v4.0 seems to work with most of packages for now, and is often recommended to resolve bugs of v0.12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "0.12"
+  - "4.0"
 before_install:
   - rvm install 2.2.2
 install:
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Lastly, if you're rolling your own setup, you can install Foundation through a v
 
 ## Documentation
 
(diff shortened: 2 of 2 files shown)


pine3ree:
...with all the benefits that comes by running a LTS version (but I shouldn't talk as I am using 6.9.1 :-D)

colin-marshall:
With this change we might want to document somewhere that you need a minimum of Node v4.0 to work with this repository. As far as I can tell, the templates should still work with Node v0.12.x since they don't use sass-lint.

For reference, when sass-lint released 1.10.0 over the weekend, it updated its ESLint dependency from v2.x to v3.x. Support was dropped for Node < v4.0 in ESLint v3.0.0.

kball:
👍 Lets add a note to the README to address @colin-marshall's point and get this baby merged.

ncoden:
@kball Done. Ready to be merged ?

PR #9337, opened 2016-11-07
Fix #9336

### LINKED ISSUE #9336: Destroying a tooltip causes console errors and does not fully remove classes and attributes from the element

1. Create a tooltip
2. Call the destroy method on the tooltip

The instance of the tooltip should be fully destroyed and should not create errors in the console.

**What happened instead?**

Console errors occurred. Additionally, the class `.has-tip` remains on the element as well as the `aria-haspopup`, `data-disable-hover`, and `data-events` attributes.

**Test case:**

[http://codepen.io/anon/pen/oYgMRN](http://codepen.io/anon/pen/oYgMRN)

madisonjmeredith:
This solves issue #9336. When the destroy method is called on a tooltip, the instance of the tooltip is not fully removed. Instead, it causes console errors and leaves behind tooltip-specific classes and attributes.

This commit:

- Corrects a typo (`.zf.tootip` to `.zf.tooltip`) that was preventing the event handler from properly being removed from the tooltip
- Removes all tooltip-specific classes, including `.has-tip`, `.top`, `.right`, and `.left`
- Removes all tooltip-specific attributes including `aria-describedby`, `aria-haspopup`, `data-disable-hover`, `data-resize`, `data-toggle`, `data-tooltip`, and `data-yeti-box`

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -339,12 +339,9 @@ class Tooltip {
    */
   destroy() {
     this.$element.attr('title', this.template.text())
-                 .off('.zf.trigger .zf.tootip')
-                //  .removeClass('has-tip')
-                 .removeAttr('aria-describedby')
-                 .removeAttr('data-yeti-box')
-                 .removeAttr('data-toggle')
-                 .removeAttr('data-resize');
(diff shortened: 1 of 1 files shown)


kball:
awesome, thank you!

PR #9342, opened 2016-11-07
Fixes #9341 - Firefox flexbox input-group-field width

### LINKED ISSUE #9341: [Input Group] Flex-mode, input overflowing container in Firefox
Foundation 6.2.4

In Firefox, with Foundation's flex-mode enabled, an input-group text input can overflow its container.

Check out http://codepen.io/simshaun/pen/MbwbbY

image

Adding `min-width: 0;` to the `.input-group-field` appears to fix it in Firefox.

# Testing in Browserstack
- Win 10 / Edge **good**
- Win 10 / IE 11 **good**
- Win 7 / IE 10 **good**
- Win 10 / FF 49 **good**
- Win 10 / Chrome 54 **good**
- Win 10 / Opera 41 **good**
- Mac Sierra / Safari 10 **good**
- Mac El Capitan / Safari 9.1 **good**
- Mac Yosemite / Safari 8 **good**

simshaun:
Fixes #9341

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -92,6 +92,7 @@ $input-prefix-padding: 1rem !default;
     @if $global-flexbox {
       flex: 1 1 0px;
       height: auto;
+      min-width: 0;
     }
     @else {
       height: $height;


kball:
@ncoden you had some conversation about this in #9341 but given this PR is targeted much more tightly than the original attempt in https://github.com/zurb/foundation-sites/pull/8511 it seems unlikely to introduce the same bug in any real usecases... would you agree? Shall we go ahead with this?

ncoden:
@kball It is the same "fix" with the same bug, but for a particular case only. That's not a real flex fix, and less maintainable. I don't think we should merge it.

simshaun:
Why is wrapping a concern in an input-group?

ncoden:
@simshaun  I understand. I only say we should search for a global fix. If we don't, of course we should fix this partial fix.

kball:
Given where we are in the release cycle, and the fact it looks like this shouldn't have any major negative consequences given how targeted it is, I think we should merge this.  I'd still like us to look for a global fix... @ncoden do you know if we have an open issue targeting this global problem?

PR #9346, opened 2016-11-08
9345 button border radius

### LINKED ISSUE #9345: [Button] $button-radius setting is duplicated and overridden in _settings.scss
Changing the `$button-radius` SCSS setting variable under the "11. Button" settings group (https://github.com/zurb/foundation-sites/blob/master/scss/settings/_settings.scss#L243) has no effect on the compiled component. 

This is because the `$button-radius` variable has been duplicated later in the file, which overrides any change you've just made.

First (correct) definition: https://github.com/zurb/foundation-sites/blame/master/scss/settings/_settings.scss#L243

Second (mistaken) definition: https://github.com/zurb/foundation-sites/blame/master/scss/settings/_settings.scss#L361

I can't imagine any need for the 2nd definition, should be safe to remove.

nandastone:
Fixes #9345.


ncoden:
Hi @nandastone, Thank you for your PR.

`scss/settings/_settings.scss` is automatically generated from the config variables or Foundation scss files. So you would rather have to remove the duplicated property `$button-radius` from `scss/forms/_text.scss`.

However, `$button-radius` is used in Forms (`scss/forms/_text.scss`) for a particular purpose different from the Button component (`scss/components/_button.scss`) one. It set the `border-radius` of form buttons elements (`[type='submit'], [type='button']`).

So I'll close this PR and open a new to rename the config variable instead of removing it. But thank you for having pointed out this bug.

nandastone:
Oh neat, wasn't aware `_settings.scss` was generated like that. Thanks @ncoden :)

PR #9347, opened 2016-11-08
Fix button-radius config naming collision

ncoden:
Fix naming collision with the `$button-radius` config variable between Button component and Form buttons : rename the config variable for Forms buttons to `$form-button-radius`.

The bug was introduced in https://github.com/zurb/foundation-sites/pull/9109, and pointed out in https://github.com/zurb/foundation-sites/pull/9346, but with a wrong fix.


kball:
👍

Issue #8301, opened 2016-03-01
[Accordion Menu] Adding a nested menu after creation does not work 


dmamolina:
**How can we reproduce this bug?**

Create an empty accordion menu

``` html
<ul id="navigation" class="vertical menu" data-accordion-menu>
</ul>
```

Add a nested menu and reInit:

``` javascript
      var liItem = '<li>' + 
              '<a href="#">Test</a>' + 
              '<ul class="menu vertical nested">' +
              '</ul>' +
            '</li>';

      $('ul#navigation').append(liItem);
      Foundation.reInit($('ul#navigation'));
```

**What did you expect to happen?**

The elements of the submenu are not hidden and the arrow to collapse/expand the accordion is not shown.

Adding the `data-submenu` attribute to the `ul.menu.nested` element and the class `is-accordion-submenu-parent` to the `li` element fixes the issue:

``` html
<li class="is-accordion-submenu-parent">
  <a href="#">Test</a>
  <ul class="menu vertical nested" data-submenu>
  </ul>
</li>
```

These attrs and classes are included in the [Feather function of the nest util](https://github.com/zurb/foundation-sites/blob/master/js/foundation.util.nest.js#L22), but it looks like this method is not called with the Foundation.reInit call, but on creation.

colin-marshall:
@Owlbertz @brettsmason can you please take a look at this? I see it was mentioned in #9348 which you guys had heavy involvement in. Thanks!

Owlbertz:
Issue seems to persist, altough as the author said it can be avoided when applying the attributes manually.

I created [this pen](https://codepen.io/Owlbertz/pen/NyWyBw) for any further analysis.
Is this actually a bug (i.e. do we want `reInit()` to take care of these attributes)?

DanielRuf:
What is the current status of this issue?

DanielRuf:
Current status is that only the arrow is missing: https://codepen.io/DanielRuf/pen/GRRNRaP

brettsmason:
@DanielRuf It's been a long time since I worked on this, but adding the class `accordion-menu` to the `UL` seems to add the arrow correctly. It does mention this in the docs: https://foundation.zurb.com/sites/docs/accordion-menu.html#submenu-toggle

DanielRuf:
@brettsmason I guess we can close it then as this is documented?

brettsmason:
@DanielRuf Yeah I'd say so.
[...]

PR #9348, opened 2016-11-08
[WIP] Add submenuToggle option to accordion menu

### LINKED ISSUE #7770: [Accordion Menu/Drilldown] Feature request/suggestions

brettsmason:
This is work in progress but I wanted to put a PR together so its not forgotten about!

This allows the parent menu link to function as a link.
It adds a separate `.submenu-toggle` button which now toggles the submenu display.

Still do to:
- [x] Decide on a styling path to define the `.submenu-toggle` visually. @andycochran I'd like to bounce some ideas I have off you on this 😄 
- [ ] Work out required keyboard navigation

This closes #7770 when complete.


andycochran:
As discussed in #7770, the best way to style these submenu toggles is as split buttons. Visually, we simply need to put a vertical line between the link and the toggle button. With split buttons that's done via margin allowing the background to show through. But that won't here. We'd need to either introduce background colors or borders. 

We _could_ style Accordion Menu with the same borders as Accordion — which would allow us to put a line between the link and toggle button. For backwards compatibility, if you don't want borders or submenu toggles, just set the new `$accordionmenu-border` variable to `none`.

brettsmason:
So I discovered by looking through the codebase that that responsive menu plugin is already setup to work with a `accordion-menu` class which isn't currently in use. This allows us to apply borders and backgrounds to correctly differentiate the submenu toggle.

I still need to tidy up the Sass and reorder it etc, but the concept now visually works.

brettsmason:
@andycochran @kball I have added a demo with the styling I added to the `.accordion-menu` class here: http://brettmason.co.uk/submenu-toggle/accordion-dropdown-test/accordion-dropdown.html

I personally think the styling now looks fine.

However running this responsive menu test revealed a few things. 

- The classes on the parent `<li>` aren't being removed - eg `has-submenu-toggle is-dropdown-submenu-parent`. I dont think this is related to this PR, but does need to be fixed.

- The `submenu-toggle` button isn't being removed when shifting to the dropdown menu

kball:
@brettsmason I think between your last push and mine the issues are addressed, but please test & confirm.

brettsmason:
So the only thing missing is the keyboard navigation needs correcting if anyone is able to take a look at this? I will also amend the docs to include an example using the toggle as well as the `.accordion-menu` class.

brettsmason:
@Owlbertz You're an accessibility wizard 😄
On the 2 points above do you have an opinion on the correct way to make this accessible?

Owlbertz:
@brettsmason:
- `aria-expanded` should be placed on the `button` (see [spec](https://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded).
- The button needs text if it does not contain any text itself. Of course it is always helpful to add additional information on what will happen.

Maybe @sethkane can correct me if I am wrong.

sethkane:
Sorry @Owlbertz you are right the aria-expanded should be on the button "Indicates whether the element, or another grouping element it controls, is currently expanded or collapsed."  Sorry for the delay work is crazy.

brettsmason:
@Owlbertz @sethkane thanks a lot for your input 😄 
I have now added text for accessibility as well as set `aria-expanded` on the button. What I now cant work out is what sets the initial `aria-expanded` on the `<li>` element so I can remove it when this option is enabled - I cant see any code that does this unless I'm missing it...

The only remaining thing then is the keyboard navigation, which currently I have no clue how to sort!

Owlbertz:
@brettsmason I pulled this down and had a look. Imho the `aria-controls` needs to be placed on the `button` rather than the `li` (as the `li` does nothing here).

brettsmason:
@owlbertz great thanks for taking a look.

I'll look at adding the aria-controls to the button. If you take a look though, there is an aria-expanded on the list element, but I can't work out where it gets applied from! 

The keyboard nav just doesn't feel like it's right at the moment, and needs changing if the button is active.

Owlbertz:
@brettsmason the `aria-expanded` that is set comes from [this line](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.util.nest.js#L32), seems like it is currently automatically applied during `reflow()`.

brettsmason:
@Owlbertz Thanks for finding this for me, I was lost!

@kball What would be the best way to remove the `aria-expanded` if this option is enabled? I can remove it on init, but is that the right route?

Also I have no clue on the keyboard navigation, so will need help to get this implemented in time for 6.3, but I really want it included to pair with off canvas 😄

kball:
Hmm good question @brettsmason.  I think this might actually tie into the discussion we were having here: https://github.com/zurb/foundation-sites/pull/9381#issuecomment-263344303 (cc @ncoden @diniscorreia @DaSchTour) about using data-attributes to specify children (and auto-applying in base cases)... right now the components pre-assume a ton about structure which works fine in base case but causes challenges in more complex cases.

It may be a little late in the cycle for this for 6.3, but what if we changed all of these menus to use data-attributes to specify child elements and then auto-applied them in the "basic" cases?  I think that would make it much easier to target the right elements for things like this aria attribute.

brettsmason:
@kball I agree its probably a bit tough to get that sorted for 6.3.
In terms of getting this PR implemented into 6.3 do you think there's any way we could get something implemented so it could get included, even if its not perfect? The only remaining issues are `aria-expanded` and the keyboard nav.

kball:
@brettsmason I think I've got the accessibility behaving properly now, though I'd love to get @Owlbertz or @sethkane to verify as they are more knowledgeable on accessibility than I am. On the keyboard nav... playing around with it now after merging develop into it, it actually seems to be working pretty well.

brettsmason:
@kball awesome thanks. I think from memory navigating from submenu back up to parent should go to button first not list element, and parent list element shouldn't expand submenu. Turned everything off for today though so can't check.

sethkane:
Sorry guys have been swamped at work, I will do my best to follow up with all my mentions tonight or tomorrow night at the latest.

tsquez:
So just to be clear, `data-parent-link="true"` still does not work with the accordion menu?

brettsmason:
@tsquez Unfortunately not. We ran out of time to get it finished (most of my time was spend on revamping off canvas) and there were still a few bugs to iron out.

I'll make sure it gets included in the next larger release though (6.4).

tsquez:
@brettsmason - no worries amigo was just checking for a few users. Keep up the good work.

nickroberts:
Great work, guys! Looking forward to being able to use this!

👍

hadl:
Awesome work!

To fix https://github.com/zurb/foundation-sites/issues/8301 could you please move the "Foundation.Nest.Feather" call from the constructor to the _init method?
If you make this change, please check for the toggle-button before appending it. Otherwise you will get multiple "togglers" on every reInit() call.

Also remove the .is-active class inside the up() method. The reInit() will open all previously opened submenus if this class is present. 
Or is this an expected behaviour?

With those changes appending new submenus via js is working.

kball:
I just brought this up to date with the current state of develop... it looks like it's working to me. @Owlbertz can you pull it down and verify the aria behavior is as we'd expect? If so I think we're good to go and can merge this for 6.4

Owlbertz:
The interactive parts (attribute toggling and stuff) look good for me, but I feel like the whole thing behaves more like a tree than like a menu ([see this page](https://www.w3.org/TR/wai-aria-practices-1.1/#TreeView)). This would mean that we need to change the `role` of the items slightly.

kball:
Brought this up to date with develop... @Owlbertz can you confirm I didn't bork the merging of your tree logic in? (it wouldn't auto merge), and then lets get this sucker merged!

Owlbertz:
@kball Looks good to me in terms of my latest PR regarding the `role="tree"` thing. However, I noticed that it is not possible to activate the links on root level via keyboard. Pressing enter just expands the sub-menu. I think the desired behavior would be the same as for mouse usage: toggling of the sub-menu only via the dedicated button and activating the links with the accordion title itself.

Owlbertz:
Just adjusted that myself. Please check and then go for the merge!

kball:
👍

kball:
Woohoo! Took since November but we finally got this sucker in. Nice work @brettsmason @Owlbertz

brettsmason:
Yay thanks guys for getting it across the line! 😁

tsquez:
So `data-parent-link="true"` is working now? Sorry  haven't had a chance to try this yet but very anxious to do so.

brettsmason:
@tsquez it will be included in 6.4! (RC out tomorrow I think)

tsquez:
HECK YES!!! Great job amigo!

cfabrice:
That's great but the pbl remains when used with responsive navigation like data-responsive-menu="accordion medium-dropdown" as mentionned in the "responsive navigation" part of the doc.
Or is it me ?

tsquez:
@cfabrice the issue was corrected. If you would like you can see it in action here https://themeawesome.com/themes/wpforge/ - granted its a WordPress theme but it still works like it should.

cfabrice:
Oh sorry, I was playing with the codepen of the official docs (https://codepen.io/IamManchanda/pen/jmxVPP) but it was calling in foundation 6.3 still !!!

PR #9353, opened 2016-11-09
Fixed the scrolling behavior when auto-focus is active

imabdulsal:
According to the [Tabs documentation](http://foundation.zurb.com/sites/docs/tabs.html#js-options), when the option `data-auto-focus` is enabled, the browser should scroll to the active (focused) tab. So this PR implements that feature.

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -69,8 +69,12 @@ class Tabs {
         'aria-labelledby': linkId
       });
 
-      if(isActive && _this.options.autoFocus){
-        $link.focus();
+      if(isActive && _this.options.autoFocus){  
+        $(window).load(function() {         
+          $('html, body').animate({ scrollTop: $elem.offset().top }, _this.options.deepLinkSmudgeDelay, () => {
+            $link.focus();
+          });
+        });
       }
 
       //use browser to open a tab, if it exists in this tabset


abdullahsalem:
@zurb/yetinauts Would appreciate it if anyone could take a look at this :)

js/foundation.tabs.js
```diff
@@ -69,8 +69,12 @@ class Tabs {
         'aria-labelledby': linkId
       });
 
-      if(isActive && _this.options.autoFocus){
-        $link.focus();
+      if(isActive && _this.options.autoFocus){  
+        $(window).load(function() {         
+          $('html, body').animate({ scrollTop: $elem.offset().top }, _this.options.deepLinkSmudgeDelay, () => {
+            $link.focus();
```
ncoden: Shouldn't the link be on focus *before* the scoll animation started ?

abdullahsalem: If the focused element is under the fold, the browser will natively scroll down to reach that element, so if the active `tab` is under the fold, then the scrolling event will be fired twice, once as a native behavior and the second is ours. So I thought that it would be good to fire the `focus` event after the browser is already there, that will prevent the browser to do anymore scrolling.
To test this scenario:
- Create an active tab under the fold.
- Open the page.
- Refresh the page.

ncoden (approved):

abdullahsalem:
As it's already reviewed by @ncoden, I marked it `ready to merge`.

kball:
Thanks @abdullahsalem!  Do you know if there were any open issues against this?

abdullahsalem:
@kball According to my previous search, it's been not reported.

PR #9356, opened 2016-11-10
modify tabs events to return content area as well as a link

### LINKED ISSUE #8952: Tabs Change Event Should Pass $targetContent
The `change.zf.tabs` event ought to pass the tabs `$targetContent` variable to the callback function. It currently only passes `$target`. 

Here is the line where I believe this change should happen: https://github.com/zurb/foundation-sites/blob/develop/js/foundation.tabs.js#L209

Without the corresponding "content" element, represented by `$targetContent`, it's a pain to affect changes in the visible tab on the change event.

kball:
Fixes #8952.  I also modified the deeplink event to have both items... given this modifies the new deep linking feature slightly, would love @ahebrank to take a look and review.

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -77,23 +77,25 @@ class Tabs {
       if (_this.options.deepLink) {
         var anchor = window.location.hash;
         //need a hash and a relevant anchor in this tabset
-        if (anchor.length && $elem.find('[href="'+anchor+'"]').length) {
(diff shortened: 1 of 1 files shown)


ahebrank:
Makes sense to me.  I'd thought (wrongly) `$target` in the original event referred to the tab content, so this is more consistent since they both now return `[$button, $content]` (right?)

kball:
correct

PR #9357, opened 2016-11-10
[WIP] New off canvas panels

brettsmason:
Rebuild of the off canvas component. Still work in progress, but getting there.

New features:

- Top/bottom off canvas option
- Overlap transition option (as well as original push)
- No overlay option (closes panel on content click, without having an overlay over the content)
- ForceTo - allows scroll to top/bottom on panel open
- No longer forces the page to top when opening a panel
- cleaner markup - no longer need `off-canvas-wrapper` and `off-canvas-wrapper-inner`

The Sass has been re-written to allow you create a custom off canvas panel. This can be done like so:

```
.my-off-canvas {
  @include off-canvas-base;
  @include off-canvas-position(top, vertical, 50px);
}
```
This will add the markup to create a custom panel, including transitions for the main content.

Still some things to tidy up and bugs to fix, as well as re-write the docs...


brettsmason:
@kball @colin-marshall I'd love some help on the JS side of things to work out whats causing the overlay to disapear if clicked before transition end 😄

kball:
@brettsmason are you still seeing the problem with the overlay?  I'm not reproducing it now... my suspicion (based on something similar previously) is it had to do with the unnecessary Foundation.Move that you have now removed.

brettsmason:
@kball No problem with the overlay now, that works as it should.
My remaining issues I think are:

- Box shadow: trying to create the same effect as live is difficult, as setting the z-index lower than the content causes issues with top/bottom. My solution at the moment is to apply the shadow as `inset`. It doesn't give the exact same result, but its what is used on Foundation for Apps. Do you think this is OK? I dont want lots of the "the shadow is different" issues... 

- @andycochran pointed out an issue with iOS which allows the content to be scrolled. Its tough for me to check without an iOS device, but we might have to re-introduce one wrapper to get around this, I'm not sure.

Would appreciate any input on these points so we can get it across the line!

rafibomb:
Not sure why the test is failing - @Owlbertz Do you see why?

Owlbertz:
```shell
1) Off Canvas init() closes Off Canvas on outside click if closeOnClick option is true:
     TypeError: 'undefined' is not an object (evaluating 'plugin.$exiter.should')
```

Seems like `$exiter` got changed to `$overlay`, which is a nice thing imo because this is more streamlined with other plugins (i.e. Reveal).
Another issue seems to be that the `.is-off-canvas-open` class is now set to the body.

I am still investigating further issues.

Owlbertz:
Update: I think I was able to track down all relevant issues. For now the problem persists that the `Foundation.transitionend` is not fired when running the tests on console. It is, however, called correctly when opening the HTML file in the browser.

brettsmason:
@Owlbertz Thanks for looking into this. Do I need to make any changes to make it pass or does the actual test need to change? Sorry - lack of knowledge on Travis 😄

Owlbertz:
@brettsmason you will need to adjust the tests to catch up with the changes you made:
- $exiter to $overlay
- changed the way autofocus works
- apply `is-off-canvas-open`-class to body instead of Offcanvas wrapper

Take a look at https://github.com/zurb/foundation-sites/commit/090a862bda098dca05d1bb0c02053e3665347f92 (available on [this branch](https://github.com/owlbertz/foundation/tree/brettsmason-new-offcanvas) which you might pull if you want).

However, test will most likely fail because the `Foundation.tranisitionend` not being called in automated tests.

paxperscientiam:
@brettsmason, looks very good. Question, is "off-canvas-wrapper" even needed? Because the menus are `position:fixed;`, this wrapper class shouldn't be needed. Unless I'm missing something.

brettsmason:
@paxperscientiam Thanks a lot for trying it out - feedback is really appreciated.

So there are 2 reasons I added the `off-canvas-wrapper` class. 

If the user changes the off-canvas panels to be absolute positioned (which in some use cases that behavior is desired), as well as the `reveal-for-large` etc classes also can be configured to be absolute positioned. An example use case was from an existing issue, here's a working example: http://brettmason.co.uk/offcanvas/double/double-canvas.html (this hasn't been updated to include the wrapper yet).

The other reason was when closing a panel you will see vertical scroll bars when closing. I can't think of another way to stop these being visible, unless you have any suggestions?

you are right though - it definitely isn't needed if people want to leave it out.
I'd really appreciate any other feedback or if you think there's a better way this could be handled!

paxperscientiam:
@brettsmason Thanks for sharing the example; that's an interesting example of using `position:absolute` on `.off-canvas`.

I will push changes with an absolute positioning example with your current code. That said, I think there ought to be a default positioning. That is, use `position:fixed` by default and introduce a data-attribute for absolute positioning.

The above is what I'm testing now.

BRB.

brettsmason:
@paxperscientiam Thanks for taking a look at this some more, I appreciate the input.

So the positioning is controlled via a variable `$offcanvas-fixed` - if that is set to false then the menu will be absolute positioned. Its also possible to set to absolute through using the mixin directly `@include off-canvas-base($fixed: true);`. I had thought about including a class (to use instead of `off-canvas`) - something like `off-canvas-absolute` (or something to that effect) that allows you to do an absolute positioned panel at the same time using just the classes out of the box. What do you think?

I like the idea of simplifying things more, however I'm not sure its going to be possible (unless you have any ideas). The first problem with not applying the class to the panel directly, is that you get a FOUC before the JS is loaded and thats not great. Would like to hear ideas on how we could do it though.

As for the wrapper - you are right it could be optional, though I personally dont like the vertical scroll bar when the panel is closing. Maybe we should just remove the wrapper, and give advice on how to remove it if the user wanted to? Though without it absolute positioned panels wouldnt work correctly.

paxperscientiam:
No problem, @brettsmason. I really enjoy working with others on Github. Being a part of the Foundation community is especially awesome.

>The first problem with not applying the class to the panel directly, is that you get a FOUC before the JS is loaded and thats not great.

Why not just use the data attribute itself as the CSS selector to target? That way, no FOUC --right? This can be leaned on even for the absolute positioning.
````CSS
#derp[data-off-canvas] {
  background:black;
  min-height:30px;
}
#derp[data-off-canvas*="top"] {
  margin:5px;
}
#derp[data-off-canvas*="overlap"] {
  border:pink 5px solid;
  min-height:30px;
}
#derp[data-off-canvas*="absolute"] {
  border-bottom-color: green;
}
````
In this way, the user can something like `data-off-canvas="['top','overlap','absolute']"` without fear of FOUC. As far as CSS processing goes, the data value is just a string, so order is immaterial.

Does that assuage some concerns?

**Comment by brettsmason** [comment]
@kball I'd be interested to see what you think about this suggestion?

**Comment by brettsmason** [comment]
@paxperscientiam So I've thought about this some more. One of the key goals of redoing off canvas way to make it as seamless as possible to move to the new version, and keep it as backward compatible as possible.

I think because of this we should keep this as it is for now, but with the plan to revisit it for version 7. This would allow us to build on this PR and rework it to be more streamlined. What do you think?

Also do you think its worth adding a `off-canvas-absolute` (or a better named?) class so people can prototype with an absolute positioned panel at the same time as a fixed panel?

**Comment by paxperscientiam** [comment]
@brettsmason 

Based on your justifiable concerns -- which have reminded me of my ignorance on how to introduce "breaking changes" -- I have some thoughts:

Conceptually speaking, let's say we have two buckets:
* Bucket A contains "behind the scenes" changes (f34f724 through 23fdb23, I think). It also contains "progressive enhancements." EG: allowing a user to choose between options`data-offcanvas` and `data-offcanvas-absolute`.
*  Bucket B goes all code that introduce changes to markup that are incompatible with current major release.

I still think classes like `position-top` are needless as CSS can use the data attributes as selectors too. Can't `data-position="top"` alone be used for initial styling, or does the JS core need the `position-top` selector? I guess this is a little of bucket A, a little of bucket B.

>Also do you think its worth adding a `off-canvas-absolute`...

Can we create an "absolute" value for the `data-position` attribute?
````HTML
<div id="myOCM" class="off-canvas position-top" data-off-canvas data-position="top absolute" data-transition="overlap">
````


What do you think?

**Comment by brettsmason** [comment]
@paxperscientiam SO i had a play around with styling the data attributes earlier and it worked OK. However the specificity of using data attributes is higher than just a class which isn't ideal.

However through this you made me realise something - I dont think the `data-position` is even needed anymore! I will check over the weekend, but I think it can be removed.

That got me thinking - maybe the transition style (push/overlap) should be controlled by classes too rather than a data option. All the data option does is add a class, so why not remove the data option and just add the class?

This would then be easier hopefully for the user, plus all can still be controlled by mixins. What do you think?

**Comment by brettsmason** [comment]
@paxperscientiam I removed the position JS options as they are no longer needed, so it should be a lot easier to setup.

@andycochran / @kball Any opinion on if an absolute positioned class should be generated? This would allow things like http://brettmason.co.uk/offcanvas/double/double-canvas.html without having to change the default, and might be useful for prototyping perhaps. On the other hand it might be an edge case, and those users could just change the default or use the mixin?

**Comment by kball** [comment]
@brettsmason playing with this today as I'm trying to hunt down the sticky related issues, found a couple of things...

1. For backwards compatibility, it feels like the contentOverlay should probably default to false

2. Not being able to scroll while off canvas is open was really surprising and offputting to me... what is the story/reason behind that?

**Comment by paxperscientiam** [comment]
@brettsmason 
>  I dont think the data-position is even needed anymore
Excellent!

>  maybe the transition style (push/overlap) should be controlled by classes too rather than a data option.

I did some reading about this and it seems the prevailing wisdom is that data attributes should be used stuff that cannot be handled otherwise -- e.g: paths to lazy loaded content or the regex used for Abide. If transition styles can be applied without data attributes, seems like it should.

Some quotes:
>Do not store content that should be visible and accessible in data attributes, because assistive technology may not access them. In addition, search crawlers may not index data attributes' values (Mozilla)
>Custom data attributes are intended to store custom data private to the page or application, for which there are no more appropriate attributes or elements. (W3C)


For some v 7 inspiration, here is a demo using OCM with InstantClick [link](https://www.paxperscientiam.com/playground/ocm-pjax.php)

Do you know is pjax is on the roadmap? Perhaps as an optional "mode" as with flex?

**Comment by brettsmason** [comment]
@kball to answer your points:

1. contentOverlay currently is on by default on live, though there is no option to turn it off, other than change the Scss variable (`$offcanvas-exit-background`) though that name no longer makes sense, as it doesn't necessarily close the off canvas anymore because of the 2 seperated options. I can turn it off though if you want to.

2. That surprises me. If you try the current off canvas (especially on a mobile device) being able to scroll the content just seems odd, to me at least. The off canvas can still be scrolled if that helps. @andycochran also mentioned a bug in Safari which allowed you to scroll the content etc before this fix. I'm happy to add it as an option and turn it off by default if you feel that would be better? That would be an easy fix.

@paxperscientiam Thinking about it some more, because we have an inheritied default transition (push as currently on live) it would be difficult to make it simple as I would need to create 2 new classes, including one for push. That may break peoples sites though unless they add the new transition classes correctly, so I think we will have to keep it as it until V7.

Hopefully I have managed to strip it back enough, but also make it a lot better to use which was the aim!

**Comment by kball** [comment]
@brettsmason re number 2 I think it needs to at least be an option - when I've seen off canvas used for desktop scrolling is an important functionality.  I'm open to broader opinions on default behavior, though my general thought is when in doubt leave it the same...

**Comment by brettsmason** [comment]
@kball I beat you to it (just) 😄 I just added the option and think it makes sense, so thanks for pointing it out.

**Comment by kball** [comment]
A point brought in is that the previous off canvas had some issues with reveal - should add a visual test and see how this works with it

**Comment by brettsmason** [comment]
@kball any particular reveal configuration or just in general? I'll get a test setup.

File: js/foundation.offcanvas.js [hunk 1]
```diff
@@ -87,8 +89,14 @@ class OffCanvas {
       'keydown.zf.offcanvas': this._handleKeyboard.bind(this)
     });
 
-    if (this.options.closeOnClick && this.$exiter.length) {
-      this.$exiter.on({'click.zf.offcanvas': this.close.bind(this)});
+    // If we have an overlay and close on click, let it close the off canvas menu.
+    if (this.options.closeOnClick && this.options.contentOverlay && this.$overlay.length) {
```
kball: Why are we using truthy/falsy here as compared to direct true/false comparison like on line 98?

js/foundation.offcanvas.js
```diff
@@ -87,8 +89,14 @@ class OffCanvas {
...
+      this.$overlay.on({'click.zf.offcanvas': this.close.bind(this)});
+    }
+
+    // If content overlay is false but close on click is true, close via click on content.
+    if (this.options.closeOnClick && this.options.contentOverlay !== true) {
+      $('[data-off-canvas-content]').on({'click.zf.offcanvas': this.close.bind(this)});
```
kball: Conceptually, we're walking down a path of 
1. Do we closeOnClick
2. What do we apply the handler to?

So wouldn't the code be clearer if we combine these two if statements into something like:

```
if (this.options.closeOnClick) {                                            
    var $target = this.options.contentOverlay ? this.$overlay : $('[data-off-canvas-content]');
    $target.on({'click.zf.offcanvas': this.close.bind(this)});                
}
```

js/foundation.offcanvas.js
```diff
@@ -146,6 +147,16 @@ class OffCanvas {
   }
 
   /**
+   * Stops scrolling of the body when offcanvas is open on mobile Safari and other troublesome browsers.
+   * @private
+   */
+  _stopScrolling(event) {
+  	event.preventDefault();
+  	event.stopPropagation();
+  	return false;
```
kball: We believe return false is sufficient:  http://stackoverflow.com/questions/1357118/event-preventdefault-vs-return-false

js/foundation.offcanvas.js
```diff
@@ -154,63 +165,52 @@ class OffCanvas {
...
-      // }
-
     this.$triggers.attr('aria-expanded', 'true');
     this.$element.attr('aria-hidden', 'false')
         .trigger('opened.zf.offcanvas');
 
-    if (this.options.closeOnClick) {
-      this.$exiter.addClass('is-visible');
+    // If `contentScroll` is set to false, add class and disable scrolling on touch devices.
+    if (this.options.contentScroll == false) {
```
kball: use ===

js/foundation.offcanvas.js
```diff
@@ -154,63 +165,52 @@ class OffCanvas {
...
+      $('body').addClass('is-off-canvas-open').on('touchmove', this._stopScrolling);
+    }
+
+    // If we have an overlay lets make it visible.
```
kball: unnecessary comment

js/foundation.offcanvas.js
```diff
@@ -154,63 +165,52 @@ class OffCanvas {
...
+    if (this.options.contentOverlay) {
+      this.$overlay.addClass('is-visible');
+    }
+
+    // If we have close on click and an overlay add a `is-closable` class.
```
kball: also excessive - the code is already extremely readable

js/foundation.offcanvas.js
```diff
@@ -154,63 +165,52 @@ class OffCanvas {
...
+    if (this.options.closeOnClick && this.options.contentOverlay) {
+      this.$overlay.addClass('is-closable');
     }
 
     if (trigger) {
       this.$lastTrigger = trigger;
     }
```
kball: This seems like it should be right at the top of the method - we're simply stashing away the argument passed in

js/foundation.offcanvas.js
```diff
@@ -224,38 +224,20 @@ class OffCanvas {
...
     focusable.off('.zf.offcanvas').on('keydown.zf.offcanvas', function(e) {
-      var key = Foundation.Keyboard.parseKey(e);
-      if (key === 'TAB' && e.target === last[0]) {
-        e.preventDefault();
-        first.focus();
-      }
-      if (key === 'SHIFT_TAB' && e.target === first[0]) {
-        e.preventDefault();
-        last.focus();
+      if (e.which === 9 || e.keycode === 9) {
```
kball: Do we need the extra `e.keycode` instead of just `e.which`?  Also, part of the benefit of the Foundation.Keyboard is it abstracts away the values of the keycodes and lets you have them named, catches combos with shift, etc... was there a reason for going away from using it?

js/foundation.offcanvas.js
```diff
@@ -266,25 +248,28 @@ class OffCanvas {
...
-    //   setTimeout(function() {
-    //     _this.$element[0].style.transform = '';
-    //     $(window).off('scroll.zf.offcanvas');
-    //   }, this.options.transitionTime);
-    // }
-    if (this.options.closeOnClick) {
-      this.$exiter.removeClass('is-visible');
+
+    // If `contentScroll` is set to false, remove class and re-enable scrolling on touch devices.
+    if (this.options.contentScroll == false) {
```
kball: ===

js/foundation.offcanvas.js
```diff
@@ -266,25 +248,28 @@ class OffCanvas {
...
+      $('body').removeClass('is-off-canvas-open').off('touchmove', this._stopScrolling);
+    }
+
+    // Remove `is-visible` class from overlay.
```
kball: excessive commenting

js/foundation.offcanvas.js
```diff
@@ -266,25 +248,28 @@ class OffCanvas {
...
+    if (this.options.contentOverlay) {
+      this.$overlay.removeClass('is-visible');
+    }
+
+    // If we have `closeOnClick` and `contentOverlay` add `is-closable` class.
```
kball: excessive comment

js/foundation.offcanvas.js
```diff
@@ -313,18 +298,13 @@ class OffCanvas {
...
-        this.$lastTrigger.focus();
-        return true;
-      },
-      handled: () => {
-        e.stopPropagation();
-        e.preventDefault();
-      }
-    });
+  _handleKeyboard(event) {
+    if (event.which !== 27) return;
```
kball: Reason not to use the keyboard utility?

js/foundation.offcanvas.js
```diff
@@ -313,18 +298,13 @@ class OffCanvas {
...
+
+    event.stopPropagation();
+    event.preventDefault();
```
kball: Any reason these can't be moved down and turned into just 'return false'?

scss/components/_off-canvas.scss
```diff
@@ -6,176 +6,298 @@
...
+    position: fixed;
+  }
+  @else {
+    position: absolute;
+  }
+
+  z-index: $offcanvas-push-zindex;
+
+  transition: transform $transition;
+  backface-visibility: hidden;
```
kball: Any particular reason this is necessary?  Do some of our transitions use rotations?

scss/components/_off-canvas.scss
```diff
@@ -6,176 +6,298 @@
...
+    left: 0;
+
     width: $size;
+    height: 100%;
+
+    transform: translateX(-$size);
+    overflow-y: auto;
+
+    // Sets the open position for the content
+    &.is-open ~ .#{$maincontent-class} {
```
kball: Why do we need the `~` selector?  Is this a specificity issue?

scss/components/_off-canvas.scss
```diff
@@ -6,176 +6,298 @@
...
+    left: 0;
+
     width: 100%;
+    height: $size;
+
+    transform: translateY(-$size);
+    overflow-x: auto;
+
+    // Sets the open position for the content
+    &.is-open ~ .#{$maincontent-class} {
```
kball: same question as line 156

scss/components/_off-canvas.scss
```diff
@@ -6,176 +6,298 @@
...
+  // Off-canvas position classes
+  .position-left    { @include off-canvas-position(left,   horizontal); }
+  .position-right   { @include off-canvas-position(right,  horizontal); }
+  .position-top     { @include off-canvas-position(top,    vertical); }
+  .position-bottom  { @include off-canvas-position(bottom, vertical); }
 
-    &.position-left   { @include off-canvas-position(left); }
-    &.position-right  { @include off-canvas-position(right); }
-    &.position-top    { @include off-canvas-position(top); }
+  // Off-canvas content
```
kball: unnecessary comment

js/foundation.offcanvas.js
brettsmason: I'm happy to adjust to whatever would prefer - whats your preference? Change everything to direct true/false? Its like this as thats what the existing off canvas used, so I went with that too.

js/foundation.offcanvas.js
brettsmason: Again should it be a direct true check? Looks good though

js/foundation.offcanvas.js
brettsmason: Will change - will need someone to check on mobile Safari to confirm though please (no longer have access to check)

scss/components/_off-canvas.scss
brettsmason: This I believe it to increase performance in certain browsers/fix flickering: http://blog.teamtreehouse.com/increase-your-sites-performance-with-hardware-accelerated-css

This was in the previous version, but happy to remove if its felt its no longer needed. Also wonder if we should use the `translateZ(0)` method mentioned in that article.

scss/components/_off-canvas.scss
brettsmason: This is because we no longer apply classes to the optional wrapper, so we need a way to trigger the content transition too, and need to cater for people using multiple off canvas sections (eg http://brettmason.co.uk/offcanvas/double/double-canvas.html). Does that make sense?

js/foundation.offcanvas.js
kball: I think a direct true check is probably better

scss/components/_off-canvas.scss
brettsmason: Also as I removed the need for `data-position` we cant do this: https://github.com/zurb/foundation-sites/blob/develop/js/foundation.offcanvas.js#L178

js/foundation.offcanvas.js
kball: ok will do

scss/components/_off-canvas.scss
kball: Ah got it. I think I misread this on first pass

kball:
Merged!

kball:
@brettsmason @rafibomb can one of you put together migration notes for this component?

PR #9358, opened 2016-11-10
Add an option to the drilldown menu to position the back button at th…

ajoy39:
Because the drilldown menus are created and sized with javascript it's pretty difficult to change the positioning of the back button in CSS, so I added an attribute to the drilldown-menu component that allows the user to define if they want the back button prepended to the ul (the previous, and default behavior) or appeneded to the end.

This is my first FOSS commit so all feedback is welcome and I expect I still have plenty of work to do, fire away!

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -85,7 +85,16 @@ class Drilldown {
       var $menu = $(this),
           $back = $menu.find('.js-drilldown-back');
       if(!$back.length){
-        $menu.prepend(_this.options.backButton);
+        switch (_this.options.backButtonPosition) {
+          case "bottom":
+            $menu.append(_this.options.backButton);
+            break;
+          case "top":
+            $menu.prepend(_this.options.backButton);
+            break;
+          default:
+            console.error("Unsupported backButtonPosition value '" + _this.options.backButtonPosition + "'");
+        }
       }
       _this._back($menu);
     });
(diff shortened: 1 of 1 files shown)


js/foundation.drilldown.js
```diff
@@ -85,7 +85,11 @@ class Drilldown {
       var $menu = $(this),
           $back = $menu.find('.js-drilldown-back');
       if(!$back.length){
-        $menu.prepend(_this.options.backButton);
+        if(_this.options.backButtonPosition == 'bottom'){
+          $menu.append(_this.options.backButton);
+        } else {
+          $menu.prepend(_this.options.backButton);
+        }
```
ncoden: I don't think this is a good idea to consider all values different from "bottom" as "top". This hide a bug if `bottom` is incorrectly written. We should check for `bottom` and `top` with `switch...case`, and throw an error on an unhandled value.

(poke @kball, what do you think ?)

js/foundation.drilldown.js
```diff
@@ -395,6 +399,12 @@ Drilldown.defaults = {
    */
   backButton: '<li class="js-drilldown-back"><a tabindex="0">Back</a></li>',
   /**
+   * You can choose to position the back button either at the top or bottom of drilldown submenus.
+   * @option
+   * @example bottom
+   */
+  backButtonPosition: 'top',
```
ncoden: Comments should be descriptive (`This component does...`) or in the imperative forms if you talk about possibilities for the users (`Do that to make that...`).

So you can write instead :
```js
/**
 * Position the back button either at the top or bottom of drilldown submenus. The button is at the top by default.
  ...
 */
```

Also, please update the `backButton` documentation above. From `Prepended to submenu lists...` to something like `Prepended or appended (see backButtonPosition) to submenu lists...`.

ncoden (commented):
Thank you @ajoy39 for your PR. :)

js/foundation.drilldown.js
ajoy39: Would adding such a statement force users to specify a position? My concern was that could cause any current usage of the drilldown menu to break on update, though I suppost I could check for "top", "bottom" or unset and only warn if there *is* a value but it's not top or bottom. Thoughts?

ncoden: Well, we *are defining* the default value (l. 406), so I don't see how this option could be `unset`. We are not introducing a breaking change. Websites that are currently using the drilldown menu will still have the back button at the top, unless they overwrite `backButtonPosition` with the `bottom` value. An error would be throw only if they overwrite it with a value different from `top` and `bottom`.

ajoy39: Right, totally forgot about that, it's been a couple days haha, hows this? 

````
switch (this.buton.backButtonPosition) {
    case "bottom":
        $menu.append(_this.options.backButton);
        break;
     case "top";
        $menu.prepend(_this.options.backButton);
        break;
     default:
        console.log("Unsupported backButtonPosition value '" + backButtonPosition + "'");
}
````

kball: Definitely agree on this approach and the increased specificity of class - relatively recently merged a similar change in some javascript.

kball:
Current state of this breaks some of the JS unit tests

ajoy39:
Yes I know, unfortunately I can't get the test suite to run locally right now (I think this is an incompatibility with macOS 10.12.1 not an issue with my code or foundations, phantomJS keeps terminating with SIGSEGV) so it may be a bit before I can fix it : /

kball:
New version looks good.  Going to go ahead and merge!

Issue #8945, opened 2016-06-17
Abide doesn't validate on blur


djohnson:
**How can we reproduce this bug?**
1. Step one
   Add abide to a form
2. Step two
   Add required fields
3. Step three
   Tab through the fields

**What did you expect to happen?**
Validation failed errors messages displayed 

**What happened instead?**
Nothing

This was the configurable default behavior in v5 but appears to be missing from the current docs.

andycochran:
To clarify the issue, Abide does validate on blur. But before doing so, it [checks `inputText.length`](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.abide.js#L335). Then if an input is blank, it checks to see if it's required. Then if the input is blank and required, it does nothing. 

However, an input *is* validated if you: 
* enter a valid value
* blur the input
* refocus the input
* delete the value
* blur the input

So Abide *will* validate a blank input *if* it has already been validated. Um. I'm not sure how that's working; we'll need on of the @zurb/yetinauts with more JS expertise to take a look.

abdullahsalem:
In fact, Abide does validate "on change" only, so that whenever you change the previous value it works, and that's how "on change" works. Therefore, I think we need to add "on blur" event too, optionally. as it was in F5.

andycochran:
Maybe the `liveValidate` option should include "on blur"?

kball:
@abdullahsalem do you want to take a crack at this?

abdullahsalem:
@kball Yes, please :)

abdullahsalem:
@andycochran `liveValidate` feature is to run the validation while the user is typing.

kball:
Implemented in #9360

PR #9360, opened 2016-11-11
Added feature on abide to enable validation on blur event

imabdulsal:
This is an implementation of the Abide's **validate-on-blur** feature that was in Foundation 5. So Abide now has a new option that's called `data-validate-on-blur`, and it is `false` by default.

The feature has been requested here #8945 .


kball:
Excellent, thanks @abdullahsalem

Issue #9361, opened 2016-11-11
Flex grid - Nested rows issue


Lindsor:
Using the _flex-grid.scss file an element with the classes of 'row' and 'column' have the float set to none and the display set to block which breaks the flex grid.

**How can we reproduce this bug?**

<div class="row column">
  <div class="small-3">Full width but should be 3 columns</div>
  <div class="small-3">Should be next to the above not below</div>
</div>

**What did you expect to happen?**
Expected the '.small-3' element to take up 3 columns.

**What happened instead?**
The '.small-3' took up 100% width.

Give us a link to a CodePen or JSFiddle that recreates the issue.

http://codepen.io/Lindsor/pen/oYbJNa/

kball:
Fixed in #9362.

PR #9362, opened 2016-11-11
Fix bug #9361

Lindsor:
The flex grid sets the display of a row that is aslo a column as 'block', remove this declaration since its not needed.


Lindsor:
Created pull request fixing the issue

Pull #9362

kball:
Seems correct to me.  Thanks @lindsor!

PR #9364, opened 2016-11-12
Fix lint problems

ncoden:
Fix various lint problems with `sass-lint` or across `.scss`/`.js`/`.md` files. Follow lint conventions with consistency.

## Leading zero
Use a leading zero for numbers.

Pro:
- css-guidelines (https://github.com/chris-pearce/css-guidelines#formatting)
- Promoted by a mathemacial point of view (http://hey.georgie.nu/leadingzero/).

Cons:
- Codestyle.co (http://codestyle.co/Guidelines/css#Syntax)
- Less concise

## No border convention
Use `border: 0;` instead of `border: zero;` to disable border

See : https://stackoverflow.com/questions/2922909/should-i-use-border-none-or-border-0

Pros:
- More concise
- Used a lot

Other convention: `border: none;`

## Brace style convention
Use "stroustrup" instead of "1TBS" as brace style convention.

See: https://en.wikipedia.org/wiki/Indent_style

Pros:
- Easily add or remove lines within a `@if`/`@else if`/`@else` block, without having to rearrange the braces.
- Able to add comments that describe a condition (`@if (...)`, `@else if (...)` or `@else`) or the whole conditional block.
- Clearly separate the different parts of the conditional block.

Cons:
- Not commonly used

## Vendor prefixes
- Remove the vendor prefixes that are automatically added by `autoprefixer` for all browsers.
- Disable `sass-lint` when a vendor prefix is required by one particular browser.

## Other minor changes:
- Update `sass-lint` from `1.2.0` to `v1.3.2` (bug fixs)
- Remove EOL spaces, fix indentation
- Partially fix various other Sass lint problems

---

## Dependencies
- [x] Require #9333


ncoden:
@kball

PR #9365, opened 2016-11-12
add block-grid-column so bottom margin matches gutters

### LINKED ISSUE #8668: Bloc Grid + Responsive Gutters
It would be rad if `margin-bottom` on columns Block Grids used Responsive Gutter sizes. Then the space between rows would match the space between columns.

andycochran:
Closes #8668. 

Adds a `.block-grid-column` class to allow block grid's columns to use responsive gutter sizes for their `margin-bottom`. Now the space between rows can easily match the space between columns.

### DIFF
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -95,9 +95,9 @@
 /// @param {Number|Keyword} $margin [auto]
 ///   The bottom margin on block grid columns, accepts multiple values:
 ///   - A single value will make the margin that exact size.
-///   - A breakpoint name will make the margin the corresponding size in the $gutters map.
-///   - "auto" will make the margin responsive, using the $gutters map values.
-/// @param {Number|Map} $margins [$grid-column-gutter] - Gutter map or single value to use. Responsive gutter settings by default.
+///   - A breakpoint name will make the margin the corresponding size in the $margins map.
+///   - "auto" will make the margin responsive, using the $margins map values.
+/// @param {Number|Map} $margins [$grid-column-gutter] - Map or single value to use. Responsive gutter settings by default.
 @mixin block-grid-column (
   $margin: auto,
   $margins: $grid-column-gutter
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -91,14 +91,14 @@
   @include grid-column-uncollapse($gutter);
 }
 
-/// Sets bottom marin on block grid columns to match gutters
+/// Sets bottom marin on grid columns to match gutters
 /// @param {Number|Keyword} $margin [auto]
-///   The bottom margin on block grid columns, accepts multiple values:
+///   The bottom margin on grid columns, accepts multiple values:
 ///   - A single value will make the margin that exact size.
 ///   - A breakpoint name will make the margin the corresponding size in the $margins map.
 ///   - "auto" will make the margin responsive, using the $margins map values.
 /// @param {Number|Map} $margins [$grid-column-gutter] - Map or single value to use. Responsive gutter settings by default.
-@mixin block-grid-column (
+@mixin grid-column-margin (
   $margin: auto,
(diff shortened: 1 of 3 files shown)


andycochran:
Now that I've submitted this PR, I'm realizing that this class might be beneficial in non-block grid columns. You might want to apply responsive `margin-bottom` to other columns. @zurb/yetinauts, do you have thoughts on this? Should I rename this something that could be repurposed—e.g. `.bottom-gutter` (ew, something more semantic)?

scss/grid/_gutter.scss
```diff
@@ -90,3 +90,37 @@
   @warn 'This mixin is being replaced by grid-col-gutter(). grid-col-uncollapse() will be removed in Foundation 6.4.';
   @include grid-column-uncollapse($gutter);
 }
+
+/// Sets bottom marin on block grid columns to match gutters
+/// @param {Number|Keyword} $margin [auto]
+///   The bottom margin on block grid columns, accepts multiple values:
+///   - A single value will make the margin that exact size.
+///   - A breakpoint name will make the margin the corresponding size in the $gutters map.
+///   - "auto" will make the margin responsive, using the $gutters map values.
```
ncoden: I see you use "single value", "breakpoint name" and "auto" like `grid-column-gutter`. I searched a way to factorize this into one mixin used by `grid-column-gutter` and `block-grid-column`, but I didn't find it. :/

Also, it's `$margins` and not `$gutters` in `in the $gutters map` and `using the $gutters map values`. Or rename the variable to `$gutters`. After all, it's vertical gutters.

scss/grid/_gutter.scss
```diff
@@ -90,3 +90,37 @@
...
+/// @param {Number|Map} $margins [$grid-column-gutter] - Gutter map or single value to use. Responsive gutter settings by default.
+@mixin block-grid-column (
```
ncoden: Whatever the name, I would have preferred to keep the `grid-` prefix. For example `grid-block-column` or `grid-column-block`.

scss/grid/_gutter.scss
```diff
@@ -90,3 +90,37 @@
...
+    }
+  } @else {
+    // breakpoint name
+    @if type-of($margin) == 'string' {
+      $margin: grid-column-gutter($margin, $margins);
+    }
+
+    // single value
+    $margin-bottom: rem-calc($margin);
+
```
ncoden: no empty line here

scss/grid/_gutter.scss
```diff
@@ -90,3 +90,37 @@
...
+    margin-bottom: $margin-bottom;
```
ncoden: empy line here

kball:
@andycochran can you make the updates requested by @ncoden? Want to get this in for 6.3! 😄

scss/grid/_gutter.scss
andycochran: I went with `@mixin grid-column-margin`. I think margin is more semantic than block. Also, this mixin could be used for adding margin to columns that aren't necessarily in block grids.

scss/grid/_gutter.scss
andycochran: You're right, I copied the technique straight from `grid-column-gutter`. Nice catch. I've fixed the notes.

andycochran:
@kball I think this should be ready. 
@ncoden Do you want to take one more look?

scss/grid/_gutter.scss
ncoden: <- *happy to see he has contributed enough to foundation to see someone pick his code*

ncoden:
https://www.diffchecker.com/uWGVy8Pi

Yes... We definitely can factorize these two mixins, using a mixin that handle the "breakpoint list + breakpoint name" stuff.
I can do it (next week) if you don't want to. For now, I think it could be merged.

rafibomb:
Merged!

PR #9369, opened 2016-11-14

simshaun:
This is an updated version of #9018 that also separates tab panel visibility from the `is-active` class.


kball:
This looks good to me. @Owlbertz you were active on #9018, what do you think?

Owlbertz:
Yeah, seems good to me.

Issue #9235, opened 2016-10-05
toggler should have a 'container' default similar to close


kball:
If something has a data-close attribute without a target id, it will close its container... I think data-toggle should have a similar mode, allowing you to build reusable components that are able to toggle states on themselves without having to have an id.

PR #9372, opened 2016-11-15
Update toggle to bubble up if no target is defined

kball:
Fixes https://github.com/zurb/foundation-sites/issues/9235


kball:
@coreysyms given you are the expert on our triggers, would love it if you can take a look and make sure I'm not missing anything obvious.

PR #9377, opened 2016-11-15
[Docs] Match headers for large and small callouts

michaellennox:
Previously the large and small callouts were headed with 'This is a secondary panel' even though it wasn't actually a secondary callout and as a visual description doesn't match the styles used in the coloring section.

This PR modifies the header text in these examples to match the style used in the coloring section and explains the example using the class names ('small callout' and 'large callout') in the same manner as with the coloring section.


kball:
👍

Issue #8391, opened 2016-03-11
Slider not showing the correct position when toggled from hidden


jspekken:
When the slider is directly shown on the page everything works just fine and the slider slides to the correct position, but when a slider is displayed after the initial page load the slider will not slide to the correct position.

You can see this behavior here: http://codepen.io/anon/pen/EKyLmW

Maybe it's an idea to incorporate a method like the 'reflow' methods from foundation 5?

gakimball:
We do have reflow plugins, although Slider appears to be missing one.

jspekken:
@gakimball Do you know where I can find an example of a reflow plugin so that I can (maybe) create a pullrequest to fix this issue?

gakimball:
@jspekken [Here's an example from Equalizer.](https://github.com/zurb/foundation-sites/blob/4c974baef093fe6c409496a81b52ac789a44a94d/js/foundation.equalizer.js#L116) It re-calls whatever internal logic is needed to set up the plugin properly. In the case of Equalizer, it's calculating the heights of every element in a set.

coreysyms:
Tagging this to #9320 so I can add in a mutation event for Slider, this will fix the issue for 6.3. Will close this issue after I add in slider mutation.

kball:
So this isn't actually solved by adding the mutation observer, at least as currently instantiated, because in this case the parent is what is changing... we'd have to put an observer on every parent for hidden elements (see e.g. this pen: http://codepen.io/seantimm/pen/lDeCv/).  I think I see the reason this is failing though, so going to fix it as well as add the mutation observers.

kball:
actually, I'm not sure the right way to fix this because it currently depends on dimensions that don't exist when the object is display: none; Do we think exposing `_reflow()` is sufficient and requiring the user to call it after unhiding?

coreysyms:
I made some notes over on #9379, I believe that passing in proper CSS values on init() will fix this. You can see in @jspekken code pen that the "fill" bar does properly layout after being toggled from display none, it's the "handle" that needs to get that value as well.

kball:
@coreysyms the tricky thing is that while fill can be a straight percentage based on value, the handle is "offset" by its size, which is determined by css and read by the JS... not sure how we can prefill that... we could do a 'rough guess' and not do the offset, but it will result in the layout being slightly different if its adjusted while the slider is hidden instead of visible.

kball:
Closing this as fixed based on #9379

Issue #9320, opened 2016-11-01
Checklist, Qualifications, and References for New Mutation Observation Trigger


coreysyms:
# Checklist, Qualifications, and References for New Mutation Observation Trigger #

This issue will act as the checklist for Foundation Plugins that require integration of DOM mutation observation, qualification requirements, and as a bug issue reference point.

## Background ##
Mutation observation has been added to 6.3 in triggers, meaning all plugins have access to async triggers for DOM mutation and can key self-functions off the event change. Types of changes we can observe are the additions and removals of child nodes, parent/child relationships, and changes to the style attribute. Practically speaking, there is a number of "display none show hide" bug issues of plugins not correctly displaying that the implementation of this will mitigate.

## References ##
To see a demo of how this [works](http://tangerineindustries.com/Foundation/feature-mutation-trigger/with_mut.html) and [doesn't / currently works](http://tangerineindustries.com/Foundation/feature-mutation-trigger/without_mut.html)
 
## What Qualifies As A Mutation Event ##
The following qualifies as events post document.load
- Elements added / removed from the DOM
- JS style attributes change
- JS show / hide

## Plugins to update ##
- [x] Tabs
- [x] Interchange
- [x] Equalizer
- [ ] Accordion
- [ ] Reveal
- [ ] Magellan
- [ ] Orbit
- [x] Sticky (PR #9462 )
- [x] Togglers and Responsive Nav (PR #9475 )

# Related Issues #
*Please consolidate bug issues against this issue and not #9126*

kball:
I'm going to take a look at this for slider and see if it will fix #8391

kball:
Found a bug in drilldown menus (surfacing in responsive menus) that I believe mutation observers will fix.  Adding them.

kball:
@coreysyms I noticed that the mutation observer implementation doesn't currently debounce... resulting in often getting lots of triggers all together.  Can you add that?

prkirby:
Was this implemented for Orbit Slider? I am running 6.3.1, and have an equalizer inside of an orbit slider that is collapsing to 0 for everything but the first slide. 

`$ourProcess.on('slidechange.zf.orbit', function() {
      setTimeout(function() {
        Foundation.reInit($ourProcess.find('.process-info'));
      }, 200);
    });`

The code above works after waiting for orbit to finish its mutations, I believe, but it is obviously not a fix and couldn't be used in production. Is there something I'm missing or a way I could fix this?

[...]

PR #9379, opened 2016-11-16
Refactor slider and provide handle position fallback for hidden sliders

kball:
Refactors slider to make it easier to do a reinit, and then adds a mutation observer to do it. Provides a way to workaround https://github.com/zurb/foundation-sites/issues/8391.  @coreysyms as the mutation observer expert, can you take a look?

### DIFF
--- a/docs/pages/slider.md
+++ b/docs/pages/slider.md
@@ -130,3 +130,12 @@ It's possible to use both the JavaScript slider and the native slider in the sam
 - To disable the slider, add `disabled` as an attribute, instead of a class.
 - No support for vertical orientation.
 - No support for two handles.
+
+## Reflow
+
+The slider takes into account the width of the handles when calculating how to display itself. This means that if the slider is initially hidden, or hidden while the value is adjusted, the resulting visual will be slightly different because the width of the handle is indeterminate.  If this is problematic, you can use JavaScript to cause the slider to reflow at the time that you change it from being hidden.  Example:
+
+```js
+$('#my-slider').show();
+$('#my-slider').foundation('_reflow');
+```
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -574,27 +574,3 @@ Foundation.plugin(Slider, 'Slider');
 
 }(jQuery);
 
-//*********this is in case we go to static, absolute positions instead of dynamic positioning********
-// this.setSteps(function() {
-//   _this._events();
-//   var initStart = _this.options.positions[_this.options.initialStart - 1] || null;
-//   var initEnd = _this.options.initialEnd ? _this.options.position[_this.options.initialEnd - 1] : null;
-//   if (initStart || initEnd) {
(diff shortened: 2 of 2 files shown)


kball:
Should note that this is for https://github.com/zurb/foundation-sites/issues/9320

coreysyms:
Originally I thought Slider would be a good candidate for a mutation observer, but now I'm rethinking it. For one. After digging through Slider, it seems that everything is based on either user interaction or initialization values. And if Slider is initialized with values, the CSS percentages would still be valid because Slider doesn't care what the physical width of the slider bar is, but rather the percentage of the total user input value. So 25% of 100px or 0px is still 25%.

To test this, I dropped in this if statement in `_setHandlePos` in the current version of Slider.
`if (isNaN(movement)) {`
`  $hndl.css(lOrT, pctOfBar * 100 + '%');`
`} else {`
`  $hndl.css(lOrT, movement + '%');`
`}`

The truth in this is presented as you see the "fill bar" is correctly the width it needs to be, all I did was give the "handle" the same value as "fill". You can see that without the above snippet, the "fill" is correctly displayed, while the handle is not, in #8391 when toggled from display none.

The second reason for not doing mutation on Slider is that I can't think of a good reason to do so. If we fix the layout in the "display:none" scenario through proper initialization, then what else is there to watch for? Who in their right mind is going to manipulate a slider's physical attributes?

Finally, a mutation event will trigger almost constantly with Slider, as an inline width is applied with every handle pull, so we would have to be super careful about not triggering a reflow when really the user is manipulating the Slider as intended.

That all being said, I do think the refactoring of reInit is better, and we should drop in some better init() CSS stylings to make it layout correctly on load like in my example above.

kball:
Makes sense to me... I actually played around with something very similar to your above code... the outcome is very slightly different in the `display:none` case because the handle width isn't taken into account, but my inclination is this is probably ok... I'll update to remove the mutation observer & add that init.

kball:
On the codepen example for #8391 the difference in layout due to not accounting for the handle is like this:

a_pen_by_kevin_ball

I think this is probably "good enough", especially with the workaround that you can now do a reflow if you really need to.

coreysyms:
That is what my tweak showed too. I think with your additions and refactoring, and an additional note to the docs that "reflow" is now an option, this will suffice. Good stuff.

kball:
@coreysyms updated with docs as well.  Thanks for helping work through this!  Go ahead and merge if this looks good to you.

coreysyms:
this looks great, but whats up with Travis bombing, I'm not a super GIT person, but your code looks and works fine to me but is it ok to merge with errors? phantomjs version number is the stated issue.

kball:
@coreysyms sadly we get a number of "phantom" (:-P)  errors from travis having trouble installing phantomjs... restarting the travis builds resolved it. Given your feedback and the now passing Travis going to merge this.

PR #9389, opened 2016-11-18
Changed the settings file name to settings-sites

imabdulsal:
This PR is related to zurb/foundation-emails#609

**Note:**
@kball Notice that I have not deployed the new settings file, so it should be deployed in the releasing process.

### DIFF
--- a/docs/assets/scss/docs.scss
+++ b/docs/assets/scss/docs.scss
@@ -7,7 +7,7 @@
 //
 // Feel free to submit pull requests to foundation-docs like you would Foundation itself!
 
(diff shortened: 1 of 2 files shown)


kball:
Hey @abdullahsalem this ends up breaking the docs build due to not being able to find the renamed variable... I think you probably need to rename `_scss/settings/_settings.scss`

abdullahsalem:
@kball according to our offline discussion:
We will keep that name as it is. And if we need to further differentiation, we will add a suffix to the other kind of Foundation editions i.e. foundation-emails, settings-emails..

PR #9393, opened 2016-11-18
Make invalid inputs opaque

### LINKED ISSUE #9376: Inconsistency with error input background opacity
In [scss/forms/_error.scss:46](https://github.com/zurb/foundation-sites/blob/a214c1b12f324ac84cce78908a68d37e1cae0621/scss/forms/_error.scss#L46), `@mixin form-input-error` (used by `.is-invalid-input` for the input error state) uses a partially transparent background.

This allows to use the `alter` color from the global color palette `$foundation-palette`, but it creates an inconsistency between the standard and the error input background color.

Preview of inputs on different background colors
https://codepen.io/ncoden/pen/eBzPBK

webdevan:
Fixes #9376 Changed colours on abide error inputs to contrast better

### DIFF
--- a/scss/forms/_error.scss
+++ b/scss/forms/_error.scss
@@ -37,14 +37,17 @@ $input-error-font-weight: $global-weight-bold !default;
 /// Styles the background and border of an input field to have an error state.
 ///
 /// @param {Color} $background [$alert-color] - Color to use for the background and border.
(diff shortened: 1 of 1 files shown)


ncoden:
Thank you @webdevan. :+1:

PR #9394, opened 2016-11-19
Test of new corrective flexbox class

paxperscientiam:
So, I've mentioned elsewhere that there is a deficiency in the flexbox version of Foundation. This fix could come as a new class or could simply baked in (the latter is probably the way to go).

A so called "beam" class is meant to ensure that a flexbox based layout fills the viewport even when content is sparse.

Why?
• Makes it easy for sticky footer to stay at bottom no matter how sparse a page's content is.
• Heights of flexbox children are not affected by image load delay, which creates the nasty side effect of sudden height increases.

Why not set the height to 100%?
````html
<body>
<header></header>
<main></main>
<footer></footer>
</body>
````

Let's say that you have just a bit of content in the main element. Let's also say that you'd like your footer to stay at the bottom. The temptation is to set the height of main to 100%; however, this causes your layout to take up more than 100% of the viewport.

So, why not use auto height for main? Not only will the footer not be at the bottom of the viewport, elements will "snap" in place as images load in.

In order to demonstrate this, I am including a new test file at (http://localhost:3000/flex-grid/column-row-beam.html).

EDIT: grammar, clarifcation

### DIFF
--- a/test/visual/flex-grid/column-row-beam.html
+++ b/test/visual/flex-grid/column-row-beam.html
@@ -1,0 +1,113 @@
+<!doctype html>
+<!--[if IE 9]><html class="lt-ie10" lang="en" > <![endif]-->
+<html class="no-js" lang="en" dir="ltr">
+ <head>
+  <meta charset="utf-8">
+  <meta http-equiv="X-UA-Compatible" content="IE=edge">
+  <title>Foundation for Sites Testing</title>
+  <link href="../assets/css/foundation-flex.css" rel="stylesheet" />
+ </head>
+ <body id="body" data-toggler=".beam">
+  <style>
+    .beam {
+     max-width: 75em;
+     margin-left: auto;
+     margin-right: auto;
+     display: flex;
+     flex-direction: column;
+     flex-wrap: nowrap;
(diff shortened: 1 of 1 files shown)


kball:
I think this is a pretty interesting idea.  @andycochran @brettsmason @rafibomb @zurbrandon would be interested in your thoughts.  (check out the branch & try the visual test)

Right now this pr is simply an example, but we could add this as a helper class... I think its pretty cool.  @zurbrandon is this something we'll get as a part of the apps grid work you're doing?

kball:
OK, awesome, had a conversation with @zurbrandon out of band and it sounds like we're going to be getting this functionality (but in a more flexible way, with a bunch of additional goodies) as a part of porting the apps grid we created for Foundation for Apps into Foundation for Sites.  We should have an initial PR for this sometime next week, targeting integration in 6.4.

kball:
With the above in mind going to close this PR.

paxperscientiam:
@kball So glad to hear that ideas like this are in the works! Thanks for the update. I guess I'll hang back and wait for the PR!

PR #9395, opened 2016-11-20
Remove redundant documentation base-typography.md / typography_base.md

karland:
It seems that base-typography has been renamed into typography-base some time ago and then the two files existed in parallel.


karland:
I don't think there is anything wrong with this PR. The Travis error is `The command "npm install" failed and exited with 1 during .`

PR #9396, opened 2016-11-20
Should't the foundation.css build process depend on settings?

karland:
`foundation.scss` pulls in all files from under `/scss/`, but ignores the `settings/settings.scss`. As a result `foundation.scss` uses the `!default` values instead of the values from the `settings.scss`. I discovered this behaviour when I was testing something and a change in the settings did not change the resulting `foundation.css`.

### DIFF
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -9,7 +9,7 @@
 @import "normalize";
 
 // Settings
-// include your own `settings` here or 
+// import your own `settings` here or 
 // import and modify the default settings through
 // @import "settings/settings";
 
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -8,6 +8,9 @@
 // Dependencies
 @import "normalize";
 
+// Settings
+@import "settings/settings";
+
 // Sass utilities
 @import 'util/util';
 
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -9,7 +9,9 @@
 @import "normalize";
 
 // Settings
-@import "settings/settings";
+// include your own `settings` here or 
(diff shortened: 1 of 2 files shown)


brettsmason:
@andycochran @ncoden @kball does this look OK to you?
Makes sense to me.

ncoden:
Well... `/settings/settings.scss` is automatically generated by the gulp `deploy:settings` task. Because of `deploy` is based on `build`, we cannot use `/settings/settings.scss` in `build:sass`.

I think `/settings/settings.scss` is intended to be copied by the user for its own sass project. I agree  the file paht and name is not clear about this purpose.

kball:
Hmm... I think this may not play nicely with setting things up in an app or other location, where we typically have you include settings and then the items you want to include... letting you move settings to wherever you want in your own app structure without mucking with the library code itself... Not saying that is the ideal setup, but I think this seemingly small change would actually have cascading impacts on existing sites and apps

karland:
I wasn't sure about the impact of this change, so I put a questionmark at the end of this PR. The docs are pretty clear about how to use the `_settings.scss`:

[http://foundation.zurb.com/sites/docs/sass.html](http://foundation.zurb.com/sites/docs/sass.html)

I can see @kball's point. If a user has included a `mysettings.scss` its values would be overriden by a following `_settings.scss`. 

If you want, we could put something like this in foundation.scss:

```SASS
// Settings
// include your own `settings` here or 
// import and modify the default settings through
// @import "settings/settings";
```

If you want to leave everything as is, I am happy for this PR to be closed.

ncoden (approved):
👍

PR #9399, opened 2016-11-21
Remove important from max-width on tooltip mixin

tadeusrox:
I was thinking about open an Issue, but I choose to use a Pull Request instead. 

I was using  tooltip element on my job and found this line of code, I can not find why on commit history. I guess the use of important will make the people who want customize the tooltip width very hard.


kball:
This makes 100% sense to me - I see no reason why !important should be on tooltip widths.

PR #9407, opened 2016-11-25
Tabs: account updateHistory only if deepLink:true.

xhezairbey:
The advent of deep links feature to F6 Tabs was a real boon to the component.
I couldn't refrain from manually updating my local copy of `foundation.tabs.js` just to face a new issue; regardless whether the deepLink option is set **true/false**, tab hashes are always being appended to the URL.

I honestly don't think that was the intended outcome from the work of @ahebrank . I think it makes sense to restrict accounting for `updateHistory` if and only if `deepLink:true`, otherwise never bother and continue to function like it used to.

P.S. I've also fixed docs depicting tabs with deepLink attribute. Wasn't aware that it should go to the master branch.

### DIFF
--- a/docs/pages/tabs.md
+++ b/docs/pages/tabs.md
@@ -156,13 +156,13 @@ Add the attribute `data-deep-link="true" to a tabstrip to allow anchoring to and
 
 ```html_example
 <ul class="tabs" data-deep-link="true" data-tabs id="deeplinked-tabs">
-  <li class="tabs-title is-active"><a href="#panel1c" aria-selected="true">Tab 1</a></li>
+  <li class="tabs-title is-active"><a href="#panel1d" aria-selected="true">Tab 1</a></li>
   <li class="tabs-title"><a href="#panel2d">Tab 2</a></li>
   <li class="tabs-title"><a href="#panel3d">Tab 3</a></li>
   <li class="tabs-title"><a href="#panel4d">Tab 4</a></li>
 </ul>
 
-<div class="tabs-content" data-tabs-content="collapsing-tabs">
+<div class="tabs-content" data-tabs-content="deeplinked-tabs">
   <div class="tabs-panel is-active" id="panel1d">
     <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
   </div>
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -246,7 +246,7 @@ class Tabs {
         history.replaceState({}, '', anchor);
       }
     } else {
-      location.hash = '';
+      history.replaceState({}, '', window.location.pathname);
     }
 
     /**
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -237,11 +237,16 @@ class Tabs {
     this._openTab($target);
 
     //either replace or update browser history
-    var anchor = $target.find('a').attr('href');
-    if (this.options.updateHistory) {
-      history.pushState({}, "", anchor);
+    if (this.options.deepLink) {
+      var anchor = $target.find('a').attr('href');
+
+      if (this.options.updateHistory) {
+        history.pushState({}, '', anchor);
+      } else {
+        history.replaceState({}, '', anchor);
+      }
     } else {
-      history.replaceState({}, "", anchor);
+      location.hash = '';
     }
 
     /**


ahebrank:
It wasn't originally my intent to have the hash always appear in the browser history, but I'm OK with it since I think it's generally good accessibility and/or UX to have the browser history reflect the state of visible content on the page.  But... that would probably mean that `deepLink` should be on by default as well, so you're right that the current setup is a bit asymmetrical.

If this goes in, I'd take out the `else` branch of the conditional since I'm not sure what it's doing, and just have:

```
if (this.options.deepLink) {
    var anchor = $target.find('a').attr('href');
    if (this.options.updateHistory) {
        history.pushState({}, '', anchor);
    } else {
        history.replaceState({}, '', anchor);
    }
}
```

xhezairi:
Right, the else block intention was to use tabs without no hash at all, just like it used to function prior to your PR, whilst the block inside the condition when `deepLink:true` will append a hash at all times, with or without history. In a nutshell there are three different situations, and that aims to account all of them.

kball:
I believe prior behavior was to not touch history at all, so unless I'm totally misunderstanding I'd agree with @ahebrank about removing that 2nd else branch

xhezairi:
@kball you're right, it didn't. The issue I brought up here is to avoid have #hash being appended to the URL at all times, regardless of the deep-link value and this is a side-effect that arose with @ahebrank 's deep-link work.

I don't wanna sound insisting or anything cuz I have the impression we're not on the same page when referencing the problem here, tho my problem is with the anchor value being passed to `replaceState()`.

If so badly you want to ditch the else block, then could we please have [line 244](../blob/develop/js/foundation.tabs.js#L244) of `foundation.tabs.js` updated to this `history.replaceState({}, '', window .location.pathname);`. With that we can even remove the if conditional I wrapped around the history block.

kball:
@xhezairi I want to make sure I'm understanding what your desired outcome is... in the case that deepLink is false, shouldn't we not touch the history api at all?

xhezairi:
Now I'm wondering why I didn't use your simple line to make my point.
But basically yes, tho I also thought for situations when we want to use deepLink (add #anchors to URL) but not alter history, hence I saw adding the else block you're not fond off handy, to offer the same Tabs functionality as before.

via Newton Mail [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=9.2.5&pv=10.1.1&source=email_footer_2]
On Mon, Nov 28, 2016 at 11:24 PM, Kevin Ball <notifications@github.com> wrote:
@xhezairi [https://github.com/xhezairi] I want to make sure I'm understanding what your desired outcome is... in the case that deepLink is false, shouldn't we not touch the history api at all?

kball:
@xhezairi I think we may be talking about different else blocks... my suggestion would be to keep the inner one, just can the outer one... so the resulting code would look like:

```
if (this.options.deepLink) {
    var anchor = $target.find('a').attr('href');
    if (this.options.updateHistory) {
        history.pushState({}, '', anchor);
    } else {
        history.replaceState({}, '', anchor);
    }
}
```
rather than
```
if (this.options.deepLink) {
 var anchor = $target.find('a').attr('href');
 
  if (this.options.updateHistory) {
    history.pushState({}, '', anchor);
  } else {
    history.replaceState({}, '', anchor);
  }
} else {
  history.replaceState({}, "", anchor);
}
````

xhezairi:
I'm sure we're talking about the additional else block following (if 
this.options.deepLink), but please see my second commit I've followed to 
this PR, that should clarify my point. Thanks.

kball:
Sorry I feel like I'm being dense... what is the goal of manipulating the path/history at all in the case this component is not using it? (to be clear, I think the current state of things is a bug - I'm not advocating we don't merge your PR, just trying to make sure we get to the best solution)


If we manipulate it in a case where we're not using it (or for that matter, even if deepLink is true if the anchor isn't a part of this set of tabs), won't that cause us to interfere with other potentially legitimate uses of the anchor?

xhezairi:
@kball I know you're aiming for the best solutions. I'm also aware there are shortcomings to the different scenarios I could think of. Here's how I see it, in user stories: 

**1)** User doesn't want #tabIDs being shown on the URL. This is the way F6 tabs used to work until recently when @ahebrank made deepLink. This however doesn't make any one tab active even if the user lands directly to an URL like this ```http://localhost:3000/tabs.html#panel2```

**2)** User wants anchors but doesn't want them stored in browser history, thus clicking the browser back button will take him to straight to the previous **page** disregarding all the tab navigation he did.

**3)** User wants anchors appended to URL and have them alter browser history, therefore clicking the back button will go through the tab panes until it reaches previous page.

---

I'm sure there can be more than this, but these are the ones I've thought of. 
In all cases tho I think that browser history should be altered only if deepLink's are enabled, otherwise it'll look a little mysterious if the back button kept the user on the same page whilst switching tabs somewhere in the document.

kball:
Ok, so took another look through this again with fresh eyes... here's my thoughts:

Case 2 and 3 are handled by your change, moving all of the behavior inside of the if for deepLink.  I'm 100% on board with this change.

Case 1, the big outstanding question for me is, if someone has an anchor in their URL, then clicks on a tab that does NOT have deeplink enabled, what should happen?

In the code as written, the anchor will be removed via replaceState.
If we remove the 2nd else, nothing would happen to anchor.

Previous versions of F6 have the second behavior, and I *think* this is "more correct".  My reasoning is - the developer may wish to use anchor tags in other ways, and if they don't tell us they want tabs to manipulate them, we should just leave them alone completely.

I'm inclined to make this change directly and merge the PR... @xhezairi anything I'm missing in this reasoning?

kball:
Merged!  I did end up removing the else based on my above comment, but @xhezairi do please keep advocating for your case if you believe this was wrong - we can get to a shared understanding.

Thanks much for trying things out ahead of the release and catching this bug.

PR #9410, opened 2016-11-28
@Import util's with relative path

emolr:
Import util's with a relative path so foundation-sites scss can be used without depending on sass include_paths.

This is needed to use foundation-sites with angular-cli at the moment because include_paths can't be set using angular-cli's webpack setup

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -40,7 +40,7 @@
 //  35. Tooltip
 //  36. Top Bar
 
-@import 'util/util';
+@import '../util/util';
 
 // 1. Global
 // ---------


DaSchTour:
I don't think this is a good idea. Changing the framework to workaround possibly temporary missing features in angular-cli.
To be honest I think angular-cli is a very bad tool that restricts developers very hard. Just stop using angular-cli and everything is fine. You'll see the world gets a lot bigger without.

emolr:
@DaSchTour I do agree with you though, but foundation then needs to work with include paths to work, otherwise it will fail. 
I can just use my fork, but thought it would be nice for others in the same boat that it could be used with angular-cli. I'm fine with both, and yeah you are right, ill close this issue, then other who has the same issue will be able to find the solution for it :)

PR #9413, opened 2016-11-28
Remove seemingly unnecessary left offset from sticky

kball:
As far as I can tell from testing our visual tests & in docs, the JavaScript applied left offset for sticky is completely unnecessary, and leads to bugs when combined with off canvas.

Found this while looking at sticky within https://github.com/zurb/foundation-sites/pull/9357


kball:
cc @colin-marshall @gehasia as other folks who've dug into sticky... can either/both of you take a look?

kball:
Going to merge this.

PR #9415, opened 2016-11-29
Use $callout-font-color-alt in _callout.scss

IllusionPerdu:
Resolve problem with font color in dark background and use  $callout-font-color-alt ;)

### DIFF
--- a/scss/components/_callout.scss
+++ b/scss/components/_callout.scss
@@ -51,8 +51,6 @@ $callout-link-tint: 30% !default;
   border: $callout-border;
   border-radius: $callout-radius;
 
-  color: $callout-font-color;
-
   // Respect the padding, fool.
   > :first-child {
     margin-top: 0;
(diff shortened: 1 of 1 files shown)


kball:
Seems reasonable to me - @andycochran what do you think?

andycochran:
Nice, @IllusionPerdu! This is a good use of the new `color-pick-contrast()` function.

andycochran:
I think this PR is okay to merge. But it makes me think of a couple things. 

- Default-colored links do not meet AA contrast requirements in Callouts with non-white backgrounds. The primary blue is just barely dark enough to meet AA requirements on white. 

- Links in Callouts should get an accessible color. @wolfgang-demeter pointed this out in #9098, for which @ncoden had some great insight.

kball:
Ok gonna go ahead and merge this.  Do we think #9098 captures enough info to address these 2 issues?

PR #9426, opened 2016-11-30
Get offcanvas tests workign with new offcanvas implementation

kball:


kball:
This gets the offcanvas tests working with the new offcanvas implementation.  Going to merge once travis verifies, but would love it if @Owlbertz and @brettsmason could collaborate on expanding this test suite to better cover all of the new options we've got.

brettsmason:
@kball Happy to help guide this, though don't really know anything about travis so maybe an advisory role for me 😄 
Do you think we need more visual tests too? That I can do.

PR #9429, opened 2016-12-01
Pass on making off canvas overlay behave properly in both absolute an…

kball:
A few last fixes on off canvas overlay for @brettsmason


brettsmason:
Thank you! As confirmed, these work great.

Issue #9412, opened 2016-11-28
[Build] Normalize should be a Bower Dependency in project and Templates


rafibomb:
Running 6.3 get's this error

```
Error: File to import not found or unreadable: normalize
       Parent style sheet: bower_components/foundation-sites/scss/foundation.scss
        on line 9 of bower_components/foundation-sites/scss/foundation.scss
>> @import "normalize";
   ^
```

Normalize needs to be added as a real dependency in the Foundation project. Also needs to be added to project templates
- zurb/foundation-sites-template
- zurb/foundation-zurb-template

rafibomb:
@ncoden Are you able to look into this one?

ncoden:
I can, but not for now. Maybe next week. :/

ncoden:
Now `normalize-scss` is imported and compiled inside `foundation-sites` as a real dependency, I don't see what you mean by "needs to be added to project templates". `foundation-zurb-template` require `"foundation-sites": "~6.2.3"`, not the foundation's dependencies. 🤔

mediabeastnz:
Is there a way to get around this to test?

When installing via npm and then compiling the SCSS I get an error saying it can't find the normalize file.

The path in `foundation.scss` has `@import "normalize";`. Which doesn't exist in that directory.

ncoden:
@mediabeastnz If you want to test the merged fix above, you can do:
```shell
$ git fetch zurb 17e1179b5f0fe1d47448dd1636f7f063fb248e64
$ git checkout FETCH_HEAD
```

But I think there is still a bug. `@import "normalize";` works for the compilation of `foundation.css` because the path to the dependency files `node_modules/normalize-scss/sass` is included in gulp. This shouldn't work with a project which include foundation partials...

ncoden:
At least _partially_ fixed in https://github.com/zurb/foundation-sites/pull/9449.

marcos-abreu:
I'm also having the same problem when installing using `npm` (`npm install foundation-sites`) and trying to use my own scripts to compile the sass files.

kball:
so... I'm a little confused about this because 

1. the current build is working for me... what am I doing differently?
2. this seems to overly subscribe how your dependencies are located - what happens if someone uses a custom node modules path, or is using bower instead of npm?  Paths should be a part of the build, not the src

kball:
oops the above should be in the PR... fixing

michaelBenin:
[...]

PR #9430, opened 2016-12-01
add normalize dependency

kball:
Since we expect normalize in order to make Foundation function (at least in SCSS builds) now, it needs to be a true dependency, not just a dev dependency

Half the fix to #9412

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,6 +28,7 @@
   ],
   "dependencies": {
     "jquery": "~2.2.0",
+    "normalize.scss": "^6.0.0",
     "what-input": "~2.0.0"
   }
 }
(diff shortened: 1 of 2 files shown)


kball:
@ncoden you did the normalize change... can you confirm I'm not crazy thinking this should be an actual dependency given we expect people to primarily use scss?

ncoden:
@kball Well... people can use the Foundation repository to:
- contribute to it -> they need standard and dev dependencies.
- use `foundation.css` in their projects -> normalize is compiled inside, they don't need any dependencies (if they import jQuery themself).
- use `foundation.scss` or partials in their projects -> they need normalize.

So yes, normalize should be a standard dependency.

PR #9433, opened 2016-12-02
update button,badge,label docs for new a11y palette

andycochran:
Per, @kball's [request](https://github.com/zurb/foundation-sites/pull/9319#issuecomment-264334071), this PR updates the docs for Buttons, Badges, and Labels to explain why text colors no longer match out-of-the box, and warns about reducing accessibility by making button text colors match.


andycochran:
I also changed the order of classes in the examples, which makes it easier to scan and notice the differences in each line's markup. Let me know if this change isn't desirable. 

In doing so, I added a disabled hollow button. This highlights a background color issue: when you hover disabled hollow buttons the background fills in.

andycochran:
@zurb/yetinauts Please let me know if these new docs are clear. We need to get this into 6.3.

docs/pages/button.md
```diff
@@ -68,11 +74,12 @@ Give a button additional meaning by adding a coloring class, or `.disabled` to c
...
-<button class="secondary hollow button" href="#">Secondary Color</button>
-<button class="success hollow button" href="#">Success Color</button>
-<button class="alert hollow button" href="#">Alert Color</button>
-<button class="warning hollow button" href="#">Warning Color</button>
+<button class="hollow button" href="#">Primary</button>
+<button class="hollow button secondary" href="#">Secondary</button>
+<button class="hollow button success" href="#">Success</button>
+<button class="hollow button alert" href="#">Alert</button>
+<button class="hollow button warning" href="#">Warning</button>
+<button class="hollow button" href="#" disabled>Disabled</button>
```
brettsmason: Should this add the class of disabled to match the example above?

brettsmason:
@andycochran I've had a read through and all looks fine to me, other than one thing I just commented on.
👍

docs/pages/button.md
andycochran: I debated that. But this way illustrates how the `disabled` attribute works for `button` elements. I'd think we want to recommend `button` elements use the attribute, not the class. Right?

brettsmason: Oh don't get me wrong, I totally understand why you did it and you are right to encourage it. Just wondered if it would cause confusion thats all. Maybe where you mention about using the `.disabled` class you could also mention the attribute and keep this as is?

andycochran: I tweaked it a little anyways. If you missed it others might too. Thanks.

ncoden (approved):
LGTM 👍

PR #9435, opened 2016-12-02
Doc fix for align-self helper. See commit notes for more information

lesliedoherty:

Referencing function: https://github.com/zurb/foundation-sites/blob/develop/scss/util/_flex.scss#L55


lesliedoherty:
D'oh, will use Master -> master for doc fix

PR #9436, opened 2016-12-02
Flexbox Helper Document Fix

lesliedoherty:
The docs read to pass in any horizontal value with an example "bottom", but the function is requiring a vertical alignment (so 'bottom' would work but 'left, center, right' would not.)

Referencing https://github.com/zurb/foundation-sites/blob/develop/scss/util/_flex.scss#L55


ncoden:
Thank you @lesliedoherty

PR #9439, opened 2016-12-03
Factorize breakpoints map handling

### LINKED ISSUE #9425: [Grid] 6.3 - Factorize breakpoint list + breakpoint name
Reference merged PR: https://github.com/zurb/foundation-sites/pull/9365

This is to factorize two mixins, using a mixin that handle the "breakpoint list + breakpoint name"

ncoden:
Fix #9425 - Factorize the handling of breakpoints map in `grid-column-gutter` and `grid-column-margin` mixins.

Changes:
- Add `@mixin -zf-breakpoint-value` which generate its content with the value `$-zf-bp-value` depending on its parameter, like `@mixin grid-column-gutter` and `@mixin grid-column-margin` did (see the `@mixin -zf-breakpoint-value` doc)
- Use `@mixin -zf-breakpoint-value` in `@mixin grid-column-gutter` and `@mixin grid-column-margin` and remove duplicated code.
- Allow `@function -zf-get-bp-val` to take a single value as responsive config `$map` and return it regardless of `$value`.
- Remove useless `@function grid-column-gutter`

### DIFF
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -60,7 +60,7 @@
   @include grid-column-uncollapse($gutter);
 }
 
-/// Sets bottom marin on grid columns to match gutters
+/// Sets bottom margin on grid columns to match gutters
 /// @param {Number|Keyword} $margin [auto]
 ///   The bottom margin on grid columns, accepts multiple values:
 ///   - A single value will make the margin that exact size.
--- a/scss/grid/_gutter.scss
+++ b/scss/grid/_gutter.scss
@@ -6,24 +6,6 @@
 /// @group grid
 ////
 
-/// Get a gutter size for a given breakpoint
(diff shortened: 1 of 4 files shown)


ncoden:
@andycochran What do you think ?

andycochran:
Nice. I think this works, so long as we're okay with breaking any apps that currently make use of the `grid-column-gutter()` function.

ncoden:
@andycochran `@function grid-column-gutter` was introduced in https://github.com/zurb/foundation-sites/pull/8528/commits/948ddd51ef7d56d6dc4a3b3703fe9b12ce4d73ba (https://github.com/zurb/foundation-sites/pull/8528) for `zurb:v6.3`. But I can depreciate with a warning instead of deleting it, and we will remove it in `6.4` ?

(poke @kball)

andycochran:
Oh, it's a new function? Nix it.

PR #9449, opened 2016-12-05
Fix #9412 - Import normalize-scss with its full path

### LINKED ISSUE #9412: [Build] Normalize should be a Bower Dependency in project and Templates

ncoden:
Import normalize-scss with its full path instead of including the assets path in gulp-sass.

Fix https://github.com/zurb/foundation-sites/issues/9412

### DIFF
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -22,6 +22,10 @@ module.exports = {
   ],
 
   // Sass
+  SASS_DEPS_PATHS: [
+    'node_modules'
+  ],
+
   SASS_DOC_PATHS: [
     'scss',
     'node_modules/motion-ui/src',
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -22,10 +22,6 @@ module.exports = {
   ],
 
   // Sass
-  SASS_DEPS_PATHS: [
-    'node_modules/normalize-scss/sass'
-  ],
-
   SASS_DOC_PATHS: [
     'scss',
     'node_modules/motion-ui/src',
--- a/gulp/tasks/sass.js
+++ b/gulp/tasks/sass.js
@@ -20,7 +20,9 @@ gulp.task('sass:foundation', function() {
   return gulp.src(['assets/*'])
     .pipe(sourcemaps.init())
     .pipe(plumber())
-    .pipe(sass().on('error', sass.logError))
+    .pipe(sass({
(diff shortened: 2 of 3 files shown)


ncoden:
Poke @kball `foundation@6.3.0-rc1` is broken. This is a hotfix.

oxyc:
This doesn't work with foundation as a NPM dependency as `normalize-scss` will be in the root `node_modules/`. I can fix it by requiring an older version of `normalize-scss` in my project root and thus installing foundations version within it's own `node_modules/` directory.

kball:
so... I'm a little confused about this because

the current build is working for me... what am I doing differently?
this seems to overly subscribe how your dependencies are located - what happens if someone uses a custom node modules path, or is using bower instead of npm? Paths should be a part of the build, not the src

oxyc:
https://github.com/npm/npm/blob/ff47bc138e877835f1c0f419fea5f5672110678a/CHANGELOG.md#flat-flat-flat

ncoden:
When Foundation is used in a Sass project via NPM or Bower, it is by importing `/scss/foundation.scss`, with the project build process. But `/scss/foundation.scss` needs its own dependencies, needs to know their path, and thus **they can't be imported via NPM or Bower as real dependencies**.

We have to install our Sass dependencies (like `normalize-scss`) via NPM only as a dev dependencies, then copy them to `/scss/_vendor` in the `sass` gulp build task. By example for normalize, it can be imported in `/scss/foundation.scss` with:
```scss
@import "_vendor/normalize-scss/sass/normalize";
```

When the package is published, the `sass` gulp build task is called too, and Sass dependencies are still available in `/scss/_vendor`.


@HugoGiraudel @kball What do you think ?

ncoden:
I close this PR. See: https://github.com/zurb/foundation-sites/pull/9458

HugoGiraudel:
That seems like a valid idea.

ncoden:
@HugoGiraudel Did you seen the PR ?

PR #9450, opened 2016-12-05
use map for responsive embed ratios

andycochran:
Per #8199, and @kball's suggestion, this PR introduces a map of responsive embed ratios. 

```scss
$responsive-embed-ratios: (
  default: 4 by 3,
  widescreen: 16 by 9,
) !default;
```

This allows you to easily add/change aspect ratio classes. You can, for example, make your default ratio 16:9, remove widescreen, and add other aspect ratios.

```scss
$responsive-embed-ratios: (
  default: 16 by 9,
  vertical: 9 by 16,
  panorama: 256 by 81,
  square: 1 by 1,
);
```


andycochran:
@ncoden, @brettsmason, I'd appreciate your thoughts on this. Would be good to squeeze it in before 6.3 (or it'll need to wait for 6.4).

scss/components/_responsive-embed.scss
```diff
@@ -29,7 +28,7 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
 
 /// Creates a responsive embed container.
 /// @param {List} $ratio [$responsive-embed-ratio] - Ratio to use for the container, formatted as `x by y`.
-@mixin responsive-embed($ratio: $responsive-embed-ratio) {
+@mixin responsive-embed($ratio: 4 by 3) {
```
ncoden: Can you use `map-get($responsive-embed-ratios,default)` here (or a null variable you override after), instead of a magic number ?

scss/components/_responsive-embed.scss
```diff
@@ -50,10 +49,14 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
 
 @mixin foundation-responsive-embed {
   .responsive-embed, .flex-video {
-    @include responsive-embed;
+    @include responsive-embed( map-get($responsive-embed-ratios,default) );
 
-    &.widescreen {
-      padding-bottom: responsive-embed($responsive-embed-ratio-widescreen);
+    @each $name, $ratio in $responsive-embed-ratios {
+      @if $name != default {
```
ncoden: I think the "map for modifiers with a default value that should be skipped" can be factorized with the breakpoint stuff :)

For example: [/scss/util/_mixins.scss:234](https://github.com/ncoden/foundation-sites/blob/6c8c184f6785478f4c70b1e7ab0bbac32c6ac3d0/scss/util/_mixins.scss#L234)

scss/components/_responsive-embed.scss
```diff
@@ -50,10 +49,14 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
...
+        &.#{$name} {
+          padding-bottom: responsive-embed($ratio);
+        }
+      }
```
ncoden: You are generating classes from the indexes of `$responsive-embed-ratios`. If it is a map intended to be customized by the user, I don't think it is safe.

For a lower specificity and to secure the `responsive-embed` modifiers inside a namespace, we could generate modifiers like `&-#{$name}` (so `.responsive-embed-panorama, .flex-video-panorama), or maybe as a generic utility with its own namespace (like `.ration-panorama`) ?

What do you think ?

ncoden (commented):
I have some remarks, but that's great 👍

ncoden:
Travis fails to download phantomjs at almost every pull request (403 http status) 😞 ...

scss/components/_responsive-embed.scss
andycochran: Fixed.

scss/components/_responsive-embed.scss
andycochran: Fixed.

scss/components/_responsive-embed.scss
```diff
@@ -29,7 +28,10 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
 
 /// Creates a responsive embed container.
 /// @param {List} $ratio [$responsive-embed-ratio] - Ratio to use for the container, formatted as `x by y`.
```
ncoden: The default parameter is not `$responsive-embed-ratio` anymore.

scss/components/_responsive-embed.scss
```diff
@@ -29,7 +28,10 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
...
-@mixin responsive-embed($ratio: $responsive-embed-ratio) {
+@mixin responsive-embed($ratio: null) {
+  @if $ratio == null {
+    $ratio: map-get($responsive-embed-ratios, default);
+  }
```
ncoden:
```scss
@mixin responsive-embed($ratio: default) {
  @if type-of ratio == 'string' {
    $ratio: map-get($responsive-embed-ratios, $ratio);
  }
  ...
}
```
So you can use:
```scss
@include responsive-embed(widescreen); // -> 16 by 9
@include responsive-embed(16 by 10);   // -> 16 by 10
```

scss/components/_responsive-embed.scss
```diff
@@ -29,7 +28,10 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
 
 /// Creates a responsive embed container.
 /// @param {List} $ratio [$responsive-embed-ratio] - Ratio to use for the container, formatted as `x by y`.
-@mixin responsive-embed($ratio: $responsive-embed-ratio) {
+@mixin responsive-embed($ratio: null) {
```
ncoden: Is it still stuff restricted to `responsive-embed` only ?
- If yes, you could name it `responsive-embed-ratio`.
- If not, you could name it `box-ratio` or (`{something}-ratio`) for a global ratio modifier utility.

ncoden (commented):
I have some more remarks:

scss/components/_responsive-embed.scss
andycochran: Thanks for catching this. I've updated this param note.

scss/components/_responsive-embed.scss
andycochran: Did you mean to comment on the function that calculates the percentage? I've abstracted that and moved it to `_math.scss`, as is could be reused for other purposes: 

```scss
/// Calculates the height as a percentage of the width for a given ratio.
/// @param {List} $ratio - Ratio to use to calculate the height, formatted as `x by y`.
/// @return {Number} A percentage value for the height relative to the width of a responsive container.
@function ratio-to-percentage($ratio) {
  $w: nth($ratio, 1);
  $h: nth($ratio, 3);
  @return $h / $w * 100%;
}
```

scss/components/_responsive-embed.scss
andycochran: Great idea! I've made this change. Now the `responsive-embed()` mixin accepts a key from the `$responsive-embed-ratios` map or a list formatted as `x by y`.

scss/components/_responsive-embed.scss
```diff
@@ -50,10 +52,14 @@ $responsive-embed-ratio-widescreen: 16 by 9 !default;
 
 @mixin foundation-responsive-embed {
   .responsive-embed, .flex-video {
-    @include responsive-embed;
+    @include responsive-embed( map-get($responsive-embed-ratios,default) );
+
+    $responsive-embed-ratios: map-remove($responsive-embed-ratios,default);
```
andycochran: I've changed this to a new local `$ratios` var. Would it be better to do the following instead of redefining the var or setting a new var? 
```scss
@each $name, $ratio in map-remove($responsive-embed-ratios,default) { ...
```

ncoden: I don't think. It's not easily readable.

scss/components/_responsive-embed.scss
ncoden: No, I talked about the whole mixin. This is a mixin that set the ratio of a any box, not only `.responsive-embed` and `.flex-video`.

kball:
There appears to be an actual bug caught by travis, not a "phantom" error

kball:
Ah I think the travis issue is that the expected usage changed, but the test did not.  It is testing for the `responsive-embed` sass function, but that appears to no longer exist... @andycochran can you please update the test for the new expected behavior, plus add a migration note?  Or if the behavior change is accidental, fix it?  Thanks

andycochran:
You're right. I broke the Sass Unit Test for component because the `responsive-embed($ratio)` function has been deleted. It was abstracted to work for components other than Responsive Embed and moved to the new `ratio-to-percentage` function in `_math.scss`. 

Should be fixed now.

PR #9454, opened 2016-12-05
Fix hamburger icon

imabdulsal:
This PR fixes a bug that has been noticed in [v6.3.0-rc1](https://github.com/zurb/foundation-sites/releases/tag/v6.3.0-rc1). The bug disappears some Hamburger icon's bars, and the full bars only appear on **hover** mode

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -73,6 +73,11 @@
     $spacing: floor($spacing);
   }
 
+  @for $i from 2 through $bars {
+    $offset: ($weight + $spacing) * ($i - 1);
+    $shadow: append($shadow, 0 $offset 0 $color, comma);
+  }
+
   // Icon container
   position: relative;
   display: inline-block;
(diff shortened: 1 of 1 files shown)


scss/util/_mixins.scss
```diff
@@ -92,14 +92,15 @@
...
-    box-shadow: $shadow;
 
     content: '';
 
     @for $i from 2 through $bars {
       $offset: ($weight + $spacing) * ($i - 1);
       $shadow: append($shadow, 0 $offset 0 $color, comma);
     }
+
+    box-shadow: $shadow;
```
ncoden: I think it would be more clear to keep `box-shadow` in the property list `background`, and move the logic stuff above, line 75 or 86.

ncoden (commented):
This bug was introduced in https://github.com/zurb/foundation-sites/pull/9333. Sorry about that.

ncoden (approved):
👍

PR #9458, opened 2016-12-06
Fix #9412 - import Sass dependencies in vendor folder

### LINKED ISSUE #9412: [Build] Normalize should be a Bower Dependency in project and Templates

ncoden:
Previous PR: https://github.com/zurb/foundation-sites/pull/9449

#### Changes:
- Revert https://github.com/zurb/foundation-sites/pull/9430: Make `normalize-scss` a dev dependency.
- Copy Sass dependencies to `/_vendor/scss` and import them in Sass from there.

Why not importing the Sass dependencies as real dependencies and from `node_modules`:
- `normalize-scss` has a different name between NPM and Bower. 😠 
- We cannot assume the path to our real dependencies. Old NPM versions store them in our `node_modules` folder, while newer versions store them in the `node_modules` folder of the root project.

### DIFF
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -6,10 +6,10 @@
  */
 
 // Dependencies
-@import "normalize";
+@import "../_vendor/scss/normalize-scss/sass/normalize";
 
 // Settings
-// import your own `settings` here or 
+// import your own `settings` here or
 // import and modify the default settings through
 // @import "settings/settings";
 
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -6,7 +6,7 @@
  */
 
 // Dependencies
-@import "../_vendor/scss/normalize-scss/sass/normalize";
+@import "../_vendor/normalize-scss/sass/normalize";
 
 // Settings
 // import your own `settings` here or
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 .sass-cache/*
 .yardoc
 _build
+_vendor
 .customizer
 
 bower_components
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,6 @@
 .sass-cache
 _build
+_vendor
 bower_components
 config
 docs
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 .sass-cache
 _build
-_vendor
 bower_components
 config
 docs
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,6 @@
   ],
   "dependencies": {
     "jquery": "~2.2.0",
-    "normalize.scss": "^6.0.0",
     "what-input": "~4.0.3"
   }
 }
--- a/gulp/config.js
+++ b/gulp/config.js
@@ -22,8 +22,8 @@ module.exports = {
   ],
 
   // Sass
-  SASS_DEPS_PATHS: [
-    'node_modules/normalize-scss/sass'
+  SASS_DEPS_FILES: [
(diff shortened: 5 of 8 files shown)


scss/foundation.scss
```diff
@@ -6,10 +6,10 @@
  */
 
 // Dependencies
-@import "normalize";
+@import "../_vendor/scss/normalize-scss/sass/normalize";
```
HugoGiraudel: This is a bit of a bummer. Ideally you’d want `../vendor/normalize`.

oxyc:
My workflow has always consisted of requiring foundation as a dependency in my project and then importing `~foundation-sites/scss/foundation` into my sass entry. This PR breaks this setup as the `_vendor` directory wont exist until gulp builds it. I guess I can live with adding normalize myself and then importing individual component files rather than `foundation.scss`.
There might be others with this setup though.

ncoden:
@oxyc The `_vendor` directory is created by the build script before publishing releases to NPM or Bower. If you import foundation-sites from them, the `_vendor` directory should exists with `normalize-scss` inside.

oxyc:
@ncoden doesn't the entry in `.npmignore` prevent that?

ncoden:
@oxyc True. I mistaken it with `.gitignore`

scss/foundation.scss
kball: I think @ncoden's second suggestion makes sense. I don't think we're going to be doing JS in this way (dependency there works a little better, if still pretty poorly), but it still makes sense to scope by package as we may end up with multiple files to selectively import

ncoden: @kball " second suggestion": do you mean this PR or my previous comment ?

kball: `../_vendor/normalize-scss/normalize`

ncoden:
@HugoGiraudel @kball Finally, I import normalize with `../_vendor/normalize-scss/sass/normalize`. I agree a shorter path like `../_vendor/normalize-scss/normalize` is more beautiful, but It is too complicated to set with the current gulp config file `gulp/config.js`.

ncoden:
@HugoGiraudel We still have issues with our Sass dependencies.

Because of does not have its own registry like NPM, and uses the Github one, it does not find `/vendor` which is ignored by git.

The solutions we see for now is to:
- Drop the Bower support: we can't (https://twitter.com/kbal11/status/805891962993713152)
- Pick up the useful part of our dependencies into the Foundation code
- Import `/vendor` in Github (but with only the required file of our dependencie moved to it)

What do you think ? Do you know how we could resolve this problem ?

HugoGiraudel:
I think I would go with cherry-picked what you need from your dependencies while preserving link to original source in your documentation (SassDoc or whatever). That would make the building process easier and is likely to be enough for now.

ncoden:
I just found a _stupidly simple_ idea and submitted it to the rest of the team on Slack:

Why do not _simply_ provide a 📦  `dist/foundation.scss` distribution file with all the dependencies included, instead of trying to make our `scss/foundation.scss` usable with an other build ?

ncoden:
See also: https://github.com/zurb/foundation-sites/pull/9469

Issue #8292, opened 2016-02-29
[SCSS 6.2] possible bug, default $foundation-palette colors overriding old variables even if @include add-foundation-colors; is not included in _settings.scss


erutan:
I was wondering why something calling $alert-color had suddenly turned salmon color (I hadn't added in `$foundation-palette` + `@include add-foundation-colors`) yet in my 6.2 migration.

An easy override, but I'd rather fix the issue, so I copied the `$foundation-palette` map from the default `_settings.scss` and edited the values inside. The element was colored as originally intended - however commenting / uncommenting `@include add-foundation-colors` had no (immediately noticeable) difference.

If I understand correctly, `$foundation-palette` is supposed to be optional and we can continue to use traditional variables, but at the moment (I assume) it's `!default` values are overriding the old traditional variables.  It'd also be nice to have some explanation of the implications of using the SASS map (I assume for autogenerating classes? hence why we'd want some things inside it and some more tertiary colors outside of it) - maybe that's up somewhere and I just haven't stumbled across it.

kball:
@JeremyEnglert can you take a look at this?

kball:
Found this while triaging outstanding v6.3.  @andycochran do you have time to take a look?

andycochran:
@kball Q: Is `$foundation-palette` still optional? Seems like that doesn't have to be the case since v6.3 is a minor version bump. Can we deprecate the old way of doing color variables now?

brettsmason:
I'd agree with removing the old method too 👍

andycochran:
Let's not let the old method hang around until v6.4. 😬 
It's now or then.

kball:
Good point. There's actually a few things that have warnings that they'll be removed in 6.3 as well that we need to get through.  Lets include this along with those and any fixes for bugs we find into a second RC mid next week.

kball:
so question - does getting rid of the old method mean getting rid of the helper variables `$primary-color`, etc?  Or getting rid of them from internal use but leaving the option to add them with `@include add-foundation-colors;` for your own use?  Or what were you thinking?

andycochran:
Good question, @kball. The more I think about it, the more complicated it gets. 

It looks like we're inconsistent in the way we're using color variables, e.g.: 
- In `_error.scss` we use `$input-background-invalid: map-get($foundation-palette, alert) !default;`
- But in `_meter.scss` we use `$meter-fill-good: $success-color !default;`

[...]

Issue #9424, opened 2016-11-30
Inline `code` tags inside `a` tags have incorrect styling


impressivewebs:
**How can we reproduce this bug?**

1. Put HTML `<code>` tags inside of a hyperink

**What did you expect to happen?**
Portions of a paragraph wrapped in  `<code>` tags that are inside links should appear like links.

**What happened instead?**
The portions of the paragraph wrapped in `<code>` tags do not appear like links.

**Test case:**

[Here is a CodePen demo using Foundation 6](http://codepen.io/impressivewebs/pen/VmMMwY)

Notice in the 2nd paragraph, the term "addEventListner" is inside `code` tags in the HTML, but it's also inside a link. It appears as if it's not inside a link, so the rest of the link looks like two links. The term "removeEventListener" in that same paragraph is also inside a link, but it does not look like a link at all because it's the only thing linked.

This can be fixed by adding the following code:

    a code {
      color: inherit;
      background: inherit;
      border: inherit;
      padding: inherit;
    }

[Here is a corrected demo](http://codepen.io/impressivewebs/pen/VmrVPp)

You could also do `all: inherit`, but I'm not sure if browser support is good enough for Foundation to use that. This also happens in Bootstrap 3 and 4, and I've filed an issue with them too. I've tested this only in Foundation 6.2.3.

ncoden:
Hi @impressivewebs,

This is a beautiful example of why it is a bad idea to apply graphical CSS properties automatically. In some case, we don't want these properties to be applied, and it is very hard to remove them (poke @kball).

But I don't think it would be a good idea to add a special case (`a code`) above it to reset the unwanted property. It would not be clean, nor maintainable.

I would advice to:
- Add `.link`, the current link automatically applied on `a`.
- Add `.code`, the current code box.
- Add a [scope](http://csswizardry.com/2015/03/more-transparent-ui-code-with-namespaces/) `.s-code` to _locally_ automatically apply `.code` on `code`.
- Remove graphical properties from `code`.

So in your example, you can use:
```html
<p>
  Some text and a <code class="code">code block</code>.
</p>

<p>
  Then a
  <a href="#">
    link with a
    <code>code tag without graphical properties</code>
  </a>.
</p>
```
Or
```html
<div class="s-code">
  <p>
    Some text and a <code>code block</code>.
  </p>

  <p>
    Then a
    <a href="#">
      link with a
      <code class="link">code tag without graphical properties</code>
    </a>.
  </p>
</div>
```

[...]

PR #9459, opened 2016-12-06
Color Variables Cleanup

andycochran:
This PR addresses some issues brought up in #8292. 

- Cleaned up the inconsistencies in how we reference color variables from `$foundation-palette`, switching instances of `map-get()` to `get-color()`. 

- New `$[badge,button,label]-classes` lists in `_settings.scss` determines which colors in `$foundation-palette` will generate color classes. If you don't need a certain class, simply remove it from the list.

- Update the description of `add-foundation-colors()` mixin. 

- Update docs about color: Global Styles, Badge, Button, Label.


scss/components/_badge.scss
```diff
@@ -18,6 +18,10 @@ $badge-color: $white !default;
 /// @type Color
 $badge-color-alt: $black !default;
 
+/// Coloring classes. The names in this list output classes in your CSS, like `.secondary`, `.success`, and so on. Each value must also be in the `$foundation-palette` map.
+/// @type List
+$badge-classes: ( secondary success warning alert ) !default;
```
ncoden: I find this a bit enclosing. Why not this ?
```scss
// Use the default palette
$badge-palette: $foundation-palette !default;

// Use its own palette
$badge-palette: (
    black: #FFF,
    red: #F00,
    rainbow: #08FF80
) !default;

// Use the default palette, plus one or several color
$badge-palette: map-merge($foundation-palette, (
    rainbow: #08FF80
)) !default;

// Use the default palette, except one or several color
$badge-palette: map-remove($foundation-palette, (
    primary,
    secondary
)) !default;
```

andycochran: @brettsmason You might have thought on this?

kball:
@brettsmason @andycochran @ncoden is this good to include in the next RC?  I'd like to package it early morning Weds PDT time.

scss/components/_badge.scss
ncoden: > But if components have their own color/class maps, there's no reason for $foundation-palette to be a map.

To iterate on it too, like we did with `$foundation-palette`, but with their own colors.

andycochran: Ooh! I see. I like this idea. I'll update the PR…

andycochran: But should the default be…
```
$badge-palette: map-remove( $foundation-palette, (primary) ) !default;
```
…so that we don't generate an unnecessary `.primary` class? That's one thing I was trying to do with the `$badge-palette` list.

ncoden: I think the default modifier is useful, but I do not find example to show it... :confused:

ncoden: Make a search of `primary` in `_build/assets/css/foundation.css`, we don't strip it out everywhere.

I was searching a case where `.primary` could be useful to _restore_ properties, like in https://github.com/zurb/foundation-sites/issues/9424. It is useful in your example only because you consider `.primary` as a non-default modifier (it is not based on the default button config) and want to _apply_ its properties...

andycochran:
@kball Hopefully I got this done in time. With @ncoden's changes, I think it's now ready for the next RC.

PR #9462, opened 2016-12-06
Mutation Observer for Sticky

coreysyms:
Sticky is receiving a mutation observer. Sticky can now listen to
elements within its DOM scope for mutation changes. This will allow the
sticky element, and it’s anchor container (if applicable) to receive a
notification if any element is changed within it’s respective space
calling setSizes() and calc() in the same manor as a resize would.


coreysyms:
See this file to test a typical set up of user enacted changes in a sticky div
[index.html.zip](https://github.com/zurb/foundation-sites/files/635182/index.html.zip)

coreysyms:
tagging @colin-marshall cause @kball said you have done some work in sticky in the past and may have some opinions / suggestions

kball:
As discussed in slack lets hold off on merging this until 6.3 goes out

js/foundation.sticky.js
```diff
@@ -138,6 +140,32 @@ class Sticky {
...
+			_this._calc(false);
+			if (_this.canStick) {
+			  if (!_this.isOn) {
+				_this._events(id);
+			  }
+			} else if (_this.isOn) {
+			  _this._pauseListeners(scrollListener);
+			}
+		});
+	});
```
coreysyms: good call, will look into

kball: @coreysyms are you going to have time to look at this in the next week or two? Would like to get this in for 6.3.1

coreysyms:
Ill take care this weekend. Sorry forgot about it.

kball:
I went ahead and fixed the styling and 1 bug I found while testing, then merged!  This already improved some of our visual tests (e.g. sticky holding an accordion that changes size) Nice work @coreysyms

mhair:
So this mutation observer aspect of sticky didn't make it into 6.3.1, right? I'm not having any luck with sticky component (sidebar application) on Win IE11 and IE10 on 6.3.1. Believe it also didn't work in a version of Safari, but good other than that so far. Not sure if this is the cause, but the container height never gets set in IE.

jinch:
I know this is an old post but I to have a Foundation 6.3.1 project with the same issues and this did not seem to fix in that version. Did this patch make it to the 6.3.1 release and if not, Is there a work around to apply a Mutation Observer for Sticky in this particular version of Foundation?

DanielRuf:
Hi @jinch, if this is not in your version you can apply it using `patch-package`.

I will try to check later, if it is included in your version.

See https://github.com/zurb/foundation-sites/pull/9462.patch or https://github.com/zurb/foundation-sites/pull/9462.diff

Please ping me here again if I forget to answer in ~6 hours.

jinch:
Thanks Daniel, unfortunately a patch update in my circumstance isn't that straight forward as the site is integrated in a back-end I don't directly administer. Ideally it would be great to find a solution for 6.3.1 if at all possible.

I was looking at triggering the height to recalculate : `$('.sticky').foundation('_calc', true);` but this did not seem to work either. In my case I have an anchor element that is initially hidden until a button is clicked. The sticky element does not calculate the anchor height initially so its not able to properly stick to it. For some reason when the browser window is resized it resets and calculates the height and the sticky element again. I also tried `$(window).trigger('resize');` to force the resize behavior but no luck.

PR #9469, opened 2016-12-07
[TMP] Fix Sass dependencies import for Bower

ncoden:
:warning: **This is a temporary solution** (see below)

Unlike NPM, Bower does not have its own registry and uses the Github repository. Like `js/`, `sass/` and `dist/`, all the files required by Bower must be published on Git.

**Changes:**
- Pin the Sass dependencies versions and update `yarn.lock`
- Move only the require files to `/vendor`
- Publish the Sass dependencies in `vendor/` on git.

**This is not a maintainable dependency management. This approach is limited to the v6.3.0 pre-releases and should be removed before any stable version.**


kball:
Committing for now for rc2 so we can keep testing.  I think we have a better solution brewing

PR #9474, opened 2016-12-08
New Off-canvas docs (WIP)

rafibomb:
This is the updated off-canvas docs for the new version from @brettsmason 

Looking for some feedback to "finalize" it.


docs/assets/scss/docs.scss
```diff
@@ -190,3 +190,7 @@ a#notice {
   0%   { opacity: 0; transform: scale(.7, .7); }
   100% { opacity: 1; transform: scale(1, 1);}
 }
+
+.menu .docs-nav-version a {
```
rafibomb: fix for docs sidebar

docs/layout/default.html
```diff
@@ -48,6 +48,7 @@
 
       <div class="expanded row">
         <div class="medium-9 large-10 medium-push-3 large-push-2 columns">
+<!-- do not indent -->
```
rafibomb: a note for future contributors

rafibomb:
@brettsmason Where should me mention the `.is-open` class that's added to the `.off-canvas`.?

rafibomb:
@brettsmason See any classes or options missing?

docs/pages/off-canvas.md
```diff
@@ -109,39 +186,47 @@ A design can have two menus: one on the left, and one on the right. Be sure that
 
 ---
 
-## Title Bar
+## Off-canvas Behavior
 
-If you need a simple title bar to toggle the off-canvas, `.title-bar` is here to help. It supports left- and right-aligned sections.
+You can switch the default behavior of the off-canvas from pushing the page over as it open to overlaping the page by adding the `.is-transition-overlap` to the `.off-canvas`. The default option is `.is-transition-push`.
```
brettsmason: Think this needs revising a little. The transition is actually applied by adding the option `data-transition="push|overlap` to a panel (push being the default). The class is added depending on what option they pick - I'm not sure the user needs to know that the class is added, thats just part of the styling that makes it work?

brettsmason (commented):
Overall this is looking brilliant, thanks @rafibomb!
There are however some things missing, and some incorrect info (eg anything with the class of `is-` is added via a JavaScript option.

I think it will be easier if I push up some changes, rather than comment on every bit if thats OK?

rafibomb:
@brettsmason Thanks for the feedback! I'll update the PR then ask you to merge it!

brettsmason:
@rafibomb I've been through and added some corrections, and reworked a few examples etc.
Could you take a look and see if its OK?

rafibomb:
Sweet! This is amazing - You are a legend!

PR #9475, opened 2016-12-08
Trigger mutates on toggler & responsive nav show

kball:
We have a set of components (drilldown being the main one I'm seeing bugs in right now) that depend on understanding their own sizes to behave properly.  This has historically broken in a couple cases:

1. When the internal content changes.
2. When they are hidden at the time the calculation is happening

We're working on addressing case 1 via the automatic data-mutate listeners.  We haven't found a performant, global solution to case 2 (which would involve watching all parents all the time.. yick)

However, we have 2 known situations _within_ the framework typically used for toggling visibility/layout... toggler and responsive nav.  This PR looks for [data-mutate] flagged components inside the targets for these toggles and triggers them.

Fixes a bug where drilldowns in a responsive nav end up being totally hidden (see `test/visual/responsive-menu/responsive-menu-right-dropdown.html`) among others.

### DIFF
--- a/js/foundation.responsiveToggle.js
+++ b/js/foundation.responsiveToggle.js
@@ -101,6 +101,7 @@ class ResponsiveToggle {
              * @event ResponsiveToggle#toggled
              */
             this.$element.trigger('toggled.zf.responsiveToggle');
+            this.$targetMenu.find('[data-mutate]').triggerHandler('mutateme.zf.trigger');
           });
         }
         else {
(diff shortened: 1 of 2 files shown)


kball:
@coreysyms am I totally crazy to do this? Any reason this would cause problems?

coreysyms:
This makes a lot of sense to me, and it is the way I did tabs. I'm glad you brought this up.
So I think moving forward we should use these guides for mutation.

1. If the framework manipulates / toggles elements in the plug in, a mutation event should fire or respond.

2. If the user initiates mutation / toggle though code of their own, provide the means to reInit() or reCalc() or whatever is needed. 

This should be merged, are you waiting on this for 6.3.1?

PR #9477, opened 2016-12-09
Correct path to JS dist in package.json

DaSchTour:
Looks like the path in dist were changed in build process but not in package.json


kball:
Good catch @DaSchTour!  Going to do something similar for bower.

PR #9478, opened 2016-12-09
Add SCSS compilation into SCSS dist file

ncoden:
📦  Use `gulp-shopify-sass` (@ncoden’s fork) to compile SCSS files and dependencies into a distribution SCSS file `dist/scss/foundation.scss`.

⚠️  **BREAKING CHANGE**: This file should be imported into external projects instead of `scss/foundation.scss`. To import Foundation partials manually, the dependencies `normalize-scss` and `sassy-lists` should be imported before.

#### Changes:
- Revert #9430, #9458, #9469
- Add `ncoden/gulp-shopify-sass` dependency.
- Add `sass:foundationSCSS` gulp task, which concat SCSS files into `_build/assets/scss/foundation.scss`.
- Rename `sass:foundation` gulp task to `sass:foundationCSS`.
- Update `deploy:dist` gulp task to handle SCSS dist file

#### Other changes:
- Add missing paths to include for Sass in `customiser` gulp task.


kball:
Trying to test this out - it doesn't look like `gulp deploy:dist` is creating a dist/scss/foundation.scss file

ncoden:
Yes. I seen `gulp deploy:dist` does not create `dist/scss`, `dist/css` and `dist/js`. But it is not creating it in `:develop` neither.

kball:
Given the breadth of change & importance of getting this right, that this references an unmerged branch of `gulp-shopify-sass`, that we're scheduled to release 6.3 next week, and the fact that our current workaround is working and transparent to end users, my inclination is to defer this to 6.4.  @ncoden does that seem reasonable to you?

ncoden:
Our current workaround ~~totally sucks~~ is not very good.
- Dependencies are pushed on Git.
- Dependencies versions are pinned.
- Internal files and relations of dependencies are assumed.
- It is difficult to partially import the components (have to import `/vendor/...` too).

We can wait for `gulp-shopify-sass` to merge the PR, to ensure there is no mistakes. But we shoudn't release `6.3` with a "workaround" in it !

I am far less confident in the current workaround than in this real proper fix.

kball:
Digging into the lack of copy turned up https://github.com/zurb/foundation-sites/pull/9485 and the fact we have not been packaging final assets properly in the later 6.3 rcs.  Going to spin out another RC fixing that, still not sure on merging this PR. Would love to hear what @andycochran @brettsmason and @HugoGiraudel think of this approach.

kball:
The gulpfile is not fully up to date... `gulp deploy:dist` now breaks...

ncoden:
Fixed

kball:
Bringing discussion from slack into this channel:

This doesn't work out of the box with e.g. the ZURB template because of the way the settings file is currently set up, especially given it imports from `util/util`.

This has highlighted the fragility in our current build approach, to quote @ncoden 

> the whole framework was build around importable sources files, with all the bad practices that goes with it. (aka we "do the work on import", and not “import > set settings > generate“)

We need to spend more time and energy on getting this right.  Going to defer to 6.4 and use our current dependency workaround (which has stood up under test so far) for 6.3.

PR #9479, opened 2016-12-09
Extend visibility classes

paxperscientiam:
I am working on a visibility test with intention to extend visibility classes in two ways:

* Boolean conjunction: EG: elements with the hypothetical class `.show-for-medium-and-landscape` would be visible *only* when the screen size is "small" *and* in "landscape" orientation. 

* Boolean exclusive disjunction: Elements with both `.show-for-medium` and `.show-for-landscape` should display *only* when either condition is true; element should be invisible when both are true and when both are false.

### DIFF
--- a/scss/components/_visibility.scss
+++ b/scss/components/_visibility.scss
@@ -29,12 +29,12 @@
   @if $upper-bound-size == null {
     @if $orientation == null {
       @media screen and (max-width: $lower-bound) {
-	display: none !important;
+	      display: none !important;
       }
     }
     @else {
       @media screen and (max-width: $lower-bound) and (orientation: $orientation) {
-	display: none !important;
+	      display: none !important;
       }
     }
   }
@@ -43,12 +43,12 @@
 
     @if $orientation == null {
       @media screen and (max-width: $lower-bound), screen and (min-width: $upper-bound) {
-	display: none !important;
+	      display: none !important;
       }
     }
     @else {
       @media screen and (max-width: $lower-bound) and (orientation: $orientation), screen and (min-width: $upper-bound) and (orientation: $orientation) {
-	display: none !important;
+	      display: none !important;
       }
     }
   }
@@ -90,10 +90,10 @@
   @each $size in $breakpoint-classes {
     @if $size != $-zf-zero-breakpoint {
       .hide-for-#{$size} {
-	@include hide-for($size);
+	      @include hide-for($size);
       }
       .show-for-#{$size} {
-	@include show-for($size);
+	      @include show-for($size);
       }
     }
     .hide-for-#{$size}-only {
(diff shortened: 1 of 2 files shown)


ncoden:
Hi @paxperscientiam 

Could you give a set of cases where these features would be useful ?
I wonder if there would be enough people using these classes to worth the increased file size.

paxperscientiam:
Hello @ncoden 

## conjunction
* show logo when small and landscape; hide logo in all other cases.
* show title-bar when small and portrait; show side-title-bar (what-have-you) when small and horizontal

## XOR
* harder to think of off the top of my head.

JeremyEnglert:
Just throwing in my two cents, but I see value in these additional classes.

Is the "and" necessary though?

.show-for-medium-landscape says the same thing in less characters. Not a deal breaker either way, just a thought.

jaysonbrown:
Deep linking doesn't seem to be working in the docs, also would this function only work on page load? I was hoping for deep-linking within the page without a refresh.

paxperscientiam:
@JeremyEnglert It's a placeholder; but I do agree that identifiers should be kept short as possible without sacrificing semantics.

paxperscientiam:
I have finally made some progress with this: "show-for-$size-osly" in conjunction with "show-for-$orientation" now works nicely (see new test hershttp://localhost:3000/visibility/index.html)

Could someone help me with ".hide-for-$size-only.hide-for-$orientation" programming? it seems like I would have to augment the breakpoint function; however, I'm unsure.

These style rules could be setup to be "opt-in" only (for sass users)s

kball:
@IamManchanda @JeremyEnglert can you help @paxperscientiam get this across the line?

IamManchanda:
Ohk @kball .... i will take a look as soon as possible from my side!

IamManchanda:
I am confused here 

> **Boolean exclusive disjunction**: Elements with both `.show-for-medium` and `.show-for-landscape` should display *only* when either condition is true; element should be invisible when both are true and when both are false.
https://en.wikipedia.org/wiki/Truth_table#Exclusive_disjunction

Do you mean,

```html
<div class="show-for-medium show-for-landscape">
 I will be only showed when i am either on medium screen or a landspace orientation. 
 If i am on both, i will be hidden. If i am on none, i will be hidden
</div>
```

Is it so ?? and if yes why so ??

DanielRuf:
> Is it so ?? and if yes why so ??

I guess because this does not change the old behavior.

paxperscientiam:
Sorry folks, I just don't know when I'll be able to work on this. It's been a while :/ 

Would anyone like to take over? If so, I'll close and you can just copy the code ... unless there's a builtin way for transferring PRs?

DanielRuf:
We can also push changes to your brsnch if that is ok.

paxperscientiam:
@DanielRuf Yes!

DanielRuf:
@ncoden 

> I wonder if there would be enough people using these classes to worth the increased file size.

Closed because of this and the scope and rare use cases?

PR #9482, opened 2016-12-09
Update resource paths for new dist paths

kball:
Following up on https://github.com/zurb/foundation-sites/pull/9477

Actually @DaSchTour do you know, should we add the `.ts` bindings to bower.json and package.js as well?

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -4,7 +4,7 @@
   "license": "MIT",
   "main": [
     "scss/foundation.scss",
-    "dist/foundation.js"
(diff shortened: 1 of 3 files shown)


ncoden (approved):

DaSchTour:
I only know about module resolution through node/package.json.
I don't think that there would be any benefit from adding it to bower.json and I don't know in which place.
I also couldn't find anything about package.js

And as I see https://github.com/moment/moment also only added it to package.json

PR #9498, opened 2016-12-13
Clarify data-toggle to use responsive-toggle

dgdblank:
When trying to use the responsive-toggle feature, I kept getting an error that '#' was not a valid method, because the element id was not being passed with data-toggle trigger. Added notes to clarify that the menu id should also be set on data-toggle: `data-toggle="MENUID"`.


ncoden:
Thank you @dgdblank.

Issue #9483, opened 2016-12-09
[Feature] Classes and mixins for Header styles


ncoden:
- Add a mixin to generate custom Header styles...
- Add classes for the default headers, like `.h1`, `.h2`, `.h3`...

ncoden:
Hum... I taked a look at typography, and I think it needs a big refactor:
- Reset and features are mixed.
- There is no one single `@mixin`.
- Everything is mixed in one big file, not organized by components.
- Config variables uses too common names and need a namespace.

About `$header-styles`:
- It is a map of properties for each header for each breakpoint (_breakpoint_ > _header_ > _properties_)
- It does not have value units and should apply them automatically
- It should apply default properties.

So I have several suggestions to factorize and simplify stuff:
**1/** For default properties, it should be applied _before_ we process the styles map, with a base mixin or placeholder for all headers, not here.

**2/** We can sort `$header-styles` by _name_ then _breakpoint_ (_header_ > _breakpoint_ > _properties_) to have a _component_ approach, and provide a mixin to generate single responsive and non-responsive headers.
```scss
// Generate styles for a header
@mixin header-style ($properties) { // ideally zf-header--style
   ... // like in scss/typography/_base.scss:330
}

// Generate header styles for each breakpoint in $map
@mixin header-responsive-style ($map) { // ideally zf-header-responsive--style
    @each $bp, $v in $map {
        @include breakpoint($bp) {
            @include header-size($v);
        }
    }
}

// Instead of scss/typography/_base.scss:330
// Generate default header styles
@each $header-name, $v in $header-styles {
    #{$header-name} {
        @include header-responsive-style($v);
    }
}
```
[scss/typography/_base.scss:330](https://github.com/ncoden/foundation-sites/blob/db1e825355925eb57aaa473995b6b848957fd9cc/scss/typography/_base.scss#L330)

**3/** We could factorize the handling of `$header-styles` into two mixins. `-zf-responsive-map` is not very useful in this case because we have process each map or properties (see below), but it could be useful elsewhere.

```scss
// Generate the given map of properties
@mixin -zf-properties-map($map) {
    @each $name, $value in $map {
        #{$name}: #{$v};
    }
}

// Generate the given responsive map: write a map of properties or a single properties for each breakpoint
@mixin -zf-responsive-map($map, $property: null) {
    @each $bp, $v in $map {
        @include breakpoint($bp) {
            // the value is a map of properties
            @if typeof $v == 'map' {
[...]

PR #9500, opened 2016-12-14
[WIP] feat: add header classes and mixins

ncoden:
See: #9483

#### Changes:
- Add a class for each header (`.h1`, `.h2`...)
- Add `@mixin header` to generate basic header styles
- Add `@mixin header-style` to generate graphic header styles
- Use these mixins in `scss/typography/base`.

### DIFF
--- a/_tmp.css
+++ b/_tmp.css
@@ -1,0 +1,3964 @@
+@charset "UTF-8";
+/**
+ * Foundation for Sites by ZURB
+ * Version 6.3.0-rc3
+ * foundation.zurb.com
+ * Licensed under MIT Open Source
+ */
+/*! normalize-scss | MIT/GPLv2 License | bit.ly/normalize-scss */
+/* Document
+       ========================================================================== */
+/**
+     * 1. Change the default font family in all browsers (opinionated).
+     * 2. Correct the line height in all browsers.
+     * 3. Prevent adjustments of font size after orientation changes in
+     *    IE on Windows Phone and in iOS.
+     */
+html {
+  font-family: sans-serif;
+  /* 1 */
+  line-height: 1.15;
+  /* 2 */
+  -ms-text-size-adjust: 100%;
+  /* 3 */
+  -webkit-text-size-adjust: 100%;
+  /* 3 */ }
+
+/* Sections
+       ========================================================================== */
+/**
+     * Remove the margin in all browsers (opinionated).
+     */
+body {
+  margin: 0; }
+
+/**
+     * Add the correct display in IE 9-.
+     */
+article,
+aside,
+footer,
+header,
+nav,
+section {
+  display: block; }
+
+/**
+     * Correct the font size and margin on `h1` elements within `section` and
+     * `article` contexts in Chrome, Firefox, and Safari.
+     */
+h1 {
+  font-size: 2em;
+  margin: 0.67em 0; }
+
+/* Grouping content
+       ========================================================================== */
+/**
+     * Add the correct display in IE 9-.
+     */
+figcaption,
+figure {
+  display: block; }
+
+/**
+     * Add the correct margin in IE 8.
+     */
+figure {
+  margin: 1em 40px; }
+
+/**
+     * 1. Add the correct box sizing in Firefox.
+     * 2. Show the overflow in Edge and IE.
+     */
+hr {
+  -webkit-box-sizing: content-box;
+          box-sizing: content-box;
+  /* 1 */
+  height: 0;
+  /* 1 */
+  overflow: visible;
+  /* 2 */ }
+
+/**
+     * Add the correct display in IE.
+     */
+main {
+  display: block; }
+
+/**
+     * 1. Correct the inheritance and scaling of font size in all browsers.
+     * 2. Correct the odd `em` font sizing in all browsers.
+     */
+pre {
+  font-family: monospace, monospace;
+  /* 1 */
+  font-size: 1em;
+  /* 2 */ }
+
+/* Links
+       ========================================================================== */
+/**
+     * 1. Remove the gray background on active links in IE 10.
+     * 2. Remove gaps in links underline in iOS 8+ and Safari 8+.
+     */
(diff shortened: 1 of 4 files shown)


kball:
@ncoden is it fair to say that these separated mixins (`header` and `header-style`) are separating the concepts of the reset & the graphical styles?   However, in the _generated css_ we still apply the graphical styles to the header elements themselves, not just the classes...

Given there is interest in moving all graphical styles into classes, do you think it makes sense to introduce a configuration setting (that we can gradually integrate/reference in more places) that restricts the application of graphical styles to only the classes and not the elements?

ncoden:
@kball 
1/ I don't think the header mixins are separated from the base typography context. But I think having all the base typography in one large file is not a good idea. This file should be split into many mixins into many files, one per component or group of closely related components. So _for now_, I put these mixins into their own file.

2/ We still apply the graphical styles to the header elements, yes. Otherwise, it would be a breaking change. I never thought about removing the default styles, not until the v7. Do you think we should ?

3/ Yes, I think it makes sense... https://github.com/zurb/foundation-sites/pull/8718 😄

brettsmason:
@ncoden somehow I didn't see this come up. Had a quick look and it looks great, amazing job!
I'll give it a proper test over the weekend 👍

kball:
@brettsmason were you able to give this a test?

DaSchTour:
I find this a bit confusing. For me it looks like styles are added for h1-h2 and than again for h1-h6 and .h1-.h6 and as header() includes header-style() it would include h1 h1, h1 h2 and so on.
I think there should be only one point where styles may possibly applied and not two.

ncoden:
@DaSchTour 
> it looks like styles are added for h1-h2...

You mean h1-h6 ?

> ...and than again for h1-h6 and .h1-.h6 and as header() includes header-style() it would include h1 h1, h1 h2 and so on...

_Base_ styles (from `@mixin header`) and _default sizes_ (from `@mixin header-style` with `$styles:(), $base: true`) are included at :315, and _Breakpoint styles_ (from `@mixin header-style`  with `$base: false`) are included at :318. Nothing else. There is no duplicated properties.

> I think there should be only one point where styles may possibly applied and not two.

I agree with you. We should dynamically generate the base styles of the headers too. But I wonder if we should generate sort of _reset_ properties for standard headers (h1-h6) if they are not in the map, since normalize does not do it.

DaSchTour:
@ncoden okay well, I see. I guess confusion came from the fact, that typography mixes resets and typography. Maybe that should be separated?

ncoden:
@DaSchTour That should be separated, for typography and for all others components.

kball:
@ncoden based on the comment above, do you want to separate for typography in this PR or is that a separate one, and this is ready to merge?

ncoden:
@kball It requires a lot of work, a refactor 😄 !  (And that's not the subject of this PR). So I'll try do it later.

kball:
@brettsmason you mentioned you were going to give this a test - did you? From your perspective are we good to merge?

brettsmason:
@ncoden @kball Sorry for the delay I forgot about this over Christmas.

So I've had a bit of a play with this. One thing I noticed - we added `.h1, .h2` etc to the heading styles section, but they aren't included in the first headings bit. This means the generated classes dont contain the font family, style, colour etc. Not sure if we should include them there or not?

I need to try out the mixins later today so I'll get back to you on those once I work out how to use them correctly.

ncoden:
> This means the generated classes dont contain the font family, style, colour etc

Right. @DaSchTour talked about a similar problem.

We can do something like this:
```scss
  %header {
    @include header();
  }

  // Headings styles
  @each $size, $headers in $header-styles {
    @include breakpoint($size) {
      @each $header, $header-defs in $headers {
        #{$header}, .#{$header} {
          @if $size == $-zf-zero-breakpoint {
            @extend %header;
          }
          @include header-style($header-defs);
        }
      }
    }
  }
```

We have to use a `%placeholder` here to do not duplicate styles across breakpoints (btw, there is a lot of places in the codebase where we should use it). However, this is just an example and there is better way to use it, like _inside_ the component with a `$extend` Boolean property.

kball:
@ncoden what is the status of this PR?

IamManchanda:
@kball @ncoden What's the state of this... I think we can close this .. No ??

oxyc:
I still think is a very useful feature and I use something similar in my own projects all the time. 

I feel heading styles shouldn't be tied to the heading level. For instance, I only want one `<h1>`, but maybe I want all section headings of a page to be styled the same way. Or maybe the `<h1>` is in the hero banner, rather than the content area on some pages. Therefore I often use BEM instead as  I prefer to use `section__title`, `page__title`, `teaser__title` etc.  Having a `@mixin header-style` makes this easy.

On several occasions I've needed to style a non-header element as a heading (`<select>`, pagination, primary menu, button). Having a `%header` placeholder helps a lot.

I've had several designs where different heading levels have different font-families. `%header` makes it possible to reset the styling when I don't want the styling tied to the level (mentioned in my first use case).

DanielRuf:
Definitely makes sense and is also helpful for good SEO.

oxyc:
Today I learned you can define placeholders multiple times, enabling stuff like this:

```scss
  %header {
    @include header();
  }

  // Headings placeholders
  @each $size, $headers in $header-styles {
    @include breakpoint($size) {
      @each $header, $header-defs in $headers {
        %#{$header} {
          @include header-style($header-defs);
        }
      }
    }
  }

%h1 {
  text-transform: uppercase;
}

  // Headings styles
  @each $size, $headers in $header-styles {
    @include breakpoint($size) {
      @each $header, $header-defs in $headers {
        #{$header}, .#{$header} {
          @if $size == $-zf-zero-breakpoint {
            @extend %header;
          }
          @extend %#{$header};
        }
      }
    }
  }


.accordion-title { @extend %h1; }
```

kball:
Closing old WIP PRs

PR #9506, opened 2016-12-15
Full list of includes for Sass docs

JeremyEnglert:
Currently users have to dig through the code or look at a starter project to get a full list of includes. This PR adds a full list of Sass includes into the Sass documentation. It uses 6.3 includes, so it should not be merged until after that launches.

@rafibomb, any objections?

### DIFF
--- a/docs/pages/sass.md
+++ b/docs/pages/sass.md
@@ -100,17 +100,50 @@ Each component has an **export mixin** which prints out the CSS for that compone
 @include foundation-everything;
 ```
 
-Our [starter projects](starter-projects.html) include the full list of imports, making it easy to comment out the components you don't need.
+Our [starter projects](starter-projects.html) include the full list of imports, making it easy to comment out the components you don't need. A full list is also included below. 
 
 ```scss
 @import 'foundation';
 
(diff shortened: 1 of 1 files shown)


docs/pages/sass.md
```diff
@@ -100,17 +100,50 @@ Each component has an **export mixin** which prints out the CSS for that compone
 @include foundation-everything;
 ```
 
-Our [starter projects](starter-projects.html) include the full list of imports, making it easy to comment out the components you don't need.
+Our [starter projects](starter-projects.html) include the full list of imports, making it easy to comment out the components you don't need. A full list is also included below. 
 
 ```scss
 @import 'foundation';
```
ncoden: `foundation` includes all the following files. So we could split the code block here and add a `Or`.

JeremyEnglert: `@import 'foundation` doesn't include those files.

`@include foundation-everything` does.

ncoden: Sorry, you're right. I thought it was imported files --'

ncoden (approved):

PR #9509, opened 2016-12-15
Fix #9499 - Add simple menu behavior for align-right

### LINKED ISSUE #9499: Simple menu with align right still has left margin
When using align-right I expected the margin to also switch from left to right, so that last child aligns with the outer edge.

```
// Align right and simple
&.simple.align-#{$global-right} {
      > li {
        margin-#{$global-right}: 0;
        margin-#{$global-left}: get-side($menu-item-padding, $global-left);
      }
}
```

Perhaps?

ncoden:
Fix #9499.

#### Changes:
- Make `@mixin menu-simple` taking a `$dir` parameter to generate margins.
- Add `@mixin menu-simple-direction($dir: $global-left)`: Direction modifier for a simple Menu.
- Add a special case for simple menu `.menu.simple` with `.align-right`. Use `@mixin menu-simple-direction` to change the menu direction to the right.

#### Other changes:
- Add `@function direction-opposite($dir)`: Return the opposite direction of $dir, between "top", "right", "bottom" and "left".
- Add documentation for `@mixin menu-simple`.
- Fix simple menu height (see https://github.com/zurb/foundation-sites/pull/9539)

### DIFF
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -289,6 +289,10 @@ $menu-border: $light-gray !default;
     // Simple
     &.simple {
       @include menu-simple;
+
+      &.align-#{$global-right} {
+        @include menu-simple-direction($global-right);
+      }
     }
 
     // Align right
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -140,15 +140,31 @@ $menu-border: $light-gray !default;
 }
 
 /// Creates a simple Menu, which has no padding or hover state.
-@mixin menu-simple {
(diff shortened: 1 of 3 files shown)


kball:
Is there any case where we'd want the current behavior?  Because unless I'm misreading this, no matter the direction passed in (including null) the output has now changed.

ncoden:
See: https://codepen.io/ncoden/pen/rWRpEM

When the menu is aligned on right, the margin between each `<li>` should be at their left.

ncoden:
Fixed

PR #9513, opened 2016-12-16
[WIP] Classes for all components

ncoden:
From: https://github.com/zurb/foundation-sites/pull/8718

### Planned changes:
- Separate *reset* properties from *component's structural and graphics* properties. Currently, this separations is unclear for all `foundation-forms` components
- Allow to make components be defined with always `.classes` instead of standard `<tags>`. Currently, lot of component define their non-reset properties on tags. It can lead to **unwanted properties** for someone using its own components.

### Previous changes:
_(30 April 2016)_
- Add `$global-namespace {Boolean} [false]` option which when enabled, make all components be defined in classes, instead of tags.
- Use `$global-namespace` for `label` and `select`

### Progress:
⚠️  Work in Progress

-

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -94,6 +94,10 @@ $global-text-direction: ltr !default;
 /// @type Boolean
 $global-flexbox: false !default;
 
+/// Always define components with classes instead of tags (besides the reset).
+/// @type Boolean
+$global-namespace: false !default;
+
 @if not map-has-key($foundation-palette, primary) {
   @error 'In $foundation-palette, you must have a color named "primary".';
 }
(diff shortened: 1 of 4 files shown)


IllusionPerdu:
I am not tested this change but i think they are a problem with the compoment "label" (not "form" version) with your commit it affects the "label" component

DanielRuf:
> I am not tested this change but i think they are a problem with the compoment "label" (not "form" version) with your commit it affects the "label" component

Can you provide a codepen / example where this is problematic?

ncoden:
@DanielRuf Don't worry, I'm closing this PR as it is a major scss refactor.

PR #9522, opened 2016-12-18
Fix #9521 - Single button in expanded button group

### LINKED ISSUE #9521: Expanded button group -- does not expand if only 1 button

```
<div class="expanded button-group">
  <a class="button">Expanded</a>  
</div>
```

the button should fill the container

**What happened instead?**

button was regular width

**Test case:**

N/a  --- see code above

ncoden:
Fix https://github.com/zurb/foundation-sites/issues/9521
A single button in a expanded button group was not expanded.

#### Changes:
- Add `width:100%` to the button when it is the only one in the button group.

**Note:** The way we are currently handling button groups is really dirty.

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -94,6 +94,12 @@ $buttongroup-radius-on-each: true !default;
       flex: 1 1 0px; // sass-lint:disable-line zero-unit
     }
     @else {
+      // One child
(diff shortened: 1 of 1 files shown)


kball:
On the note:  Yes, I believe @andycochran has also brought this up with some desire to refactor.

PR #9525, opened 2016-12-19
[Drilldown] Add an 'else' clause to not let the $wrapper property undefined

mlang38:
If the parent of the drilldown menu already has the 'is-drilldown' css class, the $wrapper property used everywhere in the plugin stays undefined. Fixes that issue.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -108,6 +108,8 @@ class Drilldown {
       this.$wrapper = $(this.options.wrapper).addClass('is-drilldown');
       if(this.options.animateHeight) this.$wrapper.addClass('animate-height');
       this.$wrapper = this.$element.wrap(this.$wrapper).parent().css(this._getMaxDims());
+    } else {
+      this.$wrapper = this.$element.parent().css(this._getMaxDims());
     }
   }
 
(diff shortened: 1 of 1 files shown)


js/foundation.drilldown.js
```diff
@@ -108,6 +108,8 @@ class Drilldown {
       this.$wrapper = $(this.options.wrapper).addClass('is-drilldown');
       if(this.options.animateHeight) this.$wrapper.addClass('animate-height');
       this.$wrapper = this.$element.wrap(this.$wrapper).parent().css(this._getMaxDims());
+    } else {
+      this.$wrapper = this.$element.parent().css(this._getMaxDims());
     }
```
ncoden: I would prefer:
```js
    // create a wrapper on element if it doesn't exist.
    if(!this.$element.parent().hasClass('is-drilldown')){
      this.$wrapper = $(this.options.wrapper).addClass('is-drilldown');
      if(this.options.animateHeight) this.$wrapper.addClass('animate-height');
      this.$element.wrap(this.$wrapper);
    }
    // set wrapper
    this.$wrapper = this.$element.parent();
    this.$wrapper.css(this._getMaxDims());
```
for clarity.

ncoden:
Hi @mlang38 

What do you think about the changes I mentioned above ?

mlang38:
No pb. 'else' clauses are evil :wink:

ncoden (approved):

PR #9528, opened 2016-12-19
updates card variable to fit standard naming

rafibomb:
We're using margin-bottom on the `$card-margin` variable. All other components that use margin-bottom follow the `$component-margin-bottom` naming convention.

This PR changes the variable name to do the same.


ncoden (approved):

rafibomb:
Retargeting to 6.4 for change that requires migration.

PR #9532, opened 2016-12-19
Tell bower & npm that it's okay to have jquery 3

### LINKED ISSUE #8975: Problems with jQuery 3
**What happend?**
Please update the deprecated method calls.

**e.g.**
`sample.load(fn())` to `sample.on('load', fn())`

1. Use Foundation 5 in a project via NodeJs
--> dependency in package.json: jQuery >= 2.x.x 
----> jQuery 3.0.0 is installed (breaking changes)

Foundation should not throw any error. Please update the code so that it don't use deprecated  methods.

**What happened instead?**
Foundation throws error.

Thanks

kball:
Fix #8975

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,7 +28,7 @@
     "yarn.lock"
   ],
   "dependencies": {
-    "jquery": "~2.2.0",
+    "jquery": ">= 2.2.0",
     "what-input": "~4.0.3"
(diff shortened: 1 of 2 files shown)


ncoden:
This PR title is so comforting :)

PR #9538, opened 2016-12-20
Fixed broken CDN links

mcvayokay:
The CDN links are pointing to version 6.3.0, which is not available on cdn.jsdelivr.net.


ncoden:
I think we should rather change the link to an other CDN for Foundation 6.3

KiC-Kilo:
Yeah I just noticed that these docs are specific to 6.3.0.  Is there another CDN where these files are available?

ncoden:
https://cdnjs.com/libraries/foundation

KiC-Kilo:
Committed the new links.

ncoden (approved):

PR #9539, opened 2016-12-20
Update _menu.scss

PixelGrinch:
It appears that centering the menu adds unnecessary/undesired height to the container.
My Proposed update to add vertical-align: top; to the inline-block applied container, fixes this issue.
Thank you
Bert (PixelGrinch)


ncoden:
Codepen: https://codepen.io/ncoden/pen/VmRMrq

ncoden (approved):

PR #9540, opened 2016-12-20
Increase the precision of the pow function when calculating luminance

natewiebe13:
When calculating the luminance of a color, the wrong value was being returned.

This was discovered when using `color-pick-contrast()` using `#ff364c` and `#313942`.
The contrast ratio calculator by Lea Verou returns 3.3, while Foundation was calculating 2.9. Foundation returns 3.3 after increasing the precision.

Link: http://leaverou.github.io/contrast-ratio/#%23313942-on-%23ff364c

### DIFF
--- a/scss/util/_color.scss
+++ b/scss/util/_color.scss
@@ -23,7 +23,7 @@
     $rgb: nth($rgba, $i);
     $rgb: $rgb / 255;
 
-    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4, 18));
+    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4));
 
     $rgba2: append($rgba2, $rgb);
   }
--- a/scss/util/_color.scss
+++ b/scss/util/_color.scss
@@ -23,7 +23,7 @@
     $rgb: nth($rgba, $i);
     $rgb: $rgb / 255;
 
-    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4));
+    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4, 18));
 
     $rgba2: append($rgba2, $rgb);
   }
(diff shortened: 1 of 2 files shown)


scss/util/_color.scss
```diff
@@ -23,7 +23,7 @@
     $rgb: nth($rgba, $i);
     $rgb: $rgb / 255;
 
-    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4));
+    $rgb: if($rgb < 0.03928, $rgb / 12.92, pow(($rgb + 0.055) / 1.055, 2.4, 18));
```
ncoden: Why 18 ?

ncoden: 👍

ncoden: Poke @andycochran, you written this function. What do you think ?

andycochran: Should we just change the `pow()` function's default in  `_math.scss`?

natewiebe13: From what I can tell the `pow()` function is only used here, so it may be better just to change the default. @ncoden, let me know if you want me to make the change.

ncoden: I do not know how the precision parameter is used by pow and what would be the advantages and disadvantages of increasing it. All I can ask is "should we expect `row` to return an approximate value by default for this range of arguments). @andycochran ?

andycochran:
I wonder if @voxpelli might have any thoughts on this issue, as the `pow()` function comes from his [Gist](https://gist.github.com/voxpelli/6304812).

natewiebe13:
The need for higher precision is most likely needed since the numbers we are dealing with are so small. When working with larger integers the values are probably accurate enough that it isn't noticeable. I would keep the default in case the `pow()` function is to be used later where such accuracy may not be needed.

Either way, since this is being used to test contrast for WCAG compliance, the values need to be correct.

andycochran:
Thanks for the feedback, @voxpelli. 

@natewiebe13 / @ncoden, currently the `pow()` function is only used for this color function. I suppose it could be useful in the future for other purposes. But I can't imagine any other use cases for which we'd need to handle decimal exponents as fractions. Can you think of any? And if so, would we need a level of accuracy of 12 decimal places? 

I'm still for changing the function's default. And only increasing that to 16 if it's accurate enough.

kball:
@natewiebe13 @andycochran is this good to go? Or are we going to change the default?

natewiebe13:
@kball, @andycochran: I've updated the default to be 16 which is where it seemed to be accurate enough as this seemed to be a good compromise.

kball:
Sweet, thanks @natewiebe13

Issue #9543, opened 2016-12-21
"tr { page-break-inside: avoid }" causing Printing problems in Firefox


simshaun:
**How can we reproduce this bug?**

1. Create a multiple-row table with one row containing multiple pages of text.
2. Open Print Preview in Firefox

**What did you expect to happen?**

The table should be printed cleanly, with no extraneous page breaks.

**What happened instead?**

The row containing pages worth of text is split to a new page.

# Test case:

http://codepen.io/simshaun/pen/rWRvaX

Change View to *Debug Mode* and open Print Preview.

This is what I see in Firefox:

![image](https://cloud.githubusercontent.com/assets/379519/21377105/13339c82-c709-11e6-8ead-71f8a0962137.png)

# Potential fix:

In [_print.scss](https://github.com/zurb/foundation-sites/blob/develop/scss/typography/_print.scss) take the tr selector out of these lines (or set it to page-break-inside: auto):

    tr,
    img { page-break-inside: avoid; }

http://codepen.io/simshaun/pen/XNGqXr

With `tr` set back to `page-break-inside: auto`, this is what I see in Firefox:

![image](https://cloud.githubusercontent.com/assets/379519/21377153/aff6e740-c709-11e6-9f59-47938b3a14a0.png)

In Chrome:

![image](https://cloud.githubusercontent.com/assets/379519/21377202/67875764-c70a-11e6-9980-7b3029a2b07b.png)

In IE11:

![image](https://cloud.githubusercontent.com/assets/379519/21377194/45714c52-c70a-11e6-8145-13afb02c4ca3.png)

In Edge:

![image](https://cloud.githubusercontent.com/assets/379519/21377219/95e73912-c70a-11e6-974e-e514cad7032a.png)

ncoden:
Hi @simshaun,

I understand why you expect your table to be split between pages. But when there is a lot of small rows, I personally expect my table to break. See: https://codepen.io/ncoden/pen/zobjWq.

On firefox, with `page-break-inside: avoid;`, the `Message` row is moved to the second page on the first table. But with `auto`, a small row is split where is would be better to not to.

So maybe we could provide a modifier (like `.print-table-break`) to allow the table to break on print when wanted. I think this behaviour is expected in very few cases. Tables are most often used with small rows.

simshaun:
I see what you're saying. 

[...]

PR #9544, opened 2016-12-21
Helper class for page breaks inside elements

simshaun:
See #9543

### DIFF
--- a/scss/typography/_print.scss
+++ b/scss/typography/_print.scss
@@ -77,5 +77,10 @@ $print-hrefs: true !default;
     // Avoid page breaks after a heading
     h2,
     h3 { page-break-after: avoid; }
+
+    // Helper to re-allow page breaks in the middle of certain elements (e.g. pre, blockquote, tr)
+    .print-break-inside {
+      page-break-inside: auto;
+    }
   }
 }


scss/typography/_print.scss
```diff
@@ -77,5 +77,10 @@ $print-hrefs: true !default;
     // Avoid page breaks after a heading
     h2,
     h3 { page-break-after: avoid; }
+
+    // Helper to re-allow page breaks in the middle of certain elements (e.g. pre, blockquote, tr)
+    .print-break-inside {
```
ncoden: `.print-break-inside` vs `.print-page-break` (I prefer this) vs `.print-page-break-inside` ?

rafibomb: Yeah - i vote `.print-page-break`

simshaun: `.print-page-break` makes me think I can use that on an element anywhere I want a page break in print (e.g. in-between two <p>s), but that's not what this feature does.

`.print-break-inside` on a containing element makes more sense to me, semantically.

ncoden (approved):

ncoden:
I understand why you opened this PR for `6.4` (this is a new feature), but be advised it probably won't be released before May.

simshaun:
That's fine. I'll add it to our styles & make a note to remove when 6.4 is released.  Thank you.

PR #9546, opened 2016-12-21
Don't listen to touchend events on dropdown element leafs.

### LINKED ISSUE #9484: Dropdown menu not working on touch devices
A few people seem to have raised this issue on the Forum but no feedback from anyone from Zurb. 

Here is a codepen to illustrate the problem (on 6.2.4):
http://codepen.io/Byzantium/pen/oYMbmx

On pc/mac the submenu items are clickable and work as you would expect (except giving a 404 within codepen)

On an ipad, if you
1. Tap one of the parent menu headings the submenu opens
2. Tap one of the submenu items the submenu closes with no other action

I have only tested on an ipad as that's all I have, but others have used other touch devices and had the same issue.

I'm still learning foundation, so if the problem is with my code then my apologies. I can't see any reason why it should not work, but it does not.

sinspiral:
Don't listen to ``touchend`` events on dropdown element leafs. This seems to be missing and was causing issues on touch devices.

Steps to reproduce using Foundation Sites 6.3.0:
 * Use the example dropdown menu code from http://foundation.zurb.com/sites/docs/dropdown-menu.html
 * Add a real link to on submenu Item 1A (i.e. http://foundation.zurb.com)
 * On iPad (or any touch device) click the link for Item 1A

Result:
On the iPad the site doesn't navigate to the link added, while on desktop it does.

Expect:
Both touch and desktop to navigate to the link added.

fix #9484

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -108,7 +108,7 @@ class DropdownMenu {
         var $elem = $(this),
             hasSub = $elem.hasClass(parClass);
         if(!hasSub){
-          _this._hide($elem);
+          _this._hide();
         }
       });
     }
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -104,7 +104,7 @@ class DropdownMenu {
 
     // Handle Leaf element Clicks
     if(_this.options.closeOnClickInside){
-      this.$menuItems.on('click.zf.dropdownmenu touchend.zf.dropdownmenu', function(e) {
+      this.$menuItems.on('click.zf.dropdownmenu', function(e) {
         var $elem = $(this),
             hasSub = $elem.hasClass(parClass);
         if(!hasSub){
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -108,7 +108,7 @@ class DropdownMenu {
         var $elem = $(this),
             hasSub = $elem.hasClass(parClass);
         if(!hasSub){
-          _this._hide();
+          _this._hide($elem);
         }
       });
     }


sinspiral:
Sorry for closing/reopening the PR. Seems that Travis CI is failing due connectivity issues :( On my end all tests are passing.

kball:
This change doesn't really make sense to me - if I'm understanding correctly, `_hide` called without any parameters hides the entire dropdown, while if you pass in an $elem it expects that $elem to contain a submenu, and it will close that submenu specifically... here we're in a case explicitly without a submenu, so I'm not clear why we want to pass in $elem.

Can you explain a little more why this addresses #9484?

sinspiral:
I followed similar approach as on the ``mouseenter.zf.dropdownmenu`` and ``mouseleave.zf.dropdownmenu`` where they are sending also the same $(this) to the hide method.

Reading your comment, you're probably right that maybe that's not correct and something like ``parentsUntil`` should be used instead.

The reason of adding it was that during debugging it looked like the touchend event behaves differently from the click, and the link navigation refuses to happen when the ``this.$element`` is hidden.

I'll refactor it to use a ``parentsUntil`` to see if it makes a difference and debug it further to find out the exact reason why it happens.

sinspiral:
After some debugging, the delay on touch devices between the ``click`` (which at least on iOS is who internally does the link navigation) and ``touchend`` events was preventing the link navigation.

It seems that browser security policies don't allow link navigation on hidden elements, so the ``touchend`` event is hiding the menu, before the internal click for the navigation is happening.

I opted for removing the ``touchend`` and let the ``click`` to handle the hiding. This way, even though there will be a 300ms delay for hiding the menu in case there is no link, any link works as expected.

sinspiral:
@kball By the way, you were right my first commit didn't make sense at all! It was actually working cause no hiding on inside clicks were happening.

kball:
Cool, will test it out, thanks @sinspiral

kball:
Awesome, works, the fix makes sense.  Merging.

Issue #9537, opened 2016-12-20
CDN Links not populated


gitwebdevseo:
This page in the documentation:
https://foundation.zurb.com/sites/docs/installation.html
refers to CDN links at jsDelivr, but the links referenced for the 6.3 code are not actually there.

ncoden:
https://github.com/zurb/foundation-sites/pull/9538

gitwebdevseo:
This is only a very minor thing, but the page
https://foundation.zurb.com/sites/docs/installation.html
under the heading 'CDN Links' states "The folks at jsDelivr host the compressed Foundation CSS and JavaScript for us."
Since the links have been changed to point to Cloudflare you should probably give them the credit now.

ncoden:
You are right. I'll update this.

PR #9549, opened 2016-12-22
Update CDN name in installation docs

ncoden:
See: https://github.com/zurb/foundation-sites/issues/9537


rafibomb:
Looks like jsdeliver didn't update to 6.3 yet - but cdnjs is great too!

PR #9552, opened 2016-12-22
Fix minor Typo "depreciated" to "deprecated"

michiruckstuhl:
Change from „depreciated“ (meaning the monetary decrease of value over
time) to „deprecated“.


rafibomb:
Thanks!

PR #9561, opened 2016-12-23
Modularizes `get-color` function

rediris:
* allows using the `get-color` function with any color map.

Examples:
* `get-color(primary)`
* `get-color(mauve, $my-color-palette)`

### DIFF
--- a/scss/util/_color.scss
+++ b/scss/util/_color.scss
@@ -94,15 +94,15 @@
 
 /// Get color from foundation-palette
 ///
-/// @param {key} color key from foundation-palette
+/// @param {key} color key from any color map
 ///
-/// @returns {Color} color from foundation-palette
-@function get-color($key) {
-  @if map-has-key($foundation-palette, $key) {
-    @return map-get($foundation-palette, $key);
+/// @returns {Color} color from any color map (foundation-palette is the default)
+@function get-color($key, $palette: $foundation-palette) {
+  @if map-has-key($palette, $key) {
+    @return map-get($palette, $key);
   }
   @else {
-    @error 'given $key is not available in $foundation-palette';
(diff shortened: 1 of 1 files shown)


ncoden:
This PR pushes color management in the same direction as https://github.com/zurb/foundation-sites/pull/9459 (cc @andycochran). Maybe the new `get-color` function behavior could be used inside Foundation.

ncoden (approved):

DaSchTour:
`get-color(mauve, $my-color-palette)` is equal to `map-get($my-color-palette, mauve)` which is not any shorter and could be used to get any value from any map. Not sure if there is any benefit from this, I don't se the use case.

ncoden:
`map-get` does not throw errors and return `null` if the index does not exists.
Also, I think we could _merge_ the given map of colors with the default one instead of replacing it.

kball:
I tend to agree with @DaSchTour on this one... what makes this color specific? Is this just a `safe-map-get`?

I too would like @andycochran's opinion as he's been thinking a lot about palettes and how we manage colors

ncoden:
> map-get does not throw errors and return null if the index does not exists.

So yes, it is a `map-safe-get`. Until we add more features like:

> Also, I think we could merge the given map of colors with the default one instead of replacing it.

So more something like: `map-safe-get(map-merge($foundation-palette, $palette), $key)`

andycochran:
I agree with @DaSchTour.

kball:
Closing based on @andycochran and @DaSchTour's feedback. I am also in agreement; I don't see the value add over map-get in this form. If we want to implement `map-safe-get`as a generic and use that instead of `get-color`, I'd be open to reopening that conversation

ncoden:
@kball See https://github.com/zurb/foundation-sites/pull/9644 😄

Issue #8554, opened 2016-04-08
Invalid US-ASCII character "\xE2"


Luckyfella73:
Hi,

after using zurb foundation 5 for about a year I'm trying to implement version 6.2.0 now. When saving my sass files I get the following error-message (SublimeText 3 console):
`
Error: Invalid US-ASCII character "\xE2"
        on line 15 of /.../vendor/bower/foundation-sites/scss/typography/_base.scss
        from line 12 of /.../vendor/bower/foundation-sites/scss/typography/_typography.scss
        from line 16 of /.../vendor/bower/foundation-sites/scss/foundation.scss
`

I searched the net for about 6 hours now but I can't find a way to solve the problem. Some suggest to add this line on top of config.rb:
`
Encoding.default_external = "utf-8"
`
But I don't use compass, so I don't have that file on my mac..

An other solution would be to put `@charset "UTF-8";` on top of each file but of course I don't want to edit the foundation files managed by bower.

I thought maybe my ruby installation is configured not to utf-8 encoding but running `ruby -e 'p Encoding.default_external'` I get:

`#<Encoding:UTF-8>`

I know that the problem is that the at least some of the foundation files have comment lines using "this longer dash" or Hyphen? I don't know the right expression for english is not my native language. But I guess you know what I mean - it's just a little bit wider that a "minus (-)" ;)

According to your issue guidelines:
>>**How can we reproduce this bug?**
hard to say because it might be a configuration issues though I have no idea what to do about that and as said I didn't have the same problem using foundation 5.

>> **What did you expect to happen?**
sass compiling the scss files like it should ..

>> **What happened instead?**
ok doesn't make sense to continue with that for this issue is not related to a specific css case/ foundation classes.

For any ideas what I could do about that I would be very thankfull!

Best regards

colin-marshall:
Have you tried putting `@charset 'utf-8';` at the top of the main SCSS file you import everything with? There's no messing with the Foundation files that way. That is how it's done in the Foundation templates:
https://github.com/zurb/foundation-zurb-template/blob/master/src/assets/scss/app.scss#L1

Luckyfella73:
@colin-marshall Yes I have that line
`@charset "utf-8";`
on top of every of my .scss files -  having double quotes like the sass docs say. Changed to single quotes to test but that doesn't help.
Thanks for your input though!

colin-marshall:
@Luckyfella73 what are you using to compile the SCSS?

[...]

PR #9570, opened 2016-12-26
Fix invalid characters in _accordion.scss and typography/_base.scss f…

zy4:
…rom issue #8554.

This bug affected compass on systems with no or wrong locale set.

### DIFF
--- a/scss/components/_accordion.scss
+++ b/scss/components/_accordion.scss
@@ -105,7 +105,7 @@ $accordion-content-padding: 1rem !default;
     }
 
     .is-active > &::before {
-      content: '-';
+      content: '\2013';
     }
   }
 }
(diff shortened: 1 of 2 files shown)


ncoden:
Can't it be fixed with `@charset "UTF-8";` at the top of each file ?

zy4:
Why would you add that on top of each file when you can clean two of them from hidden characters? :)

ncoden:
It is not hidden characters, but the minus sign showed when an accordion item is open (see http://foundation.zurb.com/sites/docs/accordion.html), different from a hyphen.

Hyphen: `-`
Long hyphen: `–` (used as minus sign)

zy4:
You are right for the accordion (the other occurrence did not affect any element). I just changed it to an ISO coded char, using `&ndash;`. Hope this will fix it.

ncoden (approved):

PR #9573, opened 2016-12-29
Fix wrong path which causes errors when settings get imported in foun…

zy4:
... especially when one comments in line 20 of foundation.scss

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -41,7 +41,7 @@
 //  36. Tooltip
 //  37. Top Bar
 
-@import 'util/util';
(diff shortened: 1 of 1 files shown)


ncoden:
`settings.scss` should not be imported in Foundation from its `scss/settings/` path, but copied to your project and imported in your own main sass file before Foundation.

zy4:
Okay. Would be nice then to remove lines 19-20 in foundation.scss.

ncoden:
`foundation.scss` is an internal file and should not be modified. Lines 19-20 are useful for when this file is copied in your own project to manually import ZF partials.

PR #9574, opened 2016-12-29
tabs markdown fix and clarify browser history updates

ahebrank:
Docs were missing a backtick, and I updated to clarify behavior since https://github.com/zurb/foundation-sites/pull/9407

### DIFF
--- a/docs/pages/tabs.md
+++ b/docs/pages/tabs.md
@@ -146,14 +146,13 @@ Add the attribute `data-active-collapse="true"` to a tabstrip to collapse active
 
 ### Browser history
 
-When the `data-deep-link="true"` attribute is set, the current state of the tabset is recorded by adding a hash with the tab panel ID to the browser URL when a tab opens. By default, tabs *update and replace* the browser history (meaning that the browser does not navigate). Modify this behavior by using attribute `data-update-history="true"` to *append* to the browser history. In the latter case the browser back button will track each click that opens a tab panel.
+When the `data-deep-link` option is set to `true`, the current state of the tabset is recorded by adding a hash with the tab panel ID to the browser URL when a tab opens. By default, tabs *replace* the browser history (using `history.replaceState()`). Modify this behavior by using attribute `data-update-history="true"` to *append* to the browser history (using `history.pushState()`). In the latter case the browser back button will track each click that opens a tab panel.
 
 By using deep linking (see below), the open state of a page's tabset may be shared by copy-pasting the browser URL.
 
 ### Deep linking
 
 Add the attribute `data-deep-link="true"` to a tabstrip to:
-
 - modify the browser history when a tab is clicked
 - allow users to open a particular tab at page load with a hash-appended URL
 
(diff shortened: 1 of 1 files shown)


docs/pages/tabs.md
```diff
@@ -146,13 +146,16 @@ Add the attribute `data-active-collapse="true"` to a tabstrip to collapse active
 
 ### Browser history
 
-The current state of the tabset is now recorded by adding a hash with the tab panel ID to the browser URL when a tab is opened. By default, tabs replace the browser history. Modify this behavior by using attribute `data-update-history="true"` to *append* to the browser history. In the latter case the browser back button will track each click that opens a tab panel.
+When the `data-deep-link="true"` attribute is set, the current state of the tabset is recorded by adding a hash with the tab panel ID to the browser URL when a tab opens. By default, tabs *update and replace* the browser history (meaning that the browser does not navigate). Modify this behavior by using attribute `data-update-history="true"` to *append* to the browser history. In the latter case the browser back button will track each click that opens a tab panel.
```
ncoden: I personally prefer: "When the `data-deep-link` option is `true`...". But the current formulation is used away and we should rather be consistent.

Also, I don't think "meaning that the browser does not navigate" is not clear enough. We could understand the URL does not change. I think "the url change but the the browser does not navigate" would be more clear.

What do you think ?

docs/pages/tabs.md
```diff
@@ -146,13 +146,16 @@ Add the attribute `data-active-collapse="true"` to a tabstrip to collapse active
...
 
 By using deep linking (see below), the open state of a page's tabset may be shared by copy-pasting the browser URL.
 
 ### Deep linking
 
-Add the attribute `data-deep-link="true" to a tabstrip to allow anchoring to and opening a tab panel at page load.
+Add the attribute `data-deep-link="true"` to a tabstrip to:
+
```
ncoden: Why an empty line ?

docs/pages/tabs.md
ahebrank: I guess "When the `data-deep-link` attribute is set to `true`" is most correct.

I'm not sure I really understand the technical distinction between 'history.updateState()' and 'history.replaceState()' either, which is what I was trying to talk around. It's probably best just to get rid of the parenthetical or link to the Mozilla documentation.

docs/pages/tabs.md
ahebrank: Habit--didn't think it did anything in markdown when between text and a list.

docs/pages/tabs.md
ncoden: You can set options in Js, without attributes: `Foundation.tabs.defaults.updateHistory`.

There is no `history.updateState()` but `history.pushState()`. And about the difference:
```
page1 --> page2 -(pushState)-> page3 -(go back)-> page2
page1 --> page2 -(replaceState)-> page3 -(go back)-> page1
```

docs/pages/tabs.md
ncoden: True.

ahebrank:
Sorry, I was conflating the js command with the option key.  These changes should clarify the actual function of the `update-history` option, and I think the UX difference was already described.

ncoden (approved):

ncoden:
@ahebrank Thank you.

ncoden:
^^ okay. Thank @rafibomb

PR #9580, opened 2016-12-31
Fix #9579 - ignore disabled input in Abide

### LINKED ISSUE #9579: [Abide] Disabled fields not being ignored in validation

1. Use the example abide form code from the documentation
2. Add a disabled attribute to the first required field
3. Submit the form for checking

I expected Abode to ignore the disabled field as it does in  in Foundation 5
**What happened instead?**
The disabled field is highlighted with an error
**Test case:**
http://codepen.io/colinmackinlay/pen/PWYZGw

colinmackinlay:
@ncoden To resolve #9579 so that disabled inputs are not validated

### DIFF
--- a/docs/pages/abide.md
+++ b/docs/pages/abide.md
@@ -28,24 +28,6 @@ These input types create a text field: `text`, `date`, `datetime`, `datetime-loc
       <p class="help-text" id="exampleHelpText">Here's how you use this input field!</p>
     </div>
     <div class="small-12 columns">
-      <label>Nothing Required!
-        <input type="text" placeholder="Use me, or don't" aria-describedby="exampleHelpTex" data-abide-ignore>
-      </label>
-      <p class="help-text" id="exampleHelpTex">This input is ignored by Abide using `data-abide-ignore`</p>
-    </div>
-    <div class="small-12 columns">
-      <label>Disabled!
-        <input type="text" placeholder="You can't use me!" aria-describedby="exampleHelpTex" disabled>
-      </label>
-      <p class="help-text" id="exampleHelpTex">This input is ignored by Abide using `disabled`</p>
-    </div>
-    <div class="small-12 columns">
-      <label>Hidden!
-        <input type="hidden" placeholder="You can't see me!" aria-describedby="exampleHelpTex" >
-      </label>
-      <p class="help-text" id="exampleHelpTex">This input is ignored by Abide using `type="hidden"`</p>
-    </div>
-    <div class="small-12 columns">
       <label>Password Required
         <input type="password" id="password" placeholder="yeti4preZ" aria-describedby="exampleHelpText" required >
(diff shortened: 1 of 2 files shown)


ncoden (approved):

ncoden:
Next time, please use a better name for commit, like `Fix #9579 - ignore disabled input in Abide`

colinmackinlay:
Sorry about that - realised after I'd moved longer description to the body.

Tried to update docs but can't see how to edit "Javascript Reference" which seems the most appropriate place. The abide.md file doesn't include this or the SASS reference. Presumably this is generated elsewhere?

ncoden:
"Javascript Reference" is generated from the comments above functions and options, like [here](https://github.com/ncoden/foundation-sites/blob/a87eb5dc02f714f80bd3648151d2c53185827548/js/foundation.abide.js#L84). But I think we should also add a section to the docs (like `Ignored inputs` with an example for `[data-abide-ignore]`, `[type="hidden"]` and [disabled]`).

docs/pages/abide.md
```diff
@@ -34,6 +34,18 @@ These input types create a text field: `text`, `date`, `datetime`, `datetime-loc
       <p class="help-text" id="exampleHelpTex">This input is ignored by Abide using `data-abide-ignore`</p>
     </div>
     <div class="small-12 columns">
+      <label>Disabled!
+        <input type="text" placeholder="You can't use me!" aria-describedby="exampleHelpTex" disabled>
```
ncoden: `placeholder="Disabled input"`

docs/pages/abide.md
```diff
@@ -34,6 +34,18 @@ These input types create a text field: `text`, `date`, `datetime`, `datetime-loc
...
+      </label>
+      <p class="help-text" id="exampleHelpTex">This input is ignored by Abide using `disabled`</p>
+    </div>
+    <div class="small-12 columns">
+      <label>Hidden!
+        <input type="hidden" placeholder="You can't see me!" aria-describedby="exampleHelpTex" >
```
ncoden: placeholder="Hidden input"

colinmackinlay:
validateInput() comment updated and I've added two further examples below the data-abide-ignore one.
Perhaps the [type="hidden"] one could be clearer but it's hard to see how!

docs/pages/abide.md
ncoden: Also, don't you think there is already too much inputs in this example, and we should rather move ignored input to their own section ?

colinmackinlay:
Section before Initial State seems to just hang without explanation. It seems to be an example of how to show error state initially both in the label and with callout. Is that right @gakimball ?

ncoden (approved):

kball:
Merged!  Thanks @colinmackinlay

PR #9595, opened 2017-01-04
Update _responsive-embed.scss

mheward:
The var in the mixin currently doesn't exist and as such including it without a set $ratio var results in build failure.

### DIFF
--- a/scss/components/_responsive-embed.scss
+++ b/scss/components/_responsive-embed.scss
@@ -60,7 +60,7 @@ $responsive-embed-ratios: (
   @include foundation-responsive-embed;
 }
 
-@mixin flex-video($ratio: $responsive-embed-ratio) {
+@mixin flex-video($ratio: default) {
   @warn 'This mixin is being replaced by responsive-embed(). flex-video() will be removed in Foundation 6.4.';
   @include responsive-embed;
 }
--- a/scss/components/_responsive-embed.scss
+++ b/scss/components/_responsive-embed.scss
@@ -62,5 +62,5 @@ $responsive-embed-ratios: (
 
 @mixin flex-video($ratio: $responsive-embed-ratio) {
   @warn 'This mixin is being replaced by responsive-embed(). flex-video() will be removed in Foundation 6.4.';
-  @include responsive-embed;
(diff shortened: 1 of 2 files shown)


ncoden (commented):
The problem here is we kept the support for a function but not for the default parameters it requires.

The PR was still breaking if:
- `$responsive-embed-ratio` is used somewhere (does not exist anymore).
- `@mixin flex-video` is included without parameters (so use `$responsive-embed-ratio`, this PR fix this case).
- `@mixin flex-video` is called with a parameter (parameter is ignored).

I would tend to say "full support or no support". So I think we should also:
- put `$responsive-embed-ratio` back
- pass the `$ratio` parameter to `@mixin responsive-embed` in `@mixin flex-video`.

AdamBarnwell:
@ncoden Good point and agreed.

andycochran:
Ack! Sorry I missed this when I created Responsive Embed. @ncoden, I think your solution works. Let's just make sure to include a comment that the `$responsive-embed-ratio` var will be removed in v6.4, along with the other deprecated functions.

kball:
@mheward it looks like there were some questions/changes requested, can you take a look so we can get this merged?  Thanks

ncoden:
Can you:
> - pass the `$ratio` parameter to `@mixin responsive-embed` in `@mixin flex-video`.

kball:
hey @andycochran @mheward @ncoden what is the status of this? The current PR doesn't appear to update anything... we're trying to wrap up 6.3.1 by next week, and would like to get this in.

mheward:
I've updated the mixin to use the $ratio parameter as requested

IamManchanda:
> we're trying to wrap up 6.3.1 by next week

Sweet :+1:

kball:
OK the one additional thing we need to do on this is make sure the default variable definition is in `_responsive-embed.scss` as _settings.scss is autogenerated upon build (I know it's confusing - it should be in dist, we'll probably move it there for 6.4)

mheward:
@kball I've moved the var as requested.

kball:
Thanks @mheward! Merged!

Issue #9603, opened 2017-01-05
Interchange using global resize handler, not resize trigger utility


kball:
Noticed while working on something else, want to capture it.  See here:

https://github.com/zurb/foundation-sites/blob/develop/js/foundation.interchange.js#L49

It also appears to never clean up that resize handler.  Double broken.  This should be fixed for 6.3.1

Issue #9438, opened 2016-12-02
Port Foundation's javascript modules to use ES2016 dependency management


kball:
We want to bring Foundation's JavaScript dependency management up to 2016.

Some challenges will include:

* Some of the utilities are not well encapsulated
* We still need to generate code that will work for folks without a build system, just inline including the various components.

However, I think the benefits will far outweigh the costs.  Among other items, this should fix https://github.com/zurb/foundation-sites/issues/7386 and make it way easier to integrate Foundation components into other systems.  It will also lay the groundwork for easier refactoring of Foundation down the road (jQuery removal, I'm looking at you!)

kball:
Not sure what the best way to track progress on this... does it make sense to break it down by component?

DaSchTour:
Maybe switch to TypeScript? Than create a bundle with systemJS? Most of the code is already in ES2015 so the switch should be easy and I could improve the code a lot and help while removing jQuery. With TypeScript working with HTMLElements and all the subclasses is a lot easier.
This may also be a great opportunity to improve integration of foundation with JS/SPA frameworks.

Owlbertz:
@DaSchTour interesting suggesting, I'd really like to head the ZURBian opinion.

kball:
I'm a little hesitant to move towards Typescript given no one at ZURB is currently using it, however I don't think I know enough at this point to say yes or no. Would be interested in some other opinions... @zurb/yetinauts are any of you on TypeScript?

@DaSchTour can you elaborate a little more on both the step from ES2015 to TypeScript and some more detail on the benefits you see from TypeScript?

DaSchTour:
@kball TypeScript is simply just ES2015+ with addition of types. First to know is, that every valid JavaScript is also valid TypeScript, so at first there is not much new to learn. The big benefit is, that you get type information and validation before runtime.

So for example if you do something like this `document.getElementsByTagName("img");` you know that there is a NodeList of HTMLImageElement and what properties this elements have. If you select elements by id, they are simply HTMLElement and you can then can make a check width `instanceof HTMLImageElement` to check if it's the type of element you expected. You can do VanillaJS with pre-runtime verification, that the code will really work.

Another big benefit is, that TypeScript creates some kind of dependency tree inside the project. So can do solid refactoring. The TypeScript transpiler sees what dependencies are use, which are missing or what may not be used anymore.

[...]

PR #9604, opened 2017-01-05
[WIP] Isolate global references to lay way to module dependencies

kball:
This is WIP working on items 2 and 3 from phase 1 of https://github.com/zurb/foundation-sites/issues/9438#issuecomment-270447118

Pieces remaining for this chunk are: 

- [x] refactor timerAndImageLoader into separate files.
- [x] refactor GetYoDigits, rtl, and transitionend into a loadable utility module
- [x] Fix https://github.com/zurb/foundation-sites/issues/9603 and remove throttle dependency
- [x] Fix global Keyboard reference within Keyboard
- [ ] Move plugin references in responsive accordion tabs and responsive menu to initialization

To be handled separately:

Rework `Foundation.plugin`, `Foundation.registerPlugin`, and `Foundation.unregisterPlugin`.

At the end of this work, the Foundation namespace should only be referenced in Foundation core, within the faked out "import blocks" within the modules, and in the plugin/register/unregister pieces mentioned above.

Depends on: #9626


kball:
I'm going to defer the changing of the responsive plugin items until I change initialization, because they'll require a change in the initialization function signature.  I think we're going to be passing Foundation into the plugin so it can register & deregister, and that will give us what we need to detect these plugins as well.

Also, because this changes some filenames and will require changes in the templates, will need to close this and target it against 6.4 instead of patch release.

PR #9610, opened 2017-01-05
Specificity first in titles on docs pages

kball:
As @MattWilcox pointed out, for usability docs page titles should have the more specific part first, global/branding last.


ncoden:
👍

Issue #9593, opened 2017-01-04
Foundation Sites 6 Typescript Definition File


stcp:
Hi  all,

It's new on Foundation Sites and first time on github. I  have made some modifications on typescript definition file "foundation.d.ts" and not sure here is the right place to post it. Attached is the file and hope someone just may find it useful.

Regards,

Steven
[foundation-sites-6.d.zip](https://github.com/zurb/foundation-sites/files/684046/foundation-sites-6.d.zip)

ncoden:
Hi @stcp. Welcome on Github !

I would advise you to create a Pull Request for Foundation.

To help you, here some guides that show how to contribute to Github projects:
- https://akrabat.com/the-beginners-guide-to-contributing-to-a-github-project/ (a very complete guide)
- http://kbroman.org/github_tutorial/pages/fork.html (more concise)
- https://guides.github.com/activities/contributing-to-open-source/ (official guide)

After you made a pull request, Foundation maintainers will review it to ensure that everything is okay, then merge your changes.

Please feel free to ask if you need help or have any question !

Nicolas

IamManchanda:
Thanks buddy for these links

stcp:
Thanks Nicolas and IamManchanda, I will try later on :)

stcp:
Thanks Nicolas and IamManchanda, I will try later on :)

IamManchanda:
Heartly welcome bro ,,,,, Foundation Community and yetinauts are always there for help.
But would urge you to delete one comment of the two as these repitions shouldn't be there

DaSchTour:
I've taken a short look on the definitions and I cant't really determine what was changed, because everything is changed so I can't see any improvement. Maybe you describe what your improvement is about.

stcp:
Hi DaSchTour,

I have added some Foundation global settings, for example:

Foundation.Abide.defaults.patterns,
Foundation.Abide.defaults.validators

The original definition file wont allow me to to that. May be there is a simpler way to do it using the original but due to my limited knowledge that is the best I can do.


Regards,

Steven

stcp:
I am writing some angular2 directives for Foundation Sites 6 just for learning both, after polish and cleanup the codes, I may submit  to the github....

DaSchTour:
Nice to hear. Would love to see the work on angular2. In the meantime I added Abide defaults to typings ;)
Maybe you could download the updated typings from my work an verify that it works as you expected.

IamManchanda:
[...]

PR #9616, opened 2017-01-08
Update foundation.d.ts

stcp:
#9593 Updated some options.
Added ResponsiveMenu and ResponsiveToggle types.
Added defaults to all plugins.

### DIFF
--- a/dist/js/foundation.d.ts
+++ b/dist/js/foundation.d.ts
@@ -12,6 +12,7 @@ declare module FoundationSites {
     requiredChecked(element: JQuery): boolean;
     findFormError(element: JQuery): JQuery;
     findLabel(element: JQuery): boolean;
+    findRadioLabels(element: JQuery): any;
     addErrorClasses(element: JQuery): void;
     removeRadioErrorClasses(groupName: string): void;
     removeErrorClasses(element: JQuery): void;
@@ -24,16 +25,8 @@ declare module FoundationSites {
     destroy(): void;
   }
 
-  interface AbideDefaults {
-    validateOn: string;
-    labelErrorClass: string;
-    inputErrorClass: string;
-    formErrorSelector: string;
-    formErrorClass: string;
-    liveValidate: boolean;
-    validateOnBlur: boolean;
-    patterns: IAbidePatterns;
-    validators: any; // TODO, maybe there is a better solution to describe how this object may look like
+  interface AbideDefaults extends IAbideOptions{
+   
   }
 
   interface IAbidePatterns {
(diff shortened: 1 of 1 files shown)


IamManchanda:
i hate this travis ci .... i must say ..... network connectivity issues always

dist/js/foundation.d.ts
```diff
@@ -62,7 +55,8 @@ declare module FoundationSites {
     formErrorSelector?: string;
     formErrorClass?: string;
     liveValidate?: boolean;
-    validators?: any;
+    validators?: {[key:string]:Function};
+    patterns?:{[key:string]:RegExp};
```
DaSchTour: I'm not sure if this will provide any already present pattern.

dist/js/foundation.d.ts
```diff
@@ -73,6 +67,10 @@ declare module FoundationSites {
     destroy(): void;
   }
 
+  interface AccordionDefaults extends IAccordionOptions{
```
DaSchTour: Empty interface? So why not reference IAccordionOptions directly. From my look on the JS I'm also not sure if Defaults and options are the same. From what I see they are not.

dist/js/foundation.d.ts
```diff
@@ -88,6 +86,10 @@ declare module FoundationSites {
     destroy(): void;
   }
 
+  interface AccordionMenuDefaults extends IAccordionMenuOptions{
```
DaSchTour: Empty interfaces?

dist/js/foundation.d.ts
```diff
@@ -333,21 +438,28 @@ declare module FoundationSites {
...
+
+  }
+
   interface ITooltipOptions {
+    disableForTouch?: boolean;
     hoverDelay?: number;
     fadeInDuration?: number;
     fadeOutDuration?: number;
-    disableHover?: boolean;
+    disableHover?: number;
```
DaSchTour: Documentation says this is a boolean. http://foundation.zurb.com/sites/docs/tooltip.html

dist/js/foundation.d.ts
```diff
@@ -24,16 +25,8 @@ declare module FoundationSites {
     destroy(): void;
   }
 
-  interface AbideDefaults {
```
DaSchTour: Why this is removed? This are the default properties.

DaSchTour (changes requested):
Why are there so many empty interfaces?

kball:
@DaSchTour @stcp what is the status of this PR?

DaSchTour:
@kball I think we shouldn't change anything here. We should do the work to shift to TypeScript and get full typings from there.

kball:
@DaSchTour Given we'll likely do at least one more release (6.4) prior to any sort of TypeScript shift, are our current typings good enough? We're likely looking at July for 6.5

DaSchTour:
@kball I work with them regularly and they serve well. I'm note sure about the changes made here and the consequences they may have and the author didn't answer any of my questions :/

kball:
Seems reasonable. Closing in lieu of any additional responses indicating why we'd need these.

PR #9620, opened 2017-01-09
Sass fix for util import in settings

digitalcrafted:
Hi! Great work!

Noticed in 6.3.0 that the file path reference for importing `util/util` in `foundation-sites/scss/settings/_settings.scss` on ine 44  

> @import 'util/util';

led to the erorr while compiling the Sass:

>  Error: File to import not found or unreadable: util/util.

Updated line 44 to

> @import '../util/util';

Thanks

### DIFF
--- a/docs/layout/default.html
+++ b/docs/layout/default.html
@@ -7,7 +7,7 @@
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <meta name="description" content="{{description}}">
   <link rel="icon" href="assets/img/icons/foundation-favicon.ico" type="image/x-icon">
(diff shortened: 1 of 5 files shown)


ncoden:
Duplicate of https://github.com/zurb/foundation-sites/pull/9573

> `settings.scss` should not be imported in Foundation from its scss/settings/ path, but copied to your project and imported in your own main sass file before Foundation. I agree this is unclear.

Foundation can works without settings file, because of the default settings are defined inside each component. This file is only a way to quickly set your own configuration. The way this file should be generated is currently in discussion.

PR #9625, opened 2017-01-09
added cthulu

ziad-saab:


ncoden:
Why ?

ncoden:
@kball Everyone needs cthulhu in its hearth.

PR #9632, opened 2017-01-12
Feature Add attribute for auto focus within off canvas

gorriecoe:
Checks for the attribute [data-autofocus] within the toggled off canvas
and focuses on the first occurrence if available.  Otherwise reverts to
standard behaviour.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -196,7 +196,12 @@ class OffCanvas {
 
     if (this.options.autoFocus === true) {
       this.$element.one(Foundation.transitionend(this.$element), function() {
-        _this.$element.find('a, button').eq(0).focus();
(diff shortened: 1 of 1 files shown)


kball:
Seems reasonable to me - @brettsmason what do you think?

kball:
ping @brettsmason

brettsmason:
@kball sorry missed this. This sounds like a really good idea. Had a quick look at the code and it looks fine to me. Maybe you could just double check from a JS point of view (not my strong point 😄 ) and merge if you are happy.

gorriecoe:
Thanks team,  how do I go about get this documented?

IamManchanda:
@gorriecoe => **/docs/pages/offcanvas.md**

Just In case you confused about how to run docs, 
see => https://github.com/zurb/foundation-sites#documentation

Issue #9587, opened 2017-01-02
JS: Smooth Scroll: Only for those links which require smooth scroll ( data attribute )


IamManchanda:
**Feature Request - Smooth Scroll** 
Something like this => http://mdbootstrap.com/javascript/smooth-scroll/
Specially handy for landing pages and Call to action anchor tags 

Codepen => http://codepen.io/IamManchanda/pen/VPZMba

Source: CSS Tricks

**Special Note**: Have proposed similar feature in bootstrap too so sorry for any duplication n all. I Just love and use both frameworks in my projects, so a bit neutral and proposing to both :)

ncoden:
I think this is a useful feature. We could add it in Foundation like this:
```html
<a href="#" data-smooth-scroll">Link</a>
```

However, I don't know if we should also provide an option to enable it by default for all, even without `data-smooth-scroll`. Could it create bugs ?

IamManchanda:
@ncoden I am using this in my company and with a class or id selector just like my example, it dont create bugs, never had any issue but i dont have idea of "data-smooth-scroll" if it create any bugs

ncoden:
@IamManchanda Using `data` attribute to apply JS behavior is proper, because of `.classes` is used for CSS styles and `#IDs` must be unique.

I wonder if there would be bugs with an option to apply smooth scoll by default for all links, because we could apply a smooth scoll where it is not expected.

IamManchanda:
No No ..... with a data-smooth-scroll only would be better .... not to every link 

@ncoden yes For navbars we can do 

Codepen => http://codepen.io/IamManchanda/pen/JEjOmN

```
<ul class="menu" id="smooth-scroll-menu">
  <li><a href="#">One</a></li>
  <li><a href="#">Two</a></li>
  <li><a href="#">Three</a></li>
  <li><a href="#">Four</a></li>
</ul>
```

```
$('#smooth-scroll-menu >li > a[href*="#"]:not([href="#"])').click(function() {
  if (location.pathname.replace(/^\//, '') == this.pathname.replace(/^\//, '') && location.hostname == this.hostname) {
    var target = $(this.hash);
    target = target.length ? target : $('[name=' + this.hash.slice(1) + ']');
    if (target.length) {
      $('html, body').animate({
        scrollTop: target.offset().top
      }, 1000);
      return false;
    }
  }
});
```

ncoden:
[...]

PR #9636, opened 2017-01-12
New SmoothScroll module based on Magellan scrollToLoc code. 

denisahac:
Implementation for #9587 

To enable smooth scroll on the internal links just add the **data-smooth-scroll** attribute to the parent container, sample code below, or via javascript initialization.

```html
<ul data-smooth-scroll data-threshold="0" data-animation-easing="swing" >
    <li><a href="#section1">Section 1</a></li>
</ul>
```

The **Magellan** module now depends on the **SmoothScroll** module for the scrolling animation functionality.

### DIFF
--- a/docs/pages/smooth-scroll.md
+++ b/docs/pages/smooth-scroll.md
@@ -1,0 +1,40 @@
+---
+title: Smooth Scroll
+description: Allows internal links smooth scrolling.
+js: js/foundation.smoothScroll.js
+tags:
+  - navigation
+---
+
+<ul class="menu vertical" data-smooth-scroll>
+  <li><a href="#setup">Setup</a></li>
+  <li><a href="#javascript-reference">Javascript Reference</a></li>
+</ul>
+
+<br>
+
+## Setup
+
+To enable SmoothScroll on internal links, just add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
+
+```html
+<ul class="horizontal menu" data-smooth-scroll>
+  <li><a href="#first">First Arrival</a></li>
+  <li><a href="#second">Second Arrival</a></li>
+  <li><a href="#third">Third Arrival</a></li>
+</ul>
+<div class="sections">
+  <section id="first">First Section</section>
+  <section id="second">Second Section</section>
+  <section id="third">Third Section</section>
+</div>
+```
+
+You can also setup SmoothScroll directly via indiviual link.
+
+```html
+<a href="#exclusive" data-smooth-scroll>Exclusive Section</a>
+<section id="exclusive">The Exclusive Section</section>
+```
+
+---
--- a/docs/pages/smooth-scroll.md
+++ b/docs/pages/smooth-scroll.md
@@ -6,9 +6,16 @@ tags:
   - navigation
 ---
 
-## Usage
+<ul class="menu vertical" data-smooth-scroll>
+  <li><a href="#setup">Setup</a></li>
+  <li><a href="#javascript-reference">Javascript Reference</a></li>
+</ul>
 
-To enable smooth scrolling on internal links add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
+<br>
+
+## Setup
+
+To enable SmoothScroll on internal links, just add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
 
 ```html
 <ul class="horizontal menu" data-smooth-scroll>
@@ -23,7 +30,7 @@ To enable smooth scrolling on internal links add the attribute `data-smooth-scro
 </div>
 ```
 
-You can also apply `data-smooth-scroll` attribute directly to the link.
+You can also setup SmoothScroll directly via indiviual link.
 
 ```html
 <a href="#exclusive" data-smooth-scroll>Exclusive Section</a>
--- a/docs/pages/smooth-scroll.md
+++ b/docs/pages/smooth-scroll.md
@@ -6,16 +6,9 @@ tags:
   - navigation
 ---
 
-<ul class="menu vertical" data-smooth-scroll>
-  <li><a href="#setup">Setup</a></li>
-  <li><a href="#javascript-reference">Javascript Reference</a></li>
-</ul>
+## Usage
 
-<br>
-
-## Setup
-
-To enable SmoothScroll on internal links, just add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
+To enable smooth scrolling on internal links add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
 
 ```html
 <ul class="horizontal menu" data-smooth-scroll>
@@ -30,7 +23,7 @@ To enable SmoothScroll on internal links, just add the attribute `data-smooth-sc
 </div>
 ```
 
-You can also setup SmoothScroll directly via indiviual link.
+You can also apply `data-smooth-scroll` attribute directly to the link.
 
 ```html
 <a href="#exclusive" data-smooth-scroll>Exclusive Section</a>
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -113,7 +113,14 @@ class Magellan {
     this._inTransition = true;
     var _this = this;
 
-    Foundation.SmoothScroll.scrollToLoc(loc, this.options, function() {
+    var options = {
+      animationEasing: this.options.animationEasing,
+      animationDuration: this.options.animationDuration,
+      threshold: this.options.threshold,
+      offset: this.options.offset
+    };
+
+    Foundation.SmoothScroll.scrollToLoc(loc, options, function() {
       _this._inTransition = false; 
       _this._updateActive();
     })
@@ -145,7 +152,7 @@ class Magellan {
       var isDown = this.scrollPos < winPos,
           _this = this,
           curVisible = this.points.filter(function(p, i){
-            return isDown ? p - _this.options.barOffset <= winPos : p - _this.options.barOffset - _this.options.threshold <= winPos;
+            return isDown ? p - _this.options.offset <= winPos : p - _this.options.offset - _this.options.threshold <= winPos;
           });
       curIdx = curVisible.length ? curVisible.length - 1 : 0;
     }
@@ -238,7 +245,7 @@ Magellan.defaults = {
    * @type {number}
    * @default 0
    */
-  barOffset: 0
+  offset: 0
 }
 
 // Window exports
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -5,6 +5,7 @@
 /**
  * Magellan module.
  * @module foundation.magellan
+ * @requires foundation.smoothScroll
  */
 
 class Magellan {
@@ -109,18 +110,13 @@ class Magellan {
    * @function
    */
   scrollToLoc(loc) {
-    // Do nothing if target does not exist to prevent errors
-    if (!$(loc).length) {return false;}
     this._inTransition = true;
-    var _this = this,
-        scrollPos = Math.round($(loc).offset().top - this.options.threshold / 2 - this.options.barOffset);
+    var _this = this;
 
-    $('html, body').stop(true).animate(
-      { scrollTop: scrollPos },
-      this.options.animationDuration,
-      this.options.animationEasing,
-      function() {_this._inTransition = false; _this._updateActive()}
-    );
+    Foundation.SmoothScroll.scrollToLoc(loc, this.options, function() {
+      _this._inTransition = false; 
+      _this._updateActive();
+    })
   }
 
   /**
--- a/js/foundation.smoothScroll.js
+++ b/js/foundation.smoothScroll.js
@@ -1,0 +1,123 @@
+'use strict';
+
+!function($) {
+
+/**
+ * SmoothScroll module.
+ * @module foundation.smooth-scroll
+ */
+class SmoothScroll {
+    constructor(element, options) {
+        this.$element = element;
+        this.options = $.extend({}, SmoothScroll.defaults, this.$element.data(), options);
+
+        this._init();
+
+        Foundation.registerPlugin(this, 'SmoothScroll');
+    }
+
+    /**
+     * Initialize the SmoothScroll plugin 
+     */
+    _init() {
+        var id = this.$element[0].id || Foundation.GetYoDigits(6, 'smooth-scroll');
+        var _this = this;
+        this.$element.attr({
+            'id': id
+        });
+
+        this._events();
+    }
+
+    _events() {
+        var _this = this;
+
+        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', function(e) {
+            e.preventDefault();
+            var arrival   = this.getAttribute('href');
+            _this.scrollToLoc(arrival);
+        });
+    }
+
+    /**
+     * Function to scroll to a given location on the page.
+     * @param {String} loc - a properly formatted jQuery id selector. Example: '#foo'
+     * @function
+     */
+    scrollToLoc(loc) {
+        this._inTransition = true;
+        var _this = this;
+        this.constructor.scrollToLoc(loc, this.options, function() {
+            _this._inTransition = false;
+        });
+    }
+
+    /**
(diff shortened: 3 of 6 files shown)


IamManchanda:
Cheers @denisahac

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,123 @@
...
+   * @option
+   * @type {number}
+   * @default 0
+   */
+  barOffset: 0
+}
+
+// Window exports
+Foundation.plugin(SmoothScroll, 'SmoothScroll');
+}(jQuery);
```
IamManchanda: **End of line** please !!

IamManchanda (commented):
One small review from my side. Please see to it :smile:

js/foundation.smoothScroll.js
ncoden: 😄

IamManchanda: I am smelling this :wink:

ncoden:
Thank you @denisahac for this PR. Sorry, I planned to work on it before but did not have the time to. However, please warn when you are working on an issue.

IamManchanda:
> However, please warn when you are working on an issue

hahahaaaaa i think have heared this before :wink:

denisahac:
Should've notified you guys! @ncoden will nudge you guys next time 😄

IamManchanda:
@denisahac end of line please ?

denisahac:
@IamManchanda I do not think the **end of line** is required, when you check the rest of the javascript codebase you'll see nothing at the end of the file. 

You should only put a comment on your code to explain what you're trying to achieve, or for documentation purposes.

js/foundation.smoothScroll.js
denisahac: @IamManchanda I do not think the **end of line** is required, when you check the rest of the javascript codebase you'll see nothing at the end of the file. 

You should only put a comment on your code to explain what you're trying to achieve, or for documentation purposes.

IamManchanda: I am just talking about adding a line break or a new blank line after line no 123...

Most (if not all) the foundation codebase has the end of line  

Yup even I don't know its benefits.... But some Coding geek has said me to do this .... So passing/forwarding that knowledge :p

kball:
This looks great - given this is a new plugin, can you do a pass on a quick docs page for this as well? E.g. `docs/page/smooth-scroll.md`

denisahac:
Got you @kball. I will do the documentation page for this new javascript module 👍

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    _events() {
+        var _this = this;
+
+        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', function(e) {
+            e.preventDefault();
+            var arrival   = this.getAttribute('href');
+            _this.scrollToLoc(arrival);
+        });
```
ncoden: It does not target a simple link, like:
```html
<a href="#target" data-smooth-scroll>Link</a>
```
We can move the code above to a public or private `onLinkClick` function, and then call:
```js
this.$element.on('click.zf.smoothScroll', _this.onLinkClick); // or onLinkClick
this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', _this.onLinkClick); // or onLinkClick
```
... with a manual check of if `[href]` is set and begin with `#`.

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,123 @@
...
+   * @default 50
+   */
+  threshold: 50,
+  /**
+   * Number of pixels to offset the scroll of the page on item click if using a sticky nav bar.
+   * @option
+   * @type {number}
+   * @default 0
+   */
+  barOffset: 0
```
ncoden: I think we can call it `offset`. An offset is not useful only to handle a sticky nav bar, but also for a better visual, to have the anchor not glued to the top of the window.

js/foundation.smoothScroll.js
ncoden: Yes, this is an old convention to end all the files with an EOL character. It allows to put the cursor after the last line and write on it, and historically to merge files without a bug which could from the concatenation of the first and last lines. It is less true with Js now.

js/foundation.smoothScroll.js
denisahac: Renaming this to `offset` would also require renaming the `barOffset` default variable name of [Magellan](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.magellan.js#L245)

The reason is, the static `scrollToLoc` method of **SmoothScroll** uses the `options.barOffset` to determine the offset.

Based on your argument, renaming the `barOffset` default to `offset` for both Magellan and SmoothScroll seems reasonable.

Let me know what do you think?

js/foundation.magellan.js
```diff
@@ -109,18 +110,13 @@ class Magellan {
...
-
-    $('html, body').stop(true).animate(
-      { scrollTop: scrollPos },
-      this.options.animationDuration,
-      this.options.animationEasing,
-      function() {_this._inTransition = false; _this._updateActive()}
-    );
+    var _this = this;
+
+    Foundation.SmoothScroll.scrollToLoc(loc, this.options, function() {
```
ncoden: Based on you comment above:

> Renaming this to offset would also require renaming the barOffset default variable name of Magellan

I do not think it is a good idea to pass an array of options from components to components:
- All Magelan options are not SmoothSettings options.
- Some options could expect the same value but under a different name.
- Some options could have the same name but expect different values.
- It is generally helpful to see which options are passed when calling a component.

And the main reason behind all the previous ones: Options redirection is an exception, not the norm, and should be explicit to ensure it is correctly maintained.

js/foundation.smoothScroll.js
ncoden: I agree. We should rename the option for both Magellan and SmoothScrolling.

js/foundation.smoothScroll.js
denisahac: This is the new implementation I'm working based on your feedback.

```javascript
// click handler function.
var handleClickFn = function(e) {
   // exit function if the event source isn't coming from an anchor with href attribute starts with '#'
   if(!$(this).is('a[href^="#"]'))  {
      return false;
   }
            
   var arrival = this.getAttribute('href');
            
   _this._inTransition = true;
   SmoothScroll.scrollToLoc(arrival, 
      _this.options.animationEasing,
      _this.options.animationDuration,
       _this.options.threshold,
       _this.options.offset,
       function() {
          _this._inTransition = false;
       });
            
   e.preventDefault();
};

this.$element.on('click.zf.smoothScroll', handleClickFn)
this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', handleClickFn);
```

js/foundation.magellan.js
denisahac: And for this case, the new implementation for `scrollToLoc` static method now accepts **6** parameters.

```javascript
static scrollToLoc(loc, animationEasing, animationDuration, threshold, offset, callback) {
...
}
```

js/foundation.smoothScroll.js
ncoden: Why not `onLinkClick` ?

js/foundation.magellan.js
ncoden: Or you can use an object of options before/after `callback` and merge it with `SmoothScroll.defaults`.

denisahac: I do not think it is a good idea of merging an object option to the SmoothScroll.defaults within this **static** method. The method doesn't have the reference to `this` instance. The reason I decided of making this method a **static** is **Magellan** module can just use it directly without creating an instance.

The merging of defaults and explicitly defined options are happened within Object instantiation. In this case in the constructor method.

ncoden: I know, this is why I am talking about `SmoothScroll.defaults` and not `this.options`. `SmoothScroll.defaults` are the default options of the Class, not of the instantiated object.

js/foundation.smoothScroll.js
ncoden: Actually, this is a handler, not a binder, so : why not `handleLinkClick` ? 😄

IamManchanda:
Awesome buddy @denisahac

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,136 @@
...
+
+    /**
+     * Function to scroll to a given location on the page.
+     * @param {String} loc - A properly formatted jQuery id selector. Example: '#foo'
+     * @param {Object} options - The options to use.
+     * @param {Function} callback - The callback function.
+     * @static
+     * @function
+     */
+    static scrollToLoc(loc, options = SmoothScroll.defaults, callback) {
```
ncoden:
```js
static scrollToLoc(loc, options, callback) {
    options = $.extend({}, SmoothScroll.defaults, options);
    ...
```

So you can do:
```js
Foundation.SmoothScroll.scrollToLoc('#myTarget', {}, function() {
   ...
});
// or
Foundation.SmoothScroll.scrollToLoc('#myTarget');
```

docs/pages/smooth-scroll.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+You can also setup SmoothScroll directly via indiviual link.
+
+```html
+<a href="#exclusive" data-smooth-scroll>Exclusive Section</a>
+<section id="exclusive">The Exclusive Section</section>
+```
+
+---
```
ncoden: Please add links to this page in:
- `docs/pages/index.md`
- `docs/partials/component-list.md`

Maybe in the "Navigation" section, I am not sure. @kball ?

docs/pages/smooth-scroll.md
```diff
@@ -0,0 +1,40 @@
...
+
+<ul class="menu vertical" data-smooth-scroll>
+  <li><a href="#setup">Setup</a></li>
+  <li><a href="#javascript-reference">Javascript Reference</a></li>
+</ul>
+
+<br>
+
+## Setup
```
ncoden: I think `## Usage` would be more appropriate.

docs/pages/smooth-scroll.md
```diff
@@ -0,0 +1,40 @@
...
+  <section id="third">Third Section</section>
+</div>
+```
+
+You can also setup SmoothScroll directly via indiviual link.
+
+```html
+<a href="#exclusive" data-smooth-scroll>Exclusive Section</a>
+<section id="exclusive">The Exclusive Section</section>
+```
```
ncoden: This example is simpler and should go first:
- Show how to enable SmoothScroll (on a simple link).
- Then, show we can enable SmoothScroll on several links at once though their parent.

docs/pages/smooth-scroll.md
```diff
@@ -0,0 +1,40 @@
...
+<ul class="menu vertical" data-smooth-scroll>
+  <li><a href="#setup">Setup</a></li>
+  <li><a href="#javascript-reference">Javascript Reference</a></li>
+</ul>
+
+<br>
+
+## Setup
+
+To enable SmoothScroll on internal links, just add the attribute `data-smooth-scroll` to the parent container like our [Menu](menu.html). Each section needs a unique ID
```
ncoden: Please use a more neutral formulation. No `just`.

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,136 @@
...
+
+        this.$element.on('click.zf.smoothScroll', handleLinkClick)
+        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', handleLinkClick);
+    }
+
+    /**
+     * Function to scroll to a given location on the page.
+     * @param {String} loc - A properly formatted jQuery id selector. Example: '#foo'
+     * @param {Object} options - The options to use.
+     * @param {Function} callback - The callback function.
```
ncoden: ` ... - Callback called after the scroll animation finished.`

js/foundation.smoothScroll.js
```diff
@@ -0,0 +1,136 @@
...
+        };
+
+        this.$element.on('click.zf.smoothScroll', handleLinkClick)
+        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', handleLinkClick);
+    }
+
+    /**
+     * Function to scroll to a given location on the page.
+     * @param {String} loc - A properly formatted jQuery id selector. Example: '#foo'
+     * @param {Object} options - The options to use.
```
ncoden: `... - Object of options, overriding SmoothScroll's defaults options.`

docs/pages/smooth-scroll.md
kball: Hmm... either 'Navigation' or 'Plugins' (which is kind of a catch all for utilities)

ncoden: Docs need a reorganization, isn't it ?

docs/pages/smooth-scroll.md
denisahac: Single link are rarely used in a project for internal navigation. The reason you create internal links is to provide a direct access to the relevant section of the page. 

The most used case scenario should go first, in this case the multiple links wrapped in a container. I don't agree that we organize the documentation based on simpleness. Provide the most relevant information directly to the user. Being able to target single link is an additional feature provided by **SmoothScroll** module.

docs/pages/smooth-scroll.md
denisahac: I think this one should go to "Plugins" by the name suggest, **SmoothScroll**.

Let me know what you guys think?

docs/pages/smooth-scroll.md
ncoden: I agree that for a complex component, the first example must be the most common use case, because we cannot expect the user to read the entire doc and fully understand the component to know how to use it.

But after the "common use case", all features of the components must be explained gradually, not from the "simpler" to the "more complex" feature, but from the fondamental features to the more complementary ones relying on them.

However, for a so simple component with only two example, I think we can expect the whole documentation to be read, and skip the "most common case" because it would be redundant. But you can disagree with that. In any case after that, we explain the core feature: not "targeting single link" but the principle itself of smooth scrolling. Then the features relying on it: applying this behavior to a whole group of links.

docs/pages/smooth-scroll.md
IamManchanda: This will and should go in plugins .... but we should link it up in navigation pages too!

ncoden: Ok for Plugins.

IamManchanda: Good boy :smile:

docs/pages/smooth-scroll.md
denisahac: I agree with you that we can expect the whole documentation to be read, as this is a minimalist module. However we should adhere to general concepts of writing documentation whether it's complex or simple one.

docs/pages/smooth-scroll.md
denisahac: Got you @ncoden

docs/pages/smooth-scroll.md
ncoden: I agree, and I described writing concept that I find not too bad, usable for both simple and complex modules.

IamManchanda:
@ncoden @denisahac ... looking at the doc changes ... i do think there should be some kind of example ( demo page) or equivalent page to reflect on the smooth scroll plugin .... because its currently its not showing the functionality

denisahac:
@IamManchanda though the documentation for **SmoothScroll** didn't explicitly provide a working functionality, however it's implicitly provided by the **Magellan** in the right sidebar navigation.

IamManchanda:
@denisahac 
But still bro .... you atleast need a example page or if that's not possible 
then a simple codepen like for eg. => http://codepen.io/IamManchanda/pen/VPZMba
or something like => https://blackrockdigital.github.io/startbootstrap-scrolling-nav/
will be great .... because this isn't showing its actual functionality!

kball:
What is the current status of this PR?

denisahac:
Waiting for merge.

kball:
@ncoden and @IamManchanda you reviewed this - do you think it's good to go at this point?

IamManchanda:
Yeah but I would like to have a PR ( master) with a codepen example linked in the docs ... Once this is released !

IamManchanda:
Because current example lacks its actual usage and fwiw as pointed earlier

kball:
Great, merging! Thanks for the work on this @denisahac @ncoden @IamManchanda

IamManchanda:
Big cheers @denisahac  !

Issue #9576, opened 2016-12-30
Branches, releases and supported versions


ncoden:
I have some remarks to make about the Foundation's release process and the way we use git branches. Poke @zurb/yetinauts.

### Scheduled releases

Currently, we we have scheduled releases, you can find them on the [roadmap](https://github.com/zurb/foundation-sites/wiki/Project-Roadmap). The next patch version (`v6.3.1`) is planned for mi-January, the next minor version (`v6.4`) for May, and the next major version (`v7`) will come in November, even if it is very difficult to estimate this.

I totally understand the benefits of planning the changes we will make and the new features we will add. It helps to make the participations more accessible, to be more synchronized between contributors, and maybe it helps the users of Foundation to be prepared for breaking changes. However, I think having this changes on a Calendar is not a good idea.
- There is a lot of new features and safe patches (sometimes very important fixes) that can't be released as soon as we have a stable version. Look at all the bugs fixes in the [list of PRs](https://github.com/zurb/foundation-sites/pulls) or that have [alreary been merged](https://github.com/zurb/foundation-sites/commits/develop). They will not be publicly available before mid-January, even if we are certain they are stable and safe.
- It sorts minor and patches releases. When the next scheduled release is a _minor_ version, all the bug fixes will not be released before we checked the minor (sometimes breaking) changes are safe, even if they are unrelated with the fixes. We experienced that recently with `v6.3.0`, where we released some fixes made more than 8 months ago.
- For the same reason, when the next scheduled release is a _patch_ version, we are pushed to include into it breaking fixes, and even sometimes breaking minor changes (a new feature or refactor required to fix a bug, or a simple change planned for the next release without checking of what type of release it would be). An example of this is: https://github.com/zurb/foundation-sites/issues/9507.
- If we discover a bug difficult to resolve just before a release (or in a Release Candidate), we are not pushed to resolve it in a proper manner. A version must be released, time passes, so we are more likely to keep a dirty work around. I think https://github.com/zurb/foundation-sites/pull/9469 is a very good example of this problem (I'm not proud of this PR at all).

For these reasons, I think we should move to a more fluid management of future fixes and features. Here some proprosals I have to do this, but I am open to any other proposition:
1. **Release patch and minor versions quickly**.
Within one to three week for patches and within one to three months for breaking minor changes and new features. Try to move out from the  "marketing release" approach for at least minor releases.
2. **Do not except some fixes or features to be introduced in the next release**.
[...]

PR #9642, opened 2017-01-13
[Abide] Add Form Error attribute

### LINKED ISSUE #9600: Inline labels with form validation.
Abide validation doesn't work with foundation inline labels.  Not sure if can add both, Ideally I would like to move the form-error span beneath the input group under the same label tag (currently abide only activates **siblings** that have class form-error).

```html
<label>
    Amount
    <div class="input-group">
        <span class="input-group-label">$</span>
        <input data-bind="value: total, attr: {required: validate}" class="input-group-field" type="number" min="0.01" step="0.01" />
        <span class="form-error">Amount is required.</span>
    </div>
</label>
```

Expected behavior is to keep same formatting of input-group when validation classes are added.

ncoden:
Closes #9600.

#### Changes:
- Add `[data-form-error-for=someId]` attribute to declare a relation between a form error and the related field.
- Add docs in Abide with explanation about the way Form Errors are found and an example for `[data-form-error-for]`

#### Other changes:
- Fix docs title in Abide (was `###` instead of `##`)

### DIFF
--- a/docs/pages/abide.md
+++ b/docs/pages/abide.md
@@ -8,7 +8,7 @@ tags:
   - validation
 ---
 
-### Abide Demo
+## Abide Demo
 
 These input types create a text field: `text`, `date`, `datetime`, `datetime-local`, `email`, `month`, `number`, `password`, `search`, `tel`, `time`, `url`, and `week`.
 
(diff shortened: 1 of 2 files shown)


kball:
Given this is a new feature, it's probably for 6.4. Will merge after I reorg the branches as discussed in https://github.com/zurb/foundation-sites/issues/9576

Issue #9596, opened 2017-01-04
System Font stacks instead of current one 


IamManchanda:
Read this article => https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/
Some more Info => https://css-tricks.com/snippets/css/system-font-stack/

Some already implemented examples 

1. Github's website
2. Medium's website
3. Bootstrap framework
4. Wordpress CMS  
and many more 

Code i would request to change 

`$body-font-family: 'Helvetica Neue', Helvetica, Roboto, Arial, sans-serif !default;`

New code 

`$body-font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif !default;`

If its get decided to not use linux specific ones and firefox specific system fonts then, 

`$body-font-family: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Helvetica Neue", Helvetica, Arial, sans-serif !default;`

ncoden:
First, I seen you opened a lot of issues these last days. Thank you a lot for your work, you help to make Foundation even better !

Concerning this issue, I agree we should at least make the font override for "Helvetica" optioned, and maybe disabled by default. But I don't know if updating `$body-font-family` would be a great idea. We are taking about low-level reset, not "styling option". In any case, we also have to disable the `font-family` Normalize set on `<html>`.

IamManchanda:
So what do you think will be a best option ?

ncoden:
I am not sure. I thought about something like:
```scss
$body-font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
$body-font-family-system: true;

...

$-zf-system-fonts: -apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans";

...

$font: $body-font-family;
@if $body-font-family-systems == true {
  $font: join($-zf-system-fonts, $font);
}
```

It allows to easily enable/disable the support for system font stacks, or to change the font family without disabled it (but this last case is not common). But maybe it is over-engineering.

IamManchanda:
Yup a bit of over engineering :smile_cat: but i liked it :+1:

kball:
Curious what @andycochran and @brettsmason think

[...]

PR #9644, opened 2017-01-13
New function: font-stack

brettsmason:
This PR adds some carefully selected font stacks to an internal map variable `$-zf-font-stack`.
A helper function has also been added `font-stack` to allow you to use one of these stacks. Simply pass in the stack key, eg `font-family: font-stack(system);` to use the system font stack.

This ensures a consistent way to use some quality font stacks, but also the ability to add your own map with stacks in too, simply pass it as the second parameter to the function.

This was the result of a discussion with @ncoden about how to sort out #9596.

### DIFF
--- a/scss/util/_typography.scss
+++ b/scss/util/_typography.scss
@@ -1,0 +1,34 @@
+// Foundation for Sites by ZURB
+// foundation.zurb.com
+// Licensed under MIT Open Source
+
+////
+/// @group functions
+////
+
+$-zf-font-stack: (
+  'georgia': (Georgia, "URW Bookman L", serif),
+  'helvetica': (Helvetica, Arial, "Nimbus Sans L", sans-serif),
+  'lucida-grande': ("Lucida Grande", "Lucida Sans Unicode", "Bitstream Vera Sans", sans-serif),
+  'monospace': ("Courier New", Courier, "Nimbus Sans L", monospace),
+  'system': (-apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif),
+  'verdana': (Verdana, Geneva, "DejaVu Sans", sans-serif),
+);
+
+/// Retrieve a font stack string from a map.
+/// @param {String} $stack - The key of the font stack from the map.
+/// @param {Map} $map - A map of font stacks to retrieve a string from.
+/// @return {List} A font stack list.
+@function font-stack($stack, $map: $-zf-font-stack) {
+  @if (type-of($map) == 'map') {
+    @if (map-has-key($map, $stack)) {
+      @return map-get($map, $stack);
+    }
+    @else {
+      @error 'Font stack $key is not available in $map';
+    }
+  }
+  @else {
+    @error '$map is not a map';
+  }
+}
--- a/scss/util/_typography.scss
+++ b/scss/util/_typography.scss
@@ -25,10 +25,10 @@ $-zf-font-stack: (
       @return map-get($map, $stack);
     }
     @else {
-      @error 'Font stack $key is not available in $map';
+      @error 'Font stack `#{$stack}` is not available in `#{$map}`';
     }
   }
   @else {
-    @error '$map is not a map';
+    @error '`#{$map}` is not a valid map';
   }
 }
--- a/scss/util/_typography.scss
+++ b/scss/util/_typography.scss
@@ -15,17 +15,19 @@ $-zf-font-stack: (
   'verdana': (Verdana, Geneva, "DejaVu Sans", sans-serif),
 );
 
-/// Retrieve a font stack string from a map.
-/// @param {String} $stack - The key of the font stack from the map.
-/// @param {Map} $map - A map of font stacks to retrieve a string from.
-/// @return {List} A font stack list.
-@function font-stack($stack, $map: $-zf-font-stack) {
+/// Return a font stack list from a map. Equivalent to `map-safe-get($name, $-zf-font-stack)`.
+///
+/// @param {String} $name - Name of the font stack.
+/// @param {Map} $map [$-zf-font-stack] - Map of font stacks to retrieve a list from.
+///
+/// @returns {List} Found font stack.
+@function font-stack($name, $map: $-zf-font-stack) {
   @if (type-of($map) == 'map') {
-    @if (map-has-key($map, $stack)) {
-      @return map-get($map, $stack);
+    @if (map-has-key($map, $name)) {
+      @return map-get($map, $name);
     }
     @else {
-      @error 'Font stack `#{$stack}` is not available in `#{$map}`';
+      @error 'Font stack `#{$name}` is not available in `#{$map}`';
     }
   }
   @else {
--- a/scss/util/_typography.scss
+++ b/scss/util/_typography.scss
@@ -17,20 +17,10 @@ $-zf-font-stack: (
 
 /// Return a font stack list from a map. Equivalent to `map-safe-get($name, $-zf-font-stack)`.
 ///
-/// @param {String} $name - Name of the font stack.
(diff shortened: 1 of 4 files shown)


scss/util/_typography.scss
```diff
@@ -0,0 +1,34 @@
...
+    @if (map-has-key($map, $stack)) {
+      @return map-get($map, $stack);
+    }
+    @else {
+      @error 'Font stack `#{$stack}` is not available in `#{$map}`';
+    }
+  }
+  @else {
+    @error '`#{$map}` is not a valid map';
+  }
```
ncoden: Could you move all this map logic to `@function map-safe-get` ?

scss/util/_typography.scss
```diff
@@ -0,0 +1,34 @@
...
+  'lucida-grande': ("Lucida Grande", "Lucida Sans Unicode", "Bitstream Vera Sans", sans-serif),
+  'monospace': ("Courier New", Courier, "Nimbus Sans L", monospace),
+  'system': (-apple-system, system-ui, BlinkMacSystemFont, "Segoe UI", "Roboto", "Oxygen", "Ubuntu", "Cantarell", "Fira Sans", "Helvetica Neue", Helvetica, Arial, sans-serif),
+  'verdana': (Verdana, Geneva, "DejaVu Sans", sans-serif),
+);
+
+/// Retrieve a font stack string from a map.
+/// @param {String} $stack - The key of the font stack from the map.
+/// @param {Map} $map - A map of font stacks to retrieve a string from.
+/// @return {List} A font stack list.
```
ncoden: I think...
```scss
/// Return a font stack list from a map. Equivalent to `map-safe-get($name, $-zf-font-stack)`.
/// (...) $name - Name of the font stack.
/// (...) $map [$-zf-font-stack] - Map of font stacks to retrieve a list from.
///
/// @returns {List} Found font stack.
```
... would be more clear.

IamManchanda:
https://css-tricks.com/snippets/css/better-helvetica/

kball:
@brettsmason @ncoden Is this ready to merge? We can potentially continue improving the stacks and/or refactoring (e.g. if we want to create `map-safe-get` and utilize it in various locations) in separate PRs

brettsmason:
From my point of view I think it's OK to go. I expect the stacks will need to be tweaked and agree that a bit of a refactoring will be a good idea down the road.

ncoden:
@brettsmason Please reply to my review.

brettsmason:
@ncoden Sorry got tied up with other things. I have changed the doc block part you commented on.
Could you give me an idea of what you want the `map-safe-get` you mentioned to be like? Then I'll get that added.

ncoden:
`@function map-safe-get` do the same thing as `@function font-stack`, but in generic (no default map).

brettsmason:
Hopefully that's now OK 😄

ncoden:
😄  The purpose of a `map-safe-get` function is to avoid to duplicate logics. So you can replace all the content of `font-stack` by `map-safe-get(...);`

brettsmason:
So you mean something like this?

```
/// Return a font stack list from a map. Equivalent to `map-safe-get($name, $-zf-font-stack)`.
///
/// @param {String} $name - Name of the font stack.
/// @param {Map} $map [$-zf-font-stack] - Map of font stacks to retrieve a list from.
///
/// @returns {List} Found font stack.
@function font-stack($stack, $map: $-zf-font-stack) {
  @return map-safe-get($stack, $map);
}
```

ncoden:
No.
```scss
@function font-stack($stack, $map: $-zf-font-stack) {
  @return map-safe-get($map, $stack);
}
```

brettsmason:
The `map-safe-get` function is written like this though (currently):
`@function map-safe-get($name, $map)...`

ncoden:
It should not :)

brettsmason:
Hopefully that's now better.

scss/util/_value.scss
```diff
@@ -138,3 +138,23 @@
...
 
 }
+
+/// Safely return a value from a map.
+///
+/// @param {Map} $map - Map to retrieve a value from.
+/// @param {String} $name - Name of the map key.
+///
+/// @returns {List} Found value.
+@function map-safe-get($map, $name) {
```
ncoden: I think `$key` would be better than `$name`.

scss/util/_value.scss
```diff
@@ -138,3 +138,23 @@
...
+  @if (type-of($map) == 'map') {
```
ncoden: We should also add a case for `()`: is is an empty map, but
```
type-of( () ); // -> 'list'
```

Why not:
```scss
@if (type-of($map) == 'map' or (type-of($map) == 'list' and length($map) == 0)) {
```
?

brettsmason: Sorry not sure I understand.
You want to replace `@if (type-of($map) == 'map') {`
with `@if (type-of($map) == 'map' or (type-of($map) == 'list' and length($map) == 0)) {`
?

ncoden: Yes. So with `()`, we have a relevant error message:

```scss
$_: map-safe-get( (), 'myKey' ); // -> @error 'Key: `myKey` is not available in `()`';
```

ncoden:
Could you add unit tests ?

brettsmason:
No idea how so you would need to point me in the right direction.
Also running out of time to get the menu code sorted for Thursday, so if that can wait then 👍 😄

ncoden (approved):

brettsmason:
Think I figured it out, take a look and let me know if that is correct 😄

test/sass/_value.scss
```diff
@@ -114,5 +114,16 @@
       'Gets a value from a nested map');
   }
 
+  @include test('Map Safe Get [function]') {
+    $map: (
+      one: 'two',
+      ),
```
ncoden: You left a useless (and breaking) parenthesis.

test/sass/_value.scss
```diff
@@ -114,5 +114,16 @@
...
+    );
+    $expect: 'two';
+    
+    @include assert-equal(map-safe-get($map, one), $expect,
+      'Safely return a value from a map');
+  }
```
ncoden: I would like to also test error cases, but unfortunately we cannot handle `@error` in sass or with sass-true.

brettsmason: Updated the stray bracket. So nothing else left for me to change now?

ncoden (approved):

ncoden:
I think it can be merged now. But before, I would like the opinion of other @zurb/yetinauts member, because I know that a `map-safe-get` function with an alias toward was previously discussed before, and there was divergent opinions (#9561).

IamManchanda:
As my didn't got any answers to my comment's i am posting it again

> https://css-tricks.com/snippets/css/better-helvetica/

```scss
font-family: "HelveticaNeue-Light", "Helvetica Neue Light", "Helvetica Neue", Helvetica, Arial, sans-serif; 
```

@brettsmason and nicolas

kball:
`map-safe-get` seems like a reasonable utility to me; not in this PR but does this potentially replace all `map-get` invocations? Do we have any situations in which we don't want the error checking?

ncoden:
@kball Only when we expect `null` as value if the key does not exist in the map. So... almost never.

ncoden:
@IamManchanda 

https://codepen.io/ncoden/pen/jyVYOq

I agree this is nicer. I heard some negative opinions about it when used on black on white or on a too small text. @kball @brettsmason Do you think we could add a font stack, or is it -this time- not our role ?

kball:
I don't have a strong opinion here; it doesn't add to the weight of the css and has minimal impact on the complexity of the scss... the question is where to draw the line, and how we decide what is in and what is out.

Are there any consensus guidelines on "best font stacks" or similar out there that we can use to determine what we want to expose in core? (cc @jnemeth @tdhartwick)

IamManchanda:
What about the docs ? @ncoden @brettsmason

Issue #9121, opened 2016-08-18
Buttons with disabled class are clickable with 6.2 (not clickable with 6.1)


EspadaV8:
This is probably an issue for the release notes rather than a chance in the SCSS but 6.2 removed the `pointer-events` rule here https://github.com/zurb/foundation-sites/commit/d8810b5466a777997e2882a3f283e430711793eb#diff-cb853bd67d0fa46de7bbe58852679eecL142

Was this intended? The release notes for 6.2 don't mention this change and we've just found that we were relying on the 6.1 functionality when someone accidently deleted a record that should've had the button disabled :-/

markadrake:
The philosophical issue with **pointer-events** is that this CSS property actually affects behavior and functionality -vs- style and design.

There is also a UX argument. If `pointer-events:none;` is applied to your disabled elements then you lose the ability to provide meaningful feedback to the user. No interaction whatsoever will trigger from both a CSS or JS perspective. I think this is a major fallback.

Just some follow-up thoughts.

EspadaV8:
Yeah. This was more of an FYI in case anyone else has the issue. Not sure what the BC should be  between point releases either. We ended up adding a little bit of JS to disable them correctly.

andycochran:
From the docs: 
> The .disabled class is a purely visual style, and won't actually disable a control. For <button> elements, you can add the disabled attribute to disable it. If you want to disable a link, you should add the aria-disabled attribute to mark it as disabled for assistive technology.

I'm gonna close this, as it's not a bug. But if y'all have any thoughts on how to improve on the way that `disabled` works — especially from a UX/a11y perspective — please continue the discussion here.

EspadaV8:
I mentioned we had some JS that got this to work, thought I'd post it in case anyone else finds it useful

```javascript
(function ($) {
    $(function () {
        $('body').on(
          'click',
          'a[disabled],a.disabled,button[disabled],button.disabled',
          function (e) {
            e.preventDefault();
          }
        );
    });
})(jQuery);
```

It just checks for any `button` or `anchor` with either the `disabled` attribute or `disabled` class and prevents the default actions. It works well for our use case.

PR #9648, opened 2017-01-14
Bug Fix #9617 - Disables hover focus on button disabled

### LINKED ISSUE #9617: Disabled buttons are hoverable
Disabled buttons should not be hoverable right? In v6.3.0 it is, see this [example](https://jsfiddle.net/ym0m4c2u/2/)

IamManchanda:
This fixes - #9617

@ncoden for review

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -158,10 +158,18 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 }
 
 /// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
-@mixin button-disabled($color: $primary-color) {
+@mixin button-disabled(
+  $background: $primary-color,
+  $color: $button-color
+) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
-  background-color: $color;
+  background-color: $background;
+  
+  $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+
+  color: $color;
+
   &:hover, &:focus {
     box-shadow:none;
   }
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -158,22 +158,22 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 }
 
 /// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
+/// @param [Color] $background [$primary-color] - Background color of the disabled button. Set to `auto` to have the mixin automatically generate a hover color.
+/// @param [Color] $color [$button-color] - Text color of the disabled button. Set to `auto` to have the mixin automatically generate a hover color.
 @mixin button-disabled(
   $background: $primary-color,
   $color: $button-color
 ) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
-  background-color: $background;
-  
-  @if $color == auto {
-    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
-  }
 
-  color: $color;
-
-  &:hover, &:focus {
+  &, &:hover, &:focus {
     background-color: $background;
+  
+    @if $color == auto {
+      $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+    }
+
     color: $color;
   }
 }
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -169,7 +169,8 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 
   &, &:hover, &:focus {
     background-color: $background;
-  
+    
+    // If color is set to `auto`, the function `color-pick-contrast` automatically generates a color based on the background color 
     @if $color == auto {
       $color: color-pick-contrast($background, ($button-color, $button-color-alt));
     }
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -164,17 +164,15 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   $background: $primary-color,
   $color: $button-color
 ) {
+  @if $color == auto {
+    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+  }
+
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
 
   &, &:hover, &:focus {
     background-color: $background;
-    
-    // If color is set to `auto`, the function `color-pick-contrast` automatically generates a color based on the background color 
-    @if $color == auto {
-      $color: color-pick-contrast($background, ($button-color, $button-color-alt));
-    }
-
     color: $color;
   }
 }
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -169,6 +169,11 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   $color: color-pick-contrast($background, ($button-color, $button-color-alt));
 
   color: $color;
+
+  &:hover, &:focus {
+    background-color: $background;
+    color: $color;
+  }
 }
 
 /// Adds a dropdown arrow to a button.
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -161,10 +161,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 @mixin button-disabled($color: $primary-color) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
-
+  background-color: $color;
   &:hover, &:focus {
-    background-color: $color;
-    color: $button-color;
+    box-shadow:none;
   }
 }
 
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -158,8 +158,8 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 }
 
 /// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
-/// @param [Color] $background [$primary-color] - Background color of the disabled button. Set to `auto` to have the mixin automatically generate a hover color.
-/// @param [Color] $color [$button-color] - Text color of the disabled button. Set to `auto` to have the mixin automatically generate a hover color.
+/// @param [Color] $background [$primary-color] - Background color of the disabled button.
+/// @param [Color] $color [$button-color] - Text color of the disabled button. Set to `auto` to have the mixin automatically generate a color based on the background color.
 @mixin button-disabled(
   $background: $primary-color,
   $color: $button-color
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -169,10 +169,6 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   $color: color-pick-contrast($background, ($button-color, $button-color-alt));
 
   color: $color;
-
-  &:hover, &:focus {
-    box-shadow:none;
-  }
 }
 
 /// Adds a dropdown arrow to a button.
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -166,7 +166,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   cursor: not-allowed;
   background-color: $background;
   
-  $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+  @if $color == auto {
+    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+  }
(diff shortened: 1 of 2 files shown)


scss/components/_button.scss
```diff
@@ -161,10 +161,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 @mixin button-disabled($color: $primary-color) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
-
+  background-color: $color;
   &:hover, &:focus {
-    background-color: $color;
-    color: $button-color;
+    box-shadow:none;
```
ncoden: Why ? There is no `box-shadow` applied in the first place.

scss/components/_button.scss
```diff
@@ -161,10 +161,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 @mixin button-disabled($color: $primary-color) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
-
+  background-color: $color;
```
ncoden: This does not change anything. If fact, like explained at https://github.com/zurb/foundation-sites/issues/9617#issuecomment-271232428, it makes the background of `[disabled]` be overridden by the one of `:focus` and `:hover` if they were defined after.

But maybe it needs more discussions to know if it is the expected behavior. We could consider `[disabled]` as a "low-level modifier" (so should have a specificity of `0.0.1` and should be overridable), or as a "low-level state" (so should have a specificity of `0.0.2` and should not be overridable).

scss/components/_button.scss
```diff
@@ -161,10 +161,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
   &:hover, &:focus {
-    background-color: $color;
-    color: $button-color;
```
ncoden: The issue with #9617 is the text color is changed from the _a11y-compliant color_ to the default `$button-color` on mouse over, because `@mixin button-disabled` force the text color on `:hover` and `focus` with the wrong color.

Removing `color: $button-color;` fix this issue in the wrong way: it makes a disabled button not a11y-compliant at all. A better (but a bit repetitive) solution would be to use:
```sass
color-pick-contrast($background, ($button-color, $button-color-alt))
```
... like in `@mixin button-style`.

ncoden (changes requested):

scss/components/_button.scss
IamManchanda: https://css-tricks.com/forums/topic/disable-hover-effect/

ncoden: Which `box-shadow` do you want to remove ???

scss/components/_button.scss
IamManchanda: This one is done !

scss/components/_button.scss
IamManchanda: forget it me removing this .... yes its not doing anything .....

IamManchanda:
Nicolas i have done no 2 & 3 ..... and i am unsure about review 1 ( confused )

ncoden:
Consider the following situation:

```sass
.my-button {
  &:hover {
    background-color: red;
  }
}
```
```html
<button class="my-button button primary" disabled>Button</button>
```

On `:hover`, should the button appear as _blue_ (primary color) or _red_ ?

Poke @kball

IamManchanda:
```scss
pointer-events: none;
```
will do the trick in that case ..... ??

IamManchanda:
How about this, a bit repititive and also is using that dreaded `!important` also but it's resolving the issue 

```scss
@mixin button-disabled(
  $background: $primary-color,
  $color: $button-color
) {
  opacity: $button-opacity-disabled;
  cursor: not-allowed;
  background-color: $background;
  
  $color: color-pick-contrast($background, ($button-color, $button-color-alt));

  color: $color;
  &:hover, &:focus {
    background-color: $background!important;
    color: $color!important;
  }
}
```

ncoden:
Nothing. The pointer event is not disabled for now.

See: https://github.com/zurb/foundation-sites/issues/9121
This behavior was introduced at: https://github.com/zurb/foundation-sites/pull/8317

ncoden:
`!important`.

Never !

This is nothing more than a hack in specificity. It is saying "whatever the context, whatever what the user really want, this will works that way". Except in some very specific case, `!important` should never be used.

And you do not need important to solve this issue. It the code you provided, it works well without.

IamManchanda:
> This is nothing more than a hack in specificity. It is saying "whatever the context, whatever what the user really want, this will works that way". Except in some very specific case, !important should never be used.

I know it already but if you gonna say dont let the user modify something they want to .... what else would i say :stuck_out_tongue: 

> Nothing. The pointer event is not disabled for now.

Ohk then let pointer events the way to go then.

ncoden:
> I know it already but if you gonna say dont let the user modify something they want to .... what else would i say 😛

I am wondering if the `:hover` color modified by the user should also override the default `[disabled]:hover` color. This is a very specific case. I am not talking about "not letting the color be overridden whatever the situation".

> Ohk then let pointer events the way to go then.

I do not understand.

IamManchanda:
i am asking whether should i add .... pointer events or not ?

ncoden:
See: https://github.com/zurb/foundation-sites/issues/9121 and #8317

> The philosophical issue with pointer-events is that this CSS property actually affects behavior and functionality -vs- style and design.

> There is also a UX argument. If pointer-events:none; is applied to your disabled elements then you lose the ability to provide meaningful feedback to the user. No interaction whatsoever will trigger from both a CSS or JS perspective. I think this is a major fallback.

And

> If there is `pointer-events: none;` for buttons we can't use tooltips on buttons

IamManchanda:
First of all,

1. Pointer events isn't the solution 
2. Important isn't surely the solution

and yes you are right morever first question would be whether this below should be red or blue... 

```html
<button class="my-button button primary" disabled>Button</button>
```

But for me the code i gave is ohk enough and there is no need to let user dont modify if they want to!

kball:
@andycochran as the architect behind our recent palette changes, what think you?

scss/components/_button.scss
```diff
@@ -158,14 +158,17 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
-@mixin button-disabled($color: $primary-color) {
+@mixin button-disabled(
+  $background: $primary-color,
+  $color: $button-color
+) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
+  background-color: $background;
+  
+  $color: color-pick-contrast($background, ($button-color, $button-color-alt));
```
ncoden: Only for `$color == default`. See `@mixin button-style`.

IamManchanda: If i do like this .... it breaks ...... the success and warning text color changes to white ....

```scss
@if $color == default {
    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
  }
```

ncoden:
A though about it a while and now I think the color should not be overridden in the example I given. `[disabled]` is definitely about the whole state of the button, including `:hover` and `:focus`, these pseudo-classes should not modify it, even when defined after.

> Let them do whatever they want to do man!!

To be clear: yes, we should always let the user do whatever ze want, it is the idea behind the idea of "low specificity" I always promote. But in this case, wanting a `:hover` color is different from wanting an `[disabled]:hover` color, so redefining the first one should not modify the second.

IamManchanda:
so nicolas ...... how can we solve this in that case then ?

scss/components/_button.scss
ncoden: Because we do not define default semantical classes to use a11y-compliant colors. See [scss/components/_button.scss:271](https://github.com/ncoden/foundation-sites/blob/282963fef240c06f83ef093c2eb5e50bd716630d/scss/components/_button.scss#L271)

Should be:
```scss
@include button-disabled($color, auto);
```
... like at [scss/components/_button.scss:236](https://github.com/ncoden/foundation-sites/blob/282963fef240c06f83ef093c2eb5e50bd716630d/scss/components/_button.scss#L236).

ncoden:
By defining `background` and `color` for `:hover, :focus`. We ensure the background and color stay the same as for `[disabled]` when over/focus.

scss/components/_button.scss
IamManchanda: I did this 

```scss
@if $color == auto {
    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
  }
```

& this 

```scss
&.disabled,
    &[disabled] {
      @include button-disabled($color, auto);

      @each $name, $color in $button-palette {
        &.#{$name} {
          @include button-disabled($color, auto);
        }
      }
    }
```

but this still breaks it up

ncoden: hmm... It should not. But two things:
- No need to write here whole part of codes.
- This is your PR, I can help you but cannot fix everything 😄 . Use `@debug` to show the values of your variables and try to understand the changes of `$color`.

IamManchanda: done

IamManchanda:
Nicolas the requested changes are done..... thanks ..... harry

Vuurvlieg:
> We ensure the background and color stay the same as for [disabled] when over/focus.

That is exactly what I want, thanks!
Thanks all of you for this fix, really love it!

scss/components/_button.scss
```diff
@@ -158,13 +158,23 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
   cursor: not-allowed;
+  background-color: $background;
+  
+  @if $color == auto {
+    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
+  }
+
+  color: $color;
 
   &:hover, &:focus {
```
ncoden: You can use `&, &:hover, &:focus {` instead of redefining `color` and `background-color` above.

scss/components/_button.scss
```diff
@@ -158,13 +158,23 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
-@mixin button-disabled($color: $primary-color) {
+@mixin button-disabled(
+  $background: $primary-color,
+  $color: $button-color
+) {
   opacity: $button-opacity-disabled;
   cursor: not-allowed;
+  background-color: $background;
+  
+  @if $color == auto {
```
ncoden: Please put logic and variable definition in the top of the mixin, before declaring any property.

IamManchanda:
Requested Changes are done, dear nicolas

IamManchanda:
@kball Is this ready to merge now ?

ncoden:
Sorry, I forgot to reply.

> Please put logic and variable definition in the top of the mixin, before declaring any property.

IamManchanda:
are you talking about anything else because this is what i did in my last commit

```scss
/// Adds disabled styles to a button by fading the element, reseting the cursor, and disabling pointer events.
/// @param [Color] $background [$primary-color] - Background color of the disabled button.
/// @param [Color] $color [$button-color] - Text color of the disabled button. Set to `auto` to have the mixin automatically generate a color based on the background color.
@mixin button-disabled(
```

ncoden:
I am talking about the _logics_ for the functions. `@if`, `@else if`, variables, `@include`... So in our case: `@if $color == auto { ... }`.

IamManchanda:
Have done just that but i wonder that in other if statements within `_button.scss` there is no such comment ?

ncoden:
Ahah, we have definitely a problem in understanding each other:

 I mean : Moving `@if $color == auto { ... }.` to _the top of the mixin_.

`@if` & cie _is the logic_, and it is better to have _the logic first_, then _CSS properties_.

IamManchanda:
Are you wanting this ?

```scss
@mixin button-disabled(
  $background: $primary-color,
  $color: $button-color
) {
  @if $color == auto {
    $color: color-pick-contrast($background, ($button-color, $button-color-alt));
  }

  opacity: $button-opacity-disabled;
  cursor: not-allowed;

  &, &:hover, &:focus {
    background-color: $background;
    color: $color;
  }
}
```

ncoden:
YES !

ncoden (approved):

IamManchanda:
done !!! 

> Ahah, we have definitely a problem in understanding each other:

I dont think so! Just happens when we all are talking codes. :smiling_imp:

IamManchanda:
ready to merge ?

kball:
Looks good, nice work you two.

Do you think this is conceptually a "fix" that should be applied to the 6.3.x branch?

ncoden:
Yes, it is.

Vuurvlieg:
A huge thanks to @IamManchanda for the PR and to @ncoden for assistance and @kball for merging it so quickly!

IamManchanda:
Pleasure is our's buddy @Vuurvlieg

PR #9650, opened 2017-01-15
Make Optional extend .column with a alias ( default : .columns )

### LINKED ISSUE #9640: Optional extend .column .columns
Hello,

In file foundation-sites/scss/grid/_classes.scss they are : 
```sass
@if $column == 'column' {
    .columns {
      // sass-lint:disable-block placeholder-in-extend
      @extend .column;
    }
  }
```

I think it's sould be good if they are a option to extend or not. Advantage : the generate file size reduce if tit's not inculde and the user who use ".columns" can include if they want .

What do you think ?

IllusionPerdu:
Fix #9640.

Add parameter $column-alias when is not false extend the .column

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -15,7 +15,7 @@ Start by adding an element with a class of `.row`. This will create a horizontal
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
 <div class="primary callout">
-  <p>By defaut `.column` have an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$grid-column-alias` option</a>) &mdash;the only difference is grammar.</p>
+  <p>By default `.column` has an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$grid-column-alias` option</a>) &mdash;the only difference is grammar.</p>
   <p>Disabling the alias can reduce the Foundation CSS file size from 3 to 5%. It is recommended if the alias is not used.</p>
 </div>
 
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -10,9 +10,14 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or its <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
+
+<div class="warning callout">
+  <p>By defaut `.column` have an <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar.</p>
+  <p>Disabling the alias can reduce the Foundation CSS file size from 3 to 5%. It is recommended if the alias is not used.</p>
+</div>
 
 ```html
 <div class="row">
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -14,7 +14,7 @@ Start by adding an element with a class of `.row`. This will create a horizontal
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
-<div class="warning callout">
+<div class="primary callout">
   <p>By defaut `.column` have an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$grid-column-alias` option</a>) &mdash;the only difference is grammar.</p>
   <p>Disabling the alias can reduce the Foundation CSS file size from 3 to 5%. It is recommended if the alias is not used.</p>
 </div>
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -10,7 +10,7 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or its <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -15,7 +15,7 @@ Start by adding an element with a class of `.row`. This will create a horizontal
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
 <div class="warning callout">
-  <p>By defaut `.column` have an <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar.</p>
+  <p>By defaut `.column` have an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$grid-column-alias` option</a>) &mdash;the only difference is grammar.</p>
   <p>Disabling the alias can reduce the Foundation CSS file size from 3 to 5%. It is recommended if the alias is not used.</p>
 </div>
 
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -167,8 +167,8 @@
     @include grid-column-margin;
   }
 
-  @if $column == 'column' {
-    .columns {
+  @if $column == 'column' and $column-alias {
+    .#{$column-alias} {
       // sass-lint:disable-block placeholder-in-extend
       @extend .column;
     }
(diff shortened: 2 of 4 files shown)


scss/grid/_classes.scss
```diff
@@ -167,8 +167,8 @@
     @include grid-column-margin;
   }
 
-  @if $column == 'column' {
-    .columns {
+  @if $column == 'column' and $column-alias {
```
ncoden: You can use `has-value($column-alias)` here.

scss/grid/_grid.scss
```diff
@@ -33,6 +33,9 @@ $block-grid-max: 8 !default;
 // Internal value to store the end column float direction
 $-zf-end-float: if($grid-column-align-edge, $global-right, $global-left);
 
+// Extend column with this alias if no alias set to false
```
ncoden: Please add a sass-doc documentation (see the comments above, or [this reference](sassdoc.com/annotations/)).

Also, I think we should avoir relative descriptors like `this`, and use a descriptive formulation. i.e:
```
// Selector used for an alias of column (with @extend). If `false`, no alias is created.
```

IllusionPerdu:
I have change with the suggestion ;)
Actualy i have not many time but i see some code example use the alias, It would be necessary to put all the documentation up to date so that in the examples codes only the version with "colunm" is used.

(sinon @ncoden apparemment tu es est (aussi) français donc si nous pouvions mettre aussi les messages en français ça m'arrangerait grandement vu mon niveau en anglais :s )

scss/grid/_grid.scss
```diff
@@ -33,6 +33,9 @@ $block-grid-max: 8 !default;
 // Internal value to store the end column float direction
 $-zf-end-float: if($grid-column-align-edge, $global-right, $global-left);
 
+// Selector used for an alias of column (with @extend). If `false`, no alias is created.
```
ncoden: The sass-doc comment must be properly formatted. Please read: sassdoc.com/annotations/

```sass
/// (...)
/// @type String
```

ncoden (commented):
> sinon @ncoden apparemment tu es est (aussi) français

("Apparently you are french too...")

Yes, but we have to talk in english here, because this review is not between the two of us. Other people will see the commits, see this PR, and must be able to understand why we have made these changes. Most of them do not speak french.

So a bad english is better than no english at all. 😄

scss/grid/_grid.scss
```diff
@@ -30,6 +30,11 @@ $grid-column-align-edge: true !default;
 /// @type Number
 $block-grid-max: 8 !default;
 
+/// Selector used for an alias of column (with @extend). If `false`, no alias is created.
+/// @since 6.4.0
```
ncoden: Hmm... @kball Should we use `@since` ? What would be the benefits of that ?

IllusionPerdu: I had add ```@since``` because the old version not have this variable and i think it's interesting to know the version this was introduce.

kball: That is interesting. If we add support for `@since` in the docs generator it would be an easier to support alternative to having docs versions for every minor release, but might still provide much of that value. @rafibomb what do you think?

ncoden: According to the docs:

> Describes the version at which the documented item has been implemented _**or updated**_.

So we describe a feature and tell from what version it is like that. It can be helpful to understand why new features described in the docs are not available in an older version, but it is confusing for updated features: is it new, it is updated ?

IllusionPerdu: I hesitated to put the "@since" because I find that the main problem is that when we do a "pull request" we do not know the version for which it will really be applied.
If the definition of the variable changes (for example in the types) I think it would change the "@since" to signal that it is no longer used as before. So as soon as there is a "@since" it can either be an addition of variable or a variable incompatibility.

IllusionPerdu:
I have one question : for update the /docs.*.md it's should be in this "pull request" or we can create a new pull request who linked with this ?

scss/grid/_grid.scss
ncoden: I agree with @IllusionPerdu:
- We can not assume the version number of a commit. We can use versions inside the code only to reference _previous versions_ (already released) or _planned changes_ (not in the current commit).
- `@since` does not really help folks that are using a previous version.

ncoden:
Ideally, docs must be updated in the same PR as the changes.

IllusionPerdu:
For the change in doc i think we sould add the note for the optional alias but i do not know how to formulate it in English: s

And i think we do an other "pull request" for the change in code exemple of ".columns" in ".column" for the pepole who make a copy & past for test ;)

What do you think ?

ncoden:
Yes, we can do an other PR to change all the `.columns` in docs to the default name `.column`.

docs/pages/grid.md
```diff
@@ -10,7 +10,7 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or its <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
```
IllusionPerdu: What do you think to add a callout to explain the alias and say if they are not alias : The file size is decreased by about 6kb when all breakpoint was use.

ncoden: 6kb: what does it give in % ?

And this information can be in the sassdoc comment: it helps to choose the value of the option, and do not need examples.

IllusionPerdu: If with have : $breakpoint-classes: (small medium large); 
* the foundation.css : 104 313 octets
* with no alias : 100 480 octets

If  $breakpoint-classes: (small medium large xlarge xxlarge);
* the foundation.css : 126 407 octets
* with no alias : 120 100 octets

the file is is no minise

All other options is defaut

ncoden: So from 3,8% and 5,2%. It is not negligible.

You can add: `Disabling the alias can reduce the Foundation CSS file size from 3 to 5%. It is recommended if the alias is not used.`

docs/pages/grid.md
```diff
@@ -10,10 +10,15 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or its <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
+<div class="warning callout">
+  <p>By defaut `.column` have an <a href="http://localhost:3000/grid.html#sass-reference">alias</a> `.columns`&mdash;the only difference is grammar.</p>
```
ncoden: If you use a callout for this, you can write:
```
...have an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$column-alias` option</a>)...
```

scss/grid/_grid.scss
```diff
@@ -30,6 +30,10 @@ $grid-column-align-edge: true !default;
 /// @type Number
 $block-grid-max: 8 !default;
 
+/// Selector used for an alias of column (with @extend). If `false`, no alias is created.
+/// @type String
+$column-alias: 'columns' !default;
```
ncoden: Please use the grid namespace: `$grid-column-alias`

docs/pages/grid.md
```diff
@@ -10,10 +10,15 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
+<div class="warning callout">
```
ncoden: Also, I think `class="callout primary"` would be better.
This is an info, not a warning.

scss/grid/_classes.scss
```diff
@@ -167,8 +167,8 @@
     @include grid-column-margin;
   }
 
-  @if $column == 'column' {
-    .columns {
+  @if $column == 'column' and has-value($column-alias) {
+    .#{$column-alias} {
```
ncoden: Please update `$column-alias` to `$grid-column-alias` everywhere.

ncoden:
After these last changes, I think it can be merged.

scss/grid/_classes.scss
IllusionPerdu: oups

docs/pages/grid.md
```diff
@@ -10,10 +10,15 @@ tags:
 
 ## Basics
 
-Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. You can use `.column` or `.columns`&mdash;the only difference is grammar. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
+Start by adding an element with a class of `.row`. This will create a horizontal block to contain vertical columns. Then add elements with a `.column` class within that row. Specify the widths of each column with the `.small-#`, `.medium-#`, and `.large-#` classes.
 
 **Foundation is mobile-first.** Code for small screens first, and larger devices will inherit those styles. Customize for larger screens as necessary.
 
+<div class="primary callout">
+  <p>By defaut `.column` have an alias `.columns` (see the <a href="http://localhost:3000/grid.html#sass-reference">`$grid-column-alias` option</a>) &mdash;the only difference is grammar.</p>
```
kball: Some typo's/English feedback:

`default` instead of `defaut`. `has` instead of `have`.

IllusionPerdu: ok it's updated

kball:
👍  Merged!

stefan2:
@IllusionPerdu, @ncoden 
Your so very very cool. Excellent work guys!

PR #9651, opened 2017-01-15
Reveal deepLinks play the same as with Tabs

xhezairbey:
The current state of Reveal deep-links alters browser history at all times (with or without hash), thus forcing numerous back button clicks to get to the previous page. Having multiple reveals with `deepLink:true` makes this a not so pleasing experience.

This changes here bring Reveal deep-link feature to the same functionality as it is with Tabs, by using **replaceState()** instead of replacing browser hash value explicitly.


js/foundation.reveal.js
```diff
@@ -184,13 +184,14 @@ class Reveal {
    * @fires Reveal#open
    */
   open() {
+    // either update or replace browser history
     if (this.options.deepLink) {
       var hash = `#${this.id}`;
 
-      if (window.history.pushState) {
-        window.history.pushState(null, null, hash);
+      if (this.options.updateHistory) {
```
ncoden: _For now_ we have to support IE 9. So we should keep the `window.history.pushState` check.

js/foundation.reveal.js
```diff
@@ -570,6 +571,12 @@ Reveal.defaults = {
    * @default false
    */
   deepLink: false,
+  /**
+   * Update the browser history with the open modal
+   * @option
+   * @example false
```
ncoden: `@default`, not `@example`.

ncoden (commented):
We should factorize the history management by the way.

xhezairi:
@ncoden right, but since we're not doing that for Tabs deepLinks I kinda assumed probably wasn't much of a concern. (https://github.com/zurb/foundation-sites/blob/develop/js/foundation.tabs.js#L243)

I'll make the updates.

ncoden:
@kball Should we continue to support IE 9 in `v6.4` and update Tabs ?

kball:
@ncoden Yes, I think so. We'll drop IE9 in `v7` but until then we need to provide something that doesn't break back to IE9.  I think it is ok if the behavior doesn't fully function for IE9, but it shouldn't throw JS errors that break other behavior.

ncoden:
So based on @kball feedback, I think we should check for `window.history` in Reveal and Tabs.

kball:
@xhezairi are you going to have time to update this PR?  Thanks!

kball:
reping @xhezairi - we're looking to get 6.4 out in the next month, will you have time to update this based on the above conversation? If not I can try to find someone else to take it across the line

xhezairi:
@kball yes absolutely, will act on this later today or at latest early tomorrow. 
P.S. sorry for staying out of the conversation here.

xhezairi:
Done @kball.

IamManchanda:
@kball .... I think this is missed .... Request you to check this out!

kball:
👍

kball:
Tested out, and it looks good. Merged!  As per the discussion above, we need the feature support check update in tabs as well... also I found magellan deep linking to be buggy during the docs update, so could use a look, maybe this logic can provide a guide?  either @IamManchanda or @xhezairi want to take a pass on those?

xhezairi:
@kball I'd gladly take that. Unless @IamManchanda has already started?!

IamManchanda:
No not started... Please go ahead

PR #9653, opened 2017-01-16
Make class on body persists if another Reveal is still visible.

Owlbertz:
When multiple Reveal modals are open and I close one of them, the class `.is-reveal-open` was removed from the `<body>` despite another modal still being open.
Now checking if another Reveal is visible before removing the class.

Added visual and automatic test.


kball:
Looks good, thanks @Owlbertz!

PR #9654, opened 2017-01-16
Keyboard mock

Owlbertz:
Added a util class for easier testing of keyboard events in Mocha.
This class generates mocked events that can be parsed by the Keyboard util.

Currently implemented as a proof of concept in Drilldown and Orbit.

@zurb/yetinauts Please have a look and give me some feedback about this idea in general. I'll then eventually implement it into further components as well.

### DIFF
--- a/test/javascript/components/drilldown.js
+++ b/test/javascript/components/drilldown.js
@@ -212,4 +212,103 @@ describe('Drilldown Menu', function() {
   });
 
 
+  describe('keyboard events', function() {
+    // Currently not testable, as triggered event won't move on focus
+    it.skip('does not trap focus on root element going down', function() {
+      $html = $(template).appendTo('body');
+      let $dummyElement = $('<a tabindex="0">Dummy</a>').appendTo('body'); // Dummy target to see if focus is not trapped
+      plugin = new Foundation.Drilldown($html, {});
+
+      // Focus last element
+      $dummyElement.focus();
+      $html.find('> li:last-child > a').trigger(window.mockKeyboardEvent('TAB'));
+
+      document.activeElement.should.not.be.equal($html.find('> li:last-child > a')[0]);
+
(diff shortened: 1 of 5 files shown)


kball:
I think this is an excellent approach.  I'd like to see the separation of concerns carried even further - it would be great if the actions that are triggered by keyboard events were functions on the plugins, so we can independently test "did the right action get triggered" from "did the action do the right DOM changes"

Owlbertz:
Yeah, would be great. Having too much logic in an event handler seems to be bad practice when writing tests. I didn't want to bloat this changes up too much without having some feedback, but I guess having a separate function for each action would be a nice thing to have.

However, I am still struggling with functions being private as most of the logic in the plugins is executed with event handlers invoking private functions - which are technically directly invokable (does this word exist?) but should not be called directly (as they are considered private...).

kball:
Yeah - we follow the practice of using a leading `_` for "private" functions.  We could make them truly private by having them not actually be on the object but captured in a closure, but that makes testing harder.  I'd prefer to make them invokable and private by convention.

Owlbertz:
Maybe we could define then that a function that is invoked in an event handler should be considered public as it is called from the outside?

kball:
I think that's reasonable... what would be ideal is if we could encapsulate these in such a way that our mouse-based interactions are going through the same functions.  This would get us a large part of the way to having a truly separable view interface, which is one of our goals for `v7`.

Owlbertz:
That is actually already the case in many places; let's have Drilldown as an example:
`_hide()` is considered private and should therefore not be invoked from the outside. But it is called from both, click- and keyboard-events.

kball:
@Owlbertz are you thinking to iterate further on this within this PR? Or should I merge and we can apply this more in additional PRs?

kball:
Going to go ahead and merge this,

Issue #9494, opened 2016-12-12
Responsive Accordion Deep Linking


mediabeastnz:
The new responsive accordion doesn't seem to play nice with deep linking. 

Events on either Tabs or Accordion only fire when the media query event is triggered on the new plugin.

It would be ideal if these events worked as expected with out having to resize or if deep linking was built into ResponsiveAccordion.

kball:
interesting, thanks for trying out 6.3 ahead of the release and filing this!  Will take a look shortly - don't suppose you have an example already coded up I can use to debug?

mediabeastnz:
No unfortunately I don't have one that's easy to share (lots of SilverStripe CMS templating logic throughout).

I just copied the example from the 6.3 docs and attempted a few different approaches to enable deep linking as well as attempting to trigger an even when a hash is in the URL but that event only fired when the browser was resized.

kball:
Oof - accordion doesn't have deeplink support at this moment, so a full solution will require implementing that.

However, I found that if I add `data-deep-link="true"` to the element that has the `data-responsive-accordion-tabs` component, it appears to work for me when the accordion-tabs is in the tabs form... both on changing tabs and then loading a page with the hash url.  The caveat is that when if the component is in accordion form either when navigating it or when coming in via a link, it doesn't work.

@mediabeastnz can you confirm?  Or does even this case not work for you?

mediabeastnz:
It doesn't seem to work - although if i have:
```html
<li class="accordion-item" data-accordion-item>
    <a href="#" id="tab3" class="accordion-title">Accordion 3</a>
</li>
```
and then link that page with the hash #tab3 it highlights the tab but doesn't open the content nor does it do anything when in accordion mode.

kball:
Try adding the ids to the hrefs... I'll see about updating the docs, but you can take a look at the testcase I was playing around with here: https://github.com/zurb/foundation-sites/commit/93c39c7e679d02885078899c5e3aa3c2bd46c35b

mediabeastnz:
Ahh I see I think we had ID's as well as href's.

So from this...

1. Deep linking is only supported on tabs (if you use accordion for mobile a deep link wont work).
2. Deep linking seems to be stripping the URL back to just the hash e.g. `/my-page#myhash` becomes `/#myhash`

At least we're aware of the limitations of the current version of responsive tabs/accordion :)

kball:
[...]

Issue #6511, opened 2015-04-29
Tabs Deep Linking href="URL+Hash" fails


designerno1:
my anchor tags have the following format

``` html
<a href="http://example.com/#hash">hash</a>
<a href="http://example.com/#hash1">hash1</a>
```

not only #hash like in dokumentation

``` html
<a href="#hash">hash</a>
<a href="#hash1">hash1</a>
```

so the handle_location_hash_change function sends undefined to the toggle_active_tab funtion wich returns in a exeption

``` html
TypeError: c.attr(...) is undefined
```

my workaround for that is to take the contains selector in line 106,112,118

``` javascript
106
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href=' + hash + ']').parent());
112
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href=#' + hash_tab_container_id + ']').parent(), hash);
118
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href=' + self.default_tab_hashes[ind] + ']').parent());
```

changed to

``` javascript
106
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href*=' + hash + ']').parent());
112
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href*=#' + hash_tab_container_id + ']').parent(), hash);
118
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href*=' + self.default_tab_hashes[ind] + ']').parent());
```

also the ends with selector for me works fine

``` javascript
106
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href$=' + hash + ']').parent());
112
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href$=#' + hash_tab_container_id + ']').parent(), hash);
118
self.toggle_active_tab($('[' + self.attr_name() + '] > * > a[href$=' + self.default_tab_hashes[ind] + ']').parent());
```

i think the contains or end with selector is the better way to check the anchor tag

gakimball:
Why are you using the entire URL instead of just the hash?

designerno1:
Interesting Question
i had upgraded a older shopping system to Responsive Design with Zurb Foundation
This system uses the base href tag to the src of images and linking
```html
<base href="http://www.example.com/" />
```
now if i'am in http://www.example.com/dir/dir/ for example and using
```html
<a href="#hash">Test</a>
```
[...]

PR #9660, opened 2017-01-16
fix Tab-Deeplinks with absolute URLs

designerno1:
fixes part of #9494

referenc to a F5 issue https://github.com/zurb/foundation-sites/issues/6511


kball:
Merged!

PR #9542, opened 2016-12-21
[WIP] Better Sass dependencies management

ncoden:
From: https://github.com/zurb/foundation-sites/pull/9478
Fix: https://github.com/zurb/foundation-sites/issues/9492

#### Changes:
- [x] Revert #9430, #9458, #9469.
- [x] Provide a "ready-to-use" SCSS dist file `dist/scss/foundation.scss` with dependencies included.
- [x] Import Sass dependencies using [Eyeglass](https://github.com/sass-eyeglass/eyeglass). (require https://github.com/theJian/gulp-shopify-sass/issues/17)
- [ ] Remove `settings.scss` deps and ensure it can be imported _after_ Foundation.
- [ ] Update documentation

*Note to myself: the `_vendor` folder was a f stupid idea. Don't do that again !*

PR #9661, opened 2017-01-17
HiDPI breakpoints

### LINKED ISSUE #9556: retina named query
Any thoughts on changing the pixel ratio for the `retina` named query?

I'm running a 4K display on Windows at 175% scaling. I think it's pretty common to run HiDPI at 150-175% scaling. Would be nice to have Interchange use high res images at those scales too, out of the box. Currently the retina named query is 200%+ scaling:

`only screen and (-webkit-min-device-pixel-ratio: 2), only screen and (min--moz-device-pixel-ratio: 2), only screen and (-o-min-device-pixel-ratio: 2/1), only screen and (min-device-pixel-ratio: 2), only screen and (min-resolution: 192dpi), only screen and (min-resolution: 2dppx)`

I see a 1.3 pixel ratio used a lot

ncoden:
~~Closes~~ Partially fix #9556.

Add `hidpi-1-5`, `hidpi-2` and `hidpi-3` breakpoints.

### Usage
```scss
@include breakpoint(hidpi-2) {
  // HiDPI x2 and up.
}

@include breakpoint(hidpi-2 only) {
  // Between HiDPI x2 and  x3 (excluded).
}

@include breakpoint(hidpi-2 down) {
  // HiDPI x3 (excluded) and down.
}
```

Any HiDPI breakpoint can be added in the `$breakpoints-hidpi` map (the order does not matter):
```scss
$breakpoints-hidpi: (
  hidpi-1: 1,
  hidpi-1-5: 1.5,
  hidpi-2: 2,
  retina: 2,
  hidpi-3: 3
) !default;
```

### Changes:
- Refactor breakpoint function.
- Add `$breakpoints-hidpi` map: A list of named HiDPI breakpoints with HiDPI ratios (without units).
- Add support for HiDPI breakpoints (up, only, down) in `@function breakpoint`.
- Add unit tests for HiDPI breakpoint with default (up), up, only and down ranges.

### Other changes:
- Add `@function -zf-map-next-number`: Find the next number in a map.
- Add `@function -zf-bp-join`: Return media query string from the given min and/or
max limits.
- Add `@function zf-str-join`: Return a join of the two given strings `$str1` and
`$str2`. If the two strings are not empty, they are separated by
`$delimiter`.

No breaking changes: HiDPI breakpoints works exactly like standard breakpoints. `retina` is an alias for `hidpi-2`.


ncoden:
So, some news: I looked at how to add HiDPI breakpoints to Interchange. It will require a refactor of Interchange and util.MediaQuery since these two files are built on the assumption there is only size breakpoint and only one breakpoint is valid at the same time.

It would also be good to pass datas from SCSS to Js properly throw JSON instead of serialised map. But I do not want to add any SCSS library (SassyJSON) before having a proper SCSS dependency management. So it will wait for https://github.com/zurb/foundation-sites/pull/9542.

@kball @nlap Let me know it you want the Interchange support to be added in this PR or an other.

nlap:
Hmm I see, I hadn't looked into it enough until now to notice the interchange retina query wasn't passed from the SCSS. 

FWIW my use for #9556 was specifically with Interchange, but I think this PR adds value as it is.

kball:
I concur, lets keep the interchange refactor separate. @brettsmason @andycochran as our other "sassy" yetinauts, can one of you review? ;)

kball:
ping @andycochran can you take a look at this?

ncoden:
Note: this PR only add support for hidpi in the `breakpoint` mixin. A refactor of the way breakpoints are handled across the whole framework is required to add support in visibility classes, grids & cie.

DanielRuf:
> Note: this PR only add support for hidpi in the breakpoint mixin. A refactor of the way breakpoints are handled across the whole framework is required to add support in visibility classes, grids & cie.
> 
> The problem mainly comes from that functions/mixins will manually get the $breakpoints map instead of relying on the breakpoint mixins. This is not hard to refactor, but it takes some time to be sure nothing has been broken.

Which is still relevant also for https://github.com/zurb/foundation-sites/pull/10979

kball (approved):
Approved, fixing merge conflicts and merging.

PR #9664, opened 2017-01-17
new feature accordion deep link

### LINKED ISSUE #9494: Responsive Accordion Deep Linking

designerno1:
Deep Linking for accordion similar to tabs deep link

```javascript
deepLink: false

deepLinkSmudge: false

deepLinkSmudgeDelay: 300

updateHistory: false
```
should fix #9494

@zurb/yetinauts please take a look

### DIFF
--- a/docs/pages/accordion.md
+++ b/docs/pages/accordion.md
@@ -140,7 +140,7 @@ Add the attribute `data-deep-link="true"` to a accordion to:
 - allow users to open a particular accordion panel at page load with a hash-appended URL
 
 ```html_example
-<ul class="accordion" data-accordion data-deep-link="true" id="deeplinked-accordion">
+<ul class="accordion" data-accordion data-deep-link="true" data-update-history="true" data-deep-link-smudge="true" data-deep-link-smudge="500" id="deeplinked-accordion">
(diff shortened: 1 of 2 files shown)


ncoden:
Reveal, Tabs, Accordion, Magellan...

I do not know these components enough to be sure, but (with this PR) they all have Deep Linking, so there is probably a way to factorize this. Maybe by making DeepLink an utility used by components and maybe usable directly.

kball:
I agree with @ncoden that we should refactor deeplinking to be a common utility in some form. I don't think that necessarily needs to be a part of this PR... the one thing I would ask for is that we add a check for if the history API exists so we don't break on IE9 (as discussed on https://github.com/zurb/foundation-sites/pull/9651) - I'm ok if deeplinking doesn't work on IE9, but it shouldn't throw an error as that can have negative side effects

kball:
Nice work @designerno1. Merged!

Issue #9421, opened 2016-11-30
Full menu rework


kball:
We want to clean up the way menus work to make them much easier to customize, make them more consistently accessible, and easier to work with.

Opening this to collect related code snippets, prs, issues, and conversations.

Some initial items of interest are: 

- https://github.com/zurb/foundation-sites/pull/9283 (temporarily reverted in https://github.com/zurb/foundation-sites/pull/9420). 

- the discussion in https://github.com/zurb/foundation-sites/pull/9381

... others? 

(cc @zurb/yetinauts @sethkane)

andycochran:
#7235

brettsmason:
So from my point of view, the main issues are:

- the default `.menu` class is too opinionated - things like padding should be left to modifier classes and specific component classes (dropdown, accordion etc) to add this sort of styling
- The specificity of the selectors makes it really difficult to override styles

A fairly brief attempt at seperating this can be seen here: http://brettmason.co.uk/menus/all-menus.html
This would allow classes to be combined, without them all overwriding eachothers styles.

ncoden:
My thoughts about menu : Your menu should with a structure exactly like `.menu > li > a` and with the default style. If you want something else, you're damn.

Things to avoid for the new menu:
- selecting elements by their tag
- mixing graphical and structural properties
- high specificity

andycochran:
A few other gathered thoughts… 

- #8504 Menu item height should be the same as button height. 

- How do we handle button/menu item radius in v6.4? There's been a lot of debate about radius on group vs radius on each button. #9031 "fixed" #8447 for v6.3. But I think the framework could/should support both cases. See also this [Codepen](http://codepen.io/andycochran/pen/reJRro) which also mentions the issues with menus/button groups wrapping to multiple lines. 

- #8239 Generate menu color classes from the palette. No need to have both `.stacked-for-small.button-group` _and_ `.vertical.medium-horizontal.menu`.

JeremyEnglert:
A big part of Foundation 6 was cutting down on the amount of variables in the settings.scss file, however, navigation items are a great example of where having "a lot" of settings greatly improves ease of use. 

As others have mentioned, it becomes quite the pain to override specific styles in a menu because they are very specific - to make this even more challenging, it's often hard to find the specific classes to target using Dev Tools as the classes are added dynamically via JS.

[...]

Issue #7095, opened 2015-11-21
[Responsive Navigation] A Couple Issues with Arrows...


Lynda333:
I've noticed a couple issues that I have with the way arrows are placed with the responsive navigation. 

First let's take a look at a menu I have setup: 

![dropdown](https://cloud.githubusercontent.com/assets/1938824/11316082/7729fe1a-8fbb-11e5-82d1-70898b6ed4d3.png)

(For reference the code for my menu is at the bottom of this post.)

I have a dropdown menu setup that turns into a drilldown menu when on a small screen. Works great but the arrows are my issue. 

First the arrow points to the right. When the menu is horizontal the arrow should point down. I see this as a UX problem myself. If the menu was vertical or even in drilldown mode, then this would be acceptable.

Second, I am using the `expanded` class for a nicely spaced menu. I love the class but I prefer the arrows to be next to the words, not all the way to the right. Depending upon the size of the menu, the arrows are placed to far from the words they go with.

Here is a screenshot for this problem: 

![arrows](https://cloud.githubusercontent.com/assets/1938824/12870752/754f3370-cd10-11e5-8fb7-2cd4266338af.jpg)

I assume some changes to the SCSS is all that is needed. I believe I can manually override the first issue with some media queries (haven't tried yet) but I am unsure how to fix the second issue.

``` html
<div class="title-bar" data-responsive-toggle="mainMenu" data-hide-for="medium">
  <button class="menu-icon" type="button" data-toggle></button>
  <div class="title-bar-title">Menu</div>
</div>

<ul class="vertical medium-horizontal menu expanded" data-responsive-menu="drilldown medium-dropdown" id="mainMenu">
  <li><a href="#">Home</a></li>
  <li class="has-submenu"><a href="#">News</a>
    <ul class="vertical submenu menu" data-submenu>
      <li><a href="#">Colorado News</a></li>
      <li><a href="#">National &amp; International News</a></li>
    </ul>
  </li>
  <li class="has-submenu"><a href="#">Information</a>
    <ul class="vertical submenu menu" data-submenu>
      <li><a href="#">City &amp; County Regulations</a></li>
      <li><a href="#">Denver Information</a></li>
      <li><a href="#">General Information</a></li>
      <li><a href="#">Case Law</a></li>
      <li><a href="#">Federal Letters</a></li>
      <li><a href="#">Colorado Patient Services</a></li>
      <li><a href="#">Power Point Presentations</a></li>
    </ul>
  </li>
[...]

PR #9670, opened 2017-01-18
Menu styling rework

brettsmason:
This is an attempt at starting a rework for the current menu system. See #9421.

## Aim of this PR

- To reduce the specificity of the current `.menu` class and make it more flexible
- Remove `table-cell` from `.menu` (as its a nightmare to work with!)
- Add more styling flexibility to each of the menu plugins (dropdown/drilldown/accordion) by adding new variables to allow for more control

## Current Problem

`.menu` includes its own padding values, along with heavy use of `>` which makes it very difficult to override with your own styles. I suggest making `.menu` almost like a base reset for menus. Then other classes like `.simple` or `dropdown` add their own margin/padding and there's no issue with overriding the inherited `.menu` styles.

This PR is still very early on and am experimenting with it a lot. There is a visual test under menu that will show you the various classes and how they work with each other.


IamManchanda:
awesome @brettsmason !!!

brettsmason:
@IamManchanda Thanks! Not to worried about conflicts at the moment as its very much work in process and an exploration of what we can do to improve the menu system.

@kball @rafibomb I think this should be ready for me to discuss with you guys on Thursday assuming that's still happening. I have a few more ideas, but probably best to check with you first.

kball:
@brettsmason checking in real quick - what is the state of this?

brettsmason:
@kball I think from memory it was in a state were it could now be tested to see how it works. I think from a previous video chat you guys were going to test it in a project to see if it worked OK. 

I'll double check though as it's been a while.

IamManchanda:
Menu with input, select and buttons is having visual issues - More with flex version
Rest looks good visually!

kball:
@brettsmason or @IamManchanda can either of you look at fixing the issues highlighted above before the first 6.4 RC (End of May)?

IamManchanda:
Kindly check now @kball .... i have updated 
and now the output....

menu-flex-styling

brettsmason:
I think this could still do with a full review before it gets merged, as far as I know it's only been quickly looked at. I'll take another look to make sure I finished everything - lost motivation on it a bit as nobody looked at it! :)

IamManchanda:
> lost motivation on it a bit as nobody looked at it

Dang .... dont loose motivation as harry is on it!

IamManchanda:
@brettsmason @kball how about creating responsive breakpoints for accordion's nested class 
https://github.com/zurb/foundation-sites/issues/10002 https://codepen.io/wecodeart/pen/xddBZz

brettsmason:
@IamManchanda I'm not sure I see the need for that? If you look at the new code there are now variables and styling to control individual navigation style components (eg `.accordion-menu`). I personally think thats enough control, as I was trying to cut back on the bloat 😄

IamManchanda:
@brettsmason But we have that one way to minimize that bloat!

```scss
/// Responsive breakpoints for nested menu
/// @type Boolean
$responsive-menu-nested: false !default;
```

```scss
.menu {
  // Nesting
  &.nested {
    @include menu-nested;

    @if ($responsive-menu-nested) {
      // Loop through Responsive Breakpoints
        @if $size != $-zf-zero-breakpoint {
          @include breakpoint($size) {
            &-for-#{$size} {
              @include menu-nested;
            }
          }
        }
        @include breakpoint($size only) {
          &-for-#{$size}-only {
            @include menu-nested;
          }
        }
      }
    }
  }
}
```

IamManchanda:
@BicanMarianValeriu being the author of that nested thread ..... What do you think ??

BicanMarianValeriu:
@IamManchanda 
You know what I think :)). It looks really ugly by default if nested class is used (mainly for vertical menus) and the margin will also remain for dropdowns. Another way of doing this is to remove that left margin of the nested class for horizontal dropdowns.

We can do that with: 
```
.nested.is-dropdown-submenu {
// When menu turns to accordion this class is transformed to is-accordion-submenu
// and we will still have that margin
margin-left:0;
}
```

I think this is better, not that much bloat but test it, you know better.

brettsmason:
@IamManchanda Looking at this properly now, I think we should add a new variable to each 'mobile' menu component (drilldown, accordion) eg: `$accordionmenu-nested-margin: null !default;` (above here: https://github.com/brettsmason/foundation-sites/blob/683b1c6639a00008373c9af77711b87bf8b9aabe/scss/components/_accordion-menu.scss#L11)

This would allow control over if the menus should have margin on them or not (null by default so nothing is output). Then a style needs adding for `.submenu` with the value of this variable.

Would that solve this do you think? I think its going to be the cleanest way.

brettsmason (commented):
Left a quick bit of feedback (added this by mistake 😆 )

scss/components/_menu.scss
```diff
@@ -52,6 +52,11 @@ $menu-item-background-hover: $light-gray !default;
 
   li {
     @include disable-mouse-outline;
+
+    @if $global-flexbox {
```
brettsmason: Not sure I understand the point of this, could you explain? If we are adding spacing between `li` items, it should be at least controllable by the user. Why not have them sitting flush together though?

IamManchanda: Yeah sitting flush together also would be good... I was wanting it to similar to float!

IamManchanda:
@brettsmason :+1: 

But man looking at @BicanMarianValeriu ,
I was looking at drilldown accordion combo

```scss
.nested.is-dropdown-submenu,
.nested.is-drilldown-submenu{
   margin-left:0;
}
```

and man i found another bug :sweat_smile: 
=> https://codepen.io/IamManchanda/pen/BRGJGL
Similar to #10031... Gonna report it :smile:

scss/components/_menu.scss
brettsmason: Remember flexbox is going to be primary from 6.4, and the space between in is more a bug with `inline-block`. I recommend we remove the margin and sit flush personally, as it would generally be better (and in 7 there would be no float anyway I expect!).

brettsmason:
>  I was advancing and looking at drilldown accordion combo

That's why I said we would have the same in the drilldown partial 😄 It would be cleaner, and the value would be totally configurable for Sass users. And for CSS users, its not hard to write the one line needed to add margin only on accordion for example (in my opinion) 👍

IamManchanda:
@brettsmason Related to our discussion in #9985 .... i added a test case 
(which is currently not working ... as we can see) .... How can we best solve this??
(Without conflicting the specificity part ofcourse)

scss/components/_menu.scss
IamManchanda: > and in 7 there would be no float anyway I expect

Then how would you get your goverment clients :sweat_smile: who uses IE all the time :wink: 
Ohk removing it!

brettsmason:
@IamManchanda Thanks for putting the test case together. I've been wrapped up in the new grid, but this evening I'll pull down this menu rework and familiarise myself with it again so I understand it properly again 😄 Then I will be able to help more 👍

scss/components/_menu.scss
brettsmason: I'll have to stick to V6 for them 😢

IamManchanda: :smile: :smile_cat: :smiling_imp: :joy: :stuck_out_tongue_closed_eyes:

IamManchanda:
@brettsmason ..... As discussed taking this up .... best of luck for the grid :wink:

IamManchanda:
Hey @brettsmason, me and @kball had discussion about #10002
Although we both liked your [idea](https://github.com/zurb/foundation-sites/pull/9670#issuecomment-302920237)

But the problem was Nested class is good for simple menu and accordion menu 
but ugly for dropdown and drilldown. 
And thus was bad responsive menu!

So here's what we got up to 

1.  Added a global variable 

```scss
$global-menu-nested-margin: 1rem !default;
```
2. Define simple menu and accordion menu the global variable 

```scss
$menu-nested-margin: $global-menu-nested-margin !default;
$accordionmenu-nested-margin: $global-menu-nested-margin !default; 
```

3. Define dropdown menu and drilldown menu to `0` .... so no nesting gets here!

```scss
$dropdownmenu-nested-margin: 0 !default; 
$drilldown-nested-margin: 0 !default; 
```

Now i added the classes like these 

```scss
.accordion-menu {
  /* my other code */
  .nested.is-accordion-submenu {
     @include menu-nested($accordionmenu-nested-margin);
   }
}
```
and same for other two! Sorry if you dont concur with this. 
Let me know your thoughts!

brettsmason:
@IamManchanda That sounds a really good plan 👍 

Just a few comments:
The `$global-menu-*` variables - should this be `$menu-global-*` instead? Would sit better under the list of variables, and other variables are all prefixed with their component name.

> .nested.is-accordion-submenu {

Cant this just be `.nested`? No need for the extra specificity unless I'm missing something.

IamManchanda:
@brettsmason 
> That sounds a really good plan 👍 

Not a plan now as implemented :smile: 

> The `$global-menu-*` variables - should this be `$menu-global-*` instead? Would sit better under the list of variables, and other variables are all prefixed with their component name.

All the variables in the global file goes with the naming convention like this.
`$global-radius` , `$global-menu-padding`, `$global-flexbox`
So i have no doubts in my mind that it should be this covention 

> > .nested.is-accordion-submenu {
>
> Cant this just be `.nested`? No need for the extra specificity unless I'm missing something.

@kball what do you think on this .... on just being `.nested` ??

kball:
Do we ever have non-accordion menus nested within accordion menus? If so, I think we need the extra specificity.

IamManchanda:
@kball We actually have it ..... yes menu nested inside a menu .... just added a test case for the same above :arrow_up:  

And this extra specificity only helping it unless it will break! 

drilldown-accordion-nested-menu

brettsmason:
@IamManchanda So I've been trying to catch up a bit on what's been going on with menu.
Just for the record I still disagree with the `.nested` stuff 😄 I don't understand why you would need to use `.nested` in the situations I've seen mentioned. Is it this that you need help with? Cant quite work it out...

IamManchanda:
There will be disagreement.... 
All I can say is you need `.nested` for simple and accordion menu 
But with drop-down and drill down, you don't need that styling and it looks ugly

IamManchanda:
@brettsmason As discussed, i have updated test case for right aligned vertical dropdown menu. Please look into it now!

For central menu, As discussed .... as unuseful and was just a test that didn't work .... i have shelved it.
**Marking it as wont fix**!

IamManchanda:
@brettsmason @abdullahsalem @kball I am able to reproduce the same right aligned vertical issues in accordion and drilldown too . Please help i added test case too 

Here are buggy test path
/menu/vertical-dropdownmenu-alignment.html
/menu/vertical-dropdownmenu-alignment-flex.html
/menu/vertical-drilldownmenu-alignment.html
/menu/vertical-drilldownmenu-alignment-flex.html
/menu/vertical-accordionmenu-alignment.html 
/menu/vertical-accordionmenu-alignment-flex.html

PS: Sorry i know i made lot of test cases :p .... but will push it to `all-menus.html` once we solve this.

kball:
I updated this PR to fix all of the arrows for right aligned case to at least pick the correct "side", and it now works nicely for the float-based menus, but I think there's a fundamental problem with the current approach for flex menus, at least for the non-simple menus (and possibly throughout), that manifests for vertical menus that have right alignment (actually global right, but I'm going to use "right" for simplicity here)

I have an idea of how to fix, but not sure it's the right thing, so want to get opinions...

The issue is that for vertical left-align menus, `a` tags are `display: block` inside a `display: block` `li` parent, which results in them being full-width, which 1. makes the whole row clickable, and 2. makes the arrow position correctly.

For vertical right-align menus, currently in flex mode we're making the `li` parent a flex parent, and using flex alignment to align the contents to the right. The problem with this is it makes the `a` tag compress to the size of the text, meaning the click target is only the text itself, and the arrow (which is attached to the `a`) positions oddly.

I think the solution is to not make the `li` a flex parent at all, and simply use `text-align` to position the text. However, I don't know the original reasoning for making the `li` a flex parent, so want to lay this out and see what @brettsmason and @IamManchanda think

andycochran:
@kball: I've suggested before that maybe arrows should just be inline-block and immediately adjacent to the text. They don't need to be aligned to the the global right of the menu item. It's actually bad UX if they are aligned to the right of menu items that are expanded—'cause the arrow is closer to the follow button's text than it is to its own. Lemme see if I can track down that old comment…

andycochran:
The arrow issue is pointed out here: https://github.com/zurb/foundation-sites/issues/9421#issuecomment-271967786 
and here: https://github.com/zurb/foundation-sites/issues/7095

kball:
@andycochran interesting... that makes complete sense to me for horizontal menus, but for vertical it feels like we should expand to fill the width of the menu

IamManchanda:
@kball .... I added it here https://github.com/zurb/foundation-sites/pull/9670/commits/1a5a6f2f63fe3e2cb0464a41860cd60dfd2354ab
I used to solve it to create vertical menu actually for flexbox.
If i remove flex from there ..... vertical `align-right, align-center` menu fails in flexbox actually (See `/menu/vertical-menu-alignment-flex.html` for testing )
Maybe we can do it in `.submenu li`?? 
But i am not sure

IamManchanda:
@brettsmason BTW ..... are we dropping `menu-centered` ??

brettsmason:
@IamManchanda If `.align-center' does what we need, then no I dont think its needed :thumbsup:

IamManchanda:
@brettsmason Same i feel .... but i am amazed we didn't add this in migration/depreciation ??
https://github.com/zurb/foundation-sites/wiki/Project-Roadmap#64-tentative-date-may

kball:
@IamManchanda we've started following a diff migration tracking approach (each PR at the end around merging if it requires a migration adds 'Migration Needed' tag, with comment about how the migration works... that way at the end we can pull them all up and note them)

With regards to the align-center, align-right... can we simply have a vertical menu with those classes apply text-align?

IamManchanda:
Done @kball ..... Your suggestion .... yup works now!

docs/pages/menu.md
```diff
@@ -64,27 +64,21 @@ By default, each item in the menu aligns to the left. They can also be aligned t
 
 <br>
 
-To align items in the middle, add a wrapping element with the class `.menu-centered`.
+To align items in the middle, use `.align-center`
```
kball: Add it to what? The `.menu`?

docs/pages/menu.md
```diff
@@ -64,27 +64,21 @@ By default, each item in the menu aligns to the left. They can also be aligned t
...
-</div>
-
 
 <div class="docs-codepen-container">
-  <a class="codepen-logo-link" href="https://codepen.io/IamManchanda/pen/BRxXxO?editors=1000" target="_blank"><img src="{{root}}assets/img/logos/edit-in-browser.svg" class="" height="" width="" alt="edit on codepen button"></a>
+  <a class="codepen-logo-link" href="https://codepen.io/IamManchanda/pen/VbOypm?editors=1100" target="_blank"><img src="{{root}}assets/img/logos/edit-in-browser.svg" class="" height="" width="" alt="edit on codepen button"></a>
 </div>
 
 ```html_example
-<div class="menu-centered">
```
kball: What will happen for folks migrating from previous versions who have "menu-centered". Will it work? Or not do anything?

docs/pages/menu.md
```diff
@@ -142,6 +136,39 @@ Add the `.vertical` class to a Menu to switch its orientation.
...
+```
+
+<br>
+
+Add `align-center` class for making the vertical menu aligned to the center.
+
+<div class="warning callout">
+  <p>
+    The above vertically left &amp; right aligned menu are supported in all types of menu's. <br>
+    But as we don't think `align-center` is useful for dropdown, accordion or a drilldown menu, we have just kept it on basic menu only.
```
kball: I think stating what we think is useful is not necessary... lets just state that `align-center` just works in basic menus.

scss/components/_menu.scss
```diff
@@ -34,40 +38,29 @@ $menu-icon-spacing: 0.25rem !default;
...
-      vertical-align: middle;
-    }
   }
 
-  // Reset line height to make the height of the overall item easier to calculate
-  > li > a {
-    display: block;
-    padding: $menu-item-padding;
+  a,
+  .button {
```
kball: All links and buttons inside a menu are this way? That feels a little heavy-handed... though I suppose this is low enough specificity we can override if we need to... @brettsmason what do you think?

scss/components/_menu.scss
```diff
@@ -218,7 +273,7 @@ $menu-border: $light-gray !default;
     }
   }
   @else if $position == top {
-    > li > a {
+    li a {
```
kball: Why do we drop the `>` specifications? This feels likely to apply too broadly now...

scss/components/_menu.scss
```diff
@@ -231,11 +286,35 @@ $menu-border: $light-gray !default;
...
           align-self: stretch;
-          margin-bottom: $menu-icon-spacing;
+          margin-bottom: $spacing;
+          text-align: center;
+        }
+      }
+    }
+  }
+  @else if $position == bottom {
+    li a {
```
kball: Same question as above on not havint `>`

kball (commented):
This is looking like a good step forward and I think we're getting close... some questions added, and I'd like to get opinions from @andycochran, @ncoden, and @brettsmason on this as well if possible.

IamManchanda:
@kball With the current code `.menu-centered` will not work!

IamManchanda:
@brettsmason Can you revert on removing `>` ?

scss/components/_menu.scss
brettsmason: Generally the aim of the whole PR was to cut down on specificity (`.menu > li > a` was a nightmare to change).
I think the way we are generating the classes, so the parent for that will be `.menu.icon-right` would make this specific enough, bit I'm open to suggestions or correction 😄

scss/components/_menu.scss
brettsmason: I think this is carried over fro the old menu system. You are right though, it maybe a bit heavy handed. Having said that, I get why its there - a mini reset. I think `text-decoration` could come out, as thats definately not needed (it should respect whatever settings you have). We could try removing it and seeing how this effects an example including buttons, but I think its needed as we dont use a class on our `a` links within menus to style them (and we needed to keep backwards compatibility).

docs/pages/menu.md
IamManchanda: Done!

docs/pages/menu.md
IamManchanda: Done!

kball:
OK, I'd love to redo the JS side of the menus as well, but we're out of time for something that complex.  Will have to tackle that for 6.5.

I think this is good to go! Going to merge.

With regards to migration, the one item I'm seeing right now is the old `.menu-center` approach has been removed... is there anything else?

kball:
Do we want to include for backwards compatibility a `.menu-centered > .menu` style? Just ran into a visual test broken b/c it hadn't been updated here, but we might want to create this option and deprecate for 1 release

PR #9674, opened 2017-01-18
Fix tabs deeplink history to work with back button within page

kball:
I noticed while working on a lesson on tabs deep linking that the history feature was a little flaky - while it added the history to the URL, clicking back did not appropriately open the previous tab.  This fixes this bug by moving the deep link logic into a function and calling it on popstate.

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -210,6 +210,7 @@ class Tabs {
   /**
    * Opens the tab `$targetContent` defined by `$target`. Collapses active tab.
    * @param {jQuery} $target - Tab to open.
+   * @param {boolean} historyHandled - browser has already handled a history update
    * @fires Tabs#change
    * @function
    */
@@ -302,6 +303,7 @@ class Tabs {
   /**
    * Public method for selecting a content pane to display.
    * @param {jQuery | String} elem - jQuery object or string of the id of the pane to display.
+   * @param {boolean} historyHandled - browser has already handled a history update
    * @function
    */
   selectTab(elem, historyHandled) {
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -78,11 +78,6 @@ class Tabs {
       }
 
     });
-    //use browser to open a tab, if it exists in this tabset
-    if (this.options.deepLink) {
-      this._checkDeepLink();
-    }
-
     if(this.options.matchHeight) {
       var $images = this.$tabContent.find('img');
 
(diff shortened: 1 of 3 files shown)


kball:
@ahebrank and @xhezairi you have both worked in this code recently, can you review this PR?  Thanks

yarn.lock
```diff
@@ -1565,16 +1565,16 @@ doiuse@^2.5.0:
     through2 "^0.6.3"
     yargs "^3.5.4"
 
-dom-serializer@0, dom-serializer@~0.1.0:
-  version "0.1.0"
-  resolved "https://registry.yarnpkg.com/dom-serializer/-/dom-serializer-0.1.0.tgz#073c697546ce0780ce23be4a28e293e40bc30c82"
+dom-serializer@0, dom-serializer@~0.0.0:
```
ncoden: We should update yarn time to time, but not in this PR.

js/foundation.tabs.js
```diff
@@ -365,6 +371,7 @@ class Tabs {
       }
     }
 
+    $(window).off('popstate', this._checkDeepLink);
```
ncoden: No check for `this.options.deepLink` ?

js/foundation.tabs.js
```diff
@@ -114,9 +93,32 @@ class Tabs {
       }
     }
 
+    this._checkDeepLink = this._checkDeepLink.bind(this);
```
ncoden: Why ?

js/foundation.tabs.js
kball: we can, but shouldn't matter. If the handler isn't there it's a noop

js/foundation.tabs.js
kball: So that we can 
1. Count on `this` within the function being a reference to the plugin
2. have a single representation of the function (rather than doing bind at the time we add the handler), letting us remove it using `off`

yarn.lock
kball: Good point, incidental, will fix.

ahebrank:
Thanks, haven't been eating my dogfood and didn't see this glitch.

Two minor armchair things:

1. I find the bound function's name reuse confusing, so I'd probably try anonymizing it.
2. move the options check to the top of the new function so you can call it with impunity and get rid of a couple of `if`s elsewhere.

So, at L96:

```
this._checkDeepLink = () => {
    if (!this.options.deepLink) { 
      return; 
    }
    var anchor = window.location.hash;
    if(anchor.length) {
      var $link = this.$element.find('[href="'+anchor+'"]');
      if ($link.length) {
        this.selectTab($(anchor));

        //roll up a little to show the titles
        if (this.options.deepLinkSmudge) {
          var offset = this.$element.offset();
          $('html, body').animate({ scrollTop: offset.top }, this.options.deepLinkSmudgeDelay);
        }

        /**
          * Fires when the zplugin has deeplinked at pageload
          * @event Tabs#deeplink
          */
         this.$element.trigger('deeplink.zf.tabs', [$link, $(anchor)]);
       }
     }
};
```

The only other thing I'm wondering is if the popstate event should check `options.updateHistory`.  I'm thinking the answer is no, because if this is false the back button shouldn't be doing anything relevant to the tabs anyway. But just in case, L137 would become:

```
    if(this.options.updateHistory) {
      $(window).on('popstate', this._checkDeepLink);
    }
```

js/foundation.tabs.js
ncoden: I think we should. We should explicitly show that a piece of code is only required by an option.

kball:
@ahebrank can you clarify a little what you meant by anonymizing it? The reason I set it up this way was to be able to remove explicitly this handler on destroy, so as not to have to mess with other handlers...

That said, maybe we should implement a global popstate handler similar to how we handle resize/scroll.  Thoughts @coreysyms?

ahebrank:
@kball, here's a diff that should clarify: https://github.com/zurb/foundation-sites/compare/fix-tabs-history...ahebrank:fix-tabs-history

kball:
Righto, makes sense. Will push up an update later today. Thanks @ahebrank

designerno1:
the updateHistory got a little problem in _handleTabChange if it get's handled by the deeplink check it shouldn't pushState again.
otherwise we got the panel hash two times in history after changing the tab on popstate or init
maybe adding a parameter to the _handleTabChange could do the job if it's set by the checkdeeplink do not pushstate

kball:
@designerno1 @ahebrank @ncoden Updated to address all of the concerns raised, please take another look thanks

designerno1:
looks good to me

js/foundation.tabs.js
```diff
@@ -297,7 +304,7 @@ class Tabs {
    * @param {jQuery | String} elem - jQuery object or string of the id of the pane to display.
    * @function
```
ahebrank: need a @param definition, like:

* @param {boolean} - browser has already handled a history updated

js/foundation.tabs.js
```diff
@@ -206,7 +213,7 @@ class Tabs {
    * @fires Tabs#change
```
ahebrank: * @param {boolean} - browser has already handled a history updated

ahebrank:
Needs a comment for the new function parameter, but otherwise thumbs up.

js/foundation.tabs.js
ncoden: One year later: _that was a stupid idea_.

Options can changes, listeners can be created in complex tests, remplaced, created dynamically. Here we only want to clear everything. No need for that check.

DanielRuf:
Hm, what was updated here about 7 hours ago?

js/foundation.tabs.js
DanielRuf: True. So do we need further changes here?

ncoden: No, don't worry. I seen that while working on https://github.com/zurb/foundation-sites/pull/11077

PR #9681, opened 2017-01-19
Form input settings

natewiebe13:
Goal:
- Add the ability to set the padding for form elements.
- Have consistent font-weight for inputs and selects

Notes: The original height calculation was using a hardcoded 1.5 and was being multiplied to the form spacing. I'm guessing this was to work like the line height and was being multiplied to the wrong variable. Let me know if it had a different purpose.


kball:
@andycochran @brettsmason @ncoden what do y'all think?

rafibomb:
The most cross-browser way to increase the size of the inputs is with height and width.

Padding is an issue on FF
screen shot 2017-02-02 at 10 00 17 am

natewiebe13:
@rafibomb, as long as the calculation for height is correct, it shouldn't hurt to update the padding to what would make sense. I don't seem to be having the issue you are. Can you provide a link?

rafibomb:
@natewiebe13 I opened the Foundation docs in FF http://foundation.zurb.com/sites/docs/forms.html

And in the inspector increased the input padding. It doesn't keep the placeholder in the middle.

natewiebe13:
@rafibomb correct, because the height has to be adjusted as well. My PR takes care of that as well. Without touching the inspector, do you notice any issues when adjusting the setting?

kball:
@andycochran can you take a look at this PR?

scss/settings/_settings.scss
```diff
@@ -375,11 +375,13 @@ $input-placeholder-color: $medium-gray;
 $input-font-family: inherit;
 $input-font-size: rem-calc(16);
 $input-font-weight: $global-weight-normal;
+$input-line-height: $global-lineheight;
```
IamManchanda: Setting file shouldn't be added .... Automatically generates!

IamManchanda:
LGTM ..... cc @kball @rafibomb for 6.4!

rafibomb:
It does work using the settings file, nice PR!

rafibomb:
Migration notes: 
new variables
- `$input-line-height: $global-lineheight;`
- `$input-padding: $form-spacing / 2;`

kball:
I think this will actually "just work" without a migration, but adding the new variable will give you the control... @rafibomb was there a particular reason you wanted a note on the new settings?

Issue #9671, opened 2017-01-18
Assign a color first and then reassign the variables to semantic color scheme


IamManchanda:
Read this article => https://davidwalsh.name/sass-color-variables-dont-suck

This will help developers with easy additional customization and also things will be better organized and most importantly this convention is **unique and memorable** 

Here's what i am suggesting for the **foundation pallete**,

**Update: Updated code as per @ncoden comments about unused colors. Removed non used color variables ( for eg: `$pink.` Let user add them themselves as per there needs.** 

_All the colors here i am using is current set of foundation colors_

```scss
// Assign color names to specific hex values.
$black: #0a0a0a !default;
$white: #fefefe !default;

$blue: #1779ba !default;
$green: #3adb76 !default;
$orange: #ffae00 !default;
$red: #cc4b37 !default;

// grayscale
$gray: #767676 !default;
$light-gray: #e6e6e6 !default;
$medium-gray: #cacaca !default;
$dark-gray: #8a8a8a !default;

// Reassign color variables to semantic colors in foundation pallete
$foundation-palette: (
  primary: $blue,
  secondary: $gray,
  success: $green,
  warning: $orange,
  alert: $red,
);
```

**Benefits ( same like **custom grid** ):** 

1. We are allowing them to modify current foundation pallete colors.
2. We are allowing them to create custom color variables for foundation pallete. 

-------------------------------------------------------------------------------------------------

**Special Note**: Final code might be different but hope you get my idea.

ncoden:
👎  We cannot pollute the Sass namespace with color global variables. If developers want a custom color palette, they can define it themselves. There is no need to rely on Foundation.

IamManchanda:
Just now you added two labels .... if you see carefully those two labels has a color different than the current foundation pallete 
or, maybe somebody wants a different primary/secondary color than the current one 
or, maybe somebody create a site with a blue color and then his boss says to change it to a variant of red

There are many scenarios out there where you want to change colors and need to add colors and surely this can be the best use case .

=> Mine is thumbs up for this :+1:

ncoden:
I agree there _are many scenarios out there where you want to change colors_. All I am saying is that if you need other colors, you can declare your own color variables and use it where it is needed.

```scss
$my-blue: #0000FF;

// On your own label
.my-label {
  background: $my-blue;
}

// On Foundation label
$label-background: $my-blue;
```

IamManchanda:
I am just asking you to be semantic.

[...]

Issue #9646, opened 2017-01-13
Social Buttons (CSS Classes)


IamManchanda:
Motivation: http://mdbootstrap.com/components/social-buttons/

Can be a perfect partner to http://zurb.com/playground/foundation-icon-fonts-3

What it will help us achieve 

```html
<button type="button" class="button facebook">
  <i class="fi-social-facebook"></i>
</button>
```

Use Case: Can be very helpful for social login, sharing and more!

{{ **Rest on you guys to decide** }}

ncoden:
This is not a bad idea, but I see a main problem: "Social Buttons" is a design modifier of Button, while social buttons can take _a lot_ of different and various designs.

What design should we provide by default ?
Would most of the users have to customize their design anyway ?

IamManchanda:
Provide a icon with white color and background with the colour of the social brand ..... simple buttons like normal foundation buttons

And let them decide to choose this 

```html
<button type="button" class="button facebook">
  <i class="fi-social-facebook"></i>Facebook 
</button>
```

or this 

```html
<button type="button" class="button facebook">
  <i class="fi-social-facebook"></i>
</button>
```

IamManchanda:
I don't think we need to create different designs .... That's not our goal I guess .... 
=> Structure and not style 
We just need to create structure .... and leave different designs on users

IamManchanda:
To anyone concerned .... moving this to building blocks!

PR #9683, opened 2017-01-20
[WIP] Separate Foundation Palette

natewiebe13:
Goal: To separate the Foundation palette into having a map that lists possible colors, and a map that assigns a color to a state.
TODO: Update documentation. (This should be done after functionality is finalized)

Fixes: #9671


IamManchanda:
I just pulled it .... i am sorry but docs is not working bro .... did you tested your code ?

scss/_global.scss
```diff
@@ -20,35 +20,50 @@ $global-width: rem-calc(1200) !default;
...
-/// Colors used for buttons, callouts, links, etc. There must always be a color called `primary`.
+/// Colors available by using `get-color()`.
 /// @type Map
-$foundation-palette: (
-  primary: #1779ba,
-  secondary: #767676,
-  success: #3adb76,
-  warning: #ffae00,
-  alert: #cc4b37,
+$foundation-colors: (
```
ncoden: I don't know if `_global.scss` is the best place for this. We should rather put it in `util/_color.scss`.

scss/_global.scss
```diff
@@ -20,35 +20,50 @@ $global-width: rem-calc(1200) !default;
 /// @type Number
 $global-lineheight: 1.5 !default;
 
-/// Colors used for buttons, callouts, links, etc. There must always be a color called `primary`.
+/// Colors available by using `get-color()`.
```
ncoden: Please precise that unline the previous `get-color`, it these colors should not be used inside components.

scss/_global.scss
```diff
@@ -94,20 +109,20 @@ $global-text-direction: ltr !default;
 /// @type Boolean
 $global-flexbox: false !default;
 
-@if not map-has-key($foundation-palette, primary) {
-  @error 'In $foundation-palette, you must have a color named "primary".';
+@if not map-has-key($foundation-states, primary) {
+  @error 'In $foundation-states, you must have a color named "primary".';
```
ncoden: `In $foundation-states, you must have a state color named "primary".`

scss/settings/_settings.scss
```diff
@@ -49,18 +49,30 @@
 $global-font-size: 100%;
 $global-width: rem-calc(1200);
 $global-lineheight: 1.5;
-$foundation-palette: (
-  primary: #1779ba,
-  secondary: #767676,
-  success: #3adb76,
-  warning: #ffae00,
-  alert: #cc4b37,
+$foundation-colors: (
```
ncoden: `scss/settings/_settings.scss` is automatically generated. You do not have to update it.

scss/util/_color.scss
```diff
@@ -92,35 +92,49 @@
...
   }
 }
 
-/// Transfers the colors in the `$foundation-palette`map into variables, such as `$primary-color` and `$secondary-color`. Call this mixin below the Global section of your settings file to properly migrate your codebase.
+/// Get color state from foundation-states
+///
+/// @param {key} state key from foundation-states
+///
+/// @returns {Color} color from foundation-states
+@function get-state($key) {
```
ncoden: You can use `@function map-safe-get()` now.

scss/_global.scss
natewiebe13: @ncoden, can you elaborate on this?

scss/_global.scss
natewiebe13: I agree with moving the states as well.

scss/util/_color.scss
```diff
@@ -114,7 +114,7 @@ $foundation-colors: (
 /// @returns {Color} color from foundation-colors
 @function get-color($key) {
   @if map-has-key($foundation-colors, $key) {
-    @return map-get($foundation-colors, $key);
+    @return map-safe-get($foundation-colors, $key);
```
ncoden: The point of using `map-safe-get` is to not having to check if the key exist or throwing an error. `get-color($key)` is an alias for `map-safe-get($foundation-colors, $key)`.

scss/_global.scss
natewiebe13: Actually, if we were to move this, the variables would need to be declared in the middle of the file. `_global.scss` may be the best place for now.

scss/_global.scss
ncoden: I mean: Please precise that unlike the previous get-color, these colors (`red`, `blue`) should not be used inside components, because component should be customizable and rely on any color.

natewiebe13:
@ncoden, the builds are failing when calling `red($color)` expecting `$color` to be a color. If I debug `$white` during testing, it echos `get-color('white')` instead of `#fefefe`. Do you have any ideas why this is happening?

scss/_global.scss
ncoden: > the variables would need to be declared in the middle of the file

Why ?

natewiebe13: The state color assignments are dependent on `get-color` which is declared further down the file.

ncoden:
@natewiebe13 In Sass, when a called function is not found, it is considered as text. This is the way Sass prevent to execute CSS functions like `rgb()`...

_I agree, that's dirty_.

So it means you called `get-color('white')` without declaring it.

scss/util/_color.scss
```diff
@@ -4,7 +4,7 @@
 
 @import 'math';
 
-/// Colors available by using `get-color()`.
+/// Colors available by using `get-color()`. These should not be used directly within components, variables should be used instead to make components customizable.
```
ncoden: `get-state() and settings should be used instead to make components customizable.`

ncoden:
Could you add unit tests ?

natewiebe13:
@ncoden, what parts would you like unit tests for? (What should be tested)

ncoden:
At least for every function you added/modified. So `get-color` and `get-state`.

test/sass/_color.scss
```diff
@@ -45,4 +45,20 @@
     @include assert-equal($test, $expect,
       'Lightens a dark color');
   }
+
+  @include test('Get Color (Blue) [function]') {
+    $test: get-color('blue');
+    $expect: #1779ba;
```
ncoden: We do not expect _always `#1779ba`_, but only _the color for blue from the color map_. So please set `$foundation-color: (blue: #1779ba);`. Same for `get-state`.

test/sass/_color.scss
```diff
@@ -47,16 +47,19 @@
...
 
     @include assert-equal($test, $expect,
       'Returns the color for blue from the color map');
   }
 
   @include test('Get State (Primary) [function]') {
-    $test: get-state(primary);
     $expect: #1779ba;
+    $foundation-colors: ('blue': $expect);
+    $foundation-states: (primary: get-color('blue'));
```
ncoden: Do not rely on other function within tests. You can simply use `$expect` instead of `get-color('blue')` here.

IamManchanda:
Tagging @DaSchTour for his views, reviews and #9646 ......

DaSchTour:
Well. Now I would just add one additional feature, that allows activating states per component without redeclaration of color and it would be nearly exact what I proposed.

scss/util/_color.scss
```diff
@@ -92,35 +107,39 @@
...
-  }
+  @return map-safe-get($foundation-colors, $key);
+}
+
+/// Get color state from foundation-states
+///
+/// @param {key} state key from foundation-states
+///
+/// @returns {Color} color from foundation-states
+@function get-state($key) {
```
DaSchTour: I would suggest to make this internal `-zf-get-state` and wrap generation of states for components. Otherwise I'm sure this will cause a lot of confusion why `get-color(green)` is equal to `get-state(primary)` and why both returns a color. Or it should be `get-state-color` to be sure, that nobody expects something else than a color.

scss/_global.scss
```diff
@@ -20,35 +20,35 @@ $global-width: rem-calc(1200) !default;
...
-/// Colors used for buttons, callouts, links, etc. There must always be a color called `primary`.
+/// Color states used for buttons, callouts, links, etc. There must always be a color called `primary`.
 /// @type Map
-$foundation-palette: (
-  primary: #1779ba,
-  secondary: #767676,
-  success: #3adb76,
-  warning: #ffae00,
-  alert: #cc4b37,
+$foundation-states: (
```
DaSchTour: I would suggest that we don't have to use get-color here but to enforce that colors used here need to be declared in `$foundation-colors`this way it is possible to introduce colors "from the side" which may cause confusion, because generated elements suddenly get colors not globally defined.

natewiebe13: Can you elaborate here? Do you mean instead of using the get-color function, we would specify the key that would be used for the `get-color()` function, then return the value using `get-color()` while calling `get-state()`?

DaSchTour: Exactly

DaSchTour:
@IamManchanda I can create a PR after this was merged ;)

IamManchanda:
I know ..... Was just reminding of both ... Thanks bro

IamManchanda:
@natewiebe13 .... just pulled it .... and there is a error and thus, docs doesn't work

```bash
Message: scss/util/_color.scss
Error: Undefined variable: "$foundation-states"
```

natewiebe13:
@IamManchanda, that's from `_setttings.scss` which is automatically generated. Plus any changes to the docs will happen after functionality is finalized.

scss/util/_color.scss
```diff
@@ -92,35 +107,39 @@
...
 
-/// Get color from foundation-palette
+/// Get color from foundation-colors
 ///
-/// @param {key} color key from foundation-palette
+/// @param {key} color key from foundation-colors
 ///
-/// @returns {Color} color from foundation-palette
+/// @returns {Color} color from foundation-colors
 @function get-color($key) {
```
DaSchTour: What also needs to be considered after looking on my own code, that this is a breaking change! So there needs to be a migration warning in Release Notes for 6.4
Or maybe we should make this backward compatible by doing
```SCSS
@if map-has-key($foundation-colors, $key) {
  @return map-get($foundation-colors, $key);
} @else if map-has-key($foundation-palette, $key) {
  @warn 'returning color from $foundation-palette which is deprecated behavior and will be remove in future versions';
  @return map-get($foundation-palette, $key);
} @else {
  @error '...'
}

**Comment by kball** [comment]
@DaSchTour @IamManchanda @ncoden @natewiebe13 what do you consider the state of this PR? Is this ready to go?

**Comment by DaSchTour** [comment]
@kball I still don't like the seperation into state-colors and color-palette with two different functions for access. That's a point that will lead to a lot of confusion. I made some suggestions in the original issue #9671 
From my point of view there should be only one place (map) with defined colors and only one function for access.

**Comment by natewiebe13** [comment]
@kball This is only missing doc changes. I was waiting for the changes to be satisfactory before updating the docs.

@DaSchTour The issue is there should be a distinction between colors that are available, and what colors are assigned to states. If something should be green, `get-color('green')` should be used. If something should be the success color, `get-state-color('success')` should be used. By default, `success` is green, but that doesn't mean that the success color may later change to something else, but green usages should remain green. States should be forced to get their color from the list of available colors. If `success` is using `green`, and the shade of green gets altered, it usually affects both. It's easier to keep track of what colors are on a site if they are defined in one map. As to what color has which meaning, that should be completely separate. Thoughts?

**Comment by DaSchTour** [comment]
@natewiebe13 I this view on colors extremely depends on type of project you work on. And as I think I already mentioned in #9671 the concept of state colors feels somehow strange and doesn't fit into many projects. For me state colors only make sense for callouts and forms. On the other side I don't need state colors for buttons because I have primary and secondary buttons, but these are not states. On the other hand I realize that tables have no state colors (just remembered bootstrap had them).
The other point I'm always concerned about is how people might use tools ensure maintainability. On a growing and long living project when you have no Single-Point-Of-Truth developers will randomly choose from the available options. Until you reach the point, that `get-color('success')` and `get-state-color('success')` and also `get-color('primary')` and `get-state-color('primary')` both are defined separately and both deliver different values and nobody know when to use which one.

**Comment by IamManchanda** [comment]
> If something should be green, get-color('green') should be used. If something should be the success color, get-state-color('success') should be used. By default, success is green, but that doesn't mean that the success color may later change to something else. Green usages should remain green. States should be forced to get their color from the list of available colors. If success is using green, and the shade of green gets altered, it is usually intended to affect both usages of green, and the state that uses that color. It's easier to keep track of what colors are on a site if they are defined in one map. As to what color has which meaning, that should be completely separate. Thoughts?

+1 @natewiebe13 i tend to concur with you! Simple but cool! Moreover can be easily transferred the idea into the docs! This is what i am looking for ....

**Comment by natewiebe13** [comment]
@DaSchTour In regards to using states with buttons, we use the alert state for delete buttons, success state for save buttons, etc. This conveys the intention, not the color. Later on you mention `get-color('success')`. Since success isn't a color, it should give an error. (Currently it would give a deprecation notice). Usage is fairly straight forward, I don't see where the confusion could come from. The color map is a list of colors (red, yellow, red, white, black, etc). States are the ability to give meaning to a color, and states **must** get the color from the color map. Again, if something should be green, `get-color()` should be used. If something should convey the success state, `get-state-color()` should be used. Can you maybe provide a specific example of when it could be confusing?

File: scss/components/_progress-bar.scss [hunk 13]
```diff
@@ -43,7 +43,9 @@
   .progress {
     @include progress-container;
 
-    @each $name, $color in $foundation-palette {
+    @each $name, $color-name in $foundation-states {
```
DaSchTour: we need to add `$progess-satets` to make this work like all other components using states.

DaSchTour:
@natewiebe13 

Well, wrote this before I realized, that we already removed the possibility to directly add colors to states. So can skip this.

> it will give a deprecation warning until somebody defines it in color-palette. I'm talking about usage inside projects where color palette is modified over several releases, iterations and between several generations of developers.
> And there is also so much wrong about named colors, they are not working when trying to use them for theming. The most important in many project is the primary/brand color which is not a state and also not a normal color. So where do I put my primary color? I want use my primary color for CTA which is different around several themes that are created in my project. So the main usage of current `get-color` is `get-color('primary')`. The current approach may fit for some projects but will not fit for other projects. It depends on the approach, the customer, the designers and the developers.
> 
> It will get confusing in the moment when you start the project and `get-color('green')` and `get-color('success')` are the same. While developing it might often not be clear if UX and Design made an element green because it should be green or if it's green because it's success state. So several releases later there are a lot of calls for both functions. In this moment design decides that success and green have to be different. So in best case there is a `success-green` added to the color palette which than is called through `get-color('success-green')` in the state palette. Further releases later somebody directly adds another green color directly to state palette and others added `get-color('success-green')` in some other places and just a year of development and you have 3 different green colors defined.

The point is to be careful about adding complexity and possibilities to a framework and which consequences this might have in every day development. In worst cases somebody changes the color of blue to a red color to achieve that primary/brand buttons are red.

Well. While writing this I also ask my self why we need different functions. In fact states are only named aliases for named colors. So we could reduce complexity a lot but using one function for accessing colors through their name and their alias/state (I really don't like the name here, primary isn't a state for me).

scss/components/_progress-bar.scss
natewiebe13: Same as callouts. I'll add these in.

natewiebe13:
@DaSchTour, I agree with avoiding unneeded complexity. We've been modifying Foundation to work with colors similar to this for about a year in a firm of ~40 people. We have found that it improves clarity in determining how developers are interpreting designs. Whether setting the color because of the color used, or identifying patterns based on intent. I'm sure this could be misused (like other things in any framework), but something blatant like making a color red when it's named blue is something that can be caught during code review/training. I don't think this change is overly complex, but perhaps I'm mistaken.

DaSchTour:
@natewiebe13 well, while talking about it and taking a deeper look on this I get that this is getting in the right direction. Still I would suggest a last change, that we reduce to have only `get-color()` which will first deliver from `$foundation-colors` and than from `$foundation-states`. For internal usage for example while getting colors for states we might avoid circular reference by adding an additional parameter to `get-color()` to just get colors from color map.

IamManchanda:
Hii Guys, 
I have to say i am loving it where its going but i have a suggestion

While creating [Building Blocks](http://foundation.zurb.com/building-blocks/) if there is anything i have noticed is that there is a social button/icon for every second ( if not every single ) component. 
Moreover if you look broader, every single site that is serious about there site have a social buttons/icons/etc  

Currently within the building blocks same variables are getting used again and again which is a bloat ...so i would suggest defining these variables within the framework .... [See Gist](https://gist.github.com/IamManchanda/f6376f8082637956e1d30409cf714956)
Yup the way i coded in the gist is not the right way and should be done like this PR .... but you get my idea??

Yup you might consider my suggestion as a bloat .... but believe me.... what's happening currently is a much bigger bloat!

What's your thoughts @kball @natewiebe13 @DaSchTour ??
Also cc / @rafibomb

kball:
I think @IamManchanda raises a potentially interesting question...

The intent of this PR if I'm understanding correctly is to create a distinct abstraction for the color palette and the choice of which states map to which colors.  Does it make sense to take that further, and allow the use of multiple palettes, allowing you to e.g. import palettes, or have different palettes for different portions of a site that are logically separated within the scss?

DaSchTour:
@kball I think this could only be achieved in an manageable way if using a map of maps. I think in this case we absolutely should have only one function to retrieve colors. On the other hand maybe there should be only one map with colors, although it might have a lot of them and than a mapping for buttons etc.
I describe a lot of different models in the original issue #9671

IamManchanda:
@DaSchTour => Are you relating with your this comment 
=> https://github.com/zurb/foundation-sites/issues/9671#issuecomment-274011251 ??

DaSchTour:
@IamManchanda yes and also the following comments. I made several suggestions.

IamManchanda:
Hmmn....
But here are palletes i am particulary interested 

1. Foundation Colors ( Can be replaced but can't be disabled ) 

2. Social Media Color Pallete ( Enabled by default )
=> https://designpieces.com/2012/12/social-media-colours-hex-and-rgb/

3. Flat Design Color Pallete ( Enabled by default )
=> https://flatuicolors.com/

4. Material Design Color Pallete ( Disabled by default )
=> https://www.materialpalette.com/colors

5. Metro UI Color Pallete ( Disabled by default )
=> https://colorlib.com/etc/metro-colors/

kball:
@joeworkman I think this may lead to a more general form of what you're proposing in emails for https://github.com/zurb/foundation-emails/pull/754

Would be interested in making sure this functionality works for the usecase of having 2 alternate palettes. Joe do you have thoughts on this approach and if something like this would achieve what you're looking for in emails? 

(cc @tdhartwick @rafibomb)

joeworkman:
@kball In my Foundation for Sites product, I actually implemented this alt color scheme as well. This is where I got the idea for Email. Its been really popular with my community of designers. 

When I update to use F6 later this year, this was on my list of things to submit a pull request for.

IamManchanda:
@DaSchTour Assigning you together with @natewiebe13 

As discussed, daniel can you modify the code within this PR ?? 
I have gone through your comments once again and i think you are going in right direction specially considering the external pallete!

```
$button-colors: (primary: blue, secondary: gray, success: green, warning: yellow, alert: red );
$callout-colors: (success: green, warning: yellow, alert: red);
$label-colors: (primary: blue, secondary: medium-gray, success: green, warning: yellow, alert: red);
```

but what is more required is external color pallete's (enabled/disabled) .... 
For *shades*, we can use sass for scale, darken and lighten .... only one color per color!
Social button is one serious need, other flat color like indigo, teal, cyan if can be added then it will be great ..... 
For building blocks i add all these colors again and again .... 
And i think same goes for end user!

kball:
@natewiebe13 @DaSchTour @IamManchanda 
Do we have clarity on what we're doing here?  It feels a little muddled to me, but I think this is a pretty cool idea and we should figure out how to move forward with it... so maybe let me see if I can state out what the situation is and where we are from trying to read through the above thread

It seems like we're trying to do a few things here

1. Separate the concepts of 'state' and 'color'.
2. Allow the possibility of multiple palettes (most common is probably a single "alternate palette")
3. Maintain ease of use for the common cases.

Are there any additional constraints I'm missing?

Looking through the conversation, it seems like keeping separate maps makes sense... what I'm not sure is if it makes sense to have the indirection from state to color palettes, especially when we start talking about alternate palettes, some arbitrary number of which may be state based. (e.g. imagine having state palette, alternate palette, social media palette, and possibly some accent colors)

I may be totally off base here, but lets walk down a thought experiment... The states matter because we use them to generate classes... so what if we instead just have an arbitrary number of palettes, and then we specify which ones create classes and which don't?

In pseudocode I'm thinking something like...
```
$brand-color-palette = (...)
$foundation-state-palette = (...)
$alternate-state-palette = (...)
$random-social-palette = (...)

$palettes-to-generate-classes = ($foundation-state-palette, $alternate-state-palette)
```

We could take this to `get-color` both by letting it accept a specific palette (e.g. `get-color('primary', $alternate-state-palette)) but also by having a precedence list for the basic lookup
```
$color-precedence = ($foundation-state-palette, $brand-color-palette, $random-social-palette)
```

Such that when `get-color` is invoked without a palette it looks through the palettes in that order to find a matching color.

Thoughts?

IamManchanda:
Any updates Guys ... I think we should surely get this in for 6.5 ??

Poke @natewiebe13 @DaSchTour 
Also tag @ncoden (as he is coming back slowly :wink: )

ncoden:
> 1. Separate the concepts of 'state' and 'color'.
> 2. Allow the possibility of multiple palettes (most common is probably a single "alternate palette")
> 3. Maintain ease of use for the common cases.

I have the feeling that we are forgetting and refinding/repeating a lot if ideas we already discussed in the issue: https://github.com/zurb/foundation-sites/issues/9671.

Please take a look at the piece of code we discussed before, they are already following the 3 points @kball given.

DaSchTour:
I think that point 1 and 2 are somehow conflicting. I think we should abandon the concept of state-colors as it always didn't feel right. I think the best would be to have one single point to define colors, a color palette including such things as 'facebook-blue' and than mappings to components. Somethink that tells the button component that it should use facebook-blue for a button with the class facebook.

ncoden:
> Somethink that tells the button component that it should use facebook-blue for a button with the class facebook.

You miswritten "custom component state". ;)

DaSchTour:
@ncoden well the problem with the term state is, that state is often something that changes and is part of a defined flow which describes the change between different states. I don't see how primary is a state. It's more of a characteristic that is set once and that is constant.

ncoden:
@DaSchTour I agree with you on this point. In a clear CSS architecture, a `state` is not a `modifier`. The first is an event, part of a different context, so should have the priority on the second.

_Generally in BEM+SMACSS, we use the declaration order to give a modifier (`.component--modifier`) the priority over the default behavior (`.component`), and an additional class to give the state (`.component.state`) the priority over the whole standard context. You already know that._

In our current ~archaic~ classic CSS architecture, "modifiers" are implemented with an additional class. So we have two solution for our "states": using two classes (`component.is.state`), or using one class and the declaration order and pray for it does not cause too much trouble. So we handle modifiers the same way we handle states. The only difference is that we must generate modifier classes before state ones.

Beside that, we also need a set of colors we can use (or not) freely inside components, without generating new dedicated classes.

So finally, we need three maps for each component (modifier, state, color), inherited from three maps at the global level (see my code examples in #9671), and an unique set of mixins the manage all of this easily without code duplication !

Also, I don't see how @kball's 1st and 2nd points are conflicting. We can add palettes globally or locally (from the config, function parameters...) by merging them into three different kind of internal palette in which we search for colors and states.

kball:
bump @ncoden @DaSchTour @natewiebe13 @IamManchanda anyone interested in picking this back up and carrying it across the finish line?

DaSchTour:
@kball I really have no idea which direction this should be carried. There are so many ideas for this topic and all of them somehow feel bad. While working on my current project I also see, that the concept of colors named with color names (blue, red and so on) as used in `$foundation-colors` is absolutely bad. I guess many pages will need colors named like (brand, secondary, highlight, alert) and maybe different shades of gray from white to black. So maybe the current concept doesn't need that many changes (or even any).

kball:
Closing this as old WIP

Issue #9695, opened 2017-01-24
[Off-canvas] toggle not working after window resize with reveal-for-large


Zausenec:
Load on large or medium size, then resize to small
or
load on small, then resize to medium

```
<div id="offCanvas" class="off-canvas position-left reveal-for-large" data-off-canvas></div>
<div class="off-canvas-content" data-off-canvas-content>
    <button class="button" data-toggle="offCanvas">Toggle</button>
</div>
```

brettsmason:
@Zausenec Thanks for reporting this issue. I have setup [a CodePen](http://codepen.io/brettsmason/pen/rjzQoQ) as you described and can confirm the bug.

@kball Any idea why this might be happening?

kball:
Not sure, but I can dig into it

PR #9700, opened 2017-01-25
Fix double event handler bug in new offcanvas

kball:
Fixes https://github.com/zurb/foundation-sites/issues/9695

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -132,7 +132,7 @@ class OffCanvas {
     } else {
       this.isRevealed = false;
       this.$element.attr('aria-hidden', 'true');
-      this.$element.on({
+      this.$element.off('open.zf.trigger toggle.zf.trigger').on({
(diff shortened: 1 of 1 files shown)


js/foundation.offcanvas.js
```diff
@@ -132,7 +132,7 @@ class OffCanvas {
     } else {
       this.isRevealed = false;
       this.$element.attr('aria-hidden', 'true');
-      this.$element.on({
+      this.$element.off('open.zf.trigger toggle.zf.trigger').on({
```
IamManchanda: Why?

kball: Because currently, if you trigger this branch in the else twice (for example by resizing across 2 sizes that both are not revealed) the trigger gets double applied, and counteracts itself, resulting in the behavior from #9695 where the 2 event handlers cancel themselves out

IamManchanda (approved):
Looks good to me!!

PR #9704, opened 2017-01-26
Grammar correction within Flexbox page

IamManchanda:
> Otherwise, we'd have two CSS classes with the same name, but different functionality.

This was so illogical as for obvious reasons **you just can't have two CSS classes with same name** and we can't say => we'd have.

So, now its 

> As we can't have two CSS classes with the same name, thus we are using different terms.

@kball - for merge or review

### DIFF
--- a/docs/pages/flexbox.md
+++ b/docs/pages/flexbox.md
@@ -130,7 +130,7 @@ Stretch alignment is the default. To set parent alignment, use these classes:
 - `.align-stretch`
 
 <div class="primary callout">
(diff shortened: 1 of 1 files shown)


rafibomb:
Thanks!

IamManchanda:
Also please merge this into develop ..... :)

------------------------------------------------------------------------------------------------------------------------------------

@rafibomb Are you the same guy from zurb videos?
Nice to meet you sir .... You post great tutorials!

Issue #9696, opened 2017-01-24
Off-Canvas v6.3 does not scroll on mobile.


jinch:
If your off-canvas menu is longer than the visible area on mobile, you can't scroll to see the rest of the content. Scrolling only seams to work on desktop.

[EXAMPLE HERE](http://inchdesigns.com/projects/OFFCANVAS/) // tested in Android Chrome as well as iOS Safari. In both cases you can't scroll to see the rest of the menu when the Off-Canvas menu is open.

brettsmason:
@jinch Thanks for reporting this issue.

I have looked into it myself and can confirm it appears to be an issue with `data-content-scroll`.
See [this CodePen](http://codepen.io/brettsmason/pen/PWKZEe) to demonstrate the bug.

I'll look into this and see if I can find a way to fix it.

kball:
Submitted PR to address this: #9707.  @brettsmason would appreciate your testing

kball:
This was fixed in #9707, should be working in 6.3.1

m-lamarre:
I have a question about this issue @kball 🤓 
I am currently using the `foundation-rails (6.4.1.3)` gem on my Rails project, and I am still experiencing this issue on Android mobile devices still.

If I set `data-content-scroll=''` to false, I am not able to scroll in `offCanvasMenu` on my android device at all. When set to `true` I am able to scroll in the OffCanvasMenu but then the background scrolling is visible from the hamburger menu.

I see that your fix was for iOS `Enable scrolling within offcanvas panel on iOS` 
Has this not been fixed for Android? 🤔

m-lamarre:
Alright, well I managed to find this link:
[Disable background scrolling when mobile menu is active](https://forum.webflow.com/t/disable-background-scrolling-when-mobile-menu-is-active-opened/38305)

and I just modified the code slightly to work for Foundation 6.
`data-content-scroll='true'` on the `#offCanvasMenu`

```
<script>
  $('.off-canvas').on('opened.zf.offcanvas', function(e) {
    e.preventDefault();
    $('body').css('overflow', 'hidden');
  });

  $('.off-canvas').on('closed.zf.offcanvas', function(e) {
    e.preventDefault();
    $('body').css('overflow', 'auto');
  });
</script>
```

jinch:
[...]

PR #9707, opened 2017-01-26
Enable scrolling within offcanvas panel on iOS even when body scrolli…

kball:
Fix issue https://github.com/zurb/foundation-sites/issues/9696

This turned out to be quite tricky, but on testing appears to work.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -147,7 +147,41 @@ class OffCanvas {
    * @private
    */
   _stopScrolling(event) {
-  	return false;
+    return false;
+  }
+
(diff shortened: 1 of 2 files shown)


brettsmason:
@kball Thanks so much for looking at this!

I've tested it and it works pretty much spot on. However I noticed if you scroll down to the bottom of an off canvas panel that's longer than the window the address bar goes out of view. Then you cant access it until you scroll to the very top of the panel.

Not sure if there's a way to stop that from happening?

kball:
Hmm dunno but I'll try to figure it out some today

kball:
@brettsmason wasn't able to reproduce the issue you highlighted; going to merge this as it is b/c it is a bugfix we'd like to include in 6.3.1, happy to revisit for 6.3.2 and/or 6.4 if you want to provide a repro case.

PR #9711, opened 2017-01-27
Make reset inherit font-family instead of settings sans-serif

### LINKED ISSUE #9668: normalize.scss is to specific for Foundation
We are using the 6.3.0 SCSS-Version of foundation-sites and ran into an issue, where `_normalize.scss` is to specific for Foundation!

If you set for example `$body-font-family: 'Times New Roman', serif;` one would assume, that the font on a `button` is in serif. Sadly this is not the case, because `_normalize.scss` explicitly sets `button, input, optgroup, select, textarea` to it's default-value which is `sans-serif`. As foundation-sites doesn't reset these Elements, this results in `sans-serif`-Buttons.

Set `$body-font-family: 'Times New Roman', serif;`

```html
<!-- Anchors (links) -->
<a href="about.html" class="button">Learn More</a>
<a href="#features" class="button">View All Features</a>

<!-- Buttons (actions) -->
<button type="button" class="success button">Save</button>
<button type="button" class="alert button">Delete</button>
```
Results in:
a-vs-button
But the Save- and Delete-Button also shoudl be in Times New Roman/serif.

#### Fix/Workaround:
You can workaround this by setting `$base-font-family` which is used by _normalize.scss to `$body-font-family;` after including your `_settings.scss` and before `foundation.scss`.

Maybe that should be the default-behaviour or maybe Foundation should reset the elements to it's desired state, if they are overwritten by `_normalize.scss`. Otherwise it could be a pitfall for some people.

natewiebe13:
Goal: Make reset less opinionated. Fixes #9668
- Reset was setting the font-family of buttons to be sans-serif, instead of inheriting. This was causing buttons to not use the font specified in the body.

### DIFF
--- a/_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss
+++ b/_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss
@@ -424,7 +424,7 @@
     optgroup,
     select,
     textarea {
-      font-family: $base-font-family; /* 1 */
+      font-family: inherit; /* 1 */
       font-size: 100%; /* 1 */
       @if $normalize-vertical-rhythm {
         line-height: ($base-line-height / $base-font-size) * 1em; /* 1 */


kball:
Makes sense to me. Any SCSS focused @zurb/yetinauts want to chime in?

IamManchanda:
+1 from my side.

rafibomb:
👍

IamManchanda (approved):
Merge?

natewiebe13:
@kball, I noticed 6.1.3-rc1 is out. Can we get this merged for 6.1.3?

kball:
Merging now, will be in 6.3.2

PR #9712, opened 2017-01-27
[Fixes #9705] Add menu-centered to flexbox mode.

### LINKED ISSUE #9705: flexbox menu alignment issue

1. menu-centered class causes list items to stack instead of center.

#### What should happen:
list items should horizontally center

#### What happened instead:
list items stacked

#### Test case:

http://codepen.io/lbeyers/pen/vgezgK

IamManchanda:
Implementation for #9705 

This adds `menu-centered` class to flexbox mode to keep the similarity beetween both modes. 
Can confirm you that i have done the testing ( both modes, flexbox and non flexbox )

Thanks for posting the issue: @LaurelBCodin 
but it was not so a bug... but instead a missing css class on flexbox mode.

--------------------------------------------------------------------------------------------------------------

@kball for merge or review.
Also please close the issue on merge as i forgot to add it on commit message.

### DIFF
--- a/docs/pages/menu.md
+++ b/docs/pages/menu.md
@@ -52,7 +52,7 @@ By default, each item in the menu aligns to the left. They can also be aligned t
 To align items in the middle, add a wrapping element with the class `.menu-centered`.
 
 <div class="primary callout">
-  <p>If you're using <a href="flexbox.html">Flexbox mode</a>, you don't need the wrapper class. Instead, you can just add the class <code>.align-center</code> to the menu.</p>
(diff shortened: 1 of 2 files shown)


IamManchanda:
Awaiting Reply/review/merge? @kball 

This is useful because 
- It doesn't break thinks while migrating in between flexbox & non flexbox mode!
- Also because users ( just like @LaurelBCodin ) expects it to be center aligned on both modes for `.menu-centered`

kball:
LGTM

LaurelBCodin:
thanks so much guys!

rafibomb:
I like the idea of including codepen examples in the docs!

IamManchanda:
Yeah you're right .... we can do that throughout foundation infact .... But in that case we should do it as target `_blank` .... what do you guys think ?? @rafibomb @kball

Btw this was just one off where I had no other choice but to include codepen!

PR #9723, opened 2017-01-29
[Advancement] Responsive variation for float classes

IamManchanda:
Hello there once again!!

=> http://foundation.zurb.com/sites/docs/float-classes.html

I Would like to add **responsive variations** to float classes,
In short, would like to add 

```scss
.medium-float-left { }
.medium-float-right { }
.medium-float-center { }
.medium-clearfix { /* Before and after */ }

.large-float-left { }
.large-float-right { }
.large-float-center { }
.large-clearfix { /* Before and after */ }
```

**Reasons for which me straight pushed it instead of posting an issue**
- It doesn't make the css bulky, not too much of the final generated css.
- Responsive helpers for things like floats are very helpful for responsive web design.
- This should be added as **default** instead of things like non-default mixins !!!

### DIFF
--- a/scss/components/_float.scss
+++ b/scss/components/_float.scss
@@ -6,22 +6,45 @@
 /// @group float
 ////
 
+@mixin float-center {
+  display: block;
+  margin-right: auto;
+  margin-left: auto;
+}
+
 @mixin foundation-float-classes {
-  .float-left {
-    float: left !important;
-  }
-
-  .float-right {
-    float: right !important;
-  }
-
-  .float-center {
-    display: block;
-    margin-right: auto;
-    margin-left: auto;
-  }
-
-  .clearfix {
-    @include clearfix;
+  @each $size in $breakpoint-classes {
+    @include breakpoint($size) {
+      // Float left and right
+      @each $float in (left, right) {
+        @if $size != $-zf-zero-breakpoint {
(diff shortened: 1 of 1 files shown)


kball:
Thoughts @brettsmason @andycochran @ncoden?

Also going to tag in @tdhartwick and @jnemeth as ya'all use prototyping classes like this much more than I do...

brettsmason:
My first reaction is why the use of `!important`? I thought we were generally trying to remove this where possible. I'll try and have a proper look through when I get some time, but thought that was worth mentioning.

IamManchanda:
`!important` was already there so that's why didn't removed it!

This is what i got from docs,

> Float Left/Right classes use `!important` to ensure they aren't overriden by more specific selectors. This framework conscientiously avoids using !important declarations. This is one of the few components that does.

I will be very happy to remove `!important` !!! @brettsmason

brettsmason:
Sorry my mistake - you are correct it was already there. Ignore me 😄 
Looks OK to me, but probably a good idea to get other opinions too.

IamManchanda:
@brettsmason In only one case i will forgive you for this and that's,

> If You will give me good sass tutorials/articles for reference :wink: :smile:

IamManchanda:
Closing ( Put on hold ), thanks to #9738 ( merged there also )

Issue #9729, opened 2017-01-30
Flexgrid not resetting nested row using 'expanded' class


chrismanderson:
#### How to reproduce this bug:

When using the flexgrid and a row with an 'expanded' class, and nesting a row below that; the nested row is not reset to contain to the grid.

    <div class="row expanded bg-pink">
      
      <div class="large-12 columns">
        <div class="row bg-purple">
          
          <div class="small-12 large-expand columns">One</div>
          <div class="small-12 large-expand columns">Two</div>
        </div>
      </div>
    </div>

This error does not occur when using the normal grid.

Result is: 
![bad](https://cl.ly/1R3j041V1Y08/Image%202017-01-30%20at%204.07.43%20PM.png)

When I should see:
![good](https://cl.ly/410M1f2j3V41/Image%202017-01-30%20at%204.08.16%20PM.png)
#### Test case:

[Error case, flexgrid](https://codepen.io/chrismanderson/pen/apEQxY)
[Successful case, non-flexgrid](https://codepen.io/chrismanderson/pen/ygpQrM)

Flexgrid version 6.2.4.

kball:
Fixed in #9729.

PR #9730, opened 2017-01-31
Fix for #9729 : Flexgrid not resetting nested row using 'expanded' class

IllusionPerdu:
Fix for #9729


IamManchanda (approved):
LGTM 
Updated codepen with this PR => https://codepen.io/IamManchanda/full/zNRYpw/  
cc / @kball for review!

Also, close #9729 on merge!

Issue #9533, opened 2016-12-20
[Card] Card Features and Improvements


rafibomb:
First off, thanks to @DaSchTour and @brettsmason for getting cards into Foundation! We're compiling feature requests for the cards (mostly flex-mode) here and outstanding issues.

### Features
- [ ] Complete Flexbox support of card in flex-mode (`.card-divider`, `card-section`)
- [ ] Add Flexbox documentation section on cards page
  - [ ] Demo equal height cards with flexbox and helper classes
  - [ ] Demo `.flex-dir-row` for horizontal card with responsive sizes

I'm sure you have suggestions for more. Please add them in the comments so we can get documentation updates into 6.3.1 and features into 6.4. Thanks!

rafibomb:
'masonry" like wrapper example:
<img width="1248" alt="screen shot 2016-12-19 at 8 10 50 pm" src="https://cloud.githubusercontent.com/assets/5652085/21338060/85e2e4d2-c627-11e6-9f9b-b1c16516dbbe.png">

IamManchanda:
http://codepen.io/IamManchanda/pen/eBoYpM

paxperscientiam:
@rafibomb Is your "masonry" example just mock-up?

dlewand691:
Would recommend adding `flex-grow: 1` to `.card` to account for short content. See screenshot for example of error.

![screen shot 2017-01-07 at 7 23 08 pm](https://cloud.githubusercontent.com/assets/1159913/21746781/cde98be6-d50e-11e6-93fb-1623315bf7ec.png)

rafibomb:
@dlewand691 Thanks for the suggestion! Not seeing that this is needed here: http://codepen.io/rafibomb/pen/JEXzGv?editors=1000

rafibomb:
@dlewand691 Got it fixed, turns out related to having to flex parent. See #9743 - Thanks!

DanielRuf:
Any updates?

PR #9743, opened 2017-02-02
Adds flex to card-divider and fixes card width for narrow content

rafibomb:
- Adds display: flex; to .card-divider so that flex-utilities can be applied (ref #9533)
- Adds flex-grow: 1; to .card to fix issue with card width collapsing when in display: flex parent. (ref #9533)

Also adds test case for both

### DIFF
--- a/scss/components/_card.scss
+++ b/scss/components/_card.scss
@@ -56,6 +56,7 @@ $card-margin: $global-margin !default;
   @if $global-flexbox {
     display: flex;
     flex-direction: column;
+    flex-grow: 1;
   }
 
   margin-bottom: $margin;
@@ -81,6 +82,7 @@ $card-margin: $global-margin !default;
 ) {
   @if $global-flexbox {
     flex: 0 1 auto;
+    display: flex;
   }
 
   padding: $padding;
--- a/test/visual/card/card.html
+++ b/test/visual/card/card.html
@@ -1,0 +1,44 @@
+<!doctype html>
+<!--[if IE 9]><html class="lt-ie10" lang="en" > <![endif]-->
+<html class="no-js" lang="en" dir="ltr">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
+    <title>Foundation for Sites Testing</title>
+    <link href="../assets/css/foundation-flex.css" rel="stylesheet" />
+  </head>
+  <body>
+    <div class="row small-up-2 medium-up-3">
+      <div class="column flex-container">
+        <div class="card">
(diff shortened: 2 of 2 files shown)


scss/components/_card.scss
```diff
@@ -56,6 +56,7 @@ $card-margin: $global-margin !default;
   @if $global-flexbox {
     display: flex;
     flex-direction: column;
+    flex-grow: 1;
```
kball: I'm confused about this... Why are we pre-assuming how a card should behave relative to its parent? AFAIK we don't have any knowledge of whether the card will be a flex child or not

scss/components/_card.scss
```diff
@@ -81,6 +82,7 @@ $card-margin: $global-margin !default;
 ) {
   @if $global-flexbox {
     flex: 0 1 auto;
+    display: flex;
```
kball: Why are we pre-assuming a card divider will be a flex parent?

kball (commented):
These changes seem to make assumptions about use; can you explain why you believe those usecases are common enough to be default?

kball:
Tagging in @DaSchTour and @brettsmason as well to review, as y'all are more familiar with cards & flex cards than I am.

kball:
I see this is targeted for 6.3.1 as a bugfix, can you explain a little more about what bugs it is fixing? The referred issue is a feature list

scss/components/_card.scss
rafibomb: If the card is not inside a flex parent, it will naturally grow because the default flex-basis is auto. When it is placed inside a flex-parent, it will shrink. This is inconsistent behavior and the largest use case is to have the card in some sort of grid or container. To curb the inconsistency, setting flex-basis to 100% (instead of flex-grow: 1) seems like the best solution.

scss/components/_card.scss
rafibomb: Since this is a shrink container, it is the best place to add actions, images, or other pieces that will be typically distributed horizontally. Making this display: flex seems natural to apply flex helpers like `.align-justify` and has no adverse affects on elements inside (images will stretch if smaller than container height but `align-items` or `.align-middle` solves it).
screen shot 2017-02-03 at 3 04 30 pm

rafibomb:
This can be considered a feature so deferring to 6.4 is fine. Waiting on feedback from others!

kball:
Going to go ahead and merge this... are there any Migration notes required?

rafibomb:
@kball Not migration notes though updating the docs to point out this feature would be great.

rafibomb:
Opened https://github.com/zurb/foundation-sites/issues/9962

PR #9744, opened 2017-02-02
Adds card columns wrapper

rafibomb:
Adds card columns wrapper to flow cards in vertical columns. 

Reference #9533

https://cloud.githubusercontent.com/assets/5652085/21338060/85e2e4d2-c627-11e6-9f9b-b1c16516dbbe.png

### DIFF
--- a/docs/pages/card.md
+++ b/docs/pages/card.md
@@ -178,3 +178,78 @@ You can either set the width of cards with custom css or add them into the Found
   </div>
 </div>
 ```
+
+## Card columns
+
+Cards can be organized into <a href="http://masonry.desandro.com/" target="_blank">Masonry-like</a> columns with just CSS by wrapping them in .card-columns. Cards are built with <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns/Using_multi-column_layouts" target="_blank">CSS column</a> properties instead of flexbox for easier alignment. Cards are ordered from top to bottom and left to right.
+
+Heads up! Your mileage with card columns may vary. To prevent cards breaking across columns, we must set them to `display: inline-block` as `column-break-inside: avoid` isn’t a bulletproof solution yet.
+
+```html_example
+<div class="card-columns">
+  <div class="card">
+    <img src="http://placehold.it/350x250">
+    <div class="card-section">
+      <h4>Card title that wraps to a new line</h4>
+      <p>This is a longer card with supporting text below as a natural lead-in to additional content. This content is a little bit longer.</p>
+    </div>
+  </div>
+  <div class="card">
+    <div class="card-section">
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere erat a ante.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+  <div class="card">
+    <img src="http://placehold.it/350x250">
+    <div class="card-section">
+      <h4>Card title</h4>
+      <p>This card has supporting text below as a natural lead-in to additional content.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+  <div class="card text-center" style="color: #fff; background: dodgerblue;">
+    <div class="card-section">
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere erat.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+  <div class="card text-center">
+    <div class="card-section">
+      <h4>Card title.</h4>
+      <p>This card has supporting text below as a natural lead-in to additional content.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+  <div class="card">
+    <img src="http://placehold.it/300x350">
+  </div>
+  <div class="card">
+    <div class="card-section">
+      <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer posuere erat a ante.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+  <div class="card">
+    <img src="http://placehold.it/350x250">
+    <div class="card-section">
+      <h4>Card title</h4>
+      <p>This card has supporting text below as a natural lead-in to additional content.</p>
+      <p><small>Updated: 10 min ago</small></p>
+    </div>
+  </div>
+</div>
+```
+
+Card columns can also be extended and customized with some additional code. Shown below is an extension of the `.card-columns` class using the same CSS we use—CSS columns— to generate a set of responsive tiers for changing the number of columns.
+
+```scss
+.card-columns {
+  @include breakpoint(large) {
+    column-count: 4;
+  }
+  @include breakpoint(xlarge) {
+    column-count: 5;
+  }
+}
+```
--- a/scss/components/_card.scss
+++ b/scss/components/_card.scss
@@ -38,13 +38,21 @@ $card-padding: $global-padding !default;
 /// @type number
 $card-margin: $global-margin !default;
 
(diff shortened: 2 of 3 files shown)


docs/pages/card.md
```diff
@@ -178,3 +178,78 @@ You can either set the width of cards with custom css or add them into the Found
   </div>
 </div>
 ```
+
+## Card columns
+
+Cards can be organized into <a href="http://masonry.desandro.com/" target="_blank">Masonry-like</a> columns with just CSS by wrapping them in .card-columns. Cards are built with <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns/Using_multi-column_layouts" target="_blank">CSS column</a> properties instead of flexbox for easier alignment. Cards are ordered from top to bottom and left to right.
+
+Heads up! Your mileage with card columns may vary. To prevent cards breaking across columns, we must set them to `display: inline-block` as `column-break-inside: avoid` isn’t a bulletproof solution yet.
```
kball: Can you clarify support levels? What browsers does this work for?

docs/pages/card.md
```diff
@@ -178,3 +178,78 @@ You can either set the width of cards with custom css or add them into the Found
   </div>
 </div>
 ```
+
+## Card columns
+
+Cards can be organized into <a href="http://masonry.desandro.com/" target="_blank">Masonry-like</a> columns with just CSS by wrapping them in .card-columns. Cards are built with <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns/Using_multi-column_layouts" target="_blank">CSS column</a> properties instead of flexbox for easier alignment. Cards are ordered from top to bottom and left to right.
```
kball: How do these interact with the grid? Can they be used with the grid? interact with it in some way?

scss/components/_card.scss
```diff
@@ -38,13 +38,21 @@ $card-padding: $global-padding !default;
...
+$card-columns-gap: $card-margin !default;
+
 /// Adds styles for a card container.
-/// @param {Color} $background - Background color of the card.
-/// @param {Color} $color - font color of the card.
-/// @param {Number} $margin - Bottom margin of the card.
-/// @param {List} $border - Border around the card.
-/// @param {List} $radius - border radius of the card.
-/// @param {List} $shadow - box shadow of the card.
+/// @param {Color} $background [$white] - Background color of the card.
```
kball: For all of these params, I think we should either use the actual value or use the name of the variable actually used... as these stand, I'm very confused because they're using variables that are 1 level up the inheritance hierarchy.

@zurb/yetinauts do y'all have an opinion on actual value vs name? Looking around i see both but more the latter, which would mean these should be something more like

```
@param {Color} $background [$card-background] - ...
```

docs/pages/card.md
rafibomb: It can go inside the grid columns or stand alone. I'll add tests to show both and docs to explain.

scss/components/_card.scss
rafibomb: I vote actual value...

andycochran:
[CSS Column](https://www.w3schools.com/css/css3_multiple_columns.asp) aren't very well supported. 😢

IamManchanda:
@andycochran with prefixed .... its exactly equal to flexbox sir 
=> **97.7%** and counting!!!

http://caniuse.com/#search=multiple%20columns
http://caniuse.com/#search=flexbox

andycochran:
But… 


Unfortunately, I have to support IE9 — as do many folks with gov clients. I'm pretty sure @brettsmason is in the same boat. We can't use Flexbox either. 

That said, I think this is a great feature. We just need to be very clear in our docs (in which say we support Versions 9+) about support for this feature if it uses CSS Columns. 

@rafibomb Could feature use Flexbox instead of CSS Columns, and only be supported if Flexbox Mode is enabled?

brettsmason:
As @andycochran said (good to see you around again by the way!) I still have to support IE9.

If there's a way it can be tied into the Flexbox mode (maybe this should be renamed/expanded?) then I'm all for it though as it looks a good idea.

IamManchanda:
Hmmnnn.... so sad that goverment sites .... are needing browser support for 0.2% and depreciating browser usage thing

> tied into the Flexbox mode

@brettsmason @andycochran 
Sorry but you guys may not have noticed but seems like that soon there will no such thing as flexbox mode
=> https://github.com/zurb/foundation-sites/wiki/Project-Roadmap#64-tentative-date-may

> 4. Default flex mode
> 5. CSS-grid based grid

IamManchanda:
Moreover for a moment, compare the depreciation with IE8 for once guys .... 
When zurb depreciated IE8 .... it had a very healthy browser usage .... but foundation decided to be advanced!!!
Then why things should change here ?? 
Forgiving web advancements just for the sake of so low browser usage is not good!

kball:
I want to correct a misconception here - we are not dropping the float grid in the 6.X series. However we will be moving to having flex be the default. @IamManchanda

brettsmason:
@IamManchanda I don't think either myself or Andy were suggesting this doesn't get implemented, just that the current version includes support for IE9, and this won't work for those users. 

It just needs to be handled a bit differently to a feature that works in all current supported browsers.

IamManchanda:
@kball Float shouldn't be dropped ... i was not misconcepted there .... the thing i am saying is that ... the time is now when developer should force goverment sites ... to forget IE9 .... If we want the web to move forward!!!

IamManchanda:
Flexbox .... CSS Columns... CSS Grids ... Calc .... CSS Transitions ... Animations .... Gradients all these advancements deserves a better state than just being rejected because of stupid IE9 

> **What is Internet Explorer?**
> It is a software to download Firefox or Chrome when you reinstall windows!!

kball:
@rafibomb can you do the docs updates mentioned above so we can merge this?

kball:
Ping @rafibomb - there's just a few docs updates necessary and then we can merge this in.

rafibomb:
@andycochran there is a flex version of the card columns as well - though it's not as flexible. I can make both as an option in the mixin if we think it helps. 

Yes - when we add future forward features we'll need a callout in the docs for browser support. I'll add this to the PR for now.

rafibomb:
Ok - so I can make a flex version along with this version. If flex mode == true then use flex? Not sure how to handle that

rafibomb:
Ok for this - today
- updating docs to reflect compatibility
- explaining how it works with columns
- add codpen example

ncoden:
@rafibomb @colin-marshall What is the status of this PR ?
Do we try to work on it for `v6.5.0`, or do we close it and reconsider in v7 ?

colin-marshall:
@ncoden hmmm. At this point I'm not sure if we want to introduce something where there is no support for IE9, even if it's documented (which is still something that needs to be done).

ncoden:
It is probable that we will drop IE9/10 in F7.

About the IE9 support in F6: IE9 does not support flexboxs and it does not prevent us from providing additional flex-only features (like the XY Grid). I think it's better to have an IE10+ card feature with a well documented browser compatibility than nothing.

colin-marshall:
@ncoden ya I was just thinking it might make things a little more complicated for going back to support F6 if we added it now. Without question I am for adding it to F7.

If you're cool with it going into F6 then I am too. We just need to add the documentation to this PR, right? I will test this once I figure out the module issues I'm having.

ncoden:
We're in a "feature freeze" mode, but that doesn't mean we cannot try to release feature that are already (or almsot) implemented. So I think we can release this in `v6.5.0` if you can provide the doc without too much work.

Also, I don't think we can plan the CSS components of v7. Relying on frameworks offers us a lot of features we will use to create components. This may radically change the way we'll want to use split, organize and use components.

kball:
@rafibomb can you add the doc comment about browser support & fi merge conflict? With that I think this should be good to merge.

PR #9747, opened 2017-02-04
Modified tolerance to be able to customize.

hikouki:
🍮

I'd like to customize `tolerance` :)

### DIFF
--- a/scss/util/_color.scss
+++ b/scss/util/_color.scss
@@ -55,10 +55,11 @@
 
 /// Checks the luminance of `$base`, and returns the color from `$colors` (list of colors) that has the most contrast.
 ///
-/// @param {Color} $color1 - First color to compare.
-/// @param {Color} $color2 - Second color to compare.
+/// @param {Color} $base - Color to check luminance.
+/// @param {List} $colors - Colors to compare.
+/// @param {Number} $tolerance [$global-color-pick-contrast-tolerance] - Contrast tolerance.
 ///
-/// @returns {Number} The contrast ratio of the compared colors.
+/// @returns {Color} the color from `$colors` (list of colors) that has the most contrast.
 @function color-pick-contrast($base, $colors: ($white, $black), $tolerance: $global-color-pick-contrast-tolerance) {
   $contrast: color-contrast($base, nth($colors, 1));
   $best: nth($colors, 1);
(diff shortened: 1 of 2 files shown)


scss/util/_color.scss
```diff
@@ -59,7 +59,7 @@
 /// @param {Color} $color2 - Second color to compare.
 ///
 /// @returns {Number} The contrast ratio of the compared colors.
-@function color-pick-contrast($base, $colors: ($white, $black), $tolerance: 0) {
+@function color-pick-contrast($base, $colors: ($white, $black), $tolerance: $global-color-pick-contrast-tolerance) {
```
IamManchanda: Can you explain a bit about this advancement ? @hikouki

hikouki: @IamManchanda 
Thank you for your review 🍣 
I want to set default `tolerance` .

`color-pick-contrast` function is used by following components:
- button
- badge
- callout
- label
- tabs

When I use this function, I thought if `tolerance` can be used more flexible,
it'll be more convenient. For example, even Button's background is a bit light color,
I want use light color to font too.

IamManchanda (approved):
Makes sense to me
@kball for decision on this!

scss/util/_color.scss
IamManchanda: Great. 
BTW.... Next time please tell this type of explanation in the  :arrow_double_up: above description! 
{{ Please dont leave description blank. It should tell what the code is doing. }}

kball:
@andycochran I believe you wrote this originally, any objections?

hikouki:
@kball @andycochran 
I'd like to know what the situation of the matter this. :bow:

kball:
Merged! Will be in 6.4

andycochran:
I went to address this and… 
Great work, @hikouki!

andycochran:
We should probably note in the documentation how this functions. 

For example, the docs for [Label](http://foundation.zurb.com/sites/docs/label.html#text-colors) says: 
> The default settings meet WCAG 2.0 level AA contrast requirements. Be sure to check the contrast when changing color variables. To give all labels the same color text, set `$label-color` and `$label-color-alt` to the same value — but know that doing so may decrease accessibility.

That last sentence should maybe be changed to, "It's possible to adjust the `$global-color-pick-contrast-tolerance` variable so that labels that do not have enough contrast use `$label-color` instead of `$label-color-alt` — but know that doing so may decrease accessibility."

^ This will need to be noted for Badges, Buttons, and Labels. 

We also might indicate somewhere how to adjust the  `$global-color-pick-contrast-tolerance` variable. Should I change it to `0.5` or `42` or a billion? And how will that affect accessibility?

PR #9749, opened 2017-02-04
Added a primary callout to kitchen sink

IamManchanda:
Callouts in kitchen sink have all variations except primary callout
Maybe this was forgotten before ... so added now .... Please add! 

--------------------------------------------------------------------------------------------------------------------

**btw .... i dont have much idea on why my merge commits (#9704) are coming here .... ???**


kball:
@IamManchanda given this is docs only, can you resubmit against master? Thanks

PR #9755, opened 2017-02-06
fixes #8404 hollow button dropdown arrow

### LINKED ISSUE #8404: Hollow dropdown buttons show white "triangle" instead of button color
Hi guys!

I'm not sure if this is expected behaviour, but if we combine the "dropdown" and "hollow" classes in a button, the dropdown triangle keeps the default colour. I think it should be displayed in the button color instead (success, secondary...). Without changes, default hollow dropdown buttons have an "invisible" triangle.

1. Create dropdown button (classes "button dropdown hollow").

The dropdown triangle should be the same highlight color as the button.

**What happened instead?**

The dropdown triangle keeps the default color (white = invisible).

**Test case:**

http://codepen.io/anon/pen/vGXzMz

canterberrie:
hollow button dropdown arrows were always white. this fixes the arrow with the appropriate dropdown arrow color

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -306,6 +306,20 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
           border-top-color: $button-background;
         }
       }
+
+      &.hollow {
+        &::after {
+          border-top-color: $button-background;
+        }
+
+        @each $name, $color in $button-palette {
+          &.#{$name} {
+            &::after {
+              border-top-color: $color;
+            }
+          }
+        }
+      }
     }
 
     // Button with dropdown arrow only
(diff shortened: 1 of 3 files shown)


kball:
Changes to `_settings.scss` should not be in the PR, as (counterintuitively, I know) that file is actually generated and should only be changed when we set up a new release.

scss/components/_button.scss
```diff
@@ -167,10 +171,18 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   color: $color;
 
   &:hover, &:focus {
-    border-color: $color-hover;
-    color: $color-hover;
+    @if $button-hollow-hover-fill == solid {
+        border-color: $color-hover;
+        color: $button-color;
+        background-color: $color-hover;
+      }
```
kball: How does this code relate to the arrow bug? This seems like a new feature...

PR #9762, opened 2017-02-09
Minor bug fix in the card-container mixin

IngoVals:
The font color parameter to the mixin was ignored and the $card-font-color variable was always used.

### DIFF
--- a/scss/components/_card.scss
+++ b/scss/components/_card.scss
@@ -67,7 +67,7 @@ $card-margin: $global-margin !default;
   box-shadow: $shadow;
 
   overflow: hidden;
-  color: $card-font-color;
+  color: $color;
 
   & > :last-child {
     margin-bottom: 0;


scss/components/_card.scss
```diff
@@ -67,7 +67,7 @@ $card-margin: $global-margin !default;
   box-shadow: $shadow;
 
   overflow: hidden;
-  color: $card-font-color;
+  color: $color;
```
IamManchanda: See -> [Line 50](https://github.com/IngoVals/foundation-sites/blob/f9e9d94a2f065ec84812ffc56731c0fcfa33e840/scss/components/_card.scss#L50). Within the mixin, this is already set with settings variable

```scss
$color: $card-font-color
```

which means if somebody writes this below .... then that color will be equal to `card-font-color`

```scss
.mycard {
   @include card-container;
}
```

But also, if you dont want that, then it can be customized within a mixin like 

```scss
.mycard {
   @include card-container($color: #ccc);
}
```

This is just how it goes to reuse mixin in sass!
See => https://www.sitepoint.com/sass-basics-the-mixin-directive/

IamManchanda (changes requested):

scss/components/_card.scss
IngoVals: if, before my suggested change, you write

`.mycard {
   @include card-container($color: #ccc);
}`

The color in `.mycard` will be equal to `$card-font-color` not to `#ccc`

I'm stating the the problem is that you set `$card-font-color` as default for `$color` parameter into the mixin, but then inside the mixin itself you totally ignore the $color parameter and apply `$card-font-color` to the color setting without looking at whatever the user passed in.

Compare the original with how `$background` and `$card-background` is handled.

How this currently works is equivalent to something like this

`@mixin headline($size, $color: red) {
  color: red; //No matter what color parameter a user set, the color will always be red
  font-size: $size;
}`

IamManchanda: So basically you would like that user can't change the color and it should always be `$card-font-color` ?? 
Am i correct in understanding you ?

brettsmason:
@IamManchanda - @IngoVals is correct with this fix. Here is the issue: https://github.com/zurb/foundation-sites/blob/develop/scss/components/_card.scss#L70

Basically if you pass a colour using the mixin its ignored as the passed value isn't used - this PR fixes that.

Thanks for the PR!

IamManchanda:
Ohhh sorry ..... now i understand ..... my mistake !

scss/components/_card.scss
IamManchanda: I am sorry @IngoVals ..... My mistake ..... i was thinking opposite!

IngoVals:
What is the normal timeline of gutting PR into release branch. Is this perhaps being overlooked because there isn't a bug/issue attached?

IamManchanda:
@IngoVals your commit has been merged and will go through quality assurance and if everything goes correctly, it will be put in release branch very soon .... most probably in 6.3.2 as its a bug fix.

You dont have to do anything else!

PR #9763, opened 2017-02-09
7548 prevent reveal scroll on iOS using position:fixed on html element

gerardkcohen:
Quick fix to address #7548, reveal scrolling on mobile, by adding `position:fixed`

Tested on iOS 10.2

### DIFF
--- a/scss/components/_reveal.scss
+++ b/scss/components/_reveal.scss
@@ -123,6 +123,7 @@ $reveal-overlay-background: rgba($black, 0.45) !default;
   html.is-reveal-open body {
     min-height: 100%;
     overflow: hidden;
+    position: fixed;
     user-select: none;
   }
 


kball:
Awesome, thanks @gerardkcohen. @zurb/yetinauts anyone have time to test this?

JeremyEnglert:
Tested on iOS 10, looks good.

If anyone else wants to test, I uploaded the files to make it easier to test on mobile devices.

With Fix: http://jointswp.com/f6-playground/
Without Fix: http://jointswp.com/f6-playground/index-old.html

pgib0811:
Is there a vanilla CSS solution for this? Background not scrolling behind full reveal modal, but modal will not fill up 100% height. Only on iOS in Safari. Can be seen in any of the modals [here](http://www.parker-gibson.com/testing/web.html)

kball:
Appears to work fine on android as well. Merging.

@pgib0811 would happily accept a PR making this simpler. It seems like it should not require the JS, but with reliable cross browser behavior has been strikingly hard to come by here.

harrington101:
I just downloaded the latest version of foundation and I'm still seeing this issue. The background is scrolling instead of the modal when the modal is longer than the screen. Also the page is showing through the bottom and top of the modal while scrolling.

pgib0811:
Still an issue across Android and iOS. I have posted to [Stack Overflow](http://stackoverflow.com/questions/43750361/foundation-modal-showing-background) with screenshots for both.

gerardkcohen:
I was unable to reproduce on iOS 10.3.1 with Safari, Chrome, or Brave. They
are all webkit so I wouldn't expect different behavior. Except Safari had a
bit of elastic scroll, but the page itself did not scroll. I don't have an
Android nearby to test.

pgib0811:
Also using Safari on iOS 10.3.1 for iPad which is what screenshot of Zurb doc page was taken on. Llinked here https://imgur.com/gallery/SLUG1

tomheaps:
I have exactly the same issue in chrome for android mobile i.e. gap at top of modal. I think its something to do with the screen height that is being taken up by mobile browsers and the vertical offset calculated by foundation js for the modal. Only solution I found was to add `data-v-offset="0"` to the modal as an option - only downside is that seems to prevent address bar from hiding when scrolling down on mobile devices!

DanielRuf:
It seems this produced a regression. The modal has a smaller height than the body but I can not scroll it unless I remove the `position: fixed´.

coreysyms:
So I'm not sure what the intentions are for the "gap at the top" of a reveal on mobile, but in my opinion, there should be no `overflow-y: scroll;` on the wrapping `.reveal-overlay`. With that removed there is no scrolling of the modal, and content is allowed to scroll. Also we should add in `-webkit-overflow-scrolling: touch;` for better content scrolling. 

If some @yetinauts can answer my question as to why there is an offset top on mobile reveal and not `top:0` I can create a PR. It seems there is room for the top bar, but you can't click on it, cause the overlay is overlaying it.

There is a 45px top gap, but you can't click on it... is this intentional?
image1

Scrolling is just fine within the content area
image2

DanielRuf:
@coreysyms exactly our issue.

> So I'm not sure what the intentions are for the "gap at the top" of a reveal on mobile, but in my opinion, there should be no overflow-y: scroll; on the wrapping .reveal-overlay. With that removed there is no scrolling of the modal, and content is allowed to scroll. Also we should add in -webkit-overflow-scrolling: touch; for better content scrolling.

I'll try that. This sounds like a possible solution.

coreysyms:
@DanielRuf yes, in the meantime, you can force override the `top` and `overflow-y` along with adding in `-webkit-overflow-scrolling: touch;` to the `.reveal` in your CSS to get you where you need to be.

DanielRuf:
@coreysyms

```
.reveal-overlay {
  overflow-y: visible;
}
.reveal {
  -webkit-overflow-scrolling: touch;
}
```

This prevents scrolling of reveals which are longer than the window.
Only
```
html.is-reveal-open,
html.is-reveal-open body {
  position: initial;
}
```
was working here so far (but there is still some height issue in our project, some reveals are scrollable even if the content is not very heigh and on iOS there is some recalculation happening).

Or how should the code look like? We also use some reveals without the overlay wrapper.

DanielRuf:
Also on iOS in Chrome 60 the whole content / body is scrolled.

coreysyms:
Ah, ok, I see what the issue is. It is not only about the space at the top, but the long standing iOS issue with overlays.

What needs to happen is a few things.... typically what I do is a mix of CSS and JS.

1. Hang an event listener for the click of or opening of the modal overlay.
2. Read the current `scrollTop` position of the `html,body`. Store this var.
3. Add a class to the` html,body` of called `noscroll`.
4. `noscroll` is a class that makes the overflow of the `html,body` : `hidden` as well as sets the height as `100px` or whatever, something very small.
5. Hang an event on close to reverse these changes. Put the body back to normal. Scroll to the previous position and close the modal.

Why you have to do this yourself and Foundation shouldn't is because it would be, in my opinion, too much manipulation to the content of the site. Frameworks can / should only take you so far, this one you will have to finish out on your own. This is a long-standing bug with iOS, and most likely will not be fixed anytime soon. We should address the top gap and addition of momentum scroll for the overlay.

@DanielRuf

PR #9769, opened 2017-02-10
Make hidden parts of submenus inaccessible without creating tab trap

kball:
Discovered during QA, the change to prevent tab trapping implemented by @Owlbertz in https://github.com/zurb/foundation-sites/pull/9634 had the side effect that if your last element on the top level menu of a drilldown had a submenu, tab would open that submenu rather than exiting the drilldown.

The reason: on the last element we dont' prevent default (to prevent tab trap), but the result is that we'd tab into the submenu.

The solution: Make the submenus actually inaccessible to the tab index when they are hidden (e.g. only open them using our JS). Use `visibility:hidden` to do this so that their heights are still available for calculation.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -100,10 +100,11 @@ class Drilldown {
       _this._back($menu);
     });
 
+    this.$submenus.addClass('invisible');
     if(!this.options.autoHeight) {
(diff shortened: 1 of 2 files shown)


Owlbertz (approved):
Yeah, seems to work out.

brettsmason:
I can also confirm this fixes the issue I raised in QA 👍

PR #9777, opened 2017-02-14
Allow table stack breakpoint to be set with a variable

mwitek:
Adding a var to set the breakpoint for stacked tables.

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -556,7 +556,6 @@ $table-foot-background: smart-scale($table-background, $table-color-scale);
 $table-foot-row-hover: darken($table-foot-background, $table-hover-scale);
 $table-head-font-color: $body-font-color;
 $table-foot-font-color: $body-font-color;
-$table-stack-breakpoint: medium;
 $show-header-for-stacked: false;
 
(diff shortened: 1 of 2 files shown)


scss/settings/_settings.scss
```diff
@@ -556,6 +556,7 @@ $table-foot-background: smart-scale($table-background, $table-color-scale);
 $table-foot-row-hover: darken($table-foot-background, $table-hover-scale);
 $table-head-font-color: $body-font-color;
 $table-foot-font-color: $body-font-color;
+$table-stack-breakpoint: medium;
```
IamManchanda: Setting files is generated automatically .....  

You need to add this variable here after this => https://github.com/mwitek/foundation-sites/blob/dbea7aa151d175f2f3db6bf7379d7bddec1d60c9/scss/components/_table.scss#L73

Please remove changes from settings!

IamManchanda (changes requested):
Poke @kball Settings again :D 
#Dist

mwitek:
@IamManchanda Ready for another look.

kball:
Looks good to me!

PR #9779, opened 2017-02-14
Fixes #9775 - Button group spacing

### LINKED ISSUE #9775: Button-Group mixin with $buttongroup-spacing variable
Sometimes it is helpful to create is own button-group. With the usage from the `include button-group()` mixin this is already possible, but sometimes there is a need for more or less button-space. The mixin takes the global `$buttongroup-spacing`, so there is no possibility to pass his own spacing to the mixin.

It would be great to have this feature implemented.

IamManchanda:
Fixes #9775 as this looks logical to me!

@kball @rafibomb for review/merge!

### DIFF
--- a/scss/components/_button-group.scss
+++ b/scss/components/_button-group.scss
@@ -11,7 +11,7 @@
 $buttongroup-margin: 1rem !default;
 
 /// Margin between buttons in a button group.
-/// @type Border
+/// @type Number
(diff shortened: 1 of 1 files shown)


kball:
Looks good to me

IamManchanda:
As always thanks @kball

PR #9819, opened 2017-02-28
Added a gentle reminder to include no-js class to html element.

Proper-Job:
It is a very important piece of information that is easy to overlook.  It isn't properly documented yet.


kball:
Merged, thank you!

PR #9828, opened 2017-03-01
ABIDE - clarification of novalidate attribute

colinmackinlay:
The need to use novalidate is new in F6. The proposed change emphasises why it is needed.


kball:
Merged, thanks @colinmackinlay

PR #9831, opened 2017-03-03
util.GetOffsets with offsets

heygambo:
I wasn't able to position the tooltip as I wanted because it was only accepting either `vOffset` or `hOffset` based on the tooltip's direction.

the lines I have commented out weren't even used.
I'm not sure if this will break other plugins.

Feedback is welcome.


kball:
I think we need to do this, but it will require thorough testing, and there are a number of other cases that need to be addressed within `util.GetOffsets`.  @nerdyglasses can you fill out the rest of the cases that are currently missing offsets, so we can do thorough testing as a part of the 6.4 QA cycle?

kball:
Going to close this in favor of the work going on in https://github.com/zurb/foundation-sites/pull/10053 which addresses all of the cases.

nerdyglasses:
sorry, i missed your comment from a month ago.

PR #9839, opened 2017-03-05
Respect $button-background for default hollow buttons.

### LINKED ISSUE #9838: Hollow button defaults to primary color

1. Override the default `$button-background` with the intention of changing what a palette-less `class="button"` looks like
2. Switch to default `$button-fill: hollow;` or simply make a palette-less `class="hollow button"`

#### What should happen:

The palette-less hollow button should be styled with your default `$button-background` color.

#### What happened instead:

It uses your palette's `$primary-color`.

#### Test case:

Hard to repeat in a fiddle that doesn't allow overriding foundation variables, but the cause is clear in the code.

1. Palette-less hollow buttons are generated [from the mixin application here](https://github.com/zurb/foundation-sites/blame/develop/scss/components/_button.scss#L278) with no arguments.
2. The hollow button style mixin [defaults to `$primary-color`](https://github.com/zurb/foundation-sites/blame/develop/scss/components/_button.scss#L160).

I think the `$primary-color` default is poorly chosen. `$button-background` (which is really more of the `$button-color` for normal buttons) seems a more logical and consistent choice. Will submit a PR if you agree.

(Technically this is a [duplicate issue](https://github.com/zurb/foundation-sites/issues/8281), but the unexpected behaviour was obscured by the fact that the submitter was really interested in customizing the palette. The issue with palette-less defaults remains.)

christhekeele:
Fixes #9838.

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -157,7 +157,7 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
 }
 
 @mixin button-hollow-style(
-  $color: $primary-color,
+  $color: $button-background,
   $hover-lightness: $button-hollow-hover-lightness, 
   $border-width: $button-hollow-border-width
 ) {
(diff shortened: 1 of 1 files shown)


IamManchanda:
I think you should do the same here [too](https://github.com/christhekeele/foundation-sites/blob/90cfa0176e6ded8c5dedab426b18d1efce3d0b5d/scss/components/_button.scss#L179) 

Reason:

```scss
/// Default background color for buttons.
/// @type Color
$button-background: $primary-color !default;
```

christhekeele:
Agreed and done. Also, related: #5277. Maybe this could be clearer in code by changing `$button-background` to `$button-default-color`? Probably a change for a later minor release but worth bringing up now.

kball:
Merged!

Issue #9822, opened 2017-03-01
Reveal Overlay Class


Knight-Yoshi:
The reveal overlay class can't be changed? It uses a static string. Kind of defeats the purpose of using your own modal overlay class. 

https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L85

kball:
Fixed in #9840. Will be in 6.4

PR #9840, opened 2017-03-05
Adding additional classes to reveal-overlay element

jefsnare:
As requested in #9822. Adding additional classes to `reveal-overlay`
element.

Not the possibility to replace the class, since many styling and some JS
handles react on this class.

### DIFF
--- a/_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss
+++ b/_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss
@@ -424,7 +424,7 @@
     optgroup,
     select,
     textarea {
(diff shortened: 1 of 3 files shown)


Owlbertz:
I see that having multiple classes on the overlay might be useful in some cases if you want to style multiple overlays differently.
Would you be able to write a short test case for your feature that checks if classes are actually added?

jefsnare:
@Owlbertz Yes ofcourse, I'll add them asap

jefsnare:
@Owlbertz Test added, and minor refactor done.

kball:
Awesome. Looks good to me!

PR #9846, opened 2017-03-06
Set the min-width value to 0 in order to limit the column width

### LINKED ISSUE #9845: Flex Grid Column Widths Not Restricted

1. Using the flex grid, add an item that is wider than the max width
2. The column is now wider than the max width

#### What should happen:

The column should be restricted by the max width

#### What happened instead:

The column was as large as the item. Related to: https://stackoverflow.com/questions/34934586/white-space-nowrap-and-flexbox-did-not-work-in-chrome

#### Test case:

http://codepen.io/natewiebe13/pen/YZGaOy

natewiebe13:
Goal: Fix the columns from expanding outside of the row when using the flex grid. Fixes #9845
The current implementation was using initial instead of 0. The comment above suggests using 0 instead of initial.
Example: http://codepen.io/natewiebe13/pen/YZGaOy

### DIFF
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -86,7 +86,7 @@
   // fixes recent Chrome version not limiting child width
   // https://stackoverflow.com/questions/34934586/white-space-nowrap-and-flexbox-did-not-work-in-chrome
   @if $columns == expand {
-    min-width: initial;
+    min-width: 0;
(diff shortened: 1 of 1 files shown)


natewiebe13:
Note: Build is failing due to an Access Denied from browserstack.

kball:
Makes sense to me.

PR #9848, opened 2017-03-07
[Reveal] Fix closing when Motion UI is used

jeromelebleu:
Thanks to #9653, when closing a modal, it now checks if there is another opened one before removing the `is-reveal-open` class from `body`. It works when jQuery is used to hide the elements - since its behavior has been modified - but no longer with Motion UI.

As the `finishUp` function is called before the modal is actually closed, the `is-reveal-open` class is never removed - since `$('.reveal:visible').length === 0` [here](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L415) will always be false.

This PR purpose one way to fix it by hiding the modal first then the overlay. Nevertheless, it will change the current closing behavior - the overlay and the modal are hidden at the same time. If it should remain like that, there are other ways to fix it, fortunately!

I believe that a unit or visual test should also be given, but I didn't learn how to write that yet...

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -368,18 +368,17 @@ class Reveal {
 
     // Motion UI method of hiding
     if (this.options.animationOut) {
-      if (this.options.overlay) {
-        Foundation.Motion.animateOut(this.$overlay, 'fade-out', finishUp);
-      }
-      else {
-        finishUp();
-      }
-
-      Foundation.Motion.animateOut(this.$element, this.options.animationOut);
+      Foundation.Motion.animateOut(this.$element, this.options.animationOut, () => {
+        if (this.options.overlay) {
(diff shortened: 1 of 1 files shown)


kball:
Hmm... @jeromelebleu I think it would probably be better to maintain the behavior of closing at the same time. Having them animate out sequentially is a little odd... you mentioned there were other ways to fix, can you take a crack at one of those?

jeromelebleu:
@kball I've updated this patch to maintain the original behavior of closing at the same time. I've just attached the `finishUp` function to the modal closing instead as it will hide the modal. Is it better?

IamManchanda:
Looks cool to me!!!

kball:
Looks good!

jeromelebleu:
Thanks!

PR #9849, opened 2017-03-07
[Tabs] Allow the target tab's id to be specified through another attribute

jeromelebleu:
I faced to some limits with tabs while building a one-page website. I'd like to explain a bit more what I'm trying to do in order to understand the goal of this PR - and maybe there is an easier solution...
So, the document is composed of sections. Each section is accessible through a hash like `#/section1` to produce some kind of navigation. In one of them, there are some tabs which I would like to access through `#/section1/panel1`. Unfortunately, it will not works if I use this hash as the `href` attribute of the link - and as the id of the panel - since it needs to be escaped to be [selected with jQuery](https://api.jquery.com/id-selector/).

This PR allows one to specify in the link the ID of the tab through a `data-tabs-target` attribute. It will be retrieved first, falling back to `href` if it's empty. What do you think about that? Thanks in advance!

### DIFF
--- a/docs/pages/tabs.md
+++ b/docs/pages/tabs.md
@@ -7,12 +7,12 @@ js: js/foundation.tabs.js
 
 ## Basics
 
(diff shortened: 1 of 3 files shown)


kball:
Seems reasonable to me. Thanks @jeromelebleu!

PR #9868, opened 2017-03-12
Fix #9866 - Hack to disable `-webkit-appearance` from getting prefixed

### LINKED ISSUE #9866: Native meters example doesn't work on chrome

1. Open Chrome (I use version 56.0.2924.87 at the moment)
2. Navigate here: http://foundation.zurb.com/sites/docs/progress-bar.html#native-meter
3. Watch as the 3 native meters are all grey, instead of with colors.
4. Open another browser (for example, Firefox).
4. Navigate to the same URL, notice the native meters work as expected.

#### What should happen:
Native meters show colors, representing their value.

#### What happened instead:
Native meters show only grey color.

IamManchanda:
Closes #9866 
Reference => http://stackoverflow.com/a/38892123/3301436

Please note that my solution is a bit of a **hack** to basically not let autoprefixer prefix `-webkit-appearance`
If somebody got a better way, then please tell and i will change!

### DIFF
--- a/scss/forms/_meter.scss
+++ b/scss/forms/_meter.scss
@@ -37,8 +37,13 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    -webkit-appearance: -hack-; // Hack to disable `-webkit-appearance: none` from getting prefixed
-    appearance: none;
+    // Specific Hack to disable `-webkit-appearance: none` from getting prefixed,
+    // We have disabled autoprefixer first and are just using `-moz-appearance: none` 
+    // as a prefix and neglecting `-webkit-appearance: none`
+    
+    /*! autoprefixer: off */
+    -moz-appearance: none;
+         appearance: none;
 
     @if has-value($meter-radius) {
       border-radius: $meter-radius;
--- a/scss/forms/_meter.scss
+++ b/scss/forms/_meter.scss
@@ -37,7 +37,7 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    -webkit-appearance: text; // Hack to disable `-webkit-appearance` from getting prefixed
+    -webkit-appearance: text; // Hack to disable `-webkit-appearance: none` from getting prefixed
     appearance: none;
 
     @if has-value($meter-radius) {
--- a/scss/forms/_meter.scss
+++ b/scss/forms/_meter.scss
@@ -37,7 +37,7 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    -webkit-appearance: text; // Hack to disable `-webkit-appearance: none` from getting prefixed
+    -webkit-appearance: -hack-; // Hack to disable `-webkit-appearance: none` from getting prefixed
     appearance: none;
 
     @if has-value($meter-radius) {
--- a/scss/forms/_meter.scss
+++ b/scss/forms/_meter.scss
@@ -37,9 +37,9 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    // Specific Hack to disable `-webkit-appearance: none` from getting prefixed,
-    // We have disabled autoprefixer first and are just using `-moz-appearance: none` 
-    // as a prefix and neglecting `-webkit-appearance: none`
+    // Disable `-webkit-appearance: none` from getting prefixed,
+    // We have disabled autoprefixer first and are just only using 
+    // `-moz-appearance: none` as a prefix and neglecting the webkit.
     
     /*! autoprefixer: off */
     -moz-appearance: none;
--- a/scss/forms/_meter.scss
+++ b/scss/forms/_meter.scss
@@ -37,6 +37,7 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
+    -webkit-appearance: text; // Hack to disable `-webkit-appearance` from getting prefixed
     appearance: none;
 
     @if has-value($meter-radius) {


ncoden (commented):
Hmm... This is a bug for a specific version of Chrome, right ? Which versions of chrome are touched by this bug: v52.x ? v56.x ? 56.0.2924.87 ? both ? all versions since v52 ?

Chrome does not support `-webkit-appearance` without its vendor prefix. This PR is disabling the reset of `appearance` for **any Chrome version** by settings an **invalid value** (appearance: text` does not exists).

So:
- Do we need `appearance:none` on Chrome ? If yes, we should focus on the behavior of not bugged versions and keep it. We cannot introduce a bug in Foundation to solve a browser bug.
- What code is generated after autoprefixer ? Is `-webkit-appearance` skipped or written but overriden ? If overriden, your hack should be _after_.
- Why `:text` ? It does not seems to be a valid `-webkit-appearance` value. It would be better to use an explicitely invalid value (i.e. `-hack-`), or to reset the property (`initial`).

What do you think ?

IamManchanda:
> Which versions of chrome are touched by this bug: v52.x ? v56.x ? 56.0.2924.87 ? both ? all versions since v52 ?

All versions, to the best of my knowledge!

> Do we need appearance:none on Chrome ?

No it is not needed for meter !

> What code is generated after autoprefixer ? Is -webkit-appearance skipped or written but overriden ? If overriden, your hack should be after.

This is the generated code before my hack

```scss
meter {
  -webkit-appearance: none;
     -moz-appearance: none;
          appearance: none;
}
```

> Why :text ? It does not seems to be a valid -webkit-appearance value. It would be better to use an explicitely invalid value (i.e. -hack-), or to reset the property (initial).

Yes you are right but `-hack-` is doing the job and `initial` is not !
So i just did that and now after putting `-hack-` , the below code is generated 

```scss
meter {
  -webkit-appearance: -hack-;
     -moz-appearance: none;
          appearance: none;
}
```

As you can see .... `-webkit-appearance: none` gets skipped now Nicolas.

IamManchanda:
> focus on the behavior of not bugged versions and keep it

@ncoden no we can't .... [around 75% web uses chrome](https://www.w3schools.com/Browsers/browsers_chrome.asp) , we simply can't keep it!

ncoden:
An other question: Do we need `appearance: none` for other browsers ?

If yes, i'll see if we can add the following feature to autoprefixer: (poke @ai)
```css
{
    /*! autoprefixer: off for webkit */
    appearance: none;
}
```

ai:
@ncoden Autoprefixer has only `off` for all browsers. But you can manually add `-moz-` prefix.

Or just write simple PostCSS plugin to clean styles:

```js
postcss.plugin(() => {
  return root => {
    root.walkDecls('-webkit-appearance', i => i.remove())
  }
})
```

IamManchanda:
I checked on firefox and it seems like `appearance: none` is not needed there also .... but i am not 100% sure for firefox ... i think we need some good testing for this PR!

ncoden:
@ai Thank you for your answer, but it seems you did not understood our problem. We want to disable `-webkit-appearance` for a specific selector, because of a bug in Chrome.

We currently have two solutions:
```css
meter {
    -webkit-appearance: -hack-; // force use of invalid value.
    appearance: none;
}
```
```css
meter {
    /*! autoprefixer: off */
    -moz-appearance: none;
         appearance: none;
}
```

It works, but the first solution is a bit dirty, and the second is not explicit about "not for webkit". `/*! autoprefixer off for webkit */` would be an useful sugar.

IamManchanda:
@ncoden @ai I think second one will be better if this is the only 2 available solutions !

ai:
@ncoden Autoprefixer code become so complicated — I will not add `off for webkit` in this year :D. Let’s keep:

```css
meter {
    /*! autoprefixer: off */
    -moz-appearance: none;
         appearance: none;
}
```

:D

IamManchanda:
done @ncoden @ai !

scss/forms/_meter.scss
```diff
@@ -37,7 +37,13 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    appearance: none;
+    // Specific Hack to disable `-webkit-appearance: none` from getting prefixed,
+    // We have disabled autoprefixer first and are just using `-moz-appearance: none` 
+    // as a prefix and neglecting `-webkit-appearance: none`
```
ncoden: That's not a hack, nor in Sass, nor in CSS. We simply do not write `-webkit-appearance`. But you can explain _why_: not necessary in Chrome, and create a bug (with a link to the issue).

However, only do this if we are sure that we _need_ appearance:none in other browsers (Chrome, Firefox, IE, Edge, Opera...)

IamManchanda:
@ncoden .... i am not fully 100% confident about is there any need of `appearance: none`
Because whatever browser i was able to test .... the meter is working even without `apperance: none` also

> ```markdown
> Specific Hack to disable
> ```
> That's not a hack nor in Sass, nor in CSS. 

Accident, removing now the hack keyword

IamManchanda:
> create a bug (with a link to the issue).

https://jsfiddle.net/t58fnan9/1/

DaSchTour:
This is reported as a bug in Chromium, so this behavior could change at any time. So we need to find a solution that will work with newer versions that might not have this bug. If `appearance: none` really isn't needed to achieve the current styling it would be best to just remove it.
@IamManchanda can you list the browsers you tested with version and OS. So we can test the browsers missing and create a list of all browsers and their behavior for meter and appearance.

IamManchanda:
@DaSchTour i dont have any testing account but with sauce labs free trial - manual testing 
i test the progress bar page on chrome and firefox ( can't inspect element )

Firefox - Its working ( tested 45+ )
Edge - Its working ( all versions, 13+ ofcourse as 12 not supported )
Safari - Working ( tested 6+ ) 
Chrome - Its working till 51, from 52-56+ (all versions) its not working ( tested 45+ )

Operating System - OSX, Windows, Ubuntu ( All OS its same, no difference whatever i have tested)

Moreoever i am quite sure, `appearance: none` is not required in chrome, 
but not sure about firefox n all !
Btw i remove `appearance: none` locally and was working without it in firefox!

Sorry i have this information only... Hope it helps!

IamManchanda:
I feel removing webkit will not hurt anything and removing all might hurt!

scss/forms/_meter.scss
```diff
@@ -37,7 +37,13 @@ $meter-fill-bad: $alert-color !default;
     height: $meter-height;
     margin-bottom: 1rem;
 
-    appearance: none;
+    // Disable `-webkit-appearance: none` from getting prefixed,
+    // We have disabled autoprefixer first and are just only using 
+    // `-moz-appearance: none` as a prefix and neglecting the webkit.
+    
+    /*! autoprefixer: off */
```
kball: Does this impact the rest of the rules for meter following this? (e.g. border-radius) down below... Do we need to separate out the appearance into a separately scoped `meter` rule?

kball: ping @IamManchanda

IamManchanda: sorry .... will look into it now!

IamManchanda:
Well i dont know why but this **[has-value function](https://github.com/zurb/foundation-sites/blob/develop/scss/util/_value.scss#L14-L25)** is not working properly because with both on and off ... For border-radius, i am setting a value (eg. 50%) to the variable instead of null or 0 but not getting anything from generated css

Here is the output 

```scss
meter {
  display: block;
  width: 100%;
  height: 1rem;
  margin-bottom: 1rem;
  -moz-appearance: none;
  -webkit-appearance: none; // This is removed on autoprefixer off 
          appearance: none;
  border: 0;
  background: #cacaca; }
  meter::-webkit-meter-bar {
    border: 0;
    background: #cacaca; }
  meter::-webkit-meter-optimum-value {
    background: #3adb76; }
  meter::-webkit-meter-suboptimum-value {
    background: #ffae00; }
  meter::-webkit-meter-even-less-good-value {
    background: #cc4b37; }
  meter::-moz-meter-bar {
    background: #1779ba; }
  meter:-moz-meter-optimum::-moz-meter-bar {
    background: #3adb76; }
  meter:-moz-meter-sub-optimum::-moz-meter-bar {
    background: #ffae00; }
  meter:-moz-meter-sub-sub-optimum::-moz-meter-bar {
    background: #cc4b37; }
```

Am i missing something @kball @ncoden  ??

kball:
Odd... I set border radius and tested and saw no issues; border radius was correctly set. Compared against generated code without the doc comment and seems fine. Going to merge.

PR #9311, opened 2016-10-31
Don't run through motion ui if orbit element is hidden

kball:
Alternate to https://github.com/zurb/foundation-sites/pull/9152 addressing second half of https://github.com/zurb/foundation-sites/issues/9103

PR #9881, opened 2017-03-16
Fix bug on Orbit resize during transition

### LINKED ISSUE #9788: motion-ui bug on orbit carousel

This is exactly what my problem is:

1. Visit: [F6 Orbit](https://foundation.zurb.com/sites/docs/orbit.html) on mobile (must be a mobile device not, resized desktop window)
2. Go to any Orbit slide
3. Swipe (left/right) the slide WHILE motioning up or down (may take several swipes to trigger error below)

#### What should happen:
The slide should go to the next slide.

#### What happened instead:
Sometimes after several "slanted" swipes the upcoming slide with disappear.

From what I can see, in the html, the `is-active` class disappears and what I assume is causing the issue.

This behavior is the same on my pages as well.

dwt47:
Added check in orbit.js to make sure slides aren't animating before hiding them when resizing the window. This should fix #9788

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -137,7 +137,7 @@ class Orbit {
       temp = this.getBoundingClientRect().height;
       $(this).attr('data-slide', counter);
 
-      if (_this.$slides.filter('.is-active')[0] !== _this.$slides.eq(counter)[0]) {//if not the active slide, set css position and display property
+      if (!/mui/g.test($(this)[0].className) && _this.$slides.filter('.is-active')[0] !== _this.$slides.eq(counter)[0]) {//if not the active slide, set css position and display property
         $(this).css({'position': 'relative', 'display': 'none'});
       }
       max = temp > max ? temp : max;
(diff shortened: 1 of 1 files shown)


RLHawk1:
Would love to have this fix merged in.

kball:
Good find here... however, not 100% sure this is the right answer... If the problem is the lingering motion ui classes, can we remove them at the same time we add display none?

My concern with this current approach is that by not hiding the slide we then may be breaking the assumption of the rest of this setup method, which may cause its own issues.

kball:
@dwt47 do you have an opinion on the feedback above? Would you like to iterate on this PR and/or make a case why this is the right solution?  Thanks!

kball:
Going to go ahead and merge this based on the updates in #10069

benjaminpaulwatson:
I just implemented the orbit slider myself and am experiencing this same issue where a slide disappears (goes blank) and then the whole slider stops functioning.  I've seen it on desktop intermittently, but it happens almost always on mobile. Has there been a reliable fix to this yet?

jinch:
+1 ~ Experiencing this intermittently on the desktop as well. It seems to be more frequent when using the visibility classes. When hiding the orbiter on small screens (class="orbit show-for-medium") and resizing the window a few times the orbit either disappears or becomes frozen.

The suggested fixes to "foundation.orbit.js" do not seem to fix this entirely so not sure there is a solution yet other than turning off the motion-ui animation entirely (data-use-m-u-i="false").

DanielRuf:
How can I apply this in 6.3.1?

DanielRuf:
Is this also related: https://github.com/zurb/foundation-sites/pull/9311 ?

PR #9885, opened 2017-03-20
[Test] verify correct tooltip aria-hidden value

dragthor:
Verify correct `aria-hidden` value in tooltip unit tests.


kball:
Merged!

PR #9887, opened 2017-03-21
adding .expanded to a row nested inside a row with class expanded wil…

### LINKED ISSUE #9886: nesting rows inside expanded row

1. Step one
create a div with class 'row expanded'

2. Step two
nest a div with class 'row' inside and then another div with class 'row' inside of that one. 

3. Step three
observe that content doesn't align.

#### What should happen:
users should have ability to expand rows nested inside of expanded rows to full width of previous row.

#### What happened instead:
padding compounds and rows nested inside an expanded row don't align.

#### Test case:
-[Click here for CodePen illustrating the issue](http://codepen.io/mryakamoto/pen/zZWBja)

mrYakamoto:
…l expand nested row to full width of previous row

closes #9886

### DIFF
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -52,6 +52,10 @@
       .#{$row} {
         margin-right: auto;
         margin-left: auto;
+
+        &.#{$expanded} {
+          @include grid-row-size(expand);
+        }
       }
     }
 


gpspake:
I updated #9886 with a codepen, I think you can get the results you're looking for by just adding `.collapse` to the nested rows

IamManchanda:
Thanks @gpspake ..... i think you should close this @mrYakamoto 
`.collapse` aka **no-gutters** is the solution for this

mrYakamoto:
@gspake, @iammanchanda, isn't alignment still an issue in this case though? The nested rows don't line up using collapse

IamManchanda:
Yup .... looked deeply and yup you have a point 
cc @kball !

gpspake:
@mrYakamoto yeah, I see what you mean now and that fix is nice. +1 on this 👍

kball:
@brettsmason @andycochran @ncoden what do one of you _sassy_ folks think? ;)

kball:
ping @andycochran can you take a look at this?

andycochran:
This seems like a specificity rabbit hole. Right, @ncoden?

kball:
Did we have a verdict @andycochran @ncoden? Is this a go or no-go?

ncoden:
Hello. Sorry for the wait ;)

So, it seems there is a problem with what you're expecting.

Actually, we have:
- `.row .row` must be aligned (negative marging on child to counter the parent padding)
- `.row.expanded .row` must be centered (auto margin)
- So: `.row.expanded .row .row` is centered instead of being aligned.

You want to make the 3rd row with `.expanded` have negative margins too, but it does not have anything to do with `.expanded`, this is a behavior you expect from any nested row.

Despite of the name (let's say we use `.forced-nested` instead of `.expanded`), the problem is we're f*** by specificity (again, huh ?). Adding an other modifier do the job for this particular situations, but will create more problems for tons of other situations.

So, for now, I only have a bad solution: Ensure both `.row .row` and `.row.expanded .row .row` is aligned (negative margins).
This increase the specificity, complexify the behavior, this is nothing but a hack for a badly declared `.row.expanded .row`, but... this is what any API base with an uncontrolled specificity will always leads to.

Let me know what you think about it.

ncoden:
@mrYakamoto @kball @andycochran Hello. Sorry for the wait ;)

So, it seems there is a problem with what you're expecting.

Actually, we have:
- `.row .row` must be aligned (negative marging on child to counter the parent padding)
- `.row.expanded .row` must be centered (auto margin)
- So: `.row.expanded .row .row` is centered instead of being aligned.

You want to make the 3rd row with `.expanded` have negative margins too, but it does not have anything to do with `.expanded`, this is a behavior you expect from any nested row.

Despite of the name (let's say we use `.forced-nested` instead of `.expanded`), the problem is we're f*** by specificity (again, huh ?). Adding an other modifier do the job for this particular situations, but will create more problems for tons of other situations.

So, for now, I only have a bad solution: Ensure both `.row .row` and `.row.expanded .row .row` is aligned (negative margins).
This increase the specificity, complexify the behavior, this is nothing but a hack for a badly declared `.row.expanded .row`, but... this is what any API base with an uncontrolled specificity will always leads to.

Let me know what you think about it.

kball:
Closing as old stranded PR

PR #9891, opened 2017-03-23
Delete dropdown click event listener on destroy

luarx:
Dropdown element does not delete **`click.zf.dropdown`** body listener when it is destroyed. This event is used to close dropdown on click on some body element.

In a **SPA application**, a dropdown with navigation links will not load a new page on navigation because it will only destroy this element.
For this reason, when old **`click.zf.dropdown`** body listener tries to access destroyed dropdown element, an error happens.
dropdown-click-listener-error

*In a non-SPA application this error happens less frequently because browser loads a new page in most navigations.


kball:
Makes sense to me

Issue #9591, opened 2017-01-03
Why Browser Support is still held for IE 9


IamManchanda:
Last few weeks, even frameworks like bootstrap have taken this big steps like 

1.) Removing IE9 Support 
2,) Make it a flexbox only grid system 
3.) RIPing out any IE9 Crutches like `X-UA-Compatible`

**We all know that semantic ui also only support IE 11+**

Web is moving forward and i feel like foundation should too ?
This will not only help make foundation better but also this will help the old lame browser to be removed from the web like what happened to IE 6-8

RamanCool:
You hit the bulls eye !

kball:
Certainly v7 should not include IE9 support.  I go back and forth about removing it in v6... my inclination would be more to make the flexbox grid the default but still maintain the ability for folks who need it to enable "compatibility mode".  @zurb/yetinauts what do ya'all think?

brettsmason:
I'm conflicted on removing Ie9. I currently have to support it at work due to lots of government clients who are unfortunately still stuck on older versions. If ie9 wasn't supported I would have to find a way to support it, or use something else.

Having said that I would love to move to newer things like flexbox and remove the old, it really is still very client dependent for me though. 

I think @andycochran is in a similar situation if I remember correctly.

JeremyEnglert:
I don't think removing it from 6 would be a good idea - as we can keep 6 available for users who need to support older browsers even after 7 has been released. 

However, I do think the percentage of IE9 users is so low it merits removing it from the next version of Foundation. Especially if this allows us to focus on Flex Box as the primary grid.

IamManchanda:
Let's do it for 7 then ..... How about adding this to roadmap ( foundation 7 ) ? 
1.) Removing IE 9 
2.) Focus on flexbox only grid system
3.) More Focus towards CSS transitions / Animations 
4.) Ripping out any IE 9 Crutches within templates or docs

IamManchanda:
cc / @kball @brettsmason @ncoden @zurb/yetinauts

IamManchanda:
Update: IE9 global browser usage is now just 0.25% ( lower than IE8 ) ...... so it makes more sense to remove IE9 Support!

kball:
I think we can conclusively say we will be dropping IE9 support in 7. For 6.X we're going to continue support at least for all existing components, and new features should degrade gracefully.

IamManchanda:
[...]

PR #9893, opened 2017-03-24
Fixed tooltip positioning for SVG elements

ASethi77:
Prior to this change, tooltips would not position correctly if I tried to create a tooltip on an SVG element. The tooltip plugin attempts to read the position to place the tooltip based on the `className` attribute of the target element. This change makes it so that the tooltip references the `className.baseVal` attribute for SVG elements, enabling tooltips to position properly.

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -72,8 +72,13 @@ class Tooltip {
    */
   _getPositionClass(element) {
     if (!element) { return ''; }
-    // var position = element.attr('class').match(/top|left|right/g);
-    var position = element[0].className.match(/\b(top|left|right)\b/g);
+
+    var elementClassName = element[0].className;
+    if (element[0] instanceof SVGElement) {
+        elementClassName = elementClassName.baseVal;
+    }
+
+    var position = elementClassName.match(/\b(top|left|right)\b/g);
         position = position ? position[0] : '';
     return position;
   };


IamManchanda:
hmmmnnn.... looks interesting 

what do you think @kball ??

dragthor:
Browser compatibility?

ASethi77:
@dragthor Looking at [SVGElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGElement) on MDN, it seems that this compatible with all versions of Chrome, FF,  Safari, Opera, and Edge. However, this change will not work on any browsers earlier than IE 9.0. As far as I can tell, that seems to be in line with the compatibility requirements in the [ZURB Foundation compatibility page](http://foundation.zurb.com/sites/docs/compatibility.html).

Is there something specific here that might raise issues with compatibility?

IamManchanda:
[Browser compatibility](http://caniuse.com/#search=svg) seems good 
Moreover we are even [welcoming ](https://github.com/zurb/foundation-sites/issues/9591#issuecomment-277500217)IE 10+ so doesn't seem an issue !

ASethi77:
Awesome! Any updates on whether or not this can be merged? Is there anything I need to do on my end?

IamManchanda:
Just wait a bit buddy!

ASethi77:
No problem! Sorry to nag, I'll try to be more patient :)

hasghari:
Why is the jQuery code commented out?
```javascript
var position = element.attr('class').match(/top|left|right/g);
```
That would automatically handle the SVG case as well.

kball:
@ASethi77 Sorry about the delay on this - was heads down on some other projects for a while. Can you provide a testcase for this, either as a visual test or in the tooltip docs page? See https://github.com/zurb/foundation-sites/wiki/Testing-Guide#visual-regression-tests

IamManchanda:
Ping @ASethi77 we are awaiting your reply

ASethi77:
@IamManchanda @kball Thanks for the feedback. Sorry I didn't reply earlier; last week was midterm week at my university. I'll take a look at the testing guide you linked and will try to get a test case for this up later this week.

kball:
@ASethi77 we're pushing towards a release candidate next Thursday - can you look at the remaining testing before then? Thanks!

DanielRuf:
Hi @ASethi77, just wanted to remind you of this PR. Let m know if you need any assistance with a testcase.

kball:
This is old and seems to have stranded waiting for a testcase. However as we're cleaning things out and this is quite simple and fully backwards compatible, I'm just going to clean up the merge conflict and merge it.

PR #9896, opened 2017-03-25
Force 'width' css attribute for sticky element (when 'is-stuck')

AlexLaforge:
The "**max-width**" attribute for a '_is-stuck_' DOM element is already forced by the Zurb Foundation Sticky library.
But the '**width**' attribute is not set, causing the element to be smaller than its original width, as it acquires the '_position:fixed_' attribute.
Forcing the '**width**' attribute ensures that the element remains visually the same in width.

### 1. The actual Foundation UNSTUCK :
1-sticky-unstuck

### 2. The actual Foundation STUCK :
2-sticky-stuck-actual-foundation

### 3. The Proposal STUCK :
3-sticky-stuck-actual-proposal

### DIFF
--- a/js/foundation.sticky.js
+++ b/js/foundation.sticky.js
@@ -310,6 +310,7 @@ class Sticky {
     }
 
     this.$element.css({
+      'width': `${newElemWidth - pdngl - pdngr}px`,
       'max-width': `${newElemWidth - pdngl - pdngr}px`
     });
 


IamManchanda:
Looks good to me 
cc @kball !

kball:
@colin-marshall can you take a look?

kball:
Upon thinking about this, I kind of think that a better solution would be to use scss to inherit the width from the container, e.g.

```
.sticky.is-stuck {
 width: 100%;
}
```

We used this approach in the recent docs revamp and it had the same outcome. The advantage of this is it is much less heavy handed than a JS solution, and easier to override if one should want to.

@AlexLaforge any reason this would not work for your usecase?

AlexLaforge:
I understand, and agree that SCSS is better than JS for this.
So why not use the same approach for the "max-width" attribute currently set in the Foundation JS ?

kball:
@AlexLaforge that can probably work, do you want to give it a try?

AlexLaforge:
Infortunately, since I posted this suggestion, I got rid of the Sticky component that was too inconsistent for my project.
But I am 99% sure that adding this static SCSS rule would have the same effect than creating it dynamically in JS ! ;)

kball:
Tested... the max-width doesn't end up working the same way (moving it from JS to scss breaks things)... I think the problem is when we move to fixed we need some (JS determined) constraint to be applied. Applying it to max width lets us then set the width to anything up to the width of the container... what we're doing with the scss here then is setting the (overridable) default.

Anyway, closing this in favor of https://github.com/zurb/foundation-sites/pull/10075

AlexLaforge:
I understantd; JS take precedence over the SCSS rules.
Thanks for your time. Hope you will have ot fixed in [#10075](https://github.com/zurb/foundation-sites/pull/10075)

PR #9915, opened 2017-03-29
Forms-reset by normalize-scss overwrite button font-family

artursopelnik:
**Problem:**
Forms-reset by normalize-scss overwrite button font-family. See on line 422, /_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss.

**Example:**
`$body-font-family: Verdana;`
`<button>Example Button</button>`
Result: Button's font-family is sans-serif.

**Solution:**
Adding $button-font-family as similar to /scss/forms/_text.scss.

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -6,9 +6,9 @@
 /// @group button
 ////
 
-/// Font family for button elements.
-/// @type String | List
-$button-font-family: $body-font-family !default;
+/// Font family of text inputs.
+/// @type Font
+$button-font-family: inherit !default;
 
 /// Padding inside buttons.
 /// @type List
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -6,7 +6,7 @@
 /// @group button
 ////
 
-/// Font family of text inputs.
+/// Font family for button elements.
 /// @type Font
 $button-font-family: inherit !default;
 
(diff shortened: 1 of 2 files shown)


scss/components/_button.scss
```diff
@@ -6,6 +6,10 @@
 /// @group button
 ////
 
+/// Font family for button elements.
+/// @type Font
+$button-font-family: inherit !default;
```
IamManchanda: Use that settings ( below review ) thing here

```scss
$button-font-family: $body-font-family;
```

scss/settings/_settings.scss
```diff
@@ -240,6 +240,7 @@ $breadcrumbs-item-slash: true;
 // 11. Button
 // ----------
 
+$button-font-family: $body-font-family;
```
IamManchanda: Setting file is generated automatically ,
You dont need to add here!

IamManchanda (changes requested):

scss/settings/_settings.scss
artursopelnik: Perfect :)

scss/components/_button.scss
artursopelnik: Perfect :)
- using $body-font-family has fix that too ...

kball:
Went ahead and made teh settings fix myself, and merged.

PR #9916, opened 2017-03-29
Fixes #9913: Removed aria-hidden and using is-active for targeting active tabs

### LINKED ISSUE #9913: [Tabs] Incorrect use of aria-hidden attribute
Visibility of individual tabs is controlled by leveraging the value of the `aria-hidden` attribute. In the current version (`6.3.0`) a tab content container is visible only if it has a `aria-hidden="false"` attribute, which represents an incorrect use of a boolean attribute.

According to the [spec](https://www.w3.org/TR/html5/infrastructure.html#boolean-attributes) all of these should be interpreted as _truthy_:
```html
aria-hidden="true"
aria-hidden="false"
aria-hidden="aria-hidden"
```

For the attribute to be _falsy_, it'd have to be either removed or:

```html
aria-hidden=""
```

Therefore the CSS rule for tabs' panels is invalid:

```CSS
.tabs-panel[aria-hidden="false"] {
  display: block;
}
```

As stated on W3C [website](https://www.w3.org/TR/wai-aria/states_and_properties#aria-hidden):

> Note: Authors are advised to avoid using aria-hidden="false" with styles or attributes that have historically prevented rendering in all modalities, such as display:none or visibility:hidden in CSS, or the hidden attribute in HTML 5. At the time of this writing, aria-hidden="false" is known to work inconsistently when used in conjunction with such features.

IamManchanda:
Hey Good evening guys
I am looking to fix #9913 

But at the moment its partial fix, 
As by default no content is showing up 
When i click on a tabs , the content get showed up and flow gets normal

Reason:
1. By default, the active content still has `aria-hidden="false"`
2. When i click on a tabs, things work normally as now the active content div has `aria-hidden` as null instead of false 

Can anyone tell me what i need to do more to solve this ??
What i want is that `aria-hidden` should be null by default too for the active content div by default!

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -280,7 +280,7 @@ class Tabs {
 
       $targetContent
         .addClass(`${this.options.panelActiveClass}`)
-        .attr({'aria-hidden': 'false'});
+        .attr({'aria-hidden': ''});
   }
 
   /**
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -65,7 +65,6 @@ class Tabs {
 
       $tabContent.attr({
         'role': 'tabpanel',
-        'aria-hidden': !isActive,
         'aria-labelledby': linkId
       });
 
@@ -280,7 +279,6 @@ class Tabs {
 
       $targetContent
         .addClass(`${this.options.panelActiveClass}`)
-        .attr({'aria-hidden': ''});
   }
 
   /**
@@ -296,7 +294,6 @@ class Tabs {
 
     $(`#${$target_anchor.attr('aria-controls')}`)
       .removeClass(`${this.options.panelActiveClass}`)
-      .attr({ 'aria-hidden': 'true' });
   }
 
   /**
--- a/scss/components/_tabs.scss
+++ b/scss/components/_tabs.scss
@@ -138,7 +138,7 @@ $tab-content-padding: 1rem !default;
   display: none;
   padding: $padding;
 
-  &[aria-hidden=""] {
+  &.is-active {
     display: block;
   }
 }
--- a/scss/components/_tabs.scss
+++ b/scss/components/_tabs.scss
@@ -138,7 +138,7 @@ $tab-content-padding: 1rem !default;
   display: none;
   padding: $padding;
 
-  &[aria-hidden="false"] {
+  &[aria-hidden=""] {
     display: block;
   }
 }


scss/components/_tabs.scss
```diff
@@ -138,7 +138,7 @@ $tab-content-padding: 1rem !default;
   display: none;
   padding: $padding;
 
-  &[aria-hidden="false"] {
+  &[aria-hidden=""] {
```
DaSchTour: wouldn't `&[aria-hidden]` be sufficient? It would just query for existence of the attribute

IamManchanda: Ohk Schuba, will look into it ... thanks a lot bro !

IamManchanda: no @DaSchTour .... its not working at all .... all div's are getting a `display: block` in that situation!

DaSchTour: Well yes. I see. It's not about existence of the attribute but about being empty. I think it would be better to fully remove it and not set it to an empty string.

IamManchanda: @DaSchTour 
No bro .... if i remove .... nothings works .... all tabs content get hidden both on click and by default
Can you pull it once and look into it??

IamManchanda:
@kball can u help me solving this sir?

kball:
Ok... so I believe the context here was that we were targeting the hidden attribute because the active class could vary, but that ends up not working due to the reasons highlighted.

Hmm...

An alternative would be to simply use the default active class to influence the hidden/unhidden, and put the burden to the user if they want to use a custom active class to make sure that it toggles the visibility.  Thoughts?

IamManchanda:
yes exactly was thinking the same @kball .... we should target `is-active` instead 

can you give me some coding tips on how do i do this ??

kball:
@IamManchanda I think it should be as simple as changing 
`&[aria-hidden=""]` to `&.is-active` in the scss... is there another piece I'm missing?

IamManchanda:
@kball Done! 

Yup stupid me :smile: , yes was so simple and i was thinking ten other things 
Also i have safely removed `aria-hidden` attribute's from javascript as doing nothing now!

I think this is good to go now .... **ready for merge**..... :smiling_imp:

js/foundation.tabs.js
```diff
@@ -65,7 +65,6 @@ class Tabs {
 
       $tabContent.attr({
         'role': 'tabpanel',
-        'aria-hidden': !isActive,
```
kball: Don't we still need to set aria-hidden in the case that we are hidden? If I'm reading the issue properly, the behavior is that we should still have aria-hidden when hidden, but have no attribute rather than a "false" when not hidden.

I think we probably need to use attr in the hidden case and removeAttr in the unhidden case, yes?

kball: or wait... I guess if default behavior is display:none that is enough to tell the browser these are hidden...

IamManchanda: >  I guess if default behavior is display:none that is enough to tell the browser these are hidden...

Yes sir, i thinked about it already and that's why i removed .... i think is good to merge if i am not missing something

kball:
Merged!  Tagging with 'Migration Needed', the migration note should be something like:

Users of `tabs` who are customizing `panelActiveClass` will need to add `display:block;` to the styles for their `panelActiveClass`.

dragthor:
Unit test failing now that `aria-hidden` is gone.

IamManchanda:
Oh that's unfortunate .... so what needs to be done now?
What would be better .... revert or adding aria hidden in js like above mentioned ??
@kball @dragthor

dragthor:
I think we should take this approach - look at the unit test, make it behave the way it should, then update the code (if necessary) to reflect proper behavior.

Owlbertz:
Just stumbled upon the failing tests as well.
The `aria-hidden` is not explicitly needed as tabs are working with `display: none`, but as most Foundation components use `aria-hidden` nonetheless, I feel like we should keep it in the codebase for Tabs as well.

So I would say we should add this back to the code.

PR #9923, opened 2017-04-02
pseudo spelling correction

dragthor:
http://www.thefreedictionary.com/pseudo

### DIFF
--- a/dist/js/foundation.js
+++ b/dist/js/foundation.js
@@ -1323,7 +1323,7 @@
 	});
 })(jQuery);
 /****************************************************
- * Method for adding psuedo drag events to elements *
+ * Method for adding pseudo drag events to elements *
  ***************************************************/
 !function ($) {
 	$.fn.addTouch = function () {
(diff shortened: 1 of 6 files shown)


dist/js/foundation.js
```diff
@@ -1323,7 +1323,7 @@
 	});
 })(jQuery);
 /****************************************************
- * Method for adding psuedo drag events to elements *
+ * Method for adding pseudo drag events to elements *
```
IamManchanda: Dist files are automatically generated 
You should be editing source files instead 

Here are the files/lines where you should edit,
=> https://github.com/zurb/foundation-sites/blob/develop/js/foundation.util.touch.js#L79
=> https://github.com/zurb/foundation-sites/blob/develop/js/foundation.sticky.js#L108
=> https://github.com/zurb/foundation-sites/blob/develop/js/foundation.sticky.js#L149

IamManchanda (changes requested):

dragthor:
Yeah, I know.  `dist` is under version control.  Maybe they should `.gitignore` it?

IamManchanda:
no,, not at all ... what about those who download css/js only version instead of compiled one ??

kball:
Dist files are updated at release time, and we dont' want them to change between releases... happy to take the comment updates for the sourcefiles alone if you can update.

IamManchanda:
@kball as @dragthor didn't reverted .... i did it .... please merge it!

kball:
👍

PR #9926, opened 2017-04-03
Set toggle button "title" to submenuToggleText

danbrellis:
When a user hovers over the toggle button, there should be a title attribute to indicate what will happen if they click the button. Adds a "title" attribute to the toggle button set to "_this.options.submenuToggleText"

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -64,7 +64,7 @@ class AccordionMenu {
 
       if(_this.options.submenuToggle) {
         $elem.addClass('has-submenu-toggle');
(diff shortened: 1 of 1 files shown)


IamManchanda (approved):
LGTM but a SEO guy might think different ... 

cc @rafibomb @kball !

kball:
Given that the text is already visible *within* the button I'm not understanding the value of adding the title attribute, can you clarify?

danbrellis:
The text inside the button is wrapper in a `span.submenu-toggle-text`, which, according to /scss/components/_accordion-menu.scss includes the `element-invisible` mixin. So, on most screens, that text is not visible.

kball:
Got it. Sounds good, merging. Thanks @danbrellis! And Congrats on first contribution!

PR #9929, opened 2017-04-04
Add browserstack api access

kball:
Adds browserstack username and key

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "4"
+  - "6"
 install:
   - npm install -g bower browserstack-runner
   - npm install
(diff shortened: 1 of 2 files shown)


IamManchanda (approved):

Deckluhm (commented):
Mistake?

kball:
Not a mistake - this is actually an open source account we got. @rafibomb set it up first which is why the accountname is that way, but it's registered under the foundation@zurb.com email address

Deckluhm:
Oh ok, that's great!

Just wanted to be sure that wasn't a mistake (since API keys and more generally credentials are pretty sensitive info).

kball:
Appreciate the check

IamManchanda:
@kball so who all can do testing thanks to this?

PR #9930, opened 2017-04-05
[Issue Template] Added a question: Browser(s) and Device(s) tested on?

IamManchanda:
I think this is important and logical!


kball:
Good call.

PR #9931, opened 2017-04-05
Fixed wrong case in code sample.

strainger:
There were two instances where the wrong case was used in the "Off-canvas Position" section of the documentation. Instead of **"Split"** the example had **"SPlit"** which would still work, but is unclean. It is also an official Off-canvas example, so it could be spreading through test code for people learning Foundation and lead to confusion. I simply made the **"P"** lowercase on lines 140 and 146. I quoted the code below.

Before Fix line 140:
`<button type="button" class="button" data-toggle="offCanvasLeftSPlit1">Open Left</button>`

Before Fix 146:
`<div class="off-canvas-absolute position-left" id="offCanvasLeftSPlit1" data-off-canvas>`


IamManchanda (approved):
Yes Sir!!
cc @kball for merge!

strainger:
Thanks!

PR #9941, opened 2017-04-09
RTL support for text alignment classes!

IamManchanda:
Inspiration => https://davidwalsh.name/text-align-start

What changes,

```scss
.text-left {
  text-align: left; // Fallback, where `start` is not supported
  text-align: start; }

.text-right {
  text-align: right; // Fallback, where `end` is not supported
  text-align: end; }
```


IamManchanda:
Closing as pushed in wrong branch locally!

PR #9942, opened 2017-04-09
[RTL] Support for text alignment classes!

IamManchanda:
Inspiration => https://davidwalsh.name/text-align-start

What changes,

```scss
.text-left {
  text-align: left; // Fallback, where `start` is not supported
  text-align: start; }

.text-right {
  text-align: right; // Fallback, where `end` is not supported
  text-align: end; }
```

### DIFF
--- a/scss/typography/_alignment.scss
+++ b/scss/typography/_alignment.scss
@@ -5,15 +5,29 @@
 @mixin foundation-text-alignment {
   @each $size in $breakpoint-classes {
     @include breakpoint($size) {
-      @each $align in (left, right, center, justify) {
-        @if $size != $-zf-zero-breakpoint {
+      @if $size != $-zf-zero-breakpoint {
+        @each $align in (center, justify) {
           .#{$size}-text-#{$align} {
             text-align: $align;
           }
         }
-        @else {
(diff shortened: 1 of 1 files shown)


strainger:
It said that `text-align: start;` was supported in IE10, would that effect the ability for Foundation to say it fully supports IE9 and newer, or would the fallback cover that?

I like the idea.

IamManchanda:
For IE9, Fallback will be covered .... @strainger 

> I like the idea.

Thanks!

kball:
@andycochran @brettsmason can one of you scss gurus review this?

andycochran:
I'm not sure what this gains us, especially since the fallback is required because of lacking [browser support](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#Browser_compatibility). The article seems to indicate that `text-align:start` is beneficial for styling elements with unknown text direction. But Foundation's alignment classes explicitly determine an alignment, regardless of the text direction. Why do the `.text-left` and `.text-right` classes need to work with RTL?

IamManchanda:
Yeah you have a very good point @andycochran .... mah bad!
I will go ahead and close this sir.

PR #9835, opened 2017-03-03
Hollow button disabled consistency fix #9832

IamManchanda:
Closes #9832

PR #9944, opened 2017-04-10
Reopen #9835, Closes #9832 - Hollow button disabled consistency fix

### LINKED ISSUE #9832: A disabled .hollow button style not consistent when hovered over
#### To reproduce:

1. Add classes `button disabled hollow` to a button element e.g. `<button type="button" class="button hollow disabled" id="mybutton">Save</button>`
2. Hover over button with the mouse

Or

1. Add classes `button hollow` to a button element and the attribute `disabled` e.g. `<button type="button" class="button hollow" id="mybutton" disabled>Save</button>`
2. Hover over button with the mouse

#### Expected:

- Button should not change in style.  The hover effect suggests an action can be taken. 
- The button should be displayed in the `hollow` style, even though disabled, to differentiate it from a standard `primary` button which is disabled

#### Happened:

__In v6.3.0__
 When the button is hovered over with the mouse, the button changes to have the background colour `$primary-color` with the text colour set to the best visibility against the new background (with `opacity: 0.25;` inherited).

__In v6.3.1__
Button background colour is set to `$primary-color` and the text colour is set to the best visibility against the new background (with `opacity: 0.25;` inherited) and doesn't change on hover (so a disabled `hollow` button is visually the same as a disabled `primary` button)

#### E.g.s:

- [Foundation Docs v6.3.0 - hollow button disabled example](http://foundation.zurb.com/sites/docs/button.html#hollow-style)
- [JSFiddle v6.3.0](https://jsfiddle.net/tymothytym/pvLqpkob/)
- [JSFiddle v6.3.1](https://jsfiddle.net/tymothytym/1yze6put/)

#### Possibly related:
#9021 and/or #9022

IamManchanda:
Reopen #9835, Closes #9832
cc / @kball @canterberrie 
Never mind kev .... no issue sir!

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -177,6 +177,11 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   &:hover, &:focus {
     border-color: $color-hover;
     color: $color-hover;
+    &.disabled,
+    &[disabled] {
+      border: $border-width solid $color;
+      color: $color;
+    }
   }
 }
 
@@ -303,11 +308,30 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
       }
     }
 
+    // Disabled style
+    &.disabled,
+    &[disabled] {
+      @include button-disabled;
+
+      @each $name, $color in $button-palette {
+        &.#{$name} {
+          @include button-disabled($color, auto);
+        }
+      }
+    }
+
     // Hollow style
     @if $button-fill != hollow {
       &.hollow {
         @include button-hollow;
         @include button-hollow-style;
+
+        &.disabled,
+        &[disabled] {
+          &, &:hover, &:focus {
+            background-color: transparent;
+          }
+        }
 
         @each $name, $color in $button-palette {
           &.#{$name} {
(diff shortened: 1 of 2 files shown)


scss/components/_button.scss
```diff
@@ -303,12 +308,31 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
       &.hollow {
         @include button-hollow;
         @include button-hollow-style;
 
+        &.disabled,
+        &[disabled] {
+          &, &:hover, &:focus {
+            background-color: transparent;
+          }
+        }
```
canterberrie: It might make more sense to move these styles to the mixin for hollow button. Is there a reason they're here instead? (Also, apologies for the confusion earlier!)

IamManchanda:
@canterberrie .... done kindly check 
Also added docs for the same!

scss/components/_button.scss
IamManchanda: No specific reason .... some hit and trial ... and it worked ... so kept it there :D 
You are right ,,,, moved it there!

canterberrie:
Thank you! Merged.

PR #9947, opened 2017-04-11
Closes #9938 - Added List Groups

### LINKED ISSUE #9938: [Component Request] Add missing List Groups!
We miss this component => https://v4-alpha.getbootstrap.com/components/list-group/

IamManchanda:
Closes #9938
This adds List Groups Component

Demo (without flexbox) => http://codepen.io/IamManchanda/pen/QvLOGx?editors=1000
Demo (with flexbox) => http://codepen.io/IamManchanda/pen/eWOeGw?editors=1000

--------------------------------------------------------------------------------------------------------

HTML for the same in the demo

```html
<ul class="list-group">
  <li class="list-group-item active">List Group Item 1 (Active)</li>
  <li class="list-group-item">List Group Item 2</li>
  <li class="list-group-item">List Group Item 3</li>
  <li class="list-group-item disabled">List Group Item 4 (Disabled)</li>
</ul>
```

-----------------------------------------------------------------------------------------------------------------

Generated CSS is just 30 lines of code!!!

### DIFF
--- a/scss/components/_list-group.scss
+++ b/scss/components/_list-group.scss
@@ -1,0 +1,148 @@
+// Foundation for Sites by ZURB
+// foundation.zurb.com
+// Licensed under MIT Open Source
+
+////
+/// @group list-group
+////
+
+/// Default background color.
+/// @type Color
+$list-group-background: $white !default;
+
+/// Default background color for active.
+/// @type Color
+$list-group-active-background: $primary-color !default;
+
+/// Default background color on hover focus.
+/// @type Color
+$list-group-hover-background: $light-gray !default;
+
+/// Default font color for list-groups.
+/// @type Color
+$list-group-font-color: $body-font-color !default;
+
+/// Default font color for disabled list-groups.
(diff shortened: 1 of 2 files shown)


kball:
@zurb/yetinauts curious your thoughts on this component... Is this fundamental enough it makes sense in the framework, or would this make more sense as a building block?

IamManchanda:
Although i have no issue porting this to building block 
but i see it more as a core component because this can be useful in more generic things instead of handful tasks !

andycochran:
This is cool. But I'd agree that it seems more appropriate as a Building Block. 

Aside: @kball Is there a path for popular Building Blocks to work their way into core?

kball:
@andycochran great question. I don't have a set answer, seems like more of a judgement call... 

That said, I'd actually lean the other way... as we make building blocks more robust and easy to integrate into a multitude of dev environments, better handle things like dependencies etc, we might consider slimming down the core framework and making more things into building blocks

PR #9948, opened 2017-04-11
[Drilldown] fixes mutateme.zf.trigger fires as often as many submenu you got beca…

designerno1:
…use it's in the _events($elem) function, and this gets bound as manu times you got submenues.

i'got a menu with 92 submenu's and wondered why the performance is so weak on clicking on the responsiveToggle, so i started searching an found that the mutateme event gets fired as many times as there are submenus, and on the drilldown then 92 times the _getMaxDims was executedthis calcs the heights and so on and is very performant, so i debugged it an moved it to the _registerEvents function

big performance improvement

@zurb/yetinauts

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -153,7 +153,6 @@ class Drilldown {
         });
       }
     });
-	  this.$element.on('mutateme.zf.trigger', this._resize.bind(this));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


kball:
Good catch

coreysyms:
First, sorry for the delay in my reply.
This is a very good catch, and something I hope more people will report on, I would have never made a submenu with 90 or 100 items and tested, real world code always is the best tester.

PR #9950, opened 2017-04-13
Dropdown menu left align in top bar right

### LINKED ISSUE #9042: Dropdown Submenu Data Alignment

```
<div class="top-bar-right">
  <ul class="dropdown menu" data-dropdown-menu data-alignment="left" >
    <li>
    <a href="#">Item 1</a>
      <ul class="menu">
     <li><a href="#">Item 1B</a></li>
      </ul>
    </li>
    <li><a href="#">Item 2</a></li>
    <li><a href="#">Item 3</a></li>
    <li><a href="#">Item 4</a></li>
  </ul>  
</div>
```

The submenu should open and the data
Item 1
Item 1A
Item 1B
should all be aligned underneath each other

**What happened instead?**

The submenu opens and data is not aligned on the left side.

Give us a link to a CodePen or JSFiddle that recreates the issue.
- [CodePen with Foundation 6 and MotionUI](http://codepen.io/rafibomb/pen/xVVGOB)
- [CodePen with Foundation 6, Flexbox grid and MotionUI](http://codepen.io/rafibomb/pen/jqqPra)

samuelmc:
Fixes #9042
I've set the default for the alignment option to 'auto' so 'left' can be used to force the alignment to the left even inside .top-bar-right.


kball:
Thanks for the PR and for including a visual test... I'm not sure the behavior I'm seeing when I try it out is what's expected though... here's what I'm seeing:

localhost_3000_dropdown-menu_left-aligned-dropdown-menu-inside-top-bar-right_html

Can we clarify what is expected to happen?

samuelmc:
It's not really intended for dropdowns on a menu item at the right edge of the screen but the carets appearing right of their item and 'Item 1B iv alpha' show that it is working. What's important is to see that the list item containing the dropdown-submenu has the 'opens-right' class instead of the 'opens-left' class even inside the 'top-bar-right'.

This is a screenshot of wat I expect see:
left-aligned-dropdown-inside-top-bar-right

To get this result I had to override the styles of the 'opens-left' class, with this PR 'Item 4' should have the 'opens-right' class instead if data-alignment="left" is specified.

kball:
Got it. Ok, I think this makes sense... I think it probably makes sense to do a rework/more explicit alignment logic similar to how I'm currently doing for Dropdown (separate from dropdown menu) in https://github.com/zurb/foundation-sites/pull/10053

However I don't know that we'll have time to do that for 6.4, and this is a step forward, so going to merge.

PR #9963, opened 2017-04-19
Home end slider

Owlbertz:
Implemented home and end keyboard events according to the specs I gathered [here](https://github.com/Owlbertz/a11y-collection/blob/master/components/slider.md#examples).

Home sets slider to minimum value.
End sets slider to maximum value.

Tests will likely fail because of #9916, but the tests I added for this specific feature should work fine.


kball:
Awesome!  Works great, tests check out locally. Merging

PR #9965, opened 2017-04-19
Move to webpack

kball:
Implements https://github.com/zurb/foundation-sites/issues/9438 

Moves all of our JS to use true ES2016 module dependencies. Uses webpack for building the full dist file, customizer generated JS files, and also a set of plugin-based distfiles for to recreate the old "drop in" JS approach.

See #9438 for more information on the design decisions and approach.


kball:
@TheLarkInn this is a first version of the ZURB Foundation moving its' JavaScript build to webpack - would you be able to take a look and see if I'm doing anything particularly stupid in my use of webpack?

DaSchTour:
I'm not quite sure about how webpack works because I never used it. But from what I've seen in module bundlers I think that entries for plugins may contain all utils imported there, if not declared as global dependencies for the module bundler.

kball:
@DaSchTour yes... that is why for the backwards-compatible (drop-innable per plugin builds), we use this section: https://github.com/zurb/foundation-sites/pull/9965/files#diff-21f8cb4967622377dfefe75eb181c026R20 :

Owlbertz:
That's a huge step in the right direction.
I will check that out during the next days.

TheLarkInn:
Hey just seeing this message. Sure!

TheLarkInn:
One thing I didn't see that I think should be called out is a `.babelrc` that has the `"modules: "false` property set so that babel doesn't transpile native imports and exports.

kball:
@TheLarkInn thanks! Updated

kball:
@DaSchTour @Owlbertz Did either of you get a chance to try this out?

js/foundation.plugin.js
```diff
@@ -0,0 +1,46 @@
...
+class Plugin {
+
+  constructor(element, options) {
+    this._setup(element, options);
+    var pluginName = hyphenate(this.constructor.name);
+    this.uuid = GetYoDigits(6, pluginName);
+
+    if(!this.$element.attr(`data-${pluginName}`)){ this.$element.attr(`data-${pluginName}`, this.uuid); }
+    if(!this.$element.data('zfPlugin')){ this.$element.data('zfPlugin', this); }
+          /**
```
Owlbertz: Indentation issue here?

js/foundation.util.keyboard.js
```diff
@@ -8,7 +8,9 @@
 
 'use strict';
 
-!function($) {
+import $ from 'jquery';
+import { rtl } from './foundation.util.core';
+let Rtl =  rtl;
```
Owlbertz: Guess such a statement could be left out?

webpack.config.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
+  entry: "./js/entries/all.js",
```
Owlbertz: Double quotes here while single quote everywhere else.

webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+  output: {
+      path: './_build/assets/js/',
+      filename: "foundation.js"
```
Owlbertz: Except here (also double quotes instead of single).

Owlbertz (approved):
I really just found a few small thing that I noticed while jumping through the files.

Overall I think that this is a good thing as it moves Foundation forward in terms of modularity.

kball:
Small items addressed, thanks for reviewing @Owlbertz! @DaSchTour did you have a chance as well? Would love to get this merged and move forward from this as a base, however a little cautious as it touches literally every JS file in the codebase. :P

js/foundation.slider.js
```diff
@@ -1,7 +1,14 @@
 'use strict';
 
-!function($) {
+import $ from 'jquery';
+import { Keyboard } from './foundation.util.keyboard';
+import { Move } from './foundation.util.motion';
+import { GetYoDigits, rtl } from './foundation.util.core';
+let Rtl = rtl;
```
DaSchTour: Not sure if this works with ES2015 and Webpack to import with `import { GetYoDigits, rtl as Rtl } from './foundation.util.core';` or if only TypeScript supports this

js/foundation.util.touch.js
```diff
@@ -3,9 +3,9 @@
 //**Done by Yohai Ararat ***************************
 //**************************************************
 
-class Touch {}
+import $ from 'jquery';
 
-(function($) {
+class Touch {}
```
DaSchTour: Should this really be closed immediately? The part there after is indented and looks like it should be part of this class.

js/foundation.util.touch.js
```diff
@@ -140,15 +140,10 @@ class Touch {}
   };
 
   Touch.init = function($) {
```
DaSchTour: Why not adding this as a class method, in class body? Can Babel handle `public static` methods?

DaSchTour (commented):
I see a lot of possibilities to refactor. But I think this should be done carefully in a later step after Transition to TypeScript.

kball:
@DaSchTour Thanks for reviewing! Totally agree that there are many more opportunities for refactoring - given how sweeping this change is already, I tried to restrict it to the changes needed to encapsulate and make things work in the new system.

I'll go through your specific points of feedback later today.

js/entries/foundation.js
```diff
@@ -0,0 +1,101 @@
...
+Foundation.MediaQuery = MediaQuery;
+Foundation.Motion = Motion;
+Foundation.Move = Move;
+Foundation.Nest = Nest;
+Foundation.Timer = Timer;
+
+// Touch and Triggers previously were almost purely sede effect driven,
+// so no // need to add it to Foundation, just init them.
+
+import { Touch } from '../foundation.util.touch';
```
gakimball: What's the reasoning behind having each file used a named export instead of a default export?

js/foundation.util.touch.js
gakimball: This module seems like it should be an object literal and not a class.

js/foundation.util.triggers.js
```diff
@@ -17,70 +18,120 @@ const triggers = (el, type) => {
     $(`#${id}`)[ type === 'close' ? 'trigger' : 'triggerHandler'](`${type}.zf.trigger`, [el]);
   });
 };
+
+var Triggers = {
```
gakimball: This is a longer-term thing, but moving to `let` and `const` would probably make some of this code more predictable. The eslint rules `pefer-const` and `no-var` can help with this, by identifying uses of `var`, or uses of `let` where the value is never reassigned.

webpack.config.js
```diff
@@ -0,0 +1,28 @@
...
+  },
+  module: {
+    loaders: [
+    ],
+    rules: [
+      // JS LOADER
+      // Reference: https://github.com/babel/babel-loader
+      // Transpile .js files using babel-loader
+      // Compiles ES6 and ES7 into ES5 code
+      {
```
gakimball: Moving to Webpack opens up the possibility to modularize Foundation more in the future. Like, move independently-useful plugins to separate repos, where they can have their own documentation and tests and so on. If we ever do that, be sure to add `exclude: /node_modules/` to this Babel rule so those dependencies aren't transpiled. That's standard practice when using babel-loader.

js/foundation.slider.js
gakimball: It's valid 👍  see <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import>

gakimball (commented):
This is fantastic! Great job pulling this together, man.

js/foundation.util.touch.js
gakimball: I think for this specific module an object literal makes more sense, because it doesn't seem like you ever actually instantiate the class itself, unless I'm mistaken.

In cases where it is a class, the `static` keyword works out of the box for methods. If you want static *properties*, you need to add the [transform-class-properties](https://babeljs.io/docs/plugins/transform-class-properties/) plugin.

js/entries/foundation.js
kball: This came out of this comment: https://github.com/zurb/foundation-sites/issues/9438#issuecomment-294953519

Essentially the idea is to be 100% explicit about our interfaces

gakimball: Ah gotcha, thanks for sending me that. I think personally I disagree with it because it makes things more verbose, but there's no need to reverse the decision. Works the same either way!

js/foundation.util.touch.js
kball: I think @gakimball is right, this is really an object literal...

These touch handlers are actually one of the more awkward feeling areas in the codebase to me. I'm leaning towards killing this and adding a hammer.js dependency (as a separate effort from this PR)

kball:
Updated based on @gakimball and @DaSchTour's feedback

js/foundation.util.touch.js
gakimball: I'd definitely advocate for this. Also, now that the codebase is in a place where the source files can (probably) take any shape, there's more options to bring in external dependencies. For Webpack users, as long as they're declared as Foundation dependencies in `package.json`, they get pulled in automatically. And then those dependencies can possibly be packed in for the bundled version.

I think at its outset, Foundation 6 had a problem with reinventing the wheel in a few cases, and one of the reasons was because we were hesitant to bring in too many non-jQuery dependencies, lest the setup process become too involved.

Hammer would be a good replacement. I'd probably also say the sticky plugin could be replaced with sticky-kit.

gakimball: *Although*, if you go down this road, consider which libraries would be included in a bundle, and which ones would still need to be included separately. For example, jQuery is not bundled in `foundation.min.js`, as its assumed you've loaded it in a separate `<script>` tag.

If you swap out some internals for things like Hammer, it's probably fine to bundle it in `foundation.min.js`. Otherwise, any time you wanted to add a new dependency, non-build-system users would need to add new `<script>` tags to load the new dependencies.

Of course, this makes it possible to have duplicate copies of libraries in your code. Like maybe you're using Foundation *and* Hammer directly, and now you have two copies of Hammer in your code. However, I'd consider this a moot point because if you really care about that stuff, you're probably using a build system.

kball: Yeah I agree - I think there's a lot of thinking we need to do about dependencies.  One of the targets with 7.0 (and I'd like to get us there via incremental change & refactor through 6 series) is to make Foundation more "pluggable", having our dependencies (including jQuery) accessed more explicitly via interfaces and wrappers, which should allow us to swap them in and out more easily.  For example, if we have a dom manipulation interface with a jquery implementation, you could then also create an angular implementation, etc.  In essence separating out the logical "this is what should change in the UI" from the implementation "this is how we do it".  Maybe we could do something similar with touch handling

gakimball (approved):

DaSchTour (approved):
Looks good :) Already looking forward to the next steps

kball:
Merged!

IamManchanda:
Wow @kball .... can't tell how awesome it is .... lets pack the web and have fun!

anselmdk:
@DaSchTour what are your reasons for wanting to move to TypeScript when already using ES6 and Webpack? Just curious :)

DaSchTour:
@anselmdk because you avoid tons of errors and mistakes by using TypeScript. I know what is happening in your application before you runtime. And it helps a lot with refactoring because you see a lot earlier in which places things may break.

anselmdk:
@DaSchTour interesting. I'm not sold yet though, as I'm not too happy going away from the JS standards, especially when they're as great as ES6, but if it's been decided upon, I guess I'll have to follow along :)

DaSchTour:
@anselmdk TypeScript is basically just ES6+ with Types and code analysis. It's not going away from JS standard. You just add definitions for the values you expect and return. You know if the code you've written actually makes sense. And the best is, you get awesome code completion in you're IDE that will never be possible with just JS.

anselmdk:
@DaSchTour ok, that sounds pretty good.

phifa:
so we are going to do a typescript rewrite? awesome!

IamManchanda:
@phifa At the moment ... That's not certain .... 
But thanks for shoutout.... This will only help us for more certainity on typescript rewrite :)

IamManchanda:
FWIW... Trying out Typescript these days and I must say @DaSchTour is right

Thanks to the type system, TypeScript helps a lot in minimizing bugs and unit tests.
+ This is not just an angular thing

- https://angular.io/guide/quickstart
- https://www.typescriptlang.org/docs/handbook/react-&-webpack.html
- https://vuejs.org/v2/guide/typescript.html

{ PS: I haven't tried angular yet, and I am not learning TypeScript for learning angular but for being better in JavaScript }

DaSchTour:
@IamManchanda I created a branch and PR for typescript migration. Take a look #10672

PR #9966, opened 2017-04-20
Add correct implementation of `transition-time` attribute.

### LINKED ISSUE #9583: [Off-canvas] Foundation 6.3 "$offcanvas-transition-length" Sass variable overlaps the "data-transition-time" plugin option.
Setting off-canvas "data-transition-time" plugin option to value other than 500. E.g : data-transition-time="300"

Off-canvas open and close transition time should follow the value set in "data-transition-time" plugin option. In this example, it should be 300ms.

**What happened instead?**
Off-canvas open and close transition time is still 500ms.

**Why this happen?**
This bug happens because the "$offcanvas-transition-length" Sass variable, which by default is 0.5s, is overlapping with the "data-transition-time" plugin option. So no matter what value we set in the "data-transition-time" plugin option, the open and close transition time will still follow the value of the "$offcanvas-transition-length" Sass variable, which is 0.5s.

Sorry I'm unable to give a link to a CodePen or JSFiddle to recreates the issue since I haven't found any Foundation 6.3 CDN

brettsmason:
This is a fix for the implementation of the `transition-time` attribute with off canvas. Also included is a visual test to confirm this works as expected.

This should close #9583

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -72,8 +72,9 @@ class OffCanvas {
       this.options.revealOn = this.options.revealOn || this.$element[0].className.match(/(reveal-for-medium|reveal-for-large)/g)[0].split('-')[2];
       this._setMQChecker();
     }
-    if (!this.options.transitionTime === true) {
-      this.options.transitionTime = parseFloat(window.getComputedStyle($('[data-off-canvas]')[0]).transitionDuration) * 1000;
+
+    if (this.options.transitionTime) {
+      this.$element.css('transition-duration', this.options.transitionTime);
     }
   }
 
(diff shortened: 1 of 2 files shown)


js/foundation.offcanvas.js
```diff
@@ -205,6 +206,12 @@ class OffCanvas {
       window.scrollTo(0,document.body.scrollHeight);
     }
 
+    if (this.options.transitionTime && this.options.transition !== 'overlap') {
+      this.$element.siblings('[data-off-canvas-content]').css('transition-duration', this.options.transitionTime);
+    } else {
+      this.$element.siblings('[data-off-canvas-content]').css('transition-duration', '');
```
kball: Does the `scss` variable still exist? If so, will this not overwrite it?

brettsmason: Yep the variable still exists. However if a user adds the data attribute, that value overrules it. Try out the visual test if you can, that will demonstrate it better than I can explain :)

kball: Got it... yeah, it works. Awesome, thanks

PR #9973, opened 2017-04-20
Update to #9711 in a way that doesn't modify underlying dependency

kball:
Addresses an issue we've been seeing where the _vendor file keeps showing diffs... the underlying issue is `_vendor` files encapsulate dependencies and should not be changed. To achieve the same effect we can utilize the cascade.


IamManchanda (approved):
LGTM .... Better to overide!

Issue #9782, opened 2017-02-15
Reveal is not closed by pressing Space


jmsfwk:
Space does not trigger a Reveal to close on Firefox Developer Edition 53.0a2 (2017-02-11) (64-bit).

#### How to reproduce this bug:

1. Create a Reveal with a Close Button
2. Focus on the Close Button
3. Press <kbd>Space</kbd>

#### What should happen:

Reveal should close and return focus to the element focused before the Reveal was opened.

#### What happened instead:

The Reveal stays open and focus is returned to the element focused before the reveal was opened.

#### Test case:

https://codepen.io/jmsfwk/pen/dNrJKa

dragthor:
Unsure about this request.  Fairly sure that Zurb 5.x did not have this behavior.

jmsfwk:
When used with a callout Close Button is activated by <kbd>Space</kbd>.

https://codepen.io/jmsfwk/pen/ggJogd

IamManchanda:
Yes callout's have these ..... i feel like this should be there in modal close too considering accessibility!

kball:
@sethkane @Owlbertz as our accessibility experts, what do you think?

sethkane:
It depends on how you are using the close element in the HTML.

If it is a <a href> then space bar does not natively using the space bar to trigger an event.
If you are using a <button> then the space bar has a native binding for that trigger.

The W3C has these guidelines for Modals https://www.w3.org/TR/wai-aria-practices/#dialog_modal

From the two examples given in this thread it appears that there might be some bad binding on the <button> elements inside of reveal.

FYI:
Buttons https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role space bar should trigger the event.

kball:
@Owlbertz do you want to tackle this one?

Owlbertz:
Should be solved via #9977.

PR #9977, opened 2017-04-21
Reveal space close

Owlbertz:
Addresses #9782.

Refactored the component to not use special keyboard listening for opening and closing.
Normal event triggering with `click` event being emitted using space and enter is used now.
Added test for clicking close button.

Asking for some feedback as I removed the whole special keyboard handling for enter and space... @kball @jmsfwk

### DIFF
--- a/docs/pages/reveal.md
+++ b/docs/pages/reveal.md
@@ -32,7 +32,7 @@ To create a modal, add the class `.reveal`, the attribute `data-reveal`, and a u
 You'll also need a way to open the modal. Add the attribute `data-open` to any element. The value of `data-open` should be the ID of the modal.
 
 ```html_example
-<p><a data-open="exampleModal1">Click me for a modal</a></p>
+<p><button class="button" data-open="exampleModal1">Click me for a modal</button></p>
 ```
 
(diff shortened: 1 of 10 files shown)


kball:
Checking out this branch, what I'm seeing is that now following the behavior in #9782 (open reveal with close button, tab to close button, hit space) what happens now is that the modal closes (good!) but I've lost focus, in fact my keyboard focus appears to be nowhere and I'm unable to navigate with the keyboard.

Owlbertz:
That's correct, I missed the part where focus should be moved back to the anchor when the modal is closed.
I came up with a solution by simply moving focus back to the anchor within `close()`, but that will cause the outline to show - which I know will annoy most people when using a mouse.

My question is now on how to progress with that. In my opinion there are these options:
- Check with JavaScript and what input if the focus should be moved to the anchor. However, what is the default case if what input is not present?
- Always move back focus on close and disable outline for links like we do for buttons already via CSS:
```CSS
 [data-whatinput='mouse'] button {
    outline: 0; }
```
- Use buttons in the docs as anchors to encourage correct usage here. As the trigger does not lead to another page, using a button instead would be semantically correct imho. This way the existing CSS (see above) would apply when always setting back focus on `close()`.

I prefer the last suggestion.

kball:
@Owlbertz I don't see any reason not to go with the last suggestion. Go for it!

Owlbertz:
@kball I updated the PR.

kball:
Beautiful, works like a charm!

PR #9981, opened 2017-04-21
Static Gutters documentation

### LINKED ISSUE #8508: [Grid] Responsive Gutters with Unresponsive Gutters

andycochran:
Closes #8508.

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -284,9 +284,7 @@ In order to work around browsers' different rounding behaviors, Foundation will 
 
 ### Gutters
 
-<div class="warning callout">
-  <p>Responsive gutters were added in Foundation 6.1.</p>
-</div>
+#### Responsive Gutters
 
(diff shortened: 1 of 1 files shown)


kball:
Awesome! Thanks @andycochran!

Issue #9852, opened 2017-03-08
Foundation 6 tooltip cursor


git-hubber:
Is it possible to re-introduce the scss variable back into settings:
$has-tip-cursor-type

I don't like that the 'help' cursor is now hardcoded and somewhat 'forced' upon us.

when I override it as 'inherrit' with my own .has-tip class,it has very mixed results, especially in FF with links with tooltips.

IamManchanda:
Can you provide a test case

You can fork these below codepen's to provide a test case.

6.3.0
- [CodePen with Foundation 6.3.0 and MotionUI](http://codepen.io/IamManchanda/pen/vgemXR)
- [CodePen with Foundation 6.3.0, Flexbox grid and MotionUI](http://codepen.io/IamManchanda/pen/xgXdRq)

6.3.1
- [CodePen with Foundation 6.3.1 and MotionUI](http://codepen.io/IamManchanda/pen/LWGZxR)
- [CodePen with Foundation 6.3.1, Flexbox grid and MotionUI](http://codepen.io/IamManchanda/pen/zZrBEv)

Deckluhm:
@IamManchanda: there is no "test case" to provide here since it's a feature request.

@git-hubber: not sure what the current policy on SASS variables is (if there is one?) but I think it's pretty safe to say that you can open a PR for this.

IamManchanda:
Yeah but i was requesting a test case of the feature only!

kball:
@git-hubber or @Deckluhm would either of you like to submit a PR?

JeremyEnglert:
Looks like this was fixed and merged. Closing this issue. 

https://github.com/zurb/foundation-sites/pull/9983

PR #9983, opened 2017-04-22
[Tooltip] Add new $has-tip-cursor variable to customize cursor

Deckluhm:
Implements #9852.


IamManchanda (approved):
I think this makes sense 
cc / @kball

kball:
Agreed. Merged! Thanks @Deckluhm

PR #9984, opened 2017-04-22
[New Feature] Closes #9870 - Added a combo pattern `website`

### LINKED ISSUE #9870: [Abide] URL Pattern not taking **www** into account
URL Pattern only takes **http** and **https** into account
but there should be a way to add most common **www** into account

**1. Step one**
Create a test case similar to
http://codepen.io/IamManchanda/pen/MpoaJY

**2. Step two**
Go to URL Pattern field

**3. Step three**
Type **www.facebook.com**

**4. Step four**
Press on that dreaded **Submit** Button

#### What should happen:
This shouldn't throw error

#### What happened instead:
This throws error

IamManchanda:
Closes #9870 - Added a combo pattern `website`
Poke @kball for review and merge!

Changes:
- Added a pattern `website` which is basically a combo of both `domain` and `url` pattern
- Also updated the docs!

website-pattern

Here are some examples which will be allowed for this `website` pattern

- [http://zurb.com](#)
- [https://zurb.com](#)
- zurb.com
- [www.zurb.com](#)


docs/pages/abide.md
```diff
@@ -10,7 +10,7 @@ tags:
 
 ## Abide Demo
 
-These input types create a text field: `text`, `date`, `datetime`, `datetime-local`, `email`, `month`, `number`, `password`, `search`, `tel`, `time`, `url`, and `week`. Note the use of the novalidate attribute to disable any browser validation that could conflict with Abide.
+These input types create a text field: `text`, `date`, `datetime`, `datetime-local`, `email`, `month`, `number`, `password`, `search`, `tel`, `time`, `url`, `website` and `week`. Note the use of the novalidate attribute to disable any browser validation that could conflict with Abide.
```
Deckluhm: `website` isn't a valid input type: https://developer.mozilla.org/fr/docs/Web/HTML/Element/Input#attr-type

Deckluhm (changes requested):

docs/pages/abide.md
IamManchanda: Oh sorry ... @Deckluhm ... mah bad ,,,,

IamManchanda:
@Deckluhm Thanks for your review ... i have updated ... check now!

Deckluhm (approved):

docs/pages/abide.md
```diff
@@ -48,8 +48,8 @@ These input types create a text field: `text`, `date`, `datetime`, `datetime-loc
   </div>
   <div class="row">
     <div class="medium-6 columns">
-      <label>URL Pattern, not required, but throws error if it doesn't match the Regular Expression for a valid URL.
-        <input type="text" placeholder="http://foundation.zurb.com" pattern="url">
```
kball: We should probably still highlight the URL pattern in the docs - can we add website as an additional rather than a replacement?

IamManchanda: Ohkk !!! Sure!

IamManchanda:
done @kball .... please check again!

kball:
Looks good! Nice work @IamManchanda

PR #9985, opened 2017-04-22
Hotfix/menu vertical align center

### LINKED ISSUE #9955: [6.3.1] Menu that is vertical will not align-center with flexbox

1. Ensure flexbox is enabled
2. Create a menu with the following classes .vertical and .align-center

#### What should happen:
Vertical Menu should be center aligned.

#### What happened instead:
Vertical Menu is aligned to the left

#### Test case:

http://codepen.io/anon/pen/eWmdJq

IamManchanda:
This closes #9955

### DIFF
--- a/scss/components/_flex.scss
+++ b/scss/components/_flex.scss
@@ -1,9 +1,15 @@
 @mixin foundation-flex-classes {
   // Horizontal alignment using justify-content
-  @each $hdir, $prop in map-remove($-zf-flex-justify, left) {
+  @each $hdir, $prop in map-remove($-zf-flex-justify, 'left') {
     .align-#{$hdir} {
-      &,
-      &.vertical.menu > li > a { // Specific for Vertical Menu
+      @include flex-align($x: $hdir);
+    }
+  }
+  
+  // Horizontal alignment Specifically for Vertical Menu
+  @each $hdir, $prop in map-remove($-zf-flex-justify, 'left', 'justify', 'spaced') {
+    .align-#{$hdir} {
+      &.vertical.menu > li > a { 
         @include flex-align($x: $hdir);
       }
     }
--- a/scss/components/_flex.scss
+++ b/scss/components/_flex.scss
@@ -2,7 +2,10 @@
   // Horizontal alignment using justify-content
   @each $hdir, $prop in map-remove($-zf-flex-justify, left) {
     .align-#{$hdir} {
-      @include flex-align($x: $hdir);
+      &,
+      &.vertical.menu > li > a { // Specific for Vertical Menu
+        @include flex-align($x: $hdir);
+      }
     }
   }
 
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -366,6 +366,9 @@ $menu-border: $light-gray !default;
         display: inline-block;
         vertical-align: top;
       }
+      &.vertical > li > a {
+        justify-content: center;
+      }
     }
   }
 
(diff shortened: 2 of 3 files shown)


IamManchanda:
Here are the Codepen's
`.menu-centered > .menu.vertical` => http://codepen.io/IamManchanda/pen/aWzpNP
`menu.vertical.align-center` => http://codepen.io/IamManchanda/pen/OmPbeM
`menu.vertical.align-right` => http://codepen.io/IamManchanda/pen/pPvRwL

-----------------------------------------------------------------------------------------------------------------

@brettsmason you are working these days with flexbox 
so can you take a quick look and review this up ??

kball:
@brettsmason could you take a quick look at this?

brettsmason:
There's some pretty nasty specificity issues going on: `.align-#{$hdir}.vertical.menu > li > a` - @ncoden would have a heart attack 😆
Not sure how this would integrate into the big WIP menu rework either?

IamManchanda:
@brettsmason So what's the best way around?? 
Do we need to create classes for li and a 
or something else.

Secondly  .... I am quite sure we need to fix this!

IamManchanda:
Also shall we remove this from a map 
And create align-right and align-center individually?

brettsmason:
@IamManchanda Apologies for the short, not helpful reply 😄 I was in a rush at the time.
So looking at the problem more in depth, the issue is that currently the menu system has really high specificity, which is why i wanted to rewrite it in the first place.

At the moment this is the problem (which I know you know!):
```
.menu.vertical > li > a {
    justify-content: flex-start;
}
```
With the reworked menu, it wont have that specified: https://github.com/brettsmason/foundation-sites/blob/683b1c6639a00008373c9af77711b87bf8b9aabe/scss/components/_menu.scss#L139 (as its not needed) -  see https://github.com/brettsmason/foundation-sites/blob/683b1c6639a00008373c9af77711b87bf8b9aabe/scss/components/_menu.scss#L392

So I'm not sure if this patch will actually be needed looking at it some more, as the rework should of solved it already (assuming it can be merged to 6.3.2 too). I'll pull my menu rework down and do some testing to confirm.

Again apologies for the bad initial reply!

IamManchanda:
Ohk closing this in that case @brettsmason and let me join forces in there with you at menu-rework!

brettsmason:
@IamManchanda 👍 Yes please!

kball:
The rework is unlikely to go into 6.3.2; patch releases are not a good place for major rework, they should be just bug fixes with low risk/surface area.

I'm going to reopen this for a point fix for 6.3.2

IamManchanda:
Ohk @kball .... For the sake of 6.3.2 .... yes we should do this
What do you think .... shall we remove it from the map or keep it there BTW ??

kball:
will need to look in more detail tomorrow - right now jotting down notes and keeping eye on the kids at the same time. ;)

IamManchanda:
Hahaahaaa ... Enjoy! I have added test case btw just now!

rafibomb:
So I think this solution is great for 6.3.2 or for a bug fix purpose. I would say let's make sure this use case is accounted for in the menu re-work that @brettsmason is doing - which be reading the code - looks like it is!

This tests out good! So merging in!

rafibomb:
@kball can you backport this to 6.3.2 please?

kball:
Done!

IamManchanda:
@brettsmason @kball @rafibomb 
Good news i solve that in 6.4 .... menu-rework too .... cheers!
https://git.io/vHIsL

screenshot from 2017-05-23 00-46-51

PR #9988, opened 2017-04-23
Nested Off-Canvas

SassNinja:
Hi,

I've been facing a great problem since the off-canvas adjustment in v6.3 what held me off upgrading.
As I already mentioned in the forum I'm using an element on the page (filters etc.) as usual page content for desktop and move it into off-canvas for mobile.
To do this I've nested the off-canvas in the content container and added some CSS to override the off-canvas styles for desktop.

But the latest version doesn't enable this nesting anymore and requires off-canvas & content to be at the same level (sibling).
Therefore I've worked over the whole feature and added support of nesting.

To achieve this I'm writing the position / state classes on the content container, so I'm able to use this within the SCSS file (since an alternative to the sibling selector is required). I also considered the cases that there are multiple off-canvas elements in the same content container sharing the same position (without overlapping if opened) and that the off-canvas element isn't nested in the content but gets bound by an ID (data-content-id).
Apart from that I've added an in-canvas mixin that overrides the off-canvas styles for specific breakpoints so the nested off-canvas feature may get used to avoid duplicate content (as mentioned in the beginning).

The things I'm still not happy with:
- position fixed doesn't seem to work on Windows if nested
- I ended up in long selectors and some redundancy in the SCSS

**It would be great if someone review my code and let me know what he/she thinks.
I really hope this feature will be part of v6.4 so I may upgrade some large projects.**

Best regards
Kai

p.s. maybe it is worth mentioning that my changes are supposed to preserve backwards compatibility so there shouldn't be any issues after upgrading an existing project


IamManchanda:
cc / @brettsmason

brettsmason:
@SassNinja Thanks for putting this together. Strangely I actually put something together just after 6.3 came out which allowed you to target different content, but never got around to finishing it!

Could you either put together a CodePen showing what this achieves, or a visual test (test/visual/off-canvas) so I can see what it does properly?

SassNinja:
@brettsmason sure, I've pushed a visual test page so you can see it in action
Besides I've fixed some bugs I spotted:
- overlay z-index issue what appeared when nesting
- transition end listener that didn't work when nesting since the off-canvas element doesn't get transformed in this case (I solved it by listening on the content container)

I'm looking forward to getting your feedback.

brettsmason:
@SassNinja Awesome thanks for putting that together, that will make it a lot easier to work everything out 👍. I'll take a full look over the next few days and get back to you.

Also for reference this is what I put together after 6.3 was released: http://brettmason.co.uk/offcanvas/content-option/

SassNinja:
ok @brettsmason , good to hear you're on it :-)

Apart from that I've noticed the git history (below your post) is quite messy now – sorry for that!
I saw there were conflicts with the current state of the develop brach and decided to pull & rebase the official repo into my fork via sourcetree (obviously a bad idea).
After solving dozens of conflicts I've noticed I just had to use the github web interface to solve it...

IamManchanda:
@SassNinja Would you mind closing this and creating a New PR instead ??

SassNinja:
@IamManchanda closing and opening another PR doesn't seem to result in a cleaner history. I think the merging/rebasing caused this.

I would like to avoid reopening this PR since all the comments are 'lost' then and I probably have to delete my fork and recommit everything from scratch.
Or do you think this should be done anyway?

brettsmason:
I'm afraid I don't know a lot about Git - @kball what would you suggest? Is there a way to fix this or is a new PR the only way to go?

kball:
Gosh, this history is kind of tangled, with what looks like the duplicates of the same commits both at the beginning and end, and then a bunch of rebased commits in the middle, which is why git is having trouble resolving it. Not entirely sure how to revise it on this branch.  It's further complicated by the fact that the branch is named develop... My recommendation is indeed to open a fresh PR (and include a link to this one so the comments/history are findable, if not visible).

In order to not have to do things from scratch, the way I'd do this is set up a new branch from the current state of develop, and then cherry-pick over the commits you made in order... something like this

```
git fetch
git checkout origin/develop
git checkout -b nested-off-canvas-new-branch
git cherry-pick 65099aa7e8eb9a30bf3e9533936e5587d1d136f5
git cherry-pick 2d8fa970036550011a0a7c3d8d32664ca13decd5
git cherry-pick 84eb54976f4f2b9a68875bc8e3bfaca471c034d5
git cherry-pick 914c02f11a61fe9cdd9b541bfcf16abb26178a5b
```

I'm not 100% sure those are the right (and complete) set of commits, as I'm finding the history hard to follow, but you could start with that and see how it looks

SassNinja:
Ok @kball I'll close this PR and open a new one.

But I think the easiest and cleanest solution is deleting my fork and reforking the repo. I'll just take all of my latest code and will put it into one commit instead of restoring everything commit-by-commit. Since many of them were fixes it shouldn't be problem to have one great clean commit.

PR #9990, opened 2017-04-26
Rewrote onImagesLoaded to prevent loading images multiple times. Fixes #9932

### LINKED ISSUE #9932: onImagesLoaded results in images loading multiple times

1. Use equalizer or something that uses an onImagesLoaded callback with images
2. If not seeing the bug, enable network throttling and disable cache

#### What should happen:
Images in a page should only be loaded once. 

#### What happened instead:
onImagesLoaded is using the following code to detect when images have loaded:
```
if (this.complete || this.readyState === 4 || this.readyState === 'complete') {
  singleImageLoaded();
}
// Force load the image
else {
  // fix for IE. See https://css-tricks.com/snippets/jquery/fixing-load-in-ie-for-cached-images/
  var src = $(this).attr('src');
  $(this).attr('src', src + (src.indexOf('?') >= 0 ? '&' : '?') + new Date().getTime());
  $(this).one('load', function () {
    singleImageLoaded();
  });
}
```

The problem is that when that's called before images have loaded (which, if it's not, it doesn't even need to exist), even on non IE browsers, it switches to using the code in the else block. This causes it to load images two times, doubling the required bandwidth for images.  If the same image is used multiple times on a slow CPU, the Date().getTime() can even return different dates, resulting in the image being loaded even more than twice.

#### Test case:

http://codepen.io/rlhawk1/full/Mpdrao/
image

RLHawk1:
Changed onImagesLoaded function to use load and error events on a detached image element to detect when images are loaded rather than changing the source on existing elements which would cause images to be loaded multiple times as described in #9932.

The method I've implemented is used in https://github.com/alexanderdickson/waitForImages and https://github.com/desandro/imagesloaded/, both popular libraries for doing this.

My version can be viewed and tested at http://codepen.io/rlhawk1/pen/OmbvGw

### DIFF
--- a/js/foundation.util.timerAndImageLoader.js
+++ b/js/foundation.util.timerAndImageLoader.js
@@ -58,19 +58,22 @@ function onImagesLoaded(images, callback){
     callback();
   }
 
-  images.each(function() {
+  images.each(function(){
     // Check if image is loaded
-    if (this.complete || (this.readyState === 4) || (this.readyState === 'complete')) {
+    if (this.complete && this.naturalWidth !== undefined) {
       singleImageLoaded();
     }
-    // Force load the image
     else {
-      // fix for IE. See https://css-tricks.com/snippets/jquery/fixing-load-in-ie-for-cached-images/
-      var src = $(this).attr('src');
-      $(this).attr('src', src + (src.indexOf('?') >= 0 ? '&' : '?') + (new Date().getTime()));
-      $(this).one('load', function() {
+      // If the above check failed, simulate loading on detached element.
+      var image = new Image();
+      // Still count image as loaded if it finalizes with an error.
+      var events = "load.zf.images error.zf.images";
+      $(image).one(events, function me(event){
+        // Unbind the event listeners. We're using 'one' but only one of the two events will have fired.
+        $(this).off(events, me);
         singleImageLoaded();
       });
+      image.src = $(this).attr('src');
     }
   });
 


kball:
Looks good, and on testing appears to work. Nice work @RLHawk1! Merging!  Congratulations on your first PR merged into Foundation

RLHawk1:
First code contribution to any public project! Feels good. Thanks for encouraging me to do it!

kball:
Awesome! It's a good one too - we'd love to do a public shoutout - do you have any public profiles other than your github account that you'd like referenced?

sanderblue:
Thanks a bunch for this PR! Just adding a note that this [line](https://github.com/zurb/foundation-sites/blob/develop/dist/js/foundation.js#L1228) of code is currently causing noticeable unnecessary load on our CDN servers due to the query string negating caching of our images. 

Regarding the inline comment in the code, see the first comment on the referenced link [here](https://css-tricks.com/snippets/jquery/fixing-load-in-ie-for-cached-images/#comment-73093). 

I'm not sure what the protocol is for expediting bug fixes, etc, but this seems like something that might fall into that category. I also understand that there's probably a solid process for this. 😃

kball:
@sanderblue We're pretty slammed on a number of fronts so this probably won't get out until we're doing 6.4 and a 6.3.2 patch in about a month, but you should be able to sub in this update ahead if you need to. Exactly how you'd do that will depend on how you're doing a build; if you'd like help making that happen email me at kball (at) zurb.com

sanderblue:
@kball No worries. Totally understand the nature of these sorts of things. We went with a temporary in-house fix for the time being. Looking forward to the 6.3.2 patch and 6.4 releases. Happy to help with some tickets if needed. Cheers.

Issue #9672, opened 2017-01-18
Off-canvas position top bug


karimramadan:
In an off canvas with a position 'top' if the content inside the off canvas is taller than 250px the bottom border (.position-top.is-transition-push::after) is floating over.

![off-canvas-bug](https://cloud.githubusercontent.com/assets/20888541/22076099/5490ef9e-ddb7-11e6-8826-8f8805cf82f9.png)

ncoden:
@brettsmason Can you take a look at this ?

brettsmason:
Yep will do!

brettsmason:
I've setup a CodePen for this issue: https://codepen.io/brettsmason/pen/VPWeGX
It's an odd one - I'm not sure how we can go about fixing this. @ncoden if you have any thoughts that would be great.

webdevan:
How bout just removing the :after styling from Foundation all together?

This gradient isn't serving any purpose beyond aesthetics and Foundation has always been very aesthetically un-opinionated (which is great) so I propose just removing the gradient from offcanvas and let users of Foundation add their own CSS gradient (if they want one).

kball:
I think @webdevan's point is a good one. I believe it was added for backwards compatibility, but at the very least this should be separated out and made optional

brettsmason:
I would agree with removing the shadow too.

@kball it is optional at the moment - see here: https://github.com/zurb/foundation-sites/blob/develop/scss/components/_off-canvas.scss#L210

However its probably not obvious. I would be happy to remove it all together, which was my plan originally as I thought it was too opinionated.

Let me know what you would like to do and I can make some changes if needed.

kball:
Lets remove it for 6.4, and add a migration note.

brettsmason:
@kball I've removed this in my branch and am putting together a PR.
Where do you think would be suitable on the off canvas page to add a migration notice/are there any similar examples?

kball:
@brettsmason I was actually thinking more a migration note for the release notes... That said, the off-canvas docs page already has a migration section for migrating to 6.3, you could add a note in there as well.

kball:
@brettsmason did you end up making this change as a PR? I thought I saw it but now I'm not finding it

SassNinja:
I've fixed this in my PR https://github.com/zurb/foundation-sites/pull/10003, precisely in commit https://github.com/zurb/foundation-sites/pull/10003/commits/0130a3fdec4cc08f1bb236e05386790b157a697f

[...]

PR #10003, opened 2017-05-03
Nested Off-Canvas II

SassNinja:
As discussed in the original PR (whose history got messed up) I've recommitted everything and opened this new PR.

For reference check the original post:
https://github.com/zurb/foundation-sites/pull/9988

@brettsmason again, would be great to get your feedback on my code :)

Best regards
Kai


brettsmason:
@SassNinja thanks for putting together a clean PR!
Sorry I'm struggling for time at the moment, but I'll do my best to review this in the next few days.

brettsmason:
@SassNinja sorry for the delay in looking over this! Thanks a lot for putting this together 😄 

So firstly, from what I understand the aim of this PR is:
- Allow targetting non sibling content when opening an off canvas (remove reliance on sibling selector)
- Add mixin/class to allow off canvas content to revert the off canvas styling

First thing I noticed is the `nested right off canvas` demo has an issue with the content not showing in the off canvas until the opening animation has finished, and vanishing as soon as the closing animation starts.

One thing I'm not sure about is making the `off-canvas-container` required again. There are situations where I think they may not be needed, so would be nice to keep it as simple as we can - it was a concious decision on my part to try and remove that requirement with the rewrite.

From the old work I did here: http://brettmason.co.uk/offcanvas/content-option/foundation.offcanvas.js you will see I took a slight different approach. Get the position class from the `off-canvas` element and add a class to the target content so we can transform it (http://brettmason.co.uk/offcanvas/content-option/_off-canvas.scss).

I'll continue to look over it as I haven't looked at off canvas for a few months so am a bit rusty again 😄 Have a look at what I put together too and see if any of that code could be used to perhaps simplify the approach a bit.

Good job so far though 👍

SassNinja:
@brettsmason thanks for your feedback so far!

You're right regarding the aims of my PR. Several times I have been facing the problem I need some parts of the page to be usual content for desktop and move into off-canvas for medium down (without duplicate content!).
So I've created the in-canvas class to undo the off-canvas behaviour for specific breakpoints and I've made the off-canvas more flexible by enabling a nesting of element in content.

The things you noticed:

1) indeed, the nested overlay element was jumping on transition end what seems to be a CSS issue – I've fixed this now (by replacing transform:none with transform:translate(0,0) )
2) I don't understand what you mean with off-canvas-container dependency since there's not such a class... Would you explain this more detailed?

Apart from that I've checked you're code.
Seems you've had similar thoughts, as e.g. adding a direction class to the content, based on the element and to offer more than the siblings selector.

But compared to your approach I've used the position class (not the transform property/matrix) and I haven't only limited the content part to the open method but initialized the content as this.$content in the setup. My intention was to fully integrate the content.
Besides I've added a third option to find the content so there are 3 possibilities to find it:
1) by content id (if specified in options)
2) by sibling (as done currently & due to backwards compatibility)
3) by closest match (fallback)

When you test my latest code you'll probabely notice some more issues:

Issue_1:
The nested off-canvas isn't fixed anymore but scrolls with the content what is probably caused by the nested structure. The only fix for this that comes in my mind is to either force a scroll to top when nesting or to read the viewport's height and set it to the element. But I'm not quite happy with such a workaround...
**Do you have another way in mind to fix this issue?**

Issue_2:
After some cross browser testing I've spotted a strange issue in the Internet Explorer where the nested push off-canvas is not working due to a different behaviour of the transform:translateX(-250px) of the .off-canvas-content .position-left.off-canvas {}
You can see this when opening the visual test (nested.html) and clicking on the 'Nested Left Push' button in IE (Windows 7). To fix this I would have to change it into transform:translateX(0px) but only for the IE!
**Any idea how to fix this without a dirty browser hack?**

Apart from that I'm quite happy with my code as it doesn't seem to break existing code (backwards compatibility was important to me) and offers much more flexibility for 'hardcore off-canvas users' cause it lets you also move several elements into the same off-canvas side. Regarding your code I haven't spotted something important that I've missed or should take over (at most the naming). In case I'm missing anything, just let me know.

I'm looking forward to getting more feedback from you once you've done reviewing my code and once the derusting is finished ;)

brettsmason:
@SassNinja I've pulled that down and the bug I mentioned is now gone 👍 

To answer your points:

1. Unfortunately there is no solution for this other than a JS hack, which I'd rather avoid if at all possible.
See: http://stackoverflow.com/questions/15194313/transform3d-not-working-with-position-fixed-children for details, but basically a `fixed` position element looses its fixed position within a transformed parent. The only way to get around this would be to change the off canvas to not use transforms but use positioning instead (eg `left: -250px) etc. However this doesn't give as good performance as transform (especially on mobile), hence why I chose to use this rather than the old positioning of pre 6.3 (there were lots of bugs to do with this).

2. I've tried the visual test in IE10 and IE11 - the off canvas doesn't seem to show for me at all. The overlap example is fine, but the other 2 don't show up.

I'm not totally sure at this point that its going to be possible to get a cross browser compatible version that allows nesting flawlessly. Removing the siblings reliance is definitely a good idea though.

Version 6.4 should be released in June, so would be good to get this included in some form!

SassNinja:
@brettsmason yeah, I worried about that. I agree regarding the benefits of transform: hardware acceleration is definitely more important!

I've adjusted the JS now so a fixed position gets emulated for a nested off-canvas element (by setting height & top value based on the viewport's height and current scrollTop).
Would be great if you or @kball can take a look at that commit (37f6dfbbe190569f092fafaf076c2dcd1ac15906) since I'm not sure if my resize listener is the best solution.
(I was thinking about the mutation feature but not sure if it's build for this purpose)

The fixed issue seems to be solved now in my eyes. I've also added a new inner box shadow for a (nested) push off-canvas element because the ::after solution doesn't work if the element's height is greater than the viewport's height (non-nested elements seem to have the same issue).
I haven't tested this cross-browser but if it's fine, it would be probably a good idea to use this inner shadow for all push elements, no matter if nested or not.

The only issue that remains is the IE issue. I haven't found a solution for this yet but I'm working on it.

June as release date sounds good to me :+1: 
I'll do my best to finish the nested feature in time since I'm needing it myself for every project. It would be really helpful for me to use the official foundation version via bower instead of locally adjusting the code.
Of course any help is appreciated!

SassNinja:
I think I've got it!
Apparently the IE has a problem with position fixed on nested off-canvas elements. Since it doesn't work and gets emulated by JS anyway, I've explicitly set it now. Seems to work :)

p.s. 
do I have to mention that I hate the fuc**** IE ?!

SassNinja:
@brettsmason did you already manage to review my latest code?
Would be great to know if you can see any issues I've missed.

brettsmason:
@SassNinja Apologies been working on getting the new grid sorted out and forgot!
I'll make sure I look this evening and come back back with feedback 😄

scss/components/_off-canvas.scss
```diff
@@ -266,16 +340,35 @@ $content: $maincontent-class
...
+  }
+
+  // backwards compatibility (prior to v6.4)
   & ~ .#{$content} {
     margin-#{$position}: $offcanvas-size;
   }
 }
 
+/// Overrides the off-canvas styles
+@mixin in-canvas() {
```
brettsmason: Is there a need for `!important` here? Generally we try and avoid the use of this whenever possible.

brettsmason:
@SassNinja I've tested the updates you have made in a variety of browsers, including IE9. I'm pleased to report everything works great 👍 

A few points/questions:
- Use of `!important` - can this work without? We try and avoid this if at all possible
- I've noticed off canvas is now set to `visibility: hidden` when closed - what was the reason for that?
- Shadow - I was actually planning on removing this because of this issue: https://github.com/zurb/foundation-sites/issues/9672 - do you think your changes would also of fixed this?

I'm going to invoke the power of @kball to take a look over the JavaScript as hes far better than me with that 😆

js/foundation.offcanvas.js
```diff
@@ -50,13 +53,30 @@ class OffCanvas extends Plugin {
...
       .attr('aria-controls', id);
+    
+    // Get position by checking for related CSS class
+    this.position = this.$element.is('.position-left, .position-top, .position-right, .position-bottom') ? this.$element.attr('class').match(/position\-(left|top|right|bottom)/)[1] : this.position;
+
+    // Find off-canvas content, either by ID (if specified), by siblings or by closest selector (fallback)
+    if (this.options.contentId) {
+      this.$content = $('#'+this.options.contentId);
+    } else if (this.$element.siblings('[data-off-canvas-content]').length) {
+      this.$content = this.$element.siblings('[data-off-canvas-content]');
```
kball: Does this need a `.first()` as well?

js/foundation.offcanvas.js
```diff
@@ -50,13 +53,30 @@ class OffCanvas extends Plugin {
...
+    } else {
+      this.$content = this.$element.closest('[data-off-canvas-content]').first();
+    }
+
+    // Assume that the off-canvas element is nested if it isn't a sibling of the content
+    this.nested = this.$element.siblings('[data-off-canvas-content]').length && !this.options.contentId ? false : true;
```
kball: I'm confused here... what is this trying to do? First off, since our outcomes are booleans I think this would be easier to read if we just invert the logic and get rid of the boolean, so:
```
this.nested = this.contentId || !this.$element.siblings('[data-off-canvas-content]');
```

That said, I don't understand why this is true... we're always nested if we have a contentId? What is the exact case we're trying to account for here?

js/foundation.offcanvas.js
```diff
@@ -65,9 +85,9 @@ class OffCanvas extends Plugin {
       overlay.setAttribute('class', 'js-off-canvas-overlay ' + overlayPosition);
       this.$overlay = $(overlay);
       if(overlayPosition === 'is-overlay-fixed') {
-        $('body').append(this.$overlay);
+        $(this.$overlay).insertAfter(this.$element);
```
kball: Why this change?

kball:
Left some feedback on the JS... agree on @brettsmason's questions as well, but broadly speaking this is looking very good. Nice work @SassNinja! let's get this across the finish line for 6.4 (first RC targeted for next Thursday)

scss/components/_off-canvas.scss
SassNinja: I've used !important here due to the CSS specificity of the things I'm overriding.
Otherwise every property I've set important wouldn't work with only one simple in-canvas selector.
I would have to fight against the specificity of the following selectors:

- transform => .off-canvas-content .off-canvas.position-[DIRECTION]
- visibility => .off-canvas.is-closed
- height => inline style set by JS (fixed position emulation)
- box-shadow => .off-canvas-content .position-[DIRECTION].is-transition-push
- position => .off-canvas-content .off-canvas

I agree !important is not cool but I think in this case it's better than including in-canvas into all the long selectors (inline style is even impossible without important).
It's worth to mention I only used !important where necessary. Some properties of my in-canvas feature are working without it (background, width, overflow, transition) and just use 'inherit'.

js/foundation.offcanvas.js
SassNinja: - ID doesn't make sense since there shouldn't be more than one match
- `siblings('[data-off-canvas-content]')` didn't use `first()` in v6.3 so the behavior wouldn't change when upgrading (backwards compatibility)
- `closest('[data-off-canvas-content]')` use `first()` but I think it should behave identical to `siblings` so either both use it or none

Since `this.$content` should only match one element I think it's a good idea to make this sure by using `first()`. I'll adjust this!

I can't imagine someone is expecting a match of two or more content elements here (as it was possible in v6.3) what doesn't make sense. So backwards compatibility shouldn't be problem here.

js/foundation.offcanvas.js
SassNinja: I'm using `this.nested` to emulate a fixed position via JS (see `_emulateFixedPosition()`) and force some options (e.g. no content scrolling) since this is necessary when nesting the off-canvas.

But it is quite difficult to determine if an off-canvas is nested – I don't want to parse the whole DOM for this!
That's why I assert the element is nested if it has no siblings content. But the usage of a contentId is a problem here since the related content element may be placed anywhere on the page (can be a sibling, can be nested, can be none of this).

Therefore I'm asserting that the off-canvas is nested if the contentId option is used. But that's obviously not a good solution as it limits the flexibility of usage.
I may ignore the contentId as it's probably not realistic to have several content elements on the same level and change it into:
`this.nested = !this.$element.siblings('[data-off-canvas-content]').length;`

What do you think?

js/foundation.offcanvas.js
SassNinja: At the beginning of this feature I had some issues with the z-index / overlapping of the overlay elements what was gone after this adjustment. I've just rechecked this and is doesn't occur anymore when I undo the adjustment so it's probably not necessary anymore.

Do you prefer the way it was done in v6.3 meaning to append it to the body?

I personally find it quite pleasant to have the overlay close to the related element but that may be a matter of taste.

SassNinja:
@brettsmason @kball many thanks for your detailed feedback!

I've answered your questions in the appropriate review notes.
Regarding your messages:

@brettsmason 
- !important is necessary due to css specificity (see review note)
- I've added `visibility:hidden` to support multiple off-canvas elements that share the same content element **and** the same position. You can see this in my visual test (nested.html) when you change the viewport to medium size and click on the two nested left buttons
- yes, my shadow uses inset box-shadow instead of `::after` which doesn't cause the scrolling issue. I've still to cross-browser test this and optimize the settings. But apart from that it's good to use imo any will probably fix https://github.com/zurb/foundation-sites/issues/9672

@kball 
It would be awesome if my feature is part of the RC! I'll do my best to finish the final adjustments in time.

SassNinja:
I've now completely replaced the off-canvas push element's shadow (::after) with my inner shadow solution and adjusted the docs & settings accordingly. Seems to work fine and fixes https://github.com/zurb/foundation-sites/issues/9672
Besides I've fixed an issue of my emulateFixedPosition function if using top or bottom direction.

But in the context of further tests I spotted another issue:
If using a wrapping grid in the off-canvas content with push & pull classes, the fixed position emulation fails because the push/pull column has position:relative and a moves to left or right.
You can see the issue yourself if you just execute `gulp` and check the off-canvas page in the docs (nested chapter).

My only idea to solve this is adding an offset option and use it for the fixed calculation.
Do you have another idea?

I also wondered about the content id option which I've added and must admit I'm not sure if this is really useful because it doesn't enable the flexibility I initially intended.
The main idea was to let the user place the off-canvas element and content anywhere without any relationship except of the ID of course (not nested, not siblings). But the CSS is actual not built for this usage and my tests didn't do the trick.

So for now I think the content id option should be reverted/removed so we get this feature finished in time. In any case is the support of nesting (even without an ID option) a huge improvement.

@brettsmason @kball what do you think?

brettsmason:
@SassNinja I'd love to keep the ID option if possible - a slight Sass restructure would allow some extra flexibility (potentially). However I need to look through the code more.

Sorry quick reply, but I'll get back to you this evening properly! 😄

js/foundation.offcanvas.js
kball: Hmm... well we've already found `$content`, so can we simply compare `$content[0]` to `this.$element.siblings('[data-off-canvas-content]')[0]` ?

js/foundation.offcanvas.js
kball: My concern about having the overlay close rather than after the body is that unless I'm mistaken any sorts of transforms applied up the tree will break the fixed positioning

brettsmason:
@SassNinja Sorry (again!) for being slow.
So I've looked over it a bit more and functionally everything seems to work great. I think there's value in keeping the content ID option, and now now longer having to have the off canvas directly before the off canvas content is great, and will open up a lot more layout options.

My only remaining "issue" is the use of !important which I already mentioned. Could we not use `.off-canvas.in-canvas-for-large` instead, so nested? I'm pretty sure that has less specificity than !important and should work?

Other than that, from my point of view this looks fantastic 👍 
Just need approval from @kball on the JavaScript side of things.

SassNinja:
@brettsmason thanks for the feedback!

I've adjusted the selector as suggested by you and got rid of two `!important` (visibility & position).
There are three `!important` left I can't remove:
- transform => I would have to use a long selector due to specifity
- height => must be `!important` due to the inline CSS (set via JS)
- box-shadow => again long selector due to specifity

Talking about the content ID option: what's the additional benefit in your eyes?

Nesting is already possible without it. My initial purpose was to let the use place the off-canvas anywhere (even not nested and not a sibling of the content). But that's currently working without issues.

Check this codepen and scroll a little down before clicking the red button.
https://codepen.io/KaiTheGreat/pen/GmdVvy?editors=1000
The problem here is that my emulateFixedPosition function is setting a top value what is required for a nested off-canvas element because `position:fixed` in this case.

Since I don't always know if an element is nested in the JS, I have to force `position:absolute` is using the content ID option. I guess the bug is gone then.

I'm honestly not 100% sure if it's a good idea to include the ID option in the next RC without much more testing and improvement...
But if you think it's already good enough for including, I'll keep it.

brettsmason:
@SassNinja Thanks for your continued work on this!

So my use case for the content ID would of been something like having a menu bar on medium and up, and transforming it into an off canvas on small (using Sass). The problem with the live version is the off canvas and off canvas content have to be siblings, making this not work.

Would the above work with your changes? Or would it be forced to have the emulated behaviour?

SassNinja:
@brettsmason yes this should be no problem even without a content ID option. Of course the emulated behavior is forced then because we're nesting here.
Check out my adjusted codepen:
https://codepen.io/KaiTheGreat/pen/GmdVvy?editors=1000

If you scale the viewport down to small the top bar will move into off-canvas and a green button will appear.
This is done by just nesting the bar in the content – no need for a content ID here.

As mentioned above the content ID option has the problem that I don't know within the JS if the element is nested or not. Therefore I would have to assert it's always nested and therefore must be forced to have position absolute even if position fixed would work.

brettsmason:
Really nice demo! That's exactly the sort of thing I meant. Could we perhaps have an option to turn emulated positioning off then (maybe off by default)? I don't personally have a need to nest content, but I do (and so do others who have contacted me about it) have a need to make the current forced sibling relationship of the Sass more flexible.

Also noticed an issue with the shadow on that CodePen - the menu bar example - the shadow starts further down the off canvas (not at the top).

SassNinja:
@brettsmason glad to hear we had the same thing in mind :)

Emulated positioning can't be turned of in this case because it's nested here. I've placed the top bar within the off-canvas content. Therefore position fixed doesn't work anymore and I have to emulate it to behave similar.
If you turn it off here via an option it will look erroneous. Therefore I don't think this option is a good idea, is it?

Within the JS I try to emulate the position only if the element is nested. Since I can't determine a nesting 100% for sure always (as mentioned above) I'm asserting it's nested if not a sibling of an off-canvas content.

Yeah I saw the shadow 'issue': it's actually not an issue of the off-canvas itself but the top bar that has a background color and is therefore overlaying the inset shadow of the element.
Since it's just a quick demo for the in-canvas feature I didn't write any CSS to make the color transparent ;)

brettsmason:
@SassNinja I see your point, sorry I didn't look at the demo in enough detail 😄 
I guess then as long as the emulation wont trigger on a typical use case we should be fine. I still wonder if an option to completely turn it off (user beware!) might be handy though, but you know a lot more about its functionality that me!

I just noticed a bug on the CodePen example actually - click the button and close without letting it finish animating:
bug

I seem to remember this was a bug we ironed out during build up to 6.3, but I can't remember what it was. @kball can you remember?

kball:
I believe the bug had to do with how we were triggering the position change - if it happened in something waiting for animation end it could get wonky in this double click case.

SassNinja:
@brettsmason nice screen capture – what program did you use for this? (just curious)

Talking about the issue:
@kball was right, it was caused by the animation end listener (transitionend) that focused a target within the element that was already closed.
I've fixed this now by checking if the element isn't already closed when it goes into the listener.

@brettsmason I guess you're happy to hear I've got the content ID option probably under control.
As I mentioned the problem is that I don't know within the JS if the element is nested or not. Therefore I've added a nested option now so the user can define if it's nested or not. Seems to work fine so far :)

The only issue that's left in my eyes is the usage of pull/push because the `position:relative` is a problem. As mentioned earlier my only idea is to use some kind of offset here as an option...
@kball @brettsmason any ideas?
You can see the issue when you clone my fork, run `gulp` and scroll down to the nested section on the off-canvas page.

IamManchanda:
Just on the screen capture , [Screencastify](https://chrome.google.com/webstore/detail/screencastify-screen-vide/mmeijimgabbpbgpdklnllpncmdofkcpn?hl=en) is a good tool!

brettsmason:
@SassNinja The program is called [Screen to Gif](http://www.screentogif.com/) - it's pretty good.

So looking at the nested example, it seems a bit messed up in Chrome:
image

However settings its position to `fixed` actually seems to fix the positioning as far as I can tell. That's because when a panel transforms, a nested fixed position element looses its fixed position from the top, but still looks like it works better than absolute from what I can see. The in-canvas example above it is also broken for me in Chrome, though again setting to fixed fixes it.

SassNinja:
@brettsmason thanks for the link but unfortunately it's Windows only – I'll take a look at the suggestion of @IamManchanda 

Regarding the issue:
yes, that's exactly what I was talking about! The problem here is the pull/push class in the docs layout (default.html)
If you remove these classes it works without problems. Just run this code in the dev tools console to see it
`$('.off-canvas-content > .expanded.row > .columns').removeClass('medium-push-3 large-push-2 medium-pull-9 large-pull-10');`

I think using position:fixed isn't the solution here although it looks better on first view. But the  text within the off-canvas element isn't visible then and in my eyes there's no alternative to position:absolute when nesting.

I'll check if I can adjust the emulated position calculation to consider this case.

brettsmason:
@SassNinja Ohhh... so this is a docs only styling issue? If that's the case, apologies I misunderstood.

So the problem is that there is a `position: relative` on a container element that's messing it up? If so couldn't that potentially be an issue for any user though, not just the docs? Not sure how you get around that other than with fixed positioning.

SassNinja:
@brettsmason exactly, it's currently only visible in the docs since I'm using pull/push neither in the visual test not in the codepen.
But of course it's a major issue because users may want to use it with pull/push (besides it must look good in the docs). 

As I said, I'll check if I can consider this in the calculation.
The position:relative is here definitely a thorn in the side...

kball:
I have not wrapped my head around all that this update is doing yet, but one thing I notice is that when I check out this branch and run it all the `sticky` components in the docs break... this is true whether or not I bring it up to date with the current state of develop...

@SassNinja any idea what might be leading to this?

kball:
I'm digging through this a bit more, and confused why we're using fixed position emulation. I see the comment:

> position fixed doesn't seem to work on Windows if nested

But no more details on why/how this is happening.

position fixed has the known limitation of not working inside of a translated parent, but I think trying to account for all of the various cases of positioned parents in some sort of emulation is setting us up for a world of hurt (the docs example being a simple one).

My preference would be we continue to allow either position fixed or position absolute... if you're trying to position something relative to your parent, use `absolute`, if you're trying to position relative to the window use `fixed`... 

@SassNinja can you explain a bit more what wasn't working for you with fixed? Maybe we can track that down instead of doom looping on fixed emulation

kball:
I just read back through this with slightly fresher eyes - sorry for the somewhat testy last comment, trying to get everything packaged up for the 6.4 RC tomorrow and a little stressed.  Now I understand the need for the emulation... the panel itself is inside the content which is getting transformed, breaking fixed...

I'm wondering if we might be trying to support too many possibilities.  The desired outcome here is I think very similar to "panels" from the (now deprecated) foundation for apps (see http://foundation.zurb.com/apps/docs/#!/panel) - in that case there was no requirement to identify the content (I think because the content was not transitioned), and probably for that reason they were able to get away with absolute and fixed positioning...

My questions then are 2fold:

1. Can we get away with not forcing every offcanvas to have a content?  This essentially puts these off-canvas elements into the "overlay" transition, implying they cannot push
2. Is it okay then to insist that "nested" off-canvas components behave in that manner, ie if you want it to be in page you cannot also transform the page when it opens, you must overlay.

If we can do both of those things, I think this becomes much simpler and we don't have to deal with emulation or anything, b/c you're not transforming the elements the off-canvas is with.

@SassNinja what do you think, would those constraints still satisfy your usecase? Because if so I think we assert them, get this way simpler, and still *dramatically* improve the power of the off-canvas element.

SassNinja:
@kball I agree, the support of all the possibilities is also making things very complicated.
As you've already spotted the transform is breaking the `position:fixed`so fixed position emulation is required in this case.

Regarding your questions:
1. removing the push transition option would of course make things much easier – but that should definitely not be done because push is really useful since the page content is still visible instead of getting overlayed. And it's one of the reasons I love the foundation off-canvas plugin!
2. same here: would make things easier but in my opinion it's a great downgrade to remove the push transition option if nesting

I've now pushed a new function called `adjustToLocalScope` that tries to consider a relative scope e.g. when using push/pull classes at the grid as done on the docs page.
To detect the relative scope I'm using `offsetParent()` and check the left css value and offset().top value.

I must admit this is some kind of hacky and may not work in all cases but so far this seems to be an acceptable solution in my eyes. I'm not abandoning the nested push option yet :)

Talking abut the issue of the `sticky`component you've mentioned:
I've no idea what's causing this! I went back in git history before my first off-canvas commit and it still occurs. So it doesn't seem to be related to my adjustments.
But I'll try to find out more...

SassNinja:
Regarding the sticky issue:
there's a sticky element in the generated `off-canvas.html` with two IDs as option
`data-top-anchor="docs-header:bottom" data-btm-anchor="footer:top"``

Neither 'docs-header' nor 'footer' do exist. That's why the sticky plugin fails. It's trying to read `offset().top` of an undefined anchor.
The live docs page contains at least 'footer'.

As far as I can see this is related to the npm foundation-docs package.
@kball do you know more about how this is set up?

SassNinja:
Good news:
I've now merged the upstream develop branch into my fork and the issues seems to be gone.

So in my eyes it looks good now.
@kball @brettsmason do you see anything else that has to be fixed?

Would be really great if this feature will be part of the final v6.4 release (or RC3?)

kball:
The emulation is still quite buggy - for example the "Multiple Panels" examples in the docs don't work, as the panels are getting positioned with `top:6869px` way offscreen.

Given how flaky this type of emulation tends to be, it's going to require very extensive testing and debugging and I honestly don't see it happening within the timeframe of 6.4... I'd love to get at least a limited version of this in, and I *think* if we limit to the nested offcanvas to be an overlay (ie no transforms to break fixed positioning) we can dramatically simplify this change and get this across the line.

I can take a pass at doing this, building on your work here, and let you continue working on this if you want for a future release.

SassNinja:
@kball you're right, I didn't notice that new issue on the docs page – sorry for this! I'll do more testing when working on this feature.

Discarding the push transition if nested is not an option for me.
But I agree to you: it's probably to complex to finish it in time for **all** scenarios.

I've now worked over the whole code and (as suggested by you) forced the overlap transition if nested.
Please have a look at this limited version.

**Apart from that I've noticed some issues that also occur on the live page:**

1. the docs are using the `.off-canvas-wrapper-inner` class that was used prior v6.3. But it seems the implementation is not correct: the content isn't nested in it (a sibling of the element) but a sibling of the wrapper inner. I've adjusted the JS to consider this but I think this is an issue, is it?
2. several push off-canvas elements on the page (e.g. `#offCanvasLeft`) are supposed to use push transition but when you click the button, the overlap transition gets used. This is related to the issue above. I've fixed this in my fork but then the sticky top bar & side navi disappear on opening a push element and comes back on opening an overlap element. This is caused by the transform scope of the push transition. What do you think? Can this be ignored?
3. I've noticed there is some CSS code in the _page.scss in a node module that is interfering with my code. Precisely there's a hardcoded off-canvas z-index 11 in `node_modules/foundation-docs/scss/_page.scss`. Is this necessary for some reason or would you mind deleting it?

@brettsmason @kball would be great if you can take a look at the issues I've listed above.
Hopefully this doesn't keep the limited version from being part of 6.4

brettsmason:
@SassNinja To answer your points:

1. That is a mistake - that class was dropped in 6.3, and is no longer needed. No reference to it or anything should be needed in the JS/Sass as this shouldn't be there and should be removed from the docs. It only applies to versions before 6.3.

2. Again I think this is a docs related issue, not off canvas itself. I checked the visual tests and it works correctly. We can file a seperate issue for the docs problems, so you dont need to make changes to cover this.

3. Not sure on this will need to be checked.

I'll let @kball come back to you on the other stuff, but generally no need to worry about the docs, so anything you added to do with the inner wrapper etc because of the docs can be removed as its not needed 😄

SassNinja:
1. ok, I've undone my wrapper inner JS fix since it doesn't make sense then
2. indeed this is not related to the off-canvas but to the erroneous implementation of the wrapper inner (content mustn't be a sibling of it but a child). I've fixed this too because I only had to remove two lines: https://github.com/zurb/foundation-sites/pull/10003/commits/4a8132df447f9f83584be6de6bfd6cbeb706c8b6
Doesn't seem to be worth opening a new issue for this one, does it?

But as previously mentioned the fixed top bar on the docs page looses its fixed state now once a push transition element gets opened.
This is caused by `transform: translate(0, 0)` on the content container. You can see the behavior I'm talking about when you toggle between overlap & push elements (e.g. in the transitions chapter).

@brettsmason what do you think, shall I always set `transform: none` once the element is closed?

I mean the fixed state would still get lost (I remember our initial talk about fixed & transform) but at least it will be restored if closed.

Some words regarding no. 3 of my list:
You can also see the code in talk here https://github.com/zurb/foundation-docs/blob/master/scss/_page.scss
I'm talking about lines 38-40 that I suggest to remove. Currently I've fixed that by increasing the CSS specificity in my code but this feels wrong because the code in _page.scss doesn't make sense for me.

```sass
@mixin off-canvas-base(

  // Set the off-canvas z-index.
  z-index: $offcanvas-push-zindex;

  // Increase CSS specificity
  &.is-transition-push {
    z-index: $offcanvas-push-zindex;
  }

}
```

SassNinja:
Update:

I've worked over the add/remove content classes part now and set `transform:none` to default as suggested in my previous post.
As a result the fixed state (e.g. the top bar in the docs) gets restored once the push element is closed.

**So for me the limited version of this off-canvas feature looks good. Would be great if you can review it and let me know whether you spot any issues I've missed.**

Apart from that I'm having some problems with the transitionend listener in another feature branch (not affecting this limited version – I'm justing working on a nested push transition support for the future): 
If a push element is nested, the element's `transform:translate` property doesn't change but only the content's property.
Therefore I need an either-or translationend listener (either element or content). Unfortunately I haven't been able to achieve this yet. I tried to unbind the other listener if one gets triggered but both of them seems to get triggered simultaneously...
Any idea how to listen on both but only execute one of them?

Apart from that I'm facing an issue with the `transitionend` function which is imported from `foundation.util.core.js`:
As far as I understand it returns the supported event name for the current browser but I wondering
1. why determining the right event over and over (every time `one()` gets called) instead of doing this only once?
2. for me (Chrome, OSX) the function returns 'webkitTransitionEnd' but while this is working for most transitions it isn't working for a nested push element (I need 'transitionend' in this case). So any reason not to listen on **all** events (transitionend, webkitTransitionEnd, ...) instead of only one specific transition end event? 

@kball maybe you can help me with this `transitionend()` function and tell me how this is supposed to be used or rather what's its purpose?

kball:
@SassNinja just checked this out, still seeing some of the docs examples not working, in particular the examples under Off-canvas Directions don't appear to work in this branch. It also seems to break the push examples in the sticky visual test...

Unfortunately with just under 1 week to final 6.4 release, I think this is looking very likely to slip out to 6.5

SassNinja:
@kball so the directions chapter is the only 'erroneous' part in the docs, isn't it?
I've increased the CSS specificity now so it behaves similar to the live version, meaning nesting a content container in another content container gets considered.

But to be honest I'm not sure if this behavior makes sense. I can't imagine why you would nest a content in another content. Besides it looks weird (in my fork & live) when the nested content and its elements doesn't act as the usual off-canvas (whitespace between both).

Regarding the off-canvas sticky visual test:
What exactly is the issue you spotted? 

The sticky bar looses it's fixed state on opening a push element and gets it back once the element is closed. As mentioned in my previous posts this is caused by the transform scope the push transition creates. 

It's a pity if at least my limited feature won't be part of 6.4 but I understand you want to be on the safe side.
If you give me faster feedback regarding issues & change request I'll do my best to fix & implement it asap.

Renkas:
I do hope this will find it's way to 6.4 at least in the minimal form. I could use the functionality as I have pretty similiar usecase as described in the initial post #9988 
I can hold off the frontend development of our app's new design a little bit and/or work on other areas in the meantime but this functionality would be massive improvement.

kball:
@SassNinja you are correct, I apologize, I reconfirmed that sticky visual test on prior releases just now and the behavior is the same.

@brettsmason can you take one final look through this and lets see if we can get this merged?

brettsmason:
@SassNinja Really sorry I've been MIA - head stuck in the XY grid.

So pulled down the latest changes and have been looking over it for any bugs I can find:
- The nested left push example (offCanvasNestedLeftPush) doesn't seem as smooth when opening/closing as the non-nested example. Any idea why that could be?
- The close button stays visible when an off canvas is in-canvas. This should probably set to `display: none` the same as the `reveal-for-x` method
- I'd still love to remove the `!important` usage some how... I need to spend a bit more time trying to understand the current Sass structure and see if we can improve it or not.

👍 though its looking good overall.

scss/components/_off-canvas.scss
```diff
@@ -118,13 +138,24 @@ $maincontent-class: 'off-canvas-content' !default;
     position: absolute;
   }
 
+  // Set the off-canvas z-index.
   z-index: $offcanvas-push-zindex;
 
+  // Increase CSS specificity
```
brettsmason: What's the purpose of this? Don't fully understand why its needed.

scss/components/_off-canvas.scss
```diff
@@ -154,10 +185,20 @@ $maincontent-class: 'off-canvas-content' !default;
 
     transform: translateX(-$size);
     overflow-y: auto;
+    
+    // Sets the position for nested off-canvas element
+    @at-root .#{$maincontent-class} .off-canvas.position-#{$position} {
```
brettsmason: This seems to be the route cause for the need for `!important` usage. You have a better understanding of the way nested works though, but is there no way we could restructure the Sass a bit to make this less specific? I understand it may not be possible.

js/foundation.offcanvas.js
```diff
@@ -300,9 +380,15 @@ class OffCanvas extends Plugin {
     this.$triggers.attr('aria-expanded', 'false');
 
     if (this.options.trapFocus === true) {
-      this.$element.siblings('[data-off-canvas-content]').removeAttr('tabindex');
+      this.$content.removeAttr('tabindex');
       Keyboard.releaseFocus(this.$element);
     }
+
+    // Listen to transitionEnd and add class when done.
+    this.$element.one(Foundation.transitionend(this.$element), function(e) {
```
kball: This will need to change to using an import rather than Foundation global (see https://github.com/zurb/foundation-sites/blob/develop/js/foundation.util.motion.js#L4 for example)

but if you like I can do this at merge time.

kball:
I think with those last changes this is probably good to go!  @brettsmason what do you think? 

I'm out of office today but should be able to merge & turn around the RC tomorrow morning.

brettsmason:
@kball If we can solve issue 2 above and hide the close button when an off canvas is in-canvas (like `reveal-for-x` then yep I think we can merge this. We can always refine the specificity later on down the line.

SassNinja:
@brettsmason thanks for the feedback & review!

Regarding your bug list:
- what browser & OS were you using? I can't reproduce this issue (Safari, FF, Chrome – all on Mac OSX). For me both, nested and not nested, look identical. From a technical point of view there shouldn't be a difference since there's no transform scope for overlap & nested elements anymore and both use the same transition.
- I wasn't sure about the close button as it's part of the content. So I wanted to leave it up to the user to take care of the element's content. But you're right, a default `display:none` is better in this case as this is probably the most desired behavior. I've adjusted this now and replaced the inline `display:none` of reveal-for-x (prior via JS) with my CSS solution in one go. I hope this is ok? You can see it here https://github.com/zurb/foundation-sites/pull/10003/commits/8c03358945f56e9395884b6456827b8d085d0298
- I've got rid of all in-canvas !important now: height works without it since there's no nested push support anymore (= no inline CSS) and box shadow + transform have an increased CSS specificity now (by using the position classes)

@kball I've adjusted this as requested by you. It's part of this commit https://github.com/zurb/foundation-sites/pull/10003/commits/5aa588dcfc831c87b9262987e51e5af2d03f3e3f

js/foundation.offcanvas.js
SassNinja: done now in https://github.com/zurb/foundation-sites/pull/10003/commits/5aa588dcfc831c87b9262987e51e5af2d03f3e3f

scss/components/_off-canvas.scss
SassNinja: I've done this to override the fix z-index that is set in https://github.com/zurb/foundation-docs/blob/master/scss/_page.scss
If this gets removed we'll be able to undo this specificity fix.

scss/components/_off-canvas.scss
SassNinja: I was facing a lot specificity problems when mixing push & overlap with the nested feature. 
But since the push support for nested elements is removed (for now) I guess this may get simplified a bit.

Would you like me to work over it and see what I can cut down for this limited version?

brettsmason:
@SassNinja Thanks for taking another look!

To answer your points:

- I think this was actually my laptop being weird as I haven't had the issue since so I would disregard this
- Great that looks better, and good call about changing the reveal option to CSS too
- Amazing! 😄 Much better

Great work and thanks for being so patient and thorough 👍 

@kball from my point of view this looks ready to merge now. You might want to check the JS changes quickly but I'm happy with the Sass.

kball:
All right! Scanned the JS again and happy with it as well. Merging. Last major feature change to sneak into the 6.4 release :) Nice work @SassNinja!

PR #10010, opened 2017-05-05
Fix for Responsive Accordion Lost Event Bindings 

rdetert:
Fixed a bug in the responsive accordion plugin where event bindings were getting lost when switching from tabs to accordion.

You can see an example of it here:
f6test.herokuapp.com

It's a simple fix. `detach` should be used instead of `remove`. The two methods are identical with the exception that `detach` keeps the jQuery event bindings in tact.


kball:
@designerno1 any concerns with this? Seems reasonable to me...

designerno1:
ok thats an inner content problem with bindings in the tabs content.
So detatch looks good to me.

if inner content should adjust to the switch this should be done be the inner content author, but bindings should be ok with detatch if it doesn't add the content twice that i have to check

kball:
Great, merged! Thanks @rdetert and congrats on your first merged PR!

Issue #9875, opened 2017-03-14
[Foundation Future] Discussion on Grid's Future ( Flex/Float/Spec - Margin/Padding )


IamManchanda:
https://github.com/zurb/foundation-sites/wiki/Project-Roadmap#other-things-to-investigateschedule

> Other Things to Investigate/Schedule
> 
> 3. CSS Grid - Not sure what the timeline is on this but we should target this, perhaps similar to flexbox mode?

CSS Grids has landed a week back first in [firefox](https://rachelandrew.co.uk/archives/2017/03/07/css-grid-lands-in-firefox-52/), and then in [chrome](https://blog.chromium.org/2017/02/chrome-57-beta-css-grid-layout-improved.html), [safari](https://developer.apple.com/library/prerelease/content/releasenotes/General/WhatsNewInSafari/Articles/Safari_10_1.html#//apple_ref/doc/uid/TP40014305-CH12-DontLinkElementID_65) and opera

I am opening this thread for the discussion on this **awesome layout monster**!
( This might be very soon, but better to be early than late! )

-------------------------------------------------------------------------------------------------------------------------------------

1. Quick Overview => https://www.slideshare.net/rachelandrew/talk-web-design-css-grid-layout

2. Complete Guide => https://webkit.org/blog/7434/css-grid-layout-a-new-layout-module-for-the-web/

3. Grid by example => http://gridbyexample.com/

4. What mozilla says => https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Grid_Layout/CSS_Grid_and_Progressive_Enhancement

5. Yes or no to IE and its old spec => https://rachelandrew.co.uk/archives/2016/11/26/should-i-try-to-use-the-ie-implementation-of-css-grid-layout/

6. Feature Queries => https://hacks.mozilla.org/2016/08/using-feature-queries-in-css/

-----------------------------------------------------------------------------------------------------------------------

**Update**: Well its so co-incidental ..... that when i posted this thread, 
Zurb was already working on this on a youtube video :smiling_imp:  => https://youtu.be/bI1wNuvuFxo   
Maybe its because both zurb and community wants the advancements! Right ??

IamManchanda:
https://rachelandrew.co.uk/archives/2016/03/30/should-i-use-grid-or-flexbox/

> This is how flexbox works. However, most of the flexbox problems I am asked about are due to developers trying to make flexbox do layout in two dimensions. This is where you need **Grid Layout**.

[...]

Issue #8024, opened 2016-01-27
Margin grids instead of padding grids


corysimmons:
Do you guys have a good reason for using padding grids?

http://codepen.io/corysimmons/pen/dGmEBW?editors=1100

Let me know if you want a PR.

corysimmons:
I don't want to duplicate this discussion in two places, and since you and Bootstrap both have the same grid system, could you guys just follow along here? https://github.com/twbs/bootstrap/issues/19026

gakimball:
There are a few quirks with the `calc()` approach, although probably the same number of quirks with our current approach. For example, you can't define the width of a column independently of its gutter, which makes the collapse/uncollapse classes weird. We also now support responsive gutters, which makes that logic *even more* confusing.

We're actually still holding out on Android 2.3, so `calc()` isn't an immediate option anyway.

I like Mark's [comment](https://github.com/twbs/bootstrap/issues/19026#issuecomment-175749492) on your issue: "I'd rather say `25%` and get exactly `25%`". Padding grids require a little extra markup in some cases, but they're *way* easier to write, maintain, and reason about.

corysimmons:
Alrighty. Let me know if you guys change your mind.

IamManchanda:
Yes Yes, we changed our mind :wink: .... Added in #10017

kball:
@gakimball wasn't wrong... the logic around responsive gutters and responsive collapse for margin classes is pretty darned confusing... but it works. Will see if we can simplify it as we dig into this more.

corysimmons:
>the logic around responsive gutters and responsive collapse for margin classes is pretty darned confusing

???

https://codepen.io/corysimmons/pen/NgKYRZ?editors=1100

kball:
Responsive gutters => a map of gutters sizes per breakpoint:

e.g.
```
$margin-gutters {
  small: 20px,
  medium: 30px
}
```
ie for small sizes have 20px gutters, medium and up have 30px gutters.

Since the grid sizes have to include the gutter, that means that that for e.g.
`.small-1` you need to redefine the size for each breakpoint where there exists an entry in the gutters map

corysimmons:
Ahhh, gotcha 👌

PR #10017, opened 2017-05-09
Flex Grid 2.0

brettsmason:
This is my vision of a brand new flex grid, based on a lot of research and input from some very knowledgeable people 😄 

**The aim:**
- Fully rewritten flexbox grid
- Padding OR margin grid option
- Vertical grid option
- Easy to use, flexible mixins
- Cleaner generated classes with less/no specificity issues

It's still very much work in progress, but I would appreciate feedback on whats present so far.

**Things to decide on:**
- How we ship this with 6.4 (how its included, what other grids are also available etc)
- How we handle the move to new classes - `.grid` and `.cell`

There are probably more things which I've forgotten, but wanted to get this up for people to look at 😄


IamManchanda:
https://github.com/zurb/foundation-sites/issues/9875#issuecomment-299731733

> Alright, I'm outta here again. Good luck team. Ping me if you have a direct question.

cc / the knowledgeable people @corysimmons for any of his views on the same!

scss/foundation.scss
```diff
@@ -116,3 +116,5 @@
     @include foundation-flex-classes;
   }
 }
+
+@include zf-grid-classes;
```
IamManchanda: End of line !!!

kball:
cc @tdhartwick @jnemeth @canterberrie @rafibomb @zurb/yetinauts

IamManchanda:
As said before in a DM i just loved the new naming convention specially considering vertical flex grid and the future - CSS Grids .... I think we should leave the classes rows and columns to the **float era** and move ahead!

IamManchanda:
Another discussed matter is the docs and the flexbox specific alignment classes!
http://frowcss.com/frow-shortcuts.html

corysimmons:
This _should_ be a major version bump since it's breaking... unless you want to have both grids side-by-side (way too confusing).

Can I get a CodePen to review?

It's worth thinking about if the grid is integrated into any Foundation components. Does this change break other components because of this?

IamManchanda:
Yes Cory. In fact Setting up visual testing would  be even better than codepen

brettsmason:
I'll try and get a demo up in the next few days. However it's missing a few pieces of the puzzle that I need to do first.

Not sure how to show both margin and padding grid in a demo, as I think it will be configurable by a variable (I'm guessing padding by default as it's the previous behaviour).

corysimmons:
Global and local params https://codepen.io/corysimmons/pen/pPagON?editors=1100

brettsmason:
Thanks Cory. Do you personally think we should ship the grid with padding and margin classes, or just pick one and allow Sass users to switch? Just worried about too much bloat!

corysimmons:
Oh yeah, I keep forgetting you need to dist css.

You could give Sass/PostCSS users global **and** local.

You could give dist css users access to a settings.css file where they could at least define global settings. https://codepen.io/corysimmons/pen/KmQRON?editors=1100

But yeah, gutterless is so common, you'd probably want to dist that as well.

If your class names are like `.cell-4`, then I'd make the gutterless classes something like `.span-4` (`span` is the term I use for these anyway).

I'd also make sure the code that generates these is decoupled so users can checkbox which one(s) they want.

There's another argument for PostCSS: dist.css is usually full of bloat to cover all the use cases where vanilla CSS falls short.

corysimmons:
Actually, why not separate these out into `dist/margin-grid.css` and `dist/padding-grid.css`?

Then people could `@import` whichever they wanted, or both.

brettsmason:
Thanks Cory, some good ideas there - I'll play around with it a bit.

The complications come from offering: padding grid, margin grid, vertical grid (works with padding and margin grids), offset classes and collapse/uncollapse classes.

I was wondering if the default `.grid` should just apply the flex properties to the grid, then modifier classes like `.padding-gutters` or `.margin-gutters` (or more suitable names) then apply the corresponding gutter to the cell, eg:

```
.padding-gutters {
    > .cell {
        // apply padding
    }
}

.margin-gutters {
    > .cell {
        // apply margin
    }
}
```

It gets complicated with things like classes for collapsing gutters and re-applying at breakpoints (for margin grid anyway) because of the use of `calc()` to have to work out the widths.

If anyone has any thoughts on how to simplify this a bit that would be great!

IamManchanda:
Sounds good Brett!

brettsmason:
I'm not sure it is 😄 But can't currently think of another better way.

IamManchanda:
The reason i like is that this way .... this way, this is also possible (damn nested nesting)  http://codepen.io/IamManchanda/pen/ybPORw?editors=1100 when no-gutters are required. Ofcourse you can't do that with those gutters classes!

When css grids gets the browser support .... we can simply replace `margin-gutters` and `padding-gutters` classes with css grids magic!

IamManchanda:
Ohk thinked about and here is my 2 cents, just remove the padding option and go with margin gutters 

When you dont want gutters (just like http://codepen.io/IamManchanda/pen/ybPORw?editors=1100 )

```pug
.grid
  .cell
    .cell
       .cell
```

When you want gutters (which is quite normal and used mainly)

```pug
.grid.gutters
  .cell
    .grid.gutters
       .cell
```

A good docs will be needed but this sounds reasonable to make understand this thing with the end user i guess!

corysimmons:
@brettsmason Maybe. I'm not sure it'd decrease the amount of bloat much, but maybe I'm just not seeing it through the right lenses right now.

It might be worth thinking about if these wrapping classes should be something like `.padding-grid`, `.margin-grid`, or something a bit more flexible like `.sm-gutter`, `.md-gutter`, etc.

IamManchanda:
`.sm-gutter, .md-gutter`

If you are talking about Responsive variations of gutters, then it should be added in variable with default to `false`!

If you are talking about sm-gutters as small gutters gap , md-gutters as medium gutters gap than i would say that responsive gutters would be still better!

corysimmons:
Sorry, ignore my previous comment.

@brettsmason Could you provide a small example demonstrating how `.padding-grid`/`.margin-grid` would cut down on bloat? I don't imagine it would.

brettsmason:
@corysimmons So I'm finding it hard to explain everything/my struggling thought process, but I'm sure you understand as you are the pro with this!

Take a look here: https://codepen.io/brettsmason/pen/qmxvWW
Basically that shows a small snippet of margin grid and padding grid styles, and the differences between them. Notice the base `.cell` styling difference between both.
Other than calling on `.cell` and one `.span` or something, then I thought we would need to split them like this:

```
.margin-gutters > .cell {
  flex: 1 1 calc(100% - 1.25rem);
  margin-right: 0.625rem;
  margin-left: 0.625rem;
}

.margin-gutters > .small-6 {
  flex: 1 1 calc(50% - 1.25rem);
}

.padding-gutters > .cell {
  flex: 1 1 100%;
  padding-right: 0.625rem;
  padding-left: 0.625rem;
}

.padding-gutters > .small-6 {
  flex: 1 1 50%;
}
```

basically there would be no base `.cell` class, as it relies on the presence of `.margin/padding-gutters` on the parent.

The issue comes when you want to do something like:
`<div class="grid margin-gutters medium-no-gutters large-margin-gutters">` (I hope someone wouldn't do that though!).

With padding grids its easy, we just remove the gutters/add the gutters on each breakpoint as its not included in the width.
However with margins, to do the above you would need to redeclare the `.[breakpoint]-[size]` classes within each breakpoint (unless I'm not seeing this right), eg:

```
.small-6 {
    flex: 0 0 calc(50% - 1.25rem);
}

.medium-no-gutters {
    .small-6 {
        flex: 0 0 50%;
        margin-left: 0;
        margin-right: 0;
    }
    ...
}

.large-margin-gutters {
    .small-6 {
        flex: 0 0 calc(50% - 1.25rem);
        margin-right: 0.625rem;
        margin-left: 0.625rem;
    }
    ...
}
```

Sorry if none of that makes sense, but hopefully it makes it a bit clearer! Would love feedback and any other suggestions on a way to do it.

andycochran:
I'm starting to think "gutter" isn't semantic. Maybe we shouldn't have a "padding OR margin grid option." Instead the grid could always have padding AND margin, regardless of if they're zero. Ditch the idea of calling the visual space a "gutter"(call it what it is). And swap the `$grid-gutters` var for… 

```scss
$grid-margin: (
  small: 20px,
  medium: 30px
) !default;
$grid-padding: (
  small: 0,
  medium: 0
) !default;
```

brettsmason:
I've made a bit of progress today. There's some issues with the generated breakpoints which I need to figure out, but on the whole its working. Still have the vertical grid, offsets and some other stuff to do though.

@corysimmons I've uploaded a rough demo here: http://brettmason.co.uk/grid/

corysimmons:
Looks pretty good. Collapsed containers aren't aligned correctly though.

I think even offering padding grids is confusing, even if it is a bit more powerful, and might prefer people just manually do this by setting margin gutters to 0 and manually adding padding.

brettsmason:
> Collapsed containers aren't aligned correctly though

For which - margin or padding example? Sorry been looking at this for hours and I cant see what's incorrect...

> I think even offering padding grids is confusing, even if it is a bit more powerful, and might prefer people just manually do this by setting margin gutters to 0 and manually adding padding.

I agree it may get a bit confusing. So what your saying is essentially the grid is a margin based grid. Then if you want padding you collapse the gutters, then either having a class to add padding, or adding the padding manually? Potentially something like `.grid-pad` for all 4 sides, `.grid-pad-horizontal` and `.grid-pad-vertical` to easily apply padding to all cells within a grid.

brettsmason:
@andycochran I think what you suggested is in a similar direction to what Cory mentioned.
If you are around at all any time would love to discuss some ideas on it.

andycochran:
> …the grid is a margin based grid. Then if you want padding you collapse the gutters, then either having a class to add padding, or adding the padding manually? Potentially something like .grid-pad for all 4 sides, .grid-pad-horizontal and .grid-pad-vertical to easily apply padding to all cells within a grid.

👍 This grid layout should be margin based. But you may want padding without collapsing the gutters. It shouldn't be either/or.

brettsmason:
@andycochran So we are saying the grid is margin based, with a `.collapse` / `.no-gutters` class to remove the margins.

Then there could be a separate set of utility classes for applying the responsive gutter sizing as padding, which doesn't effect/change the margin behaviour and will just add padding, regardless of if there are margins/gutters or not?

corysimmons:
@brettsmason 

image

Pretty much anywhere that the blue doesn't align flush needs to be changed to avoid usage confusion.

kball:
Just updated the branch referenced by this PR with all the work we've been doing... was thinking to rebase to get rid of some of the false starts but it got too complicated so just keeping it.

The docs need more work, only covering a fraction of what the new grid does ATM, but this is getting very very close

IamManchanda:
cc @corysimmons .... Your [vision](https://github.com/zurb/foundation-sites/issues/8024) that you had got successfull today!

corysimmons:
Oh damn. This actually makes me feel good. Thank you, everyone.

What's the rollout look like? Where can I find docs?

kball:
@corysimmons this was the last major thing merged in, packaging up the release candidate now, will have the docs up with this sometime this afternoon. Will update

brettsmason:
@corysimmons yeah big thanks from me too! I really appreciate all the help you provided.

corysimmons:
Before you do a hard launch, do you care if I review to make sure nothing is going to be broken? I seem to remember telling Brett something about padding grids needing to line up, and I can't remember if I was correct in that.

@brettsmason do you know what I'm talking about?

IamManchanda:
Big thanks @corysimmons from my side too!

> Before you do a hard launch, do you care if I review to make sure nothing is going to be broken?
I would love that .... @kball i remember you said me ... its not bad if we have to push release a bit late so now i that time!!!

kball:
@corysimmons absolutely would love a review; also planning in a bunch of time for multiple release candidates based on feedback. 

In particular we had to do some funky stuff to get the responsive gutter functionality from foundation to work properly, and I'm not convinced the scss api for folks who don't need that is as clean as it could be.

Would be happy to have a review of the code as it stands here... for playing with it, if you have a foundation build environment, simplest would be to play in that context... or hmm let me get a version of this up in a codepen

corysimmons:
Yeah, would appreciate a CodePen. 🙇

IamManchanda:
`fork, clone, npm i, npm start` :smile: 
It easy to run this locally though

kball:
@corysimmons here's a quick codepen: https://codepen.io/kball/pen/xrKLPM

kball:
And another one, this one doing some fun vertical/horizontal nesting using the frame classes: https://codepen.io/kball/pen/vZBeLy

kball:
@corysimmons if you want to play with the scss mixins etc, both of those are referencing a quick pen I set up containing the grid scss and its dependencies here: https://codepen.io/kball/pen/jwNLmL

kball:
Going to ahead and package up the first RC so we can get more eyes on this thing - definitely still consider this open for updating, planning in about 3 1/2 weeks of RC time to catch any problems with this or the other major major update that was in https://github.com/zurb/foundation-sites/pull/9965

corysimmons:
Thanks Kevin.

What's the purpose of the padding grid? How is it different from margin grids with `0` gutter?

I'm not opposed to including the padding grid, I'd just love to see a real world use cases for padding grids and padding + margin grids.

kball:
@corysimmons I'm also positive there is a usecase... I'm tasking some of the folks who were waxing on about padding grids to help us document that. @rafibomb @andycochran :)

brettsmason:
@corysimmons Glad you approve! 😄 
Would welcome your feedback on the actual Sass code. As @kball mentioned, it got a little messier than we would of liked in order to generate the complicated classes (pretty much all down to responsive gutters).

andycochran:
> What's the purpose of the padding grid? How is it different from margin grids with 0 gutter?

Semantics. Confusing semantics. The concept of "gutter" is leading to confusion about the applications of margin and padding in the grid. Grid columns should just have margin and padding. How you apply background colors to the rows/columns will determine the visual appearance of a "gutter."

A better question might be: Why are we using the term "gutter?"

IamManchanda:
Personally i am fine with `gutter` naming convention!

andycochran:
It's a holdover from print design. Books have gutters. Screens do not.

IamManchanda:
Gutters === the blank space between facing pages of a book or **between adjacent columns** of type or stamps in a sheet. (Source: Google)
Why can't we say this spacing as gutters when the whole web including [other frameworks](https://v4-alpha.getbootstrap.com/layout/grid/#no-gutters) mark them as **gutters** ??

corysimmons:
@andycochran Interesting. I see your point.

@IamManchanda He's saying the word "gutter" is confusing because there are now multiple ways to create that space between columns: padding and/or margin.

This is kind of my issue with bundling padding up with this at all...

`margin: 0` between columns + custom padding = easy "padding grids" without bloating the api.

This is why I was curious if anyone could provide real world use cases for the padding api.

andycochran:
We can use the term "gutter." We probably should; as you mention, it _is_ common parlance. But it's confusing if "gutter" can mean margin and/or padding. 

But I don't think it _can_ be either. If "gutter" is the space between two DOM elements, then `gutter === margin`. Always. If "gutter" is a more abstract way of describing whitespace between visual elements which results from margin and/or padding and applied backgrounds, then the term is not semantic as a settings variable. 

Let's call a horse a "horse" — which I think is what @corysimmons was getting at with asking about the purpose of the padding grid. A "margin grid" can have padding and zero margin. That shouldn't make it a "padding grid." 

> I'd just love to see a real world use cases for padding grids and padding + margin grids.

Here's all the ways to use the abstract concept of "gutter":

1. margin: no | padding: no
image
2. margin: yes | padding: no
image
3. margin: no | padding: yes
image
4. margin: yes | padding: yes
image

Any of these methods could be used with or without background colors to get the desired abstract "gutter" concept. But I see no reason to call option 3 a "padding grid." If we want to use the term "gutter," it should always mean "the `margin` between columns." It should not refer to the visual whitespace between columns' content. This whitespace should be referred to explicit as a combination of two things — e.g. "gutter" and "column padding."

andycochran:
> there are now multiple ways to create that space between columns

This is false if "columns" literally means `.column` elements. There are multiple ways to add whitespace between content. 

😀 See? It's hard to talk about because the terms aren't clear.

andycochran:
> This is why I was curious if anyone could provide real world use cases for the padding api.

I can't. 

I just want a mixin that says, "Create a grid with `n` columns. Set column margin (aka 'gutter') to `m`. Set column padding to `p1 p2 p3 p4`."

tdhartwick:
Hello everybody! The designers over here at ZURB had a chance to play with the new grid and it is freakin' sweet. Really well received overall!!

One comment that we'd heard during this was that `-horizontal` and `-vertical` might get pretty tiring and cumbersome to type over and over again when building. We were thinking... since this branded as the XY Grid -- meaning the `(horizontal, vertical)` coordinates -- does it make sense to use `x` and `y` to indicate direction?

We were thinking of having the classes named `.grid` and `grid-x` to indicate the typical horizontal grid and `.grid-y` to indicate a vertical grid.

Some immediate benefits:
- It's faster to type
- It has consistent branding with the XY Grid
- Less language barrier / translation of terms since it's universal math terminology
- It looks cool (ok this might be a personal opinion...)

Anyway, interested in hearing your thoughts!

rafibomb:
@andycochran brings up a good point - one I have't considered yet. I can get behind the breaking out the naming of gutters to more specific - it falls in line with the box model intuitively.

rafibomb:
+ 1 for `.grid-x` and `.grid-y`

brettsmason:
@tdhartwick Glad it was well received 😄 
I actually started off the process using x and y instead of horizontal and vertical, but switched in case it caused confusion. I'm all for it though 👍 Less typing is always good.

tdhartwick:
@brettsmason ah, awesome to hear that's how you were originally thinking about it! I also think naming the grid as "XY Grid" will help with understanding faster.

lgcallahan:
The new grid is super impressive and exciting! Looking forward to using it! Using horizontal and vertical is more straight forward, but I think after people understand that x and y is replacing horizontal and vertical, it would be faster overall. I’m a fan! :)

kball:
OK Based on discussion today and the above feedback, I'm going to put together a PR that does the following:

- Removes distinction of `.cell-vertical` and `.cell` and has all vertical/horizontal come from grid parent (we got quickly confused by having it both at grid and cell)
- changes `.grid` and `.grid-vertical` to `.grid-x` and `.grid-y`

I think there is something still to be figured out on the margin/padding... @andycochran do you want to put together a more precise proposal or PR?

canterberrie:
Love this new grid! Super excited to use these new cell-blocks. And to echo some of the ZURBians here... I'm also pro `.grid-x` `.grid-y`!

IamManchanda:
Dear Designers ( @tdhartwick, @lgcallahan , @canterberrie , and ofcourse @rafibomb )

sorry to be so prototype holic .... but i think the same should apply for [Margin](http://foundation.zurb.com/sites/6.4.0-rc1/docs/prototyping-utilities.html#margin-helpers) and [Padding](http://foundation.zurb.com/sites/6.4.0-rc1/docs/prototyping-utilities.html#padding-helpers) Classes

`m-x-0` or `mx0` instead of `margin-horizontal-0` ??

IamManchanda:
@kball @brettsmason would like your thoughts too for above :arrow_double_up:  ??

tdhartwick:
@IamManchanda Ooooh I'm diggin that! Would want to probably namespace `m-x-0` over `mx0` for potentially future classes like `mx-h` for max height?

IamManchanda:
Yup there are [sizing](http://foundation.zurb.com/sites/6.4.0-rc1/docs/prototyping-utilities.html#sizing) classes too 
For them i would like 
`w-25`, `w-50`... and `mx-w`
`h-25`, `h-50`... and `mx-h`

canterberrie:
@IamManchanda Glad those prototyping classes are finally here! Been wanting official padding & margin classes for a while

kball:
@IamManchanda @canterberrie @tdhartwick plz move prototype conversation into a different issue and leave this to grid.

IamManchanda:
I am sorry :smile: @kball

corysimmons:
@andycochran Thanks for the effort post: https://github.com/zurb/foundation-sites/pull/10017#issuecomment-305856660

I'd prefer we just abolish the term "gutter" and just refer to things as "cell margin" or "cell padding".

To me "gutter" has always meant margin between columns, but padding grids ruined it with their faux space between columns.

I suppose we could use "gutter" for margin and "faux gutter" for padding, but I think "cell margin/padding" is simpler/cleaner terminology.

tbh I don't care what terminology you guys want to go with, but it would be nice to solidify something for docs and support if nothing else.

---

@brettsmason Welcome to the part of hell where you get feedback faster than you can refactor.

I prefer `.grid-horizontal` because it's more explicit and [explicit > implicit](https://en.wikipedia.org/wiki/Zen_of_Python).

`.grid-x` and `.grid-y` aren't too bad, but I'd be careful going down that path too far.

All the `m-x-0` stuff reminds me a lot of AtomicCSS—which the general public (your audience) spent weeks RT'ing insults about. Nowadays that exact same approach is making a resurgence in the form of frameworks like Tachyons, but it's still mostly just useful for super CSS experts (again, not most of your audience).

In the short-run, these aliases work fine, but when a new dev comes into a project, it'll be particularly confusing. You _could_ offer a verbose and shorthand API (like vanilla CSS does with some properties), but I think it would just fragment the API/docs and lead to more confusion. If people want aliases or even namespacing, they're simple enough to create with preprocessors:

```scss
.zf-col {
  @include column($args...);
}
```

`.grid-horizontal` doesn't have to sacrifice its explicitness to change to `.grid-x`.

`.grid-x` is explicit because people know what "x" means in relation to a grid.

`.g-x` is not explicit. `.m-x-0` is not explicit.

```html
<section class="grid-x">
  <div class="column m-x-0">...</div> <!-- wtf? -->
</section>
```

Anyway, terse aliasing **should be a separate issue/discussion**. `.grid-x` doesn't really qualify so we can go ahead and make this change.

@brettsmason Also, kudos! It passed human trials! :D

kball:
The more I think about it, the more I think @andycochran is going in the right direction... we should be conceptualizing this as cell padding and cell margin, and then we can say that you can "use cell margin to create gutters"...

Additionally, I think we should be specifying these in all 4 dimensions (or at least 2... horizontal/vertical) letting us do things a la CSS Grid's `grid-row-gap` and `grid-column-gap`... 

I think we'd probably change the names to something like

```
.grid.padded-cells
```
and
```
.grid.margin-cells
```

or something similar... not exactly sure. thoughts @andycochran @brettsmason @corysimmons ?

corysimmons:
We can't* do 2D from one element at all. This is the whole appeal of Grid Layout.

What would `.grid` in `.grid.padded-cells` offer?

I'd rather make "disposable" class names that encouraged people manually upgrading. That is, I would _not_ include `.grid` right now. Just release `.grid-x` and `.grid-y`, then, when you go to add Grid Layout support, do so with the `.grid` class and degrade `.grid-x`/`y`.

> \* I can, but it'd require a ton of unpaid work for a fairly hacky outcome.

kball:
Well, we can't fully do 2D, but in e.g. the block grid it can be super helpful to have a bottom/top margin even as the cells are laying out horizontally... and I could definitely see applying padding on all sides as shown in @andycochran's comment above... 

The `.padded-cells` was essentially what padding-gutters does now, but experimenting with nomenclature to get away from the "gutters" concept a bit and see if that made things clearer... doesn't look like it, may need to noodle on it more.

Using the gutters nomenclature for now b/c I don't have a better offer, wonder if it's too verbose to have `.padding-gutters-x`, `.padding-gutters-y`, `.margin-gutters-x`, `.margin-gutters-y`

brettsmason:
So I've been trying to catch up on all the discussion - glad to see its a popular topic 😄 
I'm happy to move away from the term gutters all together, as @andycochran @corysimmons make some good points. I also agree with you Cory about not turning this into prototyping sounding classes - thats not a direction I want to take this, but I think `.grid-x` etc works in this scenario as it is still clear what its for.

@kball Now that we have changed it to `x/y` should I change all the mixin arguements from direction to axis, just to make it clearer?

So my thoughts on moving away from gutters:

The classes should be `.grid-margin/grid-padding` - reason being that we apply them to the grid and its grid-wide. Mentioning cell could cause confusion thinking they can be applied to cells individually.

We could then have `grid-margin-x/grid-margin-y` and `grid-padding-x/`grid-padding-y` too. It might be nice to be able to apply this to all 4 sides (at least in the case of padding), but what about for example a block grid - might be nice to be able to do left/right/bottom margin - not sure how we could handle that in terms of naming (though the current gutter mixin could handle that fine).

kball:
@brettsmason I think applying the nomenclature change to axis makes a ton of sense. Go for it!

My one hesitation on `.grid-margin` and `.grid-padding` as nomenclature is that while they are grid wide-wide applications, the elements they are actually being applied are the cells. That said, its better than anything I've come up with. @andycochran what do you think?

I think agree the mixin can handle this case just right, but we should by default create classes for the common cases. Do we ever want 4 sides margin, or left/right/bottom is the most common?

rafibomb:
Does anyone know - 

1. when, if ever, we will need a padding gutter rather than margin? 

2. Should the grid handle cell padding and gutter? What's the benefit?

I'd like to reduce confusion. It might be hard for people new to grid systems to know when/why they would use padding vs margin. If we don't need padding, let's drop it as this is a new grid.

andycochran:
In regards to margin/padding nomenclature… 
* `.grid-[padding,margin]` makes more sense if applied to the containing element which globally sets consistent spacing on the entire grid. 
* `.cell-[padding,margin]` makes more sense if applied to individual cells—which could be very useful particularly for padding cells various amounts.

Timoti:
@rafibomb re: "when, if ever, we will need a padding gutter rather than margin?"

I've had several situations I've need to do that - http://journal.phecanada.ca is an old project where the top tabs needed padding grid while content required margin grid.  Not understanding the breadth of susy's flexibility at the time I 'magic-numbered' the tab margins - but in any case, that is one such use case.

FWIW I would want to put an emphatic plug in for "gutters" as designer-friendly nomenclature.  It's universal - it's states the result (rather than the means) with what is still an appropriate metaphor. Ain't broken and don't need fixin'. 

To @brettsmason [request for grid help feedback](https://github.com/zurb/foundation-sites/issues/10136#issuecomment-306192225):

In terms of an overarching understanding of XY grids - it's not immediately clear from the docs how these work in 2 dimensions.  With all the buzz, and and opportunity around cssgrid and its 2D capability, I had hoped/expected XY grids were cssgrid with flexbox fallback (and I think there were hints that this was the direction).  But my inspector says otherwise.  Not sure there's an intension to shield newbs from the underpinnings but it would be good if the docs outlined the overarching principles and use cases for the different systems now included.  eg. I assume that the old flexbox grid is only retained for backwards compatability, but the new shiny XY grid is best for most new projects. Right?

Beyond this I was/am hopeful that the XY grid syntax – even if flexbox powered, WILL  be able to seamlessly support cssgrid going forward - ie the engine may change, but the controls will be the same. Was that a goal? if so it would be great to reassure users how where this fits going into a cssgrid-centric future.  (Here, in New Zealand, cssgrid is already supported on 80% of browsers - :) ). 

@brettsmason re the classes you asked about  - will feedback more when I've taken XY for proper spin - but again - this is where the dist/css centric docs make diving into the sass an extra layer - have to 'translate' classes to mixins etc. 

And as Judge John Hodgman would say… "that Is all" (for now).

andycochran:
> FWIW I would want to put an emphatic plug in for "gutters" as designer-friendly nomenclature. It's universal - it's states the result (rather than the means) with what is still an appropriate metaphor. Ain't broken and don't need fixin'.

I disagree that it states the result. If "gutters" are the space between things, padding gutter does not [always] achieve this. 

If you put background colors on the cells in a grid that uses padding gutter… 

image

…there's no space between the cells. There is space _within_ cells. There's nothing here that designers would call "gutter."

Issue #10019, opened 2017-05-10
Opening AJAX reveal modal by keyboard with data-open in link fails


xinspire:
#### How to reproduce this bug:

1. Go here: https://codepen.io/rafibomb/pen/kXEwWq?editors=1010
2. Add `data-open="reveal1"` to the button tag.
3. Now click outside the button in the preview area, and press `Tab` on the keyboard so that the button gets the focus, and then press Enter to "click" the button.

#### What should happen:

The modal should pop up with "I'm loaded with Ajax!" and a picture, just like it would if the `data-open` attribute is not in there.

#### What happened instead:

The modal pops up with no content.   The content shows up if you use the mouse to click, but not when you click using the Enter key.  The reason lies in the block of code at line 151 of js/foundation.reveal.js.

#### Browser(s) and Device(s) tested on:

Chrome 58.0.3029.96 on Mac OS X 10.12.4.

I've tested with Foundation 6.3.1 and 6.2.3.

#### Test case link:

- [CodePen with the data-open attribute added, based off of article at http://zurb.com/university/lessons/ajaxing-dynamic-content-with-foundation](http://codepen.io/anon/pen/Njybdd?editors=1010)

Owlbertz:
Hi @xinspire, thank you for that detailed issue.

I also think that the [keydown listener](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L151-L159) here is causing the AJAX click handler to not fire.

A possible workaround until this is patched would be to unbind this specific listener using
```js
$('#trigger-reveal').unbind('keydown.zf.reveal');
```

xinspire:
Thanks!

Owlbertz:
Hi again @xinspire,
I just notices something while attempting to fix this issue.
As you do manually open the Reveal in your AJAX callback, you could easily just remoe the `data-open` attribute from your button. This way the listener will not behave the wrong way and interrupt the event propagation.

xinspire:
Ah, yes.  That was my initially solution before you commented, but wasn't sure if it was considered "hacky."  Does that mean that data-open doesn't really serve a purpose when reveal is used in AJAX mode?

Owlbertz:
[...]

PR #10023, opened 2017-05-12
Reveal ajax keydown

Owlbertz:
Removed keydown handler that used to trigger open.
Will still trigger for space and enter automatically for buttons as Reveal anchors (i.e. triggers).
However, space will not trigger the click event for links (which is the default behavior).
As buttons should be used as anchors, this follows the specs.

Relevant for issue #10019.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -148,16 +148,6 @@ class Reveal extends Plugin {
       }
     });
 
-    if (this.$anchor.length) {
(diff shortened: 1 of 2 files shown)


kball:
👍

kball:
@Owlbertz just to confirm, this is a "bug fix" in the sense that it should be backported to the 6.3 series as well as being in place for 6.4?

Issue #10031, opened 2017-05-16
Responsive menu small to medium broken in 6.3.1


kball:
This appears to be a regression introduced in 6.3.1 from 6.3.0.  If you set up a responsive menu with drilldown on small, then dropdown for medium up, and resize to small and then back to medium the dropdowns will not work.

See this example:

https://codepen.io/kball/pen/GmBYjx?editors=1000

kball:
This is related to this commit: https://github.com/zurb/foundation-sites/commit/893a5d71589a77505c4aefb2f6639c4a97024dbe

and a workaround is here: https://codepen.io/kball/pen/rmrqWO?editors=1010

Thanks to @ttyson99 for the report. Will get this fixed for 6.4 & 6.3.2

kball:
Just was checking into this, it turns out it is already fixed for 6.4 and 6.3.2 in this PR: #9949 
Both should ship middle of June.

PR #10033, opened 2017-05-17
Add a Patch for a dropdown-drilldown bug within docs.

IamManchanda:
#10031 Add a Patch for a dropdown-drilldown bug within docs.

I have also created a 6.3.1 specific pen to reproduce the issue specifically for that same version 
and have linked it within docs... **I think this will be helpful for the community.**
=> http://codepen.io/IamManchanda/pen/mmGOgG?editors=1000

cc @kball 

bug-patch-drilldrop


kball:
👍

PR #10039, opened 2017-05-18
Need help: Looking to create a toggle on allowing multi expand 

IamManchanda:
As per => http://zurb.com/article/1470/the-new-foundation-docs-learn-your-way#comment-3311190013

cc @kball for help!

What we need i feel is a checkbox "Allow nav to multi-expand"
What's the best way to go about it??


kball:
@IamManchanda oop didn't see this before putting in a PR myself here: https://github.com/zurb/foundation-sites/pull/10042

PR #10042, opened 2017-05-18
Add persistent expand/contract state for docs nav

kball:
Adds a persistent expand/contract state based on top link for docs.

### DIFF
--- a/docs/assets/scss/docs.scss
+++ b/docs/assets/scss/docs.scss
@@ -243,3 +243,17 @@ a[data-open-video] {
 .mb1 {
   margin-bottom: 1rem;
 }
+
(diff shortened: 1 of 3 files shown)


docs/assets/scss/docs.scss
```diff
@@ -243,3 +243,17 @@ a[data-open-video] {
...
   margin-bottom: 1rem;
 }
+
+.docs-nav-category {
+  position: relative;
+}
+
+.docs-nav.menu {
+  .docs-expand-all {
+    font-size: 0.5rem;
```
IamManchanda: Too Small .... i think 0.6rem is the correct value!

Issue #9575, opened 2016-12-30
[Dropdown] Vertical scrollbar is still visible after dropdown pane is closed


jimmy72:
I added the following style to a dropdown pane to make it scroll vertical in case the content gets too long:

`<div id="j1" style="max-height: 250px; overflow-y: scroll;" class="dropdown-pane" data-dropdown data-auto-focus="false" data-close-on-click="true">`

Everything work as expected but after closing the dropdown, the scrollbar is still visible and does not disappear.

I also tried to add these styles to a container inside this dropdown (not on the dropdown itself) but that doesn't work either.

I cannot figure out if this is a bug or me doing something wrong here?

Thanks

Jimmy

ncoden:
Hi @jimmy72,

I can not reproduce the bug you are describing: https://codepen.io/ncoden/pen/YpmMOP.
Can you provide a codepen ?

ncoden:
There is a bug between `overflow: scoll` and `visibility: hidden`. The component is hidden but not the scrollbar. I was able to reproduce it in [MDN](https://developer.mozilla.org/fr/docs/Web/CSS/overflow-x) but not in codepen, this is weird.

![dec-30-2016 14-51-49](https://cloud.githubusercontent.com/assets/9939075/21566069/94ec341c-ce9f-11e6-8d3f-7e6e98d73247.gif)

I do not know why we use `visibility:hidden` and not `display:none`. @kball Any thoughts on this ?

jimmy72:
Ok I see...

I'll wait for the bug being fixed...

In the meantime I worked around it through jQuery....

$dropdowns.on('show.zf.dropdown hide.zf.dropdown', function(event) {
    event.preventDefault();
    var $thisDropdown = $(this);
    if(event.type == 'show'){
        $thisDropdown.css({
          "max-height": "300px",
          "overflow-y": "scroll"
        });
    }
    if(event.type == 'hide'){
      $thisDropdown.css({
          "max-height": "",
          "overflow-y": ""
        });
    }
  });

Jimmy

2016-12-30 14:53 GMT+01:00 Nicolas Coden <notifications@github.com>:

kball:
I also don't know why we use visibility hidden instead of display:none.. i think I encountered a different bug related to this the other day. Will take a crack on this for 6.4

kball:
Fixed in #10049. Will be in both 6.4 and 6.3.2

Issue #9945, opened 2017-04-11
Dropdown causing horizontal scrollbar


mitja-p:
#### How to reproduce:
1. Add a dropdown on the right side of the webpage (float:right will do)
2. Open and close it
3. Resize to smaller width

On smaller browser width sometimes even default dropdown position causes a horizontal scrollbar.

#### Example (borrowed from another issue):
http://codepen.io/tw42/pen/dNEvOJ
1. Just resize to smaller width and notice the empty space on the right.
2. Click the right most button to open and close the menu on desktop width. Now resize to make it mobile width. Horizontal scrollbar appears, because dropdown is only visibility:hidden and therefore stays on the right. Dropdown should move to the left when closed so it does not cause the scrollbar to appear.

#### Partial workaround for one part of the issue:
CSS: 
`// Initially move dropdown to the left and set max width to 100%`
`.dropdown-pane:not(.is-open){`
`  left: 0px; // do not set !important as it won't work on larger screens`
`  max-width: 100%;`
`}`
Problem of opening a dropdown on larger screen and resizing to smaller size still remains.

mitja-p:
Another side effect. When opening the dropdown, the dropdown gets moved to multiple positions to find a position that fits the available space. However on Android (also desktop Chrome in mobile development mode) this causes the page to scroll up. You can try manually settings dropdown element left value to 1500px and then 10px and see the effect.

kball:
Overflow/scrollbar issue should be fixed by #10049, which will be in both 6.4 and 6.3.2. Positioning improvements are also coming for 6.4 with #10053

PR #10049, opened 2017-05-19
Fix dropdown issue related to odd interactions with visibility:hidden

kball:
Fixes https://github.com/zurb/foundation-sites/issues/9575


kball:
Going to go ahead and merge this. It also should fix  https://github.com/zurb/foundation-sites/issues/9945

Issue #9989, opened 2017-04-24
[Tooltip] Considering Tether type functionality?


IamManchanda:
I think my comment was missed so here it is ??

https://github.com/zurb/foundation-sites/issues/9618#issuecomment-292786762

> i think we should seriously consider **[Tether](http://tether.io/)**
> 
> I mean just look at this PR twbs/bootstrap#22264 and changes
> Demo => http://jsbin.com/lijalowata/edit?html,output
> 
> With this all 12 positioning elements can be used
> Default: top | right | bottom | left
> Optional: top left | top right | right top | right bottom | bottom left | bottom right | left top | left bottom

IamManchanda:
if this is a stupid idea or can be done through current JS then let me know!

ompmega:
Support! Please implement this, or if there's a way to implement it ourselves, that would be awesome! +1

IamManchanda:
cc @kball!

IamManchanda:
@ompmega We have started work on it but with current resources instead of tether!
Thanks for the shoutout!

IamManchanda:
Added in #10053 .... closing this!

Issue #9799, opened 2017-02-21
[Improvement] dropdown position configuration


SassNinja:
I like the framework's dropdown component. But very often its configuration is much to limited for me and I can't use it. The main issue in my eyes is the limited position configuration.

On the one hand I think it's inappropriate to bind the horizontal position to a float class (I have to use a custom float-none class if I don't need the float but only the position). On the other hand (much more important) I sometimes want to define a horizontal AND a vertical position.
Currently it's either or.

In my opinion it would be better if you can define both positions at the dropdown element. So for example 'left bottom' or 'right top'. Perfect solution would be if you can define a position related to the element (left|top|right|bottom) and as second parameter the horizontal or vertical alignment for it (left|top|right|bottom|center). Attached you can find a quick mock of what I mean.
This would heavily improve the dropdown module and make it super flexible.

I can try to adjust the code to achieve at least parts of my described solution. But since I'm not a javascript ninja this can take a while and may need some reworking.
Did anyone else face the limited options of the dropdown or has maybe already started to work on an improvement? (I couldn't find an issue or PR yet)

Any help on this is appreciated :-)

![dropdown-improvement](https://cloud.githubusercontent.com/assets/19912475/23163513/07d04b00-f833-11e6-9660-1b2bd9436c6b.jpg)

kball:
I agree that dropdown could definitely use an update... do you want to take this on? Happy to consider a much more sass-based approach than the current as well... the current approach uses JS to do a lot that I think could probably be done with SCSS.

SassNinja:
Yes, I'll try to work on this asap.
But talking about more sass-based: do have something particular in mind?

I've seen the whole positioning is done via JS (absolute) instead of CSS that is relative to the dropdown toggle button. But on the other hand this allows more flexibility in the markup since the button mustn't wrap the dropdown pane...
Did you mean this or something else?

kball:
@SassNinja fyi working on this in #10053. If you're interested in taking a look/helping push it forward, I'd be happy for the help

kball:
Implemented and merged in #10053

SassNinja:
@kball sorry for my late reply (I was quite busy with the nested off-canvas feature and lost track on this one)

Thanks for taking care of the dropdown improvement!
Explicit positioning looks good :+1:

PR #10053, opened 2017-05-20
Dropdown explicit positioning

kball:
Create a fully explicit positioning model for Box a la Tether (CC issue #9989). Initially inspired by https://github.com/zurb/foundation-sites/issues/9799 for dropdown but should also work for tooltips.

This is WORK IN PROGRESS do not merge... checklist below

- [x] Set up explicit positioning and alignment
- [x] Create visual tests for full position and alignment
- [x] Get dropdown working when fully positioned and aligned
- [x] Add configuration for whether dropdown can overflow container or not.
- [x] Create visual tests for adjustments on overflows (IN PROGRESS)
- [x] Get dropdown to adjust properly when positioned and alignment overflows (IN PROGRESS)
- [x] Get dropdown to adjust properly when not positioned or aligned and defaults overflow
- [x] Create/verify tests for prior behavior of dropdown
- [ ] Create tests for Box.ImNotTouchingYou to make sure the new implementation over Box.OverlapArea is 100% compatible.
- [x] Get as many cases of prior behavior to be 'automatically' determined as possible, to ease migration
- [x] Update tooltip to use new Box methods.
- [x] Add tooltip positioning visual tests
- [x] Document new positioning approaches


IamManchanda:
@kball i tweaked a bit with visual testing and i think overflow has its issues .... apart from it looks good !

IamManchanda:
@kball i added a test case ... is this the right behaviour what's happening??

kball:
@IamManchanda good question... as currently constructed, the overflow/repositioning only checks 1 way if you set the position. IE if you set position top or bottom it will only look for left/right overflow, and if you set position left/right it will only look for top/bottom overflow.

I'm not sure that's the right behavior... however it is more consistent with past behavior.

I haven't yet implemented the all position all alignment overflow checking (that I want for cases where position is auto anyway), so once I do maybe we can look at how that behavior works and see if it's better...

One other outstanding question is... what if there is *no* way to lay out the dropdown without an overflow? What should behavior be?  I could make cases for defaulting to the original position/alignment, or for trying to find the "least overflow" case... what do you think?

IamManchanda:
@kball => Trying to find the "least overflow" case

kball:
OK I think to do least overflow case we'll just need to modify the box ImNotTouchingYou to return the overlap amount rather than just true/false. I don't think I'll have time today to push this forward but probably tomorrow

kball:
OK I just pushed updates that get all of the overflow adjusting working "as expected". It feels magical, and is kind of awesome.  Need to add a bunch of tests for previous behavior and make sure that we're backwards compatible as much as possible... 

One outstanding question for me is... I'm confident we want to be backwards compatible for dropdowns themselves... if you've gone through work to get your dropdown to lay out the way you want, it should continue to do so... and our "base cases" should work right...

What I'm wondering is how much to worry about to maintaining backwards compatibility for Box.GetOffsets (which was always kinda hokey and automagical) or if we can just update tooltip as well (the only internal client that uses Box) and more or less wave our hands with Box.GetOffsets... with a "decent attempt" at backwards compatibility.  

Or maybe put it back to the old code, which has a lot of duplication/overlap with the new Box.GetExplicitOffsets, and put a deprecation warning in that it's obsolete and will be removed in 6.5.  Actually, after writing through these, I think that is probably the best option... update tooltip to also use GetExplicitOffsets, revert GetOffsets to original state, and add a deprecation warning.

Any @zurb/yetinauts want to chime in on this approach?

IamManchanda:
Will test it out tonight for sure... Yup magical coz we didn't had to rely on other plugins like tether and create one more dependency for the users!

IamManchanda:
Updating tooltip also make sense

IamManchanda (approved):
Approving current changes .... tested it out .... had some browser testing too
1. Windows 10 (IE, Edge, Safari, Firefox, Chrome, Opera, Yandex)
2. Mac Sierra (Safari, Firefox, Chrome, Opera, Yandex)
3. Ubuntu 16 (Chrome, Firefox, Chromium, Ubuntu)

This looks good and awesome!

kball:
OK, got a first pass of tooltip working with the new positioning model... or at least, the new positioning logic (refactored to share logic with dropdown). Need to put together some visual tests for tooltip the same way we did on dropdown, but it's at least working for the cases in the docs already.

Also fixed up some dropdown cases based on our other tests and docs. I think we're looking good, but @rafibomb has some existing implemented cases we want to pull into the docs to make sure they behave correctly backwards.

IamManchanda:
Hey @kball i added test case but it seems like buggy .... fetching `data-alignment` in the opposite direction!
Can you check!

kball:
Ah interesting... there's two pieces here. 

The first is that tooltips are treating vOffset and hOffset differently than dropdowns (I ran into this a little bit and hacked around it for the centered case, but these tests highlight that I actually need to separate the concept of the offset that leaves space for the nub (applied in your *position* axis, but not in the your *alignment* axis) from the concept of hOffset and vOffset which are a way to manually nudge things around.

The second is that we need to actually reposition the nub based on alignment (probably by applying `align-x` classes.

kball:
OK, fixed tooltip and got it working as we'd like it. The outstanding pieces of this are to test against some more existing cases (particularly in dropdown) to verify backwards compatibility and document the explicit usecase.

@rafibomb can you help get those two pieces across the line?

kball:
We could probably merge for the RC even without those 2 pieces, as functionality-wise I believe this is now complete.

IamManchanda:
Ohk @kball .... have updated the docs! 
When that nub is fixed ... good to merge i guess!!

kball:
nub aligned... @rafibomb is working on getting me another good testcase and then I think this is good

kball:
I think this is good to go! Merging

Issue #9991, opened 2017-04-26
Documentation for Title Bar component is missing


Proper-Job:
There is a mixin called `foundation-title-bar` located in foundation-sites/scss/components/_title-bar.scss.  However there is no documentation related to that component.

IamManchanda:
@Proper-Job Do you mean `top-bar-title` ??

Proper-Job:
@IamManchanda No, I don't mean top-bar-title.
In several places the ``.title-bar`` class is mentioned, for instance in:
- [Sticky Navigation](http://foundation.zurb.com/sites/docs/sticky.html#sticky-navigation)
- [Off-canvas](http://foundation.zurb.com/sites/docs/off-canvas.html#combining-with-title-bar)
- [Kitchen Sink](http://foundation.zurb.com/sites/docs/kitchen-sink.html#title-bar)
- [Responsive navigation](http://foundation.zurb.com/sites/docs/responsive-navigation.html#responsive-toggle)
- [foundation.scss](https://github.com/zurb/foundation-sites/blob/v6.3.1/scss/foundation.scss#L109)

However, there is no documentation as to what the title-bar component actually does or how to properly configure it.

IamManchanda:
Sorry... this got missed...
So you basically want the docs for the container of the hamburger menu ??

https://codepen.io/IamManchanda/pen/LymroM?editors=1100
```html
<div class="title-bar" data-responsive-toggle="example-menu" data-hide-for="medium">
  <button class="menu-icon" type="button" data-toggle="example-menu"></button>
  <div class="title-bar-title">Menu</div>
</div>
```

Proper-Job:
@IamManchanda thanks for getting back to me.
It would be nice to have a section explaining what the Title Bar component does and when it should be used.  Hints are scattered throughout the docs, but like I said, searching for "Title Bar" yields no results.
For instance there are the ``.title-bar-left`` and ``.title-bar-right`` classes that, I assume, align content left and right.  However, this information I glean from a section in the docs called: "Sticky Navigation".
I'm also wondering what the difference is between Top Bar and Title Bar and when should I use which.
Forgive me if I'm being obtuse, but I believe a doc section covering the Title Bar component is merited.

IamManchanda:
cc @kball .... shall we have it?? 

PS: I also think the same that we should have it!

IamManchanda:
https://github.com/zurb/foundation-sites/pull/10056#issuecomment-305307281

[...]

PR #10056, opened 2017-05-21
Closes #9871: Flex Topbar

### LINKED ISSUE #9871: 6.3.0 Advanced top-bar with flex-box enabled causes the responsive-menu to render to the right
adjust configuration to enable global flexboxes and import required flex box classes in sass.
Add the advanced top bar menu example per the docs. http://foundation.zurb.com/sites/docs/top-bar.html

#### What should happen:
the top-bar-left container should align to left and the top-bar-right should align to right.

#### What happened instead:
the top-bar-left container is stacked on top of the top-bar-right container.  the entire responsive-menu is aligned to the right with a constrained width. (See code pen)

#### Test case:

http://codepen.io/anon/pen/vxZWed

IamManchanda:
This closes #9871.
The new docs [codepen example](https://codepen.io/IamManchanda/pen/LymroM?editors=1000) already fixes this. 

This just adds 

- [x] Titlebar tweaks to make it look similar to float one
- [x] Add a test case for new docs example!
- [x] Docs update with new codepen example

Here is the flex version codepen (with titlebar tweaks) => https://codepen.io/IamManchanda/pen/OmBVZE?editors=1100

### DIFF
--- a/scss/components/_title-bar.scss
+++ b/scss/components/_title-bar.scss
@@ -42,7 +42,7 @@ $titlebar-icon-spacing: 0.25rem !default;
 
     @if $global-flexbox {
       display: flex;
-      justify-content: space-between;
+      justify-content: flex-start;
       align-items: center;
     }
     @else {
--- a/docs/pages/top-bar.md
+++ b/docs/pages/top-bar.md
@@ -54,53 +54,41 @@ In the below example, our top bar includes a [dropdown menu](dropdown-menu.html)
 
 ## Advanced Layout
 
-You can further divide a top bar into a title area and content area. Use the `.top-bar-title` class to create a title/branding area. Next to that can be any element, which is used for the rest of the content.
+<p>To set up a Responsive menu with toggle, first give your menu a unique ID. Next, add a title bar with the class <code>.title-bar</code> and the attribute <code>data-responsive-toggle</code>. The value of <code>data-responsive-toggle</code> should be the ID of the menu you're toggling. Lastly, add <code>data-toggle</code> to the element that will trigger the toggle. The value of <code>data-toggle</code> should also be the ID of the menu you're toggling.</p>
 
-```html
-<div class="top-bar">
-  <div class="top-bar-title"></div>
-  <div>
-    <div class="top-bar-left"></div>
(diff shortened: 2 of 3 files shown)


IamManchanda:
@kball ,,,, i think the docs update should go in master (live) also!

kball:
@rafibomb can you take a look at this?

scss/components/_title-bar.scss
```diff
@@ -42,7 +42,7 @@ $titlebar-icon-spacing: 0.25rem !default;
 
     @if $global-flexbox {
       display: flex;
-      justify-content: space-between;
+      justify-content: flex-start;
```
rafibomb: I don't see why this should be start (left align)

By default I think it should be space-between is you had 2 sides like:
http://foundation.zurb.com/sites/docs/off-canvas.html#combining-with-title-bar

rafibomb: Interestingly, this component can be manipulated with the flex helper classes. I think title bar should get it's own docs page.

rafibomb: Ok I see when you dont use title bar left or right then flex start makes sense. I've added to your test case to demonstrate.

IamManchanda: So shall we insert this in a variable ? 
@rafibomb?  
and keep flex start? 
And some docs about it?

IamManchanda: On title bar docs,  yes I agree infact having same discussion about it at #9991

IamManchanda:
@rafibomb @kball Awaiting reply for above
Can we close this before 6.4 RC?

rafibomb:
I don't know how I feel about removing the all in one responsive toggle docs example - it's nice to have but I get it needs work for flex version

rafibomb:
It looks good from a bug fix perspective - merging!

What I'd like to see in the future:

- [] title-bar has it's own docs page within navigation
- [] title-bar inherits top-bar styles (height, color, background)
- [] or eliminate the thing altogether and use top-bar on mobile and larger

Thoughts? @IamManchanda @kball @brettsmason @andycochran @colin-marshall

PR #10062, opened 2017-05-22
Call $(window).on in init only once

bpceee:
fix some some memory leak here.

### DIFF
--- a/js/foundation/foundation.js
+++ b/js/foundation/foundation.js
@@ -293,6 +293,27 @@
     this.query = '';
   }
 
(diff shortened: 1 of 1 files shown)


dragthor:
Which issue does this address?  I am unable to reproduce.

Included in Zurb 5 is debounce and throttle.  Maybe these will you?  Additionally, you could even "bring" in underscore's _.once() function.

bpceee:
Ok, it's long time ago, maybe I was wrong, sorry about this.

Issue #9968, opened 2017-04-20
select icon in chrome


captain-xu:
There two icons in chrome  , the select should add -webkit-apperance: none

Deckluhm:
As stated [here](http://foundation.zurb.com/sites/docs/sass.html#autoprefixer-required), vendor prefixes aren't included in Sass sources.

kball:
We can if necessary to fix a bug... can you put in a reproduction codepen?

Deckluhm:
> We can if necessary to fix a bug... can you put in a reproduction codepen?

The bug is only visible if you use the Sass version of Foundation (since there is no vendor prefixes in sources) and as far as I know, there is no easy way to import a full Sass project in CodePen.

I reproduced the bug @captain-xu is talking about by using Foundation CSS version and by setting `select` `-webkit-appearance` to the default WebKit value (`menulist`):

https://codepen.io/anon/pen/EmKZgW?editors=1100

captain-xu:
@Deckluhm that's right,   there are no problems if use the css. And I think, the select -webkit-appearance show be 'none'

kball:
@IamManchanda I think this is similar to the fix you did in https://github.com/zurb/foundation-sites/pull/9868 - do you want to take a pass on it?

IamManchanda:
Ohk @kball will take a look!

IamManchanda:
@captain-xu @Deckluhm 
I couldn't reproduce in the sass build codepen! 
https://codepen.io/IamManchanda/pen/wdQVoB

Apart from that i have added test case here (where also i cant see the bug)
https://github.com/zurb/foundation-sites/pull/10063 
Can you pull it and let me know with screenshot and browser details ??

rafibomb:
@Deckluhm Hmm, as far as I know, Autoprefixer is running on the ZURB Stack and Basic Template. You can see it here,
https://github.com/zurb/foundation-zurb-template/blob/master/gulpfile.babel.js#L84
https://github.com/zurb/foundation-sites-template/blob/master/gulpfile.js#L17

If you are not auto-prefixing, then manually adding the vendor prefixes would be good.

I am also seeing the correct declarations in my 6.3.1 project:
<img width="971" alt="screen shot 2017-05-22 at 10 22 42 am" src="https://cloud.githubusercontent.com/assets/5652085/26320703/b55a1ed6-3ed8-11e7-9040-fb774c542098.png">

Unless I'm missing something, let me know - we'll close this.

[...]

PR #10063, opened 2017-05-22
Just added a test case for select icon

IamManchanda:
Test case for #9968


IamManchanda:
Merging this test case for keeping it in records for #9968
Although bug cant be repro till date! Thanks!

PR #10065, opened 2017-05-23
remove useless lines for padding props

benjamin-schneider:
We did a css selector performance test which suggest us to remove the padding-top and bottom declarations, as they are overwritten by the padding shorthand property.


Owlbertz:
That seems reasonable.

kball:
Merged! Thanks @artebenji, and congrats on first PR merged

PR #10068, opened 2017-05-23
Docs installation page updates

rafibomb:
Updates content with more informative copy, videos, video deep links and external supporting links


docs/assets/scss/docs.scss
```diff
@@ -432,3 +432,26 @@ a[data-open-video] {
     color: #aaa;
   }
 }
+
+.button.clear {
```
kball: Clear buttons are coming, so we don't want to conflict... lets scope this to docs & add comment that this will come in the next release and these styles should be removed then.

docs/pages/installation.md
```diff
@@ -1,59 +1,177 @@
 ---
 title: Installation
-description: There are many ways to install Foundation, but if you're just getting started, we have a few suggestions.
+description: There are many ways to install Foundation from downloading the compiled CSS, automated Sass installs to popular package managers. Each one serves a different skill level or preferred workflow. If this is all new to you, don’t worry; The video below will give you an overview of each method and help you decide which is best for you.
```
kball: This is super long, to the point I think it doesn't look great as it goes above the vid. Can we reduce this and move more of the content below?

rafibomb: yeah

docs/pages/installation.md
```diff
@@ -1,59 +1,177 @@
...
+
+  <li><a href="#manual-setup"><strong>Manual Setup</strong></a> - You can Install from either the ZURB Stack or Basic Template manually. <span style="vertical-align: top; border: 0; border-radius: 3px;" class="label">Sass</span></li>
+
+  <li><a href="#package-managers"><strong>Package Managers</strong></a> - Install foundation through your favorite package manager like npm, Bower, Meteor, RubyGems, Composer or NuGet. <span style="vertical-align: top; border: 0; border-radius: 3px;" class="label">Sass</span></li>
+
+  <li><a href="#yeoman-generator"><strong>Yoeman Generators</strong></a> A generator is basically a plugin to scaffold complete projects or useful parts.
+ <span style="vertical-align: top; border: 0; border-radius: 3px;" class="label">Sass</span></li>
+
+  <li><a href="#download-the-sass-files"><strong>Download the Sass files</strong></a> - Already have a workflow you like? No problem, download Foundation's core SCSS files and you can go your own way. <span style="vertical-align: top; border: 0; border-radius: 3px;" class="label">Sass</span></li>
```
kball: This link doesn't appear to be working

docs/pages/installation.md
```diff
@@ -158,6 +250,21 @@ Here's what comes in the package.
```
kball: I think this could use some additional work... if someone is brand new, we should clarify to them when they'd use SASS, when they'd use CSS, and when they'd use dist vs the other files, and maybe link them to resources on those. Right now if I didn't have prior knowledge I'd be totally confused about why there's both `js` and `dist/js`, as well as whether I was supposed to use `scss` or `dist/css` or both

docs/pages/installation.md
rafibomb: done

docs/assets/scss/docs.scss
rafibomb: done

docs/pages/installation.md
rafibomb: added this section! I think we need to package up a Sass download with the essentials rather than the whole framework @kball

docs/pages/installation.md
```diff
@@ -248,6 +270,12 @@ Here's what comes in the package.
   - `js/`: Concatenated JavaScript files. Includes minified and unminified files.
     - `plugins/`: Standalone JavaScript plugins.
 
+Once installed with one of the above commands, Foundation has been downloaded via git and installed into the `assets/` directory.
```
kball: This is not correct. Each package manager uses its own install mechanism, of those only bower is actually using git, and none of them put Foundation into `assets`.  I think we can probably drop this... lets focus this section on explaining what the pieces are and when you'd use them, as below...

docs/pages/installation.md
```diff
@@ -248,6 +270,12 @@ Here's what comes in the package.
...
+
+If you are using Foundation Sass, you'll `@import` Foundation from the `scss/` folder. You will also link to the `js/` folder where all the uncompressed JS files live.
```
kball: These are two separate concepts.  The Sass statement is correct... the JS folder piece should be for if you're using a JS build system and integrating this into an ES2016 project.

docs/pages/installation.md
kball: @rafibomb why? The point of SCSS is you can import only what you need

docs/pages/installation.md
```diff
@@ -1,59 +1,195 @@
...
-### Command-Line Tool
+## How to choose the right version for you?
 
-Not a fan of GUIs? The Node-powered Foundation CLI can install the same template projects for you. Install it with npm:
+Each of these different ways to get started serves a different skill level or preferred workflow. If this is all new to you, don’t worry; The video above will give you an overview of each method and help you decide which is best for you.
 
-```bash
-npm install --global foundation-cli
-```
+Here’s a breakdown of the options you’ll see on this page:
```
kball: This breakdown is way better, but I think a little bit intimidating, especially if you're brand new to web development. Can we highlight which one we recommend for a complete beginner?

DanielRuf:
It seems there are still many things open and not changed as recommended. Personally I would prefer a much simpler overview and texts in the form `tldr; run these commands (with descriptions)` or something similar.

ncoden:
This PR is out of date. See https://github.com/zurb/foundation-sites/pull/10847

DanielRuf:
I guess we can close this here then or does it just need a rebase?
https://github.com/zurb/foundation-sites/pull/10847 is what I've meant, better structure and better overview and easy to understand information.

DanielRuf:
@ncoden is this still a useful addition in general? I think we have to restructure this a bit.

ncoden:
@DanielRuf Closed, thanks for the reminder

PR #10081, opened 2017-05-28
Closes #10080 - Hollow Button Group and `no-gaps`

### LINKED ISSUE #10080: [Feature Request] Hollow button group

1. Step one

Create a BUTTON GROUP - The docs provide for All basic types of coloring - Alert, warning etc.
However - an important basic button coloring --  "HOLLOW" is not available for the BUTTON GROUP.

IT IS AVAILABLE for granular button - which is cool, But it would be nice to be able to
assign it to entire button group also if desired without having to insert a class = "button primary hollow" to every button in the group.

#### What should happen:

class = "button-group primary hollow"

Primary Colored Border - Transparent fill

#### What happened instead:

No Hollow Available for button group that renders like a HOLLOW SHOULD

ALL current

#### Test case link:

- [CodePen with Foundation 6.3.0 and MotionUI](http://codepen.io/IamManchanda/pen/vgemXR)
- [CodePen with Foundation 6.3.0, Flexbox grid and MotionUI](http://codepen.io/IamManchanda/pen/xgXdRq)

- [CodePen with Foundation 6.3.1 and MotionUI](http://codepen.io/IamManchanda/pen/LWGZxR)
- [CodePen with Foundation 6.3.1, Flexbox grid and MotionUI](http://codepen.io/IamManchanda/pen/zZrBEv)

IamManchanda:
Closes #10080

Changes
- Adds Hollow Button Group
- Adds a class `no-gaps` for no spacing between buttons within a button group ( for both hollow and simple buttons)
- Update docs!

cc @kball .... as i think that for hollow buttons i did wrong coding (should be using some code from `_button.scss` which i am unclear about it)


Deckluhm:
See my comment: https://github.com/zurb/foundation-sites/issues/10080#issuecomment-304624376

kball:
As per the comment above, deferring this until 6.4.x

ncoden:
> Some cleaning and refactor seems required here.
>
> At first, there is already some logics to generate buttons styles (solid, hollow and clear) classes. We should use reuse it for button groups.
    Also, we should not generate these classes when their styles is already the default one.
    Finally, I think there is some CSS duplication we could remove between buttons and button groups.
>
> And as you already know, I love refactoring ;)
> -- https://github.com/zurb/foundation-sites/issues/10080#issuecomment-354623701

Working on it

ncoden:
So I made some cleaning around button and button-group:
- [x] move all button filling (and no-gap) styles to mixins and use them to generate CSS
- [x] process the same way `hollow` and `clear` filling (in both button and button-group)
- [x] add some doc
- [x] fix the border problem with `no-gap`

To do:
- [ ] more factorization: process the same way all filling types (`solid`, `hollow` and `clear`)
- [ ] add doc about all these new features
- [ ] allow `no-gap` to be used with any border size (currently only 1px)

Poke @IamManchanda @blynnc @developer-lindner @Deckluhm

ncoden:
I move this to Foundation v6.5.0. This introduce new backward-compatible functionnalities and according to SemVer this should not be released in a patch version.

```
Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API. It MUST be incremented if any public API functionality is marked as deprecated. It MAY be incremented if substantial new functionality or improvements are introduced within the private code. It MAY include patch level changes. Patch version MUST be reset to 0 when minor version is incremented.
```

Foundation did not followed SemVer very carefully since then, but I'll try to be more strict about that now.

ncoden:
Poke @IamManchanda. Could you review this ?

ncoden:
Poke also @IamManchanda @Owlbertz @brettsmason @colin-marshall if you are interested.

IamManchanda:
Superb @ncoden

PR #10086, opened 2017-05-30
Slight tweak for combined row/column description

adrianpauly:
I had never fully understood what the row/column combination on a single element did exactly. Today I clicked on the "Edit in browser" link and found a different explanation in codepen. Key words here: "If you have just a single column, you can save some markup" -- A _single column_! This shed some light on the concept. 

Maybe it's just me, but the current text "If you need a full-width column to use as a container," had never been clear to me which width it was referring to. Also, the "save some markup" part also cleared up the purpose of this technique. So basically I'm just proposing changing the current description to be a little more like the Codepen description.

- For bug fixes or new features compatible with the next version, use `develop`.
- For bug fixes incompatible with `develop`, use the support branch of the latest compatible version, which will be formatted `support/6.x`.

Happy coding! :)

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -168,7 +168,7 @@ Medium sized screens will inherit styles from small, unless you specify a differ
 
 ### Combined Column/Row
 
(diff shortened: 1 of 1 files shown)


rafibomb:
I like it! Thanks @adrianpauly

PR #9738, opened 2017-02-01
[Advancement] Prototype/sepration from core

IamManchanda:
**Update**: This is Moved to https://github.com/zurb/foundation-sites/pull/10091

---------------------------------------------------------------------------------------------------------------

Implementation for #9736 

This adds helper classes for prototyping ( with responsive breakpoints ) but with a separate folder `scss/prototype` and is **disabled by default** (so, it doesn't heavy the framework) and is easily available with a simple 

```scss
@include foundation-prototype-classes;
```

To include only the specific prototype class ( eg, display ) you are looking for, you can use

```scss
@include foundation-prototype-display;
```

**Moreover you can also enable responsive breakpoints for prototype classes via sass variables.**
Responsive breakpoints are **disabled by default even if prototype mode is enabled**. You can turn it on through sass variables.

Special Note: For immutability purpose, Many Prototype classes use `!important` to ensure that these they aren't overriden by more specific selectors. As we all know that foundation framework conscientiously avoids using `!important` declarations. Also, note that i have only inserted `!important` on those specific CSS properties which i was fully sure that values of those properties shouldn't be overridden.

----------------------------------------------------------------------------------------------------------------------------

**Here is the task list**,
This prototype advance will gonna include:

**1. Current Classes** 
Added a depreciation notice as this will be added(moved) to prototypes in **v6.5** instead of **v6.4**. 
This is basically inline with the discussion i had with @kball in a **DM** on how to go about the current prototypes classes that are already there in the docs.

- [x] Float classes & clearfix ( #9723, Merged within this ) 
- [x] Visibility Classes
- [x] Text alignment classes

**2. New Classes**
Simply will be added in **v6.4** .... There are no second thoughts what so ever. 

- [x] Component Styling (Rounded, Bordered & Shadow)
- [x] Font Styling ( weight, italic, wide )
- [x] Title Bar Utility
- [x] Typescale
- [x] List Styling ( ordered, unordered )
- [x] Margin Helpers ( #9607 )
- [x] Padding Helpers ( #9607 )
- [x] Sizing Utilities 
- [x] Border box
- [x] Display classes ( #9719 )
- [x] Position Helpers
- [x] Overflow ( visibile, hidden, scroll )
- [x] Image Replacement, Text HIde ( #9718, Merged within this  )
- [x] Text transformation classes ( #9718, Merged within this  )
- [x] Text decoration classes
- [x] Text Truncate ( #9718, Merged within this  )
- [x] Text Wrapping ( #9718, Merged within this  )

[...]

PR #10091, opened 2017-05-31
Prototype Utilities - Successor of #9738

IamManchanda:
Successor of #9738 
cc @rafibomb ... lets extend the work here :)


scss/prototype/_spacing.scss
```diff
@@ -0,0 +1,204 @@
...
+    .margin-#{$spacer} {
+      @include margin($spacer, $spacer, $spacer, $spacer);
+    }
+
+    .padding-#{$spacer} {
+      @include padding($spacer, $spacer, $spacer, $spacer);
+    }
+
+    // Top Side
+    .margin-top-#{$spacer} {
```
IamManchanda: @rafibomb .... We were to rename the convention here if you remember ??
Shall i do it ??

rafibomb: I'm not sure - we need to ask more people - I say leave it, and during RC we can gather feedback

IamManchanda: Ohk no issue .... But if you ask me in any day i would like `mx0` instead of `margin-horizontal-0` for these!
Yes foundation convention is not `mx0` but then .... its prototype i guess (fast coding) ??

rafibomb: I use the shorter ones in my work too - let's put to the public in RC - then we can quickly adjust

docs/pages/prototyping-utilities.md
```diff
@@ -0,0 +1,797 @@
...
+	@include box(1rem); // Square
+}
+
+.baz {
+	@include box(1rem, $circle: true); // Circle
+}
+```
+
+### Rotate Mixin
+These Rotate mixins lets you rotate an element to a certain degree. Clockwise is the default direction but adding a `-` in front of the degrees will make it counter-clockwise.
```
IamManchanda: Here in East, we call it Anti clockwise!

PR #10110, opened 2017-06-02
fixed the usage of transitionend in motion util

swey:
The ES6 + Webpack usage of `foundation.util.motion.js` was broken because of a missing `import`. The previous version will only work if `Foundation` is exposed not if it's encapsulated.


kball:
Thanks @swey! And congrats on first merged PR

swey:
@kball Thank your work and the fast merging!

PR #10122, opened 2017-06-02
[Prototypes] Rename of spacing sizing classes

IamManchanda:
cc @tdhartwick  @kball for review!


IamManchanda:
I have updated test case for sizing too plus i can confirm that everything is working properly!

IamManchanda:
Closing this as it seems like we are nearing 6.4 and no such plans till now to change class name convention!

PR #10133, opened 2017-06-04
Add opacity mixin, No CSS Added!

IamManchanda:
I think this is a useful mixin as i don't think adding classes would be great on this infact bloat!

### DIFF
--- a/scss/prototype/_opacity.scss
+++ b/scss/prototype/_opacity.scss
@@ -1,0 +1,15 @@
+// Foundation for Sites by ZURB
+// foundation.zurb.com
+// Licensed under MIT Open Source
+
+////
+/// @group prototype-opacity
+////
+
+/// Opacity Mixin: Easily create opacity to an element
+/// @param {Number} $opacity[] Opacity/Transparency of an element
+@mixin opacity($opacity) {
+  opacity: $opacity;
+  $opacity-ie: $opacity * 100;
+  filter: alpha(opacity=$opacity-ie); //IE8
+}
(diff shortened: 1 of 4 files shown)


rafibomb:
Yes to mixins!

brettsmason:
I don't understand why we need this? Ie9 supports default opacity setting so why the need for the filter, as we don't support older browsers? It's quicker to type the style than to use the mixin :smile:

scss/prototype/_opacity.scss
```diff
@@ -0,0 +1,19 @@
...
+
+/// Opacity Mixin: Easily create opacity to an element
+/// @param {Number} $opacity[] Opacity/Transparency of an element
+@mixin opacity($opacity) {
+  @if ($opacity >= 0 and $opacity <= 1) {
+    opacity: $opacity;
+    $opacity-ie: $opacity * 100;
+    filter: alpha(opacity=$opacity-ie); //IE8
+  } @else {
+    @warn 'Opacity should be between 0.0 and 1.0';
```
IamManchanda: @brettsmason But this is yelling too , if user makes a mistake!

brettsmason: @IamManchanda you could say that about every single CSS property though (eg `float: top;`) 😄

kball:
Guys I don't get this... the only value this creates is for a browser we don't support... why would we do this?

IamManchanda:
Ohk please revert :disappointed: .... thinked about it ... yes you and @brettsmason are true!

PR #10135, opened 2017-06-05
[XY-Grid] propose swapping wrap logic to nowrap

rafibomb:
Just setting a basic `.grid-x` revealed an unexpected behavior. 

Using this markup
```
<div class="grid-x margin-gutters">
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
</div>
```
Results in this
5ydznka4ce

But what is expected is 
aowda3ez6s

Since it's a mixin option - this PR proposes we default to nowrap which means the grid follows prior convention. 

The opposite allows more than 12 cells per `.grid` but results can be inconsistent.

### DIFF
--- a/scss/xy-grid/_grid.scss
+++ b/scss/xy-grid/_grid.scss
@@ -19,10 +19,10 @@
 /// Creates a container for your flex cells.
 ///
 /// @param {Keyword} $direction [horizontal] - Either horizontal or vertical direction of cells within.
-/// @param {Boolean} $wrap [true] - If the cells within should wrap or not.
+/// @param {Boolean} $wrap [false] - If the cells within should wrap or not.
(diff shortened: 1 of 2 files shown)


brettsmason:
@rafibomb I haven't properly looked at this yet, but won't setting it to nowrap break all sorts of stuff, including block grids? The current live behaviour is actually to wrap.

rafibomb:
Hmmm, that's right. So there must be something else causing this bug. I'll close this and open an issue.

PR #10139, opened 2017-06-05
Set correct auto flex properties

### LINKED ISSUE #10138: [XY Grid] Unexpected wrapping issue.
Just setting a basic `.grid-x` revealed an unexpected behavior. 

Using this markup
```
<div class="grid-x margin-gutters">
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
  <div class="medium-auto cell">
    <img src="http://placehold.it/300" alt="" />
  </div>
</div>
```
Results in this
5ydznka4ce

But what is expected is 
aowda3ez6s

brettsmason:
Fixes #10138
See https://github.com/philipwalton/flexbugs#4-flex-shorthand-declarations-with-unitless-flex-basis-values-are-ignored - must include `px`.

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -53,7 +53,7 @@
     min-width: 0px;
   }
   @elseif ($size == 'auto') {
-    flex: 1 1 auto;
+    flex: 1 1 0px; // sass-lint:disable-line zero-unit
   }
   @elseif ($size == 'shrink') {
     flex: 0 0 auto;


kball:
So... this ends up to cause a number of problems. Having flex basis auto lets other sizing classes set width and height... setting 0px means flex ends up taking priority, breaking the ability to use sizing classes with anything that also has auto.

If we do this, we need to have all of our sizing classes explicitly reset the flex basis, which is going to be a LOT of css... we switched this to `auto` as a part of the IE10 support + css reduction process

What if instead we continued to have auto be flex `1 1 auto;` but also have `auto` set `width: 0px;` (Taking advantage of the fact that for flex-basis `auto` it will use width as flex basis)? I think this will unbreak our sizing classes without forcing us to have them all specify auto, and I don't *think* there are any negative side effects... I'll put together a PR to this effect

PR #10151, opened 2017-06-07
Fix auto in a way it doesn't break sizing classes

kball:
Fixes issues introduced by https://github.com/zurb/foundation-sites/pull/10139

CC @brettsmason @rafibomb

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -76,12 +76,25 @@
 // Sets our cell widths or heights depending on gutter type.
 @mixin -xy-cell-properties($size, $margin-gutter, $vertical) {
   $direction: if($vertical == true, height, width);
+
+  // y grid needs to explicitly reset flex basis here to overrule auto.
+  // See: https://github.com/zurb/foundation-sites/pull/10139 and https://github.com/zurb/foundation-sites/pull/10151
+  @if ($vertical) {
+    flex-basis: auto;
+  }
+
   @if($size == 'full') {
     $val: if($margin-gutter == 0, 100%, calc(100% - #{rem-calc($margin-gutter)}));
     #{$direction}: $val;
   }
   @elseif ($size == 'auto') {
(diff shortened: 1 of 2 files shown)


kball:
Testing this further, this breaks Safari for vertical auto grids (Safari has a wonderful flexbug where if height is 0, even though flexbox resizes, all children are set to height 0), so this solution won't work... sadly I'm starting to think we might need to explicitly set flex basis in sizing classes

kball:
Hmmm... digging into this more I realized - if auto is changing the flex basis, it means that to work properly responsively in interaction with auto, ALL of our responsive classes will have to change the flex basis.

Given what I was seeing above, I think we can limit this to the y-grid, which helps cut down our outputted CSS a bit, but... GRRR... would love to find a solution that lets us leave flex basis auto throughout.

kball:
Ok, took another pass on this in a way that uses the `width:0px` approach on horizontal and resets flex basis for vertical... this appears to work across Chrome, Safari (major problem child on these grids), IE10, and Firefox. It increases CSS size (pre-minification) by about 9k, all of it in the vertical grid (so if you don't need vertical you can drop the weight)... applying the "right" flex basis change across the board increased weight by 30k pre-minification.

@brettsmason definitely need your eyes & thoughts on this one.

brettsmason:
@kball Will take a look tonight - feel free to prod me then!

brettsmason:
I'm wondering if we can do this a different way with some clever CSS... Not sure what yet, but I was thinking something along these lines (sizes generated from the breakpoints):
```
.grid-y > .cell {
  flex-basis: 0px;
}

.grid-y > [class^='small'],
.grid-y > [class^='medium'],
.grid-y > [class^='large'] {
  flex-basis: auto;
}
```
Also I think there might be a bit of confusion here, in that the base cell class uses the `full` attributes, rather than `auto` unless I'm misreading everything. Lets discuss later!

Issue #10152, opened 2017-06-07
y-grid margin-gutters bad reset


kball:
As highlighted in https://github.com/zurb/foundation-sites/issues/10141#issuecomment-306924371 the `.y-grid` with margin-gutters is still getting negative margin on left and right...

I think this can be taken care of as a part of the margin/padding gutter changes we've been talking about @brettsmason - `.grid-margin-x` and `.grid-margin-y` should be responsible for those negative margins, regardless of the orientation of the grid itself..

brettsmason:
@kball So I think what I've gathered is we want to do:

`.grid-margin/padding-x > .cell` and `.grid-margin/padding-y > .cell` - these classes would purely apply margin/padding to the cells. This would also apply negative margins to the grid in that axis (for margin grids anyway) so that the cells line up flush.

Then `.grid-x`/`.grid-y` would just be responsible for setting the width or height of the `.cell`, as well at the `flex-direction` - is that correct? I'll take a pass at this tonight.

@oxyc Does the above sound correct for your use case?

oxyc:
Sounds like that would solve it yes. 👍 Great work on this new grid btw, it'll solve a lot of the earlier pains I lets me drop 3-4 grid helpers I don't need anymore.

brettsmason:
@oxyc I've put together a PR to sort this out. If you get a chance could you try it out and see if it works for you? https://github.com/zurb/foundation-sites/pull/10157
Don't worry if not 😄

brettsmason:
Closing as this is now included 😄

Issue #10141, opened 2017-06-05
XY Grid - Feedback and Discussion


brettsmason:
Now that the new XY Grid has been merged and is included in the RC lets keep all discussion surrounding this in one place, and gather as much feedback to improve it as much as possible before the full release.

You’ll find a Codepen to the new grid here: 
https://codepen.io/ZURBFoundation/pen/owvqYp?editors=1000
 
And the docs:
http://foundation.zurb.com/sites/6.4.0-rc1/docs/xy-grid.html

For previous discussion history see:
- https://github.com/zurb/foundation-sites/pull/10017
- https://github.com/zurb/foundation-sites/issues/9875

brettsmason:
Currently the plan is to rename `margin/padding-gutters` classes to `.grid-margin/padding` classes.
However feedback on how we can make this more flexible (eg allow padding to be added on all 4 sides, margin on bottom too etc) would be welcomed!

rafibomb:
Padding to create vertical spacing in the grid - my opinion is that padding as a "gutter" might now be obsolete so combining the `.grid-margin` with `.grid-padding` might be the way to achieve this - using the `$global-padding` variable

oxyc:
Is this a bug? https://codepen.io/anon/pen/owjYyR?editors=1100

```html
<div class="grid-y margin-gutters">
  <div class="cell">cell</div>
  <div class="cell">cell</div>
</div>
```

Seems something isn't getting overridden correctly as the grid overflows the container.

Also, would be great if this would add vertical gutters only between the cells (not above or below the grid itself). I often use such a pattern to separate teasers in a content listing where I dont want any trailing whitespace.

kball:
@oxyc can you specify browser? It looks fine in Chrome, but I'm seeing an issue in Safari that looks like this:

![xy_grid_gutters___foundation_6_4_docs](https://user-images.githubusercontent.com/44007/26901910-380261c6-4b8c-11e7-8a8f-c10351a779ea.png)

@brettsmason can you look into it?

oxyc:
In your screenshot you can see that the cells background color is not aligned with the heading, which I think it should be.

kball:
Ah, yes, I see it. The left-right negative margin from margin gutters is getting applied even though this is a vertical grid. Good catch!

kball:
@oxyc with regards to the vertical gutters not applying on the outside, I believe that actually is working but the h4 has a margin-bottom that is offsetting. Can you verify?

[...]

PR #10157, opened 2017-06-08
Xy grid gutter class fix

brettsmason:
This changes the old margin/padding gutter classes to the new `.grid-margin/padding-x/y` syntax.
This allows margin/padding to be applied to all 4 sides, regardless of the grid direction.

One remaining issue that I can think of with this. The base `.cell` class is linked to `.grid-margin-x/y` - specifically their width/height whish is outputted at 100% (minus the margin). Could this cause issues in certain configurations?

### DIFF
--- a/docs/assets/scss/examples/_grid.scss
+++ b/docs/assets/scss/examples/_grid.scss
@@ -78,11 +78,15 @@
 [id^="docs-xy"].docs-component .docs-code-live {
   @include foundation-xy-grid-classes;
 
-  .grid {
+  .grid-x,
+  .grid-y {
     font-size: 12px;
-    margin-bottom: 10px;
     line-height: 2rem;
     margin-bottom: 1.5rem;
+  }
+
+  .grid-y .grid-x {
+    margin-bottom: 0;
   }
 
   .cell {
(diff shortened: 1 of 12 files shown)


oxyc:
The bug described here https://github.com/zurb/foundation-sites/issues/10152 is now fixed. But the height change is causing issues when there's no size set for the cells. It spans the browsers entire height.

```html
<div class="grid-y grid-margin-y">
  <div class="cell"><div class="demo">cell</div></div>
  <div class="cell"><div class="demo">cell</div></div>
</div>
```
screen shot 2017-06-08 at 2 58 08 pm

---

Allowing the gutter to be added to all sides simultaneously seems to work. It would still be nice if `.grid-x.grid-y` was supported as described here https://github.com/zurb/foundation-sites/issues/10141#issuecomment-306936167. Fix would be to either add

```css
.grid-x.grid-y {
  flex-flow: row grid;
}
```

or move the `grid-y` generated css before `grid-x` so that `grid-x`'s flex-flow takes precedence.

I guess this is a different issue though :)

oxyc:
Oops, I forgot to declare the doctype in my test case, that's why the height went buggy. Sorry about that! The fix works.

kball:
This is freaking sweet @brettsmason!  Added a visual test playing around with the combos some... but it's very very nice. We'll need to do some more work to better document the power of this approach, but I'm PUMPED! :D

oxyc:
>  it's very very nice.

Seconded! thanks for the combo demonstration.

PR #10160, opened 2017-06-09
the cb is missed in this._setWrapperHeight() so the current cb logic …(foundation.orbit.js)

simongcc:
Sorry for previous failed attempts, finally I understand how to PR it right.
Since it was already a while ago, it is related to my previous opened PR #8755 
The callback logic is being added but the callback option and parameter is missed so I am here to add it back.

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -131,7 +131,8 @@ class Orbit extends Plugin {
   */
   _prepareForOrbit() {
     var _this = this;
-    this._setWrapperHeight();
+    var cb = _this.options.callback || function(){};
+    this._setWrapperHeight( cb );
   }
 
   /**


kball:
Looking at this, I think I understand what you're trying to accomplish, but would actually rather we fire an event so that folks using an indirect instantiation (e.g. using data attributes in the html) can still set up a listener.  Can you replace the callback option with a new event being fired at the same point in the lifecycle?

DanielRuf:
> Can you replace the callback option with a new event being fired at the same point in the lifecycle?

Definitely a good idea.

@kball how should we name the event?

DanielRuf (changes requested):
We should trigger an event to support unobstrusive and asynchronous, reactive implementations.

kball:
This PR is stale and the requested changes were not made. Just going to close it out.

simongcc:
@kball Sorry for having not been a while, what should I do with the stale PR?

kball:
@simongcc no worries; the entire project has all been stale for a while, but some of the community is making a push to revive it. 

Based on the feedback from the original PR, if you'd like to pick it back up please rework it to trigger an event instead of use a callback, and then reopen/open a new PR.

PR #10163, opened 2017-06-09
Make XY Grid cell block containers responsive as well

kball:
Fixes responsiveness of frame cell block containers


brettsmason:
Excellent thanks! 👍

PR #10168, opened 2017-06-11
Closes #10167 - Mobile First Gutters Issue!

### LINKED ISSUE #10167: [XY Grid] Cells improperly wrap when relying on cell size inheritance from smaller screens
**v6.4.0-rc3**

CodePen: https://codepen.io/simshaun/pen/rweyxg

Coming from the flex grid, I'm used to being able to declare the column width for the smallest effective screen size, and the larger screens would simply inherit from the smaller. e.g. `.row>.small-8.columns+.small-4.columns`. I'm trying to do the same thing with a grid-x, grid-margin-x grid. e.g.

    <div class="grid-x grid-margin-x">
      <div class="small-4 cell">X</div>
      <div class="small-4 cell">Y</div>
      <div class="small-4 cell">Z</div>
    </div>

However, the last cell wraps to the next line when viewed from any larger screen than small, so 640+ px.

This is happening because of the cell width calculation:

    .grid-margin-x > .small-4 {
        width: calc(33.33333% - 1.25rem);
    }

On small screens, `-1.25rem` is correct. The problem is that it stays `-1.25rem` on medium+ screens, where the gutter size increases to 30px. Because of the larger gutter, the cell width calculation needs to have `-1.875rem`, but it doesn't.

IamManchanda:
This PR fixes  #10167

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -80,13 +80,15 @@
   > .#{$prefix}shrink {
     @include xy-cell-static(shrink, false, $breakpoint: $gutter-breakpoint, $vertical: $vertical);
   }
-
-  @for $i from 1 through $grid-columns {
-    // Sizing (percentage)
-    $classname: if($vertical, '.#{$class-breakpoint}-#{$i}', '.#{$class-breakpoint}-#{$i}');
-
-    > #{$classname} {
-      @include xy-cell-static($i, false, $breakpoint: $gutter-breakpoint, $vertical: $vertical);
+  
+  @include -zf-each-breakpoint() {
+    @for $i from 1 through $grid-columns {
+      // Sizing (percentage)
+      $classname: if($vertical, '.#{$class-breakpoint}-#{$i}', '.#{$class-breakpoint}-#{$i}');
+      
+      > #{$classname} {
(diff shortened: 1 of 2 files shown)


IamManchanda:
cc @kball @brettsmason

brettsmason:
@IamManchanda I've managed to have a very quick look at the generated CSS. Unfortunately I don't think this is the correct fix at the moment. If you look at what's generated, its creating `.grid-x > .small-x` classes for the large breakpoint, but it only needs to do it for the breakpoints in the gutter map.

Its also messed up some of the media queries - eg `@media print, screen and (min-width: 40em) and (min-width: 40em) {`. @kball will probably know about the nested media query issue as I know he worked on that a lot.

I'll have a proper look tonight though. Also we need to solve block grids too, as the same issue will apply.

IamManchanda:
Ohk no issue ... will await @kball reply :smile:

kball:
Yeah this is not the correct fix. Going to close this. It would be super helpful to a "kitchen sink" visual test for the XY-grid to quickly scan through... @IamManchanda @brettsmason either of you have time to put that together?

brettsmason:
Yep no problem I can put that together. Do we literally want everything in it? That's will be pretty big :smile:

kball:
I'd say at least all of the common cases, with quick titles of what expected behavior should be so it's easy to scan through

PR #10172, opened 2017-06-12
Fix a typo that led to responsive gutters being missed in margin gutters

### LINKED ISSUE #10167: [XY Grid] Cells improperly wrap when relying on cell size inheritance from smaller screens

kball:
Fixes #10167

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -119,7 +119,7 @@
       // This is purely for responsive gutters - the margin grid has to go back and adjust widths (or heights)
       // for prior breakpoints based on the responsive gutter.
       @if(type-of($grid-margin-gutters) == 'map' and map-has-key($grid-margin-gutters, $-zf-size)) {
-        @each $bp in -xy-breakpoints-less-than($-zf-size) {
+        @each $bp in -zf-breakpoints-less-than($-zf-size) {
           @if(map-has-key($grid-margin-gutters, $bp)) {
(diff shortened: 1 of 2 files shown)


IamManchanda (approved):
LGTM ... Added a visual test for the same!

brettsmason:
Good spot 😄 I didn't think I imagined it already worked this way at some point...
Looks good.

IamManchanda:
$-zf-xy
Gonna be pretty confusing for v7 😞

PR #10173, opened 2017-06-12
XY Grid - Add kitchen sink/flex utilties visual tests

brettsmason:
This should hopefully provide a single visual test that shows all use cases in one go.
Let me know if you think there's any other examples I should add.


brettsmason:
I've also added a flex utilities visual test so we can confirm these cases work OK too.

brettsmason:
I'm going to merge this in myself as I need it for testing.

PR #10185, opened 2017-06-14
Add Foundation.MediaQuery.is to the docs

gnuletik:


kball:
Great, thanks @GNUletik!

Issue #8131, opened 2016-02-10
[Feature][Breadcrumb] Use $bread-crumb-slash settings


SalahAdDin:
I have using the last feature $crumb-slash and i can't find in the new Foundation version. 

This option is very useful if we want customizing a lot the breadcrumbs:
![seleccion_084](https://cloud.githubusercontent.com/assets/5034215/12957809/a3e9b18a-cffc-11e5-86ce-6ffea51fa0b2.png) 

We have to this:
![seleccion_106](https://cloud.githubusercontent.com/assets/5034215/13022173/9e19e178-d1ad-11e5-9a00-259b7b85a843.png)

I think that this feature will be useful for us, i want implements it but i don't understand the code:

``` scss
@if $breadcrumbs-item-slash {
      // Need to escape the backslash
      $slash: if($global-text-direction == 'ltr', '/', '\\');

      &:not(:last-child)::after {
        color: $medium-gray;
        content: $slash;
        margin: 0 $breadcrumbs-item-margin;
        position: relative;
        top: 1px;
        opacity: 1;
      }
    }
```

This is the actual code.

``` scss
&:before {
    color: $crumb-slash-color;
    content: "#{$crumb-slash}";
    margin: 0 $crumb-side-padding;
    position: relative;
    top: 1px;
  }
```

And this is the old code for it.

Can you hang back this feature for the new version?

SalahAdDin:
:+1:

PR #10192, opened 2017-06-15
add breadcrumb separator variable

bitbruder:
Adds variable for custom breadcrumb separators instead of a slash.

Reference #8131

### DIFF
--- a/scss/components/_breadcrumbs.scss
+++ b/scss/components/_breadcrumbs.scss
@@ -34,13 +34,21 @@ $breadcrumbs-item-margin: 0.75rem !default;
 /// @type Boolean
 $breadcrumbs-item-uppercase: true !default;
 
-/// If `true`, adds a slash between breadcrumb links.
+/// If `true`, adds a seperator between breadcrumb links.
 /// @type Boolean
-$breadcrumbs-item-slash: true !default;
+$breadcrumbs-item-separator: true !default;
 
-/// Color of breadcrumb slash.
+/// Used character for the breadcrumb separator.
+/// @type Content
+$breadcrumbs-item-separator-item: '/' !default;
+
+/// Used character for the breadcrumb separator in rtl mode.
+/// @type Content
+$breadcrumbs-item-separator-item-rtl: '\\' !default;
+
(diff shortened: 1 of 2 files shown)


IamManchanda (approved):
LGTM...

IamManchanda:
Checked it out locally... This is looking good...
@kball for merge/review!

IamManchanda:
@digitalfrood... By the way congrats on your first time contribution ;)

scss/components/_breadcrumbs.scss
```diff
@@ -34,13 +34,21 @@ $breadcrumbs-item-margin: 0.75rem !default;
 /// @type Boolean
 $breadcrumbs-item-uppercase: true !default;
 
-/// If `true`, adds a slash between breadcrumb links.
+/// If `true`, adds a seperator between breadcrumb links.
 /// @type Boolean
-$breadcrumbs-item-slash: true !default;
+$breadcrumbs-item-separator: true !default;
```
kball: Can we do a check for if `$breadcrums-item-slash` is defined and use that if so? That would enable backwards compat with older settings files

kball:
Let 1 item of feedback, other than that LGTM, can we get that addressed and then will merge?

scss/components/_breadcrumbs.scss
```diff
@@ -34,13 +34,21 @@ $breadcrumbs-item-margin: 0.75rem !default;
...
+$breadcrumbs-item-separator-item: '/' !default;
+
+/// Used character for the breadcrumb separator in rtl mode.
+/// @type Content
+$breadcrumbs-item-separator-item-rtl: '\\' !default;
+
+/// Color of breadcrumb item.
 /// @type Color
-$breadcrumbs-item-slash-color: $medium-gray !default;
+$breadcrumbs-item-separator-color: $medium-gray !default;
```
IamManchanda: Same for this also @kball 

Again warn @digitalfrood i am taking this up!

IamManchanda:
done @kball ... Anything else needed or good to go??

PR #10195, opened 2017-06-16
paramenter description fixed

derweili:
the element description references the dropdown element instead of the  responsive accordion tabs


kball:
Merged, thanks!

PR #10200, opened 2017-06-16
XY grid fixes

### LINKED ISSUE #10174: XY Grid - vertical grid with padding wraps
Noticed a bug with the vertical grid using padding gutters. See the kitchen sink visual test to see what happens. This doesn't happen with margin, so is likely a specificity issue. Setting `flex-wrap` to `nowrap` solves the issue, but we need to decide if this is the right behaviour or not.

### LINKED ISSUE #10180: XY Grid - align-stretch has no effect with grid-margin-y

1. Create a horizontal grid with vertical margins: `grid-x grid-margin-y`.
2. Add `align-stretch`.

#### What should happen:

The cells height should stretch.

#### What happened instead:

The height is enforced to by `.grid-margin-y > .cell` and therefore do not stretch.

Chrome

`develop`-branch

#### Test case link:

https://codepen.io/anon/pen/xrOMNP?editors=1100

Related: https://github.com/zurb/foundation-sites/pull/10157

> @brettmason One remaining issue that I can think of with this. The base .cell class is linked to .grid-margin-x/y - specifically their width/height whish is outputted at 100% (minus the margin). Could this cause issues in certain configurations?

screen shot 2017-06-13 at 11 53 51

brettsmason:
This is an attempt to fix multiple XY grid issues I've discovered whilst constructing visual tests as well as other outstanding issues.

This fixes:
- `[size]-auto/shrink` classes overriding sizing classes with a bigger breakpoint widths
- Vertical padding grid wrapping issue (closes #10174 
- `align-stretch` not applying to `grid-margin-y` as that class was overwriting the height (closes #10180)

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -40,7 +40,6 @@
     > .shrink {
       @include xy-cell-static(shrink, false);
     }
-
   }
 
   // Auto width
@@ -228,7 +227,7 @@
   }
 
   .grid-y {
-    @include xy-grid(vertical);
+    @include xy-grid(vertical, false);
 
 
     > .cell {
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -58,13 +58,11 @@
         @include xy-cell-base(shrink);
         @include xy-cell-static(shrink, false);
       }
-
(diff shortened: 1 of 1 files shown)


kball:
Excellent work

Issue #10182, opened 2017-06-13
aria-expanded not accepatble with role="menuitem" with F6 drilldown menus


getvijayji:
I am using Foundation 6 Drilldown menus for off-canvas. It seems that foundation6 is injecting the ARIA attributes dynamically, where it is putting aria-expanded and role="menuitem" together in "li" element which have a "ul" element as its sub element. But role="menuitem" and aria-expanded is not allowed togather by the ARIA spec. Is there any fix for this?

I have seen that there is fix done for foundation6 data-accordion-menu plugin which had the same issue. the fix is done in upcoming 6.4 release, see here: #10064.

Can we put some fix for the above issue too?

Thanks.

getvijayji:
any update on this, please?

rafibomb:
Sounds good to me! Are you able to help with a PR on this @getvijayji ?

@Owlbertz What do you think?

Owlbertz:
I guess that for Drilldown the same rules apply as for the AccordionMenu, i.e. that it should have a role of tree. If so, than the changes from #10064 would need to be applied for Drilldown, which would solve this issue here.

getvijayji:
yes the same fix will work. but as far as submitting a PR is concerned, i guess i will not be able to submit a PR for this.
@Owlbertz , Can you help is by submitting this as a PR against the develop branch?

Owlbertz:
Yep

Owlbertz:
Should be closed with #10204.

getvijayji:
Thanks @Owlbertz

Daltyn06:
This is now getting flagged by lighthouse audits. it seems to expect menubar now instead of tree. if you run the light house audit on the demo page in the docs on mobile this fails.

PR #10064, opened 2017-05-22
Accordion menu tree

Owlbertz:
Addresses #10061.

PR #10204, opened 2017-06-19
Make Drilldown use tree role.

Owlbertz:
Drilldown should - like AccordionMenu - have the role tree as it consists of a hierarchical list of elements.
Sub-menus should have role group and items should have role treeitem.

Pretty much the same as #10064.

Addresses #10182.


IamManchanda (approved):
LGTM if i am not missing something :)

kball:
Excellent work, as always @Owlbertz

PR #10205, opened 2017-06-19
Closes #10203, update docs with current drilldown menu examples!

### LINKED ISSUE #10203: [Drilldown Menu] not working at all in 6.4
http://foundation.zurb.com/sites/6.4.0-rc/docs/drilldown-menu.html drilldown is not working at all in 6.4.

IamManchanda:
Closes #10203 
Update docs as per #9670 !

PS: The newly created Visual tests are running fine!

cc @kball

### DIFF
--- a/docs/pages/drilldown-menu.md
+++ b/docs/pages/drilldown-menu.md
@@ -23,15 +23,19 @@ Any `<a>` without a submenu will function like a normal link.
 </div>
 
 ```html
-<ul class="vertical menu" data-drilldown>
+<ul class="vertical menu drilldown" data-drilldown>
+  <li><a href="#">One</a></li>
   <li>
-    <a href="#Item-1">Item 1</a>
-    <ul class="vertical menu">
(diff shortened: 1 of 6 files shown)


kball:
@IamManchanda can you take a look at the drilldown visual tests too?

And I'm pondering... since drilldown requires the class to work, should we have a (configurable, but default on) JS option that will auto-add drilldown class if it doesn't exist?

IamManchanda:
> @IamManchanda can you take a look at the drilldown visual tests too?
Doing it now 

> And I'm pondering... since drilldown requires the class to work, should we have a (configurable, but default on) JS option that will auto-add drilldown class if it doesn't exist?

Yes it will be great sir!

IamManchanda:
There you go @kball ... i have updated the visual tests :)

brettsmason:
@IamManchanda Does the drilldown menu now require the `nested` class to work?

IamManchanda:
Nopes not needed i just added this for the sake of responsive menu! @brettsmason

PR #10207, opened 2017-06-19
Rework media query init

kball:
Fixes initialization of media query plugin to work properly in new module environment.

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -75,6 +75,11 @@ var MediaQuery = {
    */
   _init() {
     var self = this;
+    var $meta = $('meta.foundation-mq');
+    if(!$meta.length){
+      $('<meta class="foundation-mq">').appendTo(document.head);
+    }
+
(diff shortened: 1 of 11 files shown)


IamManchanda (approved):
LGTM

js/foundation.util.mediaQuery.js
```diff
@@ -173,7 +178,7 @@ var MediaQuery = {
    * @private
    */
   _watcher() {
-    $(window).on('resize.zf.mediaquery', () => {
+    $(window).off('resize.zf.mediaquery').on('resize.zf.mediaquery', () => {
```
IamManchanda: BTW ... it will be great if you can explain this one above code to a js beginner like me :wink:

kball: sure... `.zf.mediaquery` is putting a namespace on this so that we can target it specifically, and we run off and on to make this idempotent (ie we only ever have one listener, not a new one for every time this fn is called)

PR #10222, opened 2017-06-20
Add testcase and add qualifiers on basis overrides

kball:
This is a little bit of a hackfix, but we're running into a situation where the wildcard-style flex-basis resets we introduced in https://github.com/zurb/foundation-sites/pull/10164 collide with other responsive flex or grid classes like `[size]-order-*` and `[size]-offset-*`.

The "pure" fix would be to go back to having every size class reset its flex basis, but that generates truly absurd amounts of HTML. Other fixes explored prior to #10164 ran into various browser flexbugs that broke different browsers.  Thus, I propose this workaround which explicitly excludes the 2 sets of responsive classes we generate that make sense to apply to cells and are not grid sizing classes.

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -44,6 +44,9 @@
 
   // Auto width
   @include -zf-each-breakpoint() {
+    // This is a bit of a hack/workaround, see these PRs for the backstory:
+    // https://github.com/zurb/foundation-sites/pull/10222 and
+    // https://github.com/zurb/foundation-sites/pull/10164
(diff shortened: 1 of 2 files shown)


IamManchanda (approved):
For me this is a good hack!!!

brettsmason:
@kball I wouldn't call this a hackfix, I'd call this a super clever solution to code bloat 😉 
The great thing about `:not` is it doesn't add to specificity (I think) so this will work great.

Works great from my end... merging.

oxyc:
This also applies to `auto`. https://codepen.io/anon/pen/pwrXmx?editors=1000

Issue #10117, opened 2017-06-02
Vision - Flexbox Utilities (XY Grid)


IamManchanda:
Since the rework of grid .... i was very certain that the docs need to improve with grid and its integration with flexbox helper classes. But at the same time, working with prototypes I have realized that these flexbox helper classes are not good css practices and really they are just only one step away from inline styles. 
At the same time, flexbox is not supported in every browser and for css only projects, people need these classes inbuilt as they give those vendor prefixes.
Moreover, flexbox is still new to the web and i think we need to support these helpers atleast for now 

So, Here is my Vision
1. Rename the docs page title to **Flexbox Utilities** (Url can remain the same as its old/SEO/Bookmarks) 
2. From Scss, Remove `flex.scss` and create a separate folder `flexbox` and move all those classes there.
3. Revamp them to the same values that we followed at [Prototype PR](https://github.com/zurb/foundation-sites/pull/10091)... Eg. Responsive breakpoints.
4. Keep them **Enable by default** but allow an option to disable 
5. Add some more helper classes as discussed in #10109 and #9814
6. **Most Importantly**, update that docs page/codepen examples to (on par) with something like http://frowcss.com/frow-shortcuts.html. Here is the example codepen with an example of these helper classes with the grid => https://codepen.io/IamManchanda/pen/WONvGz?editors=1000

Thoughts @kball @brettsmason @rafibomb @andycochran @zurb/yetinauts ??

brettsmason:
@IamManchanda You raise some good points here. I'll properly go through this in more detail tomorrow when I'm a bit fresher...

1. Agreed
2. Why not just rename the partial to `flexbox-utilities.scss`?
3. Don't understand, but I havent caught up on your excellent work on the prototype stuff yet 😄 
4. As long as it doesn't add bloat then I think there may be use in doing that
5. There's a bit of grid crossover with `.auto/`shrink` classes, though order classes might be helpful
6. I think the difficulty is that's not really how our docs are setup to work, but I agree they work really well and are very clear. I personally prefer looking at the examples first and if I'm interested in it I'll look at the code, so the code is just taking up space to me unless I want to look at that particular example.

IamManchanda:
@brettsmason 

> Why not just rename the partial to flexbox-utilities.scss?

[...]

PR #10224, opened 2017-06-21
Flexbox utilities docs & Some Sass Changes!

IamManchanda:
This is a ~~WIP~~ PR for flexbox utilities and relating them to xy grid!

### DIFF
--- a/docs/assets/scss/examples/_grid.scss
+++ b/docs/assets/scss/examples/_grid.scss
@@ -74,7 +74,8 @@
     }
   }
 }
-// Flex Grid
+
+// XY Grid
 [id^="docs-xy"].docs-component .docs-code-live {
   @include foundation-xy-grid-classes;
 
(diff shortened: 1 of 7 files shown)


IamManchanda:
@brettsmason @kball 
Just for this page i have add colors to make it interactive ... is it good idea or a bad one ??

make-flex-util-more-interactive

IamManchanda:
@kball @brettsmason 
While moving flex helper classes i have a thought... we are having responsive classes
Why not make the responsive classes customizable by letting user disable it 
(by this we can disable it in later release like v7 also)

```scss
$flexbox-responsive-breakpoints: true !default;

.flex-child-auto {
  flex: 1 1 auto;
}

@if ($flexbox-responsive-breakpoints) {
  // Loop through Responsive Breakpoints
    @include breakpoint($size) {
      @if $size != $-zf-zero-breakpoint {
        .#{$size}-flex-child-auto {
          flex: 1 1 auto;
        }
      }
    }
  }
}
```

kball:
Honestly think that's probably too much, especially since primary context for this type of thing will be in the grid and it's already handled there.

kball:
oop I misinterpreted - we have this already, and the question is to make it optional. I thought it was to add new... yes, by all means make it optional. :)

IamManchanda:
Ohk @kball i am making it optional and also adding a **depreciation notice** in docs that responsive breakpoints will be disabled by default for these classes from 6.5 (though people can reenable it by that same var) ??

Is that sounds good ??

IamManchanda:
Just for records, this closes https://github.com/zurb/foundation-sites/issues/10117 on merge!

PR #10241, opened 2017-06-23
Changes the Abide pattern to support Mastercard 2-digits BIN

### LINKED ISSUE #10240: New Mastercard cards do not pass validation of the Abide plugin
Hey, guys.
Mastercard had released new cards starting with 2. Current regex don't let them pass ATM.

Place to fix -> https://github.com/zurb/foundation-sites/blob/develop/js/foundation.abide.js#L538

Thanks.

slipets-a:
This PR closes #10240 by adding a regex to the Abide default pattern "card".


kball:
Tested this out against a range of numbers, looks good! Would be great to add some unit tests for number testing, but that doesn't need to block this.

PR #10245, opened 2017-06-24
Add missing viewport tag on all test cases!

IamManchanda:
This is as per the discussion that me and @kball had that we misses the viewport tags 

Frankly this kinda sucked that tests dont include viewport tag. Its simply unforturnate that we miss this tag on tests!

cc @kball for merge!


IamManchanda:
I am merging this myself as its imp for mobile testing!

PR #10247, opened 2017-06-24
Add backwards compatibility for menu-state-active & menu-centered

IamManchanda:
Keep `active` class but with option to disable the same!

cc @kball for merge/review!

### DIFF
--- a/scss/components/_menu.scss
+++ b/scss/components/_menu.scss
@@ -462,7 +462,6 @@ $menu-centered-back-compat: true !default;
     .menu-centered {
       > .menu {
         @if $global-flexbox {
-          display: flex;
           justify-content: center;
         }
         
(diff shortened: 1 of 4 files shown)


scss/components/_menu.scss
```diff
@@ -437,6 +458,19 @@ $menu-item-background-hover: $light-gray !default;
     }
   }
 
+  @if $menu-centered-back-compat {
+    .menu-centered {
+      > .menu {
+        @if $global-flexbox {
+          display: flex;
```
kball: Why do we need to add display flex? In global flexbox case won't we already be display flex?

IamManchanda:
For test case 

http://localhost:3000/menu/active-menu-flex.html
http://localhost:3000/menu/active-menu-float.html

http://localhost:3000/menu/menu-centered-flex.html
http://localhost:3000/menu/menu-centered-float.html

IamManchanda:
You are right @kball ... removed! 
KIndly check now ... above are the links for test case!

kball:
👍

PR #10251, opened 2017-06-26
Fixes #10244 - Add breakpoint auto to not selector and `flex` tweak

### LINKED ISSUE #10244: [XY-Grid] responsive auto classes do not evenly distribute the space.

1. Create a `grid-x`
2. Add 2 cells, both with `large-auto` classes.
3. Add varying content to the cells.

#### What should happen:

They should split 50/50 just like `auto` does.

#### What happened instead:

The cell with longer content takes up more space than the other.

Chrome 59

v6.4.0-rc5

#### Test case link:

https://codepen.io/anon/pen/pwrXmx?editors=1000

Related to https://github.com/zurb/foundation-sites/pull/10222

IamManchanda:
This fixes #10244!

@kball can you check it ??
I have added test case too!

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -47,8 +47,9 @@
     // This is a bit of a hack/workaround, see these PRs for the backstory:
     // https://github.com/zurb/foundation-sites/pull/10222 and
     // https://github.com/zurb/foundation-sites/pull/10164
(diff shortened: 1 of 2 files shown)


kball:
This fix doesn't work... checking on kitchen sink the later autos still fail.

I believe the underlying problem is our current `:not` based implementation assumes `:not` does not add specificity, but it appears to, resulting in this rule overriding things later that it should not.

IamManchanda:
@kball so what are the options we can look around ??

kball:
brute force: https://github.com/zurb/foundation-sites/pull/10257

PR #10257, opened 2017-06-26
Brute force the auto override issue

### LINKED ISSUE #10244: [XY-Grid] responsive auto classes do not evenly distribute the space.

kball:
Fixes #10244

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -44,12 +44,22 @@
 
   // Auto width
   @include -zf-each-breakpoint() {
-    // This is a bit of a hack/workaround, see these PRs for the backstory:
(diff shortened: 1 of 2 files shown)


IamManchanda:
Superb @kball ... This seems good to go!

PR #10261, opened 2017-06-26
Add  class to ease padding containers below max-width

kball:
This implements the `grid-container-padding` class discussed in https://github.com/zurb/foundation-sites/issues/10141#issuecomment-310694090


IamManchanda (approved):
:+1:

PR #10262, opened 2017-06-27
Remove background color from styled <legend>

### LINKED ISSUE #10254: [Forms] Unnecessary background color on <legend>?
A styled `fieldset > legend` element currently [has the background color set to `$body-background`](https://github.com/zurb/foundation-sites/blob/develop/scss/forms/_fieldset.scss#L35). This is confusing as I don't expect a legend to pick up the page background colour. The form may be in an area with a different background colour, for example.

1. Add a `<div>`
2. Inside the `div` add `<fieldset class="fieldset"><legend>Test</legend></fieldset>`
3. Change the background colour of the div.

#### What should happen:

Background colour of legend should be "none".

#### What happened instead:

Background colour is the same as body background colour.

Chrome 59, Safari 10.1.1

6.3.1

#### Test case link:

https://codepen.io/Odaeus/pen/QgqzGB

Odaeus:
Fixes #10254.

### DIFF
--- a/scss/forms/_fieldset.scss
+++ b/scss/forms/_fieldset.scss
@@ -32,7 +32,6 @@ $legend-padding: rem-calc(0 3) !default;
     margin: 0;
     margin-#{$global-left}: rem-calc(-3);
     padding: $legend-padding;
-    background: $body-background;
   }
 }
 


rafibomb:
Thanks!

PR #10264, opened 2017-06-27
Fix Interchange when used on images path containing parenthesis

yannick-milanetto:
Hi,

we faced an issue with Interchange plugin when the Interchange data contain some images with parenthesis in their URL.

The issue happens only for background style replacement.
It comes from the fact that Interchange concatenates the image path with 'url( )' which does already contain some parenthesis:
https://github.com/zurb/foundation-sites/blob/develop/js/foundation.interchange.js#L160

To fix that, i suggest escaping the parenthesis in the image path.

**Example of use case**
We do use Thumbor to manipulate our images.
As shown below, Thumbor's way of applying on-the-fly filter to images is by adding some keywords containing parenthesis on the picture url called:
http://thumbor.readthedocs.io/en/latest/strip_icc.html

**Note for the reviewers**
I've updated the test case to reflect this fix biut i'm not sure whether you prefer having a dedicated test for this case or simply, as I did, having the data used in all Interchange tests using URL containing parenthesis (this ensures that all scenarii work fine with parenthesis). Please let me know if you do prefer a dedicated test.

Cheers

### DIFF
--- a/js/foundation.interchange.js
+++ b/js/foundation.interchange.js
@@ -157,6 +157,7 @@ class Interchange extends Plugin {
     }
     // Replacing background images
     else if (path.match(/\.(gif|jpg|jpeg|png|svg|tiff)([?#].*)?/i)) {
+      path = path.replace(/\(/g, '%28').replace(/\)/g, '%29');
       this.$element.css({ 'background-image': 'url('+path+')' })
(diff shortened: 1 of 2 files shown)


kball:
Looks good to me.

PR #10291, opened 2017-06-29
bump up the version

gormus:
Not sure about the versions for listed other packages though.


.versions
```diff
@@ -11,4 +11,4 @@ modules-runtime@0.6.3
 promise@0.6.7
 random@1.0.9
 underscore@1.0.8
-zurb:foundation-sites@6.3.0
+zurb:foundation-sites@6.4.0
```
IamManchanda: 6.4.1 :wink:

PR #10294, opened 2017-06-29
TASK: extend docs with hint for a single cell in the xy grid

marvinhuebner:
This PR extends the docs with a hint for a single cell in the xy grid. 

Based on the question here: https://github.com/zurb/foundation-sites/issues/10141#issuecomment-311686892

This could be helpful for others , if they search for something similar.

### DIFF
--- a/docs/pages/xy-grid.md
+++ b/docs/pages/xy-grid.md
@@ -117,6 +117,19 @@ below that. If you want to add padding below the `$max-width`, simply add the
   </div>
 </div>
 ```
+
+If you have just a single cell for a padding grid, you can save some markup by combining the `.grid-container`, `.grid-x` and `.grid-padding-x` classes together on the same element. You can still nest more grids inside this container like usual.
+
+<div class="warning callout">
+  <p>For a margin grid it needs to stay separated.</p>
+</div>
+
+```html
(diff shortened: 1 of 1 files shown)


docs/pages/xy-grid.md
```diff
@@ -117,6 +117,19 @@ below that. If you want to add padding below the `$max-width`, simply add the
   </div>
 </div>
 ```
+
+If you have just a single cell for a padding grid, you can save some markup by combining the `.grid-container`, `.grid-x` and `.grid-padding-x` classes together on the same element. You can still nest more grids inside this container like usual.
```
kball: This should actually work for any padding grid, regardless of cells inside... the need to have the container be a separate div for the outermost container is a margin grid requirement.  This essentially comes from the fact that the container uses `margin:auto;` to center it on the page, but the margin grid itself influences margins

marvinhuebner:
@kball maybe i was a little bit rushy with this PR. In the command here (https://github.com/zurb/foundation-sites/issues/10141#issuecomment-312205066) i can read the view that cells and grids are seperated elments and so they should not be merged together. 

This makes totally sense for me. So should i close the PR?

kball:
Yeah lets close this

PR #10298, opened 2017-06-29
Custom Pattern and Validator code fix

wulfsolter:
In `Adding Custom Pattern and Validator` the JS example is broken.

Foundation is initialized before the defaults are set.

Changed code example to set params first, then init Foundation.


kball:
Thanks @wulfsolter!

PR #10341, opened 2017-07-04
Remove `flex-grid-column()` duplicate `max-width`

LeoColomb:
Very small fix inside `flex-grid-column()` mixin: `max-width` fix is added twice: call to `flex-grid-size()` already import this fix once.

Also, the `min-width` fix should be inside `flex-grid-size()` for really fixing things for any size change.

### DIFF
--- a/scss/grid/_flex-grid.scss
+++ b/scss/grid/_flex-grid.scss
@@ -88,10 +88,6 @@
   @if $columns == expand {
     min-width: 0;
   }
-  // max-width fixes IE 10/11 not respecting the flex-basis property
-  @if $columns != expand and $columns != shrink {
-    max-width: grid-column($columns);
-  }
 }
 
 /// Creates a block grid for a flex grid row.


scss/grid/_flex-grid.scss
```diff
@@ -119,6 +109,11 @@
 
   flex: flex-grid-column($columns);
 
+  // fixes recent Chrome version not limiting child width
+  // https://stackoverflow.com/questions/34934586/white-space-nowrap-and-flexbox-did-not-work-in-chrome
+  @if $columns == expand {
+    min-width: 0;
+  }
```
ncoden: Even if I understand the logic behind this move (all fixes in the same place), this Chrome fix should stay in the component mixin `flex-grid-column`, because it does not need to be re-generated when we apply a different size.

For example:
```sass
.my-flex-col {
    @include flex-grid-column(12);
    // ✓ The IE fix is generated for 12 columns.
    // ✓ The Chrome fix is generated.
}

.my-flex-col--demi {
    @include flex-grid-size(6);
    // ✓ The IE fix must be generated for 6 columns.
    // ☓ The Chrome fix still work, so is useless here.
}
```

Please move back the Chrome fix to `flex-grid-column`.

ncoden (changes requested):
Hi @LeoColomb. Thank you for your ✨ **first contribution** ✨ on Foundation !

I have some remarks to make in your PR. What do you think ?
If you agree on these change, will you be able to provide them more or less quickly ?

scss/grid/_flex-grid.scss
LeoColomb: I see your point, fix moved back! 😉

ncoden (approved):

ncoden:
@kball Ready for merge I think.

IamManchanda (approved):
LGTM :+1:

IamManchanda:
Hey @LeoColomb ... Congrats on your first merged PR :wink: 
Thanks for digging into it @ncoden :wink:

PR #10345, opened 2017-07-05
Fixed #10344

### LINKED ISSUE #10344: Dropdown Arrow Position Issues when using Responsive Navigation
When you use a class such as `vertical medium-horizontal dropdown menu`, the dropdown arrow is mis-positioned. It appears the styles from `vertical` override the default horizontal styles.

#### What should happen:
The dropdown arrow should remain to the right of the navigation items on medium+.

#### What happened instead:
The dropdown arrows are showing to the left of the navigation item. 

6.4.1

#### Test case link:
https://codepen.io/anon/pen/qjKrqp

hikouki:
Fixed because left was not reset ✈️


kball:
Perfect, thanks @hikouki!

PR #10347, opened 2017-07-05
Use tabindex of -1 to make an element focusable

Vinnl:
If we add `tabindex="0"`, an element does become focusable, but it is also inserted in the tab order - even if there's no way to interact with it.

By setting the tabindex to -1, the element becomes focusable, but it does not receive focus when tabbing through the page.

See https://stackoverflow.com/questions/44909339/why-does-foundation-recommend-adding-a-tabindex-0-to-main-content

### DIFF
--- a/docs/pages/visibility.md
+++ b/docs/pages/visibility.md
@@ -120,7 +120,7 @@ To hide text from assistive technology, while still keeping it visible, add the 
 
 If your site has a lot of navigation, a screen reader will have to read through the entire navigation to get to your site's content. To remedy this, you can add a *skip link* at the very top of your page, which will send the user farther down the page, past the navigation when clicked on.
 
-Use the class `.show-on-focus` to hide an element, except when it has focus. Adding tabindex="0" to the target element makes if focusable.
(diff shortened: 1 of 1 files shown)


Vinnl:
The build is failing, but I don't think that's due to my change:

```
3.02s$ npm install

npm ERR! Linux 3.13.0-115-generic

npm ERR! argv "/home/travis/.nvm/versions/node/v6.11.0/bin/node" "/home/travis/.nvm/versions/node/v6.11.0/bin/npm" "install"

npm ERR! node v6.11.0

npm ERR! npm  v3.10.10

npm ERR! code ETARGET

npm ERR! notarget No compatible version found: gulp-postcss@^6.4.1
```

kball:
Yeah build is failing due to a dependency issue (a package got unpublished) that is fixed in develop, will get that into master shortly... 

I think it would be worth elaborating a little bit on this section, as the two choices (0 vs -1) have different behavior and could conceivably be used in different scenarios - I like the explanation in the stackoverflow post. Can you update this section to explain what the different behavior would be for 0 and -1 and why you'd choose one or the other?

Vinnl:
@kball I've now added the suggestion that you might want to add a `tabindex` of 0. That said, as far as I'm aware in the large majority of use cases you wouldn't want to add a tabindex of 0: you're usually referring to the location of content (i.e. not something you'd directly want to interact with), and even if we're referring to an element that one can interact with, most of the time it will already be focusable (e.g. in-build interactive elements such as `<a>`). Thus, I'd personally leave it out, but it's up to you :)

kball:
👍  Thanks @Vinnl!

PR #10350, opened 2017-07-05
Move typescale utility from prototype to typography

### LINKED ISSUE #10301: Base-level rules cannot contain the parent-selector-referencing character '&'.
Just upgraded my rails project to use Foundation 6.4.
It was a new project I hadn't modified any code so replacing the settings.scss was easy and I was previously only selecting what components  I wanted to load in.

I thought it would be seamless but I'm getting this :
`Base-level rules cannot contain the parent-selector-referencing character '&'.`

It's showing me this error for `@include foundation-button-group;`

I have foundation-sites installed via yarn, but I'm compiling the scss through rails asset pipeline.

Any help please, I'm dead in the water here.

IamManchanda:
Ad discussed,
Still keeping it disabled by default
This Fixes #10301

@kball please check it once ... if this works will update the docs!


IamManchanda:
Closing this based on #10350

PR #10351, opened 2017-07-05
Hotfix: Move typescale from prototype to typography and enable default!

### LINKED ISSUE #10301: Base-level rules cannot contain the parent-selector-referencing character '&'.

IamManchanda:
Another try to fix #10301

Reason: 

- We found out that extend is failing in ruby sass ... it sucks btw on **my part** that i didnt knew this will fail in ruby sass 
- This is not adding any extra css ... Basically its just `h1, .h1` instead of `h1` so no such bloat that could harm the code if enable by default!


IamManchanda:
Another thing to add in for records 

Considering Accessibility, 
This is a very Good helper class!

januszm:
Could you please release this also as a Ruby gem? It's not possible to use `@include foundation-prototype-classes;` with Ruby on Rails + `foundation-rails` gem because of this issue.

ncoden:
Related to https://github.com/zurb/foundation-rails/pull/254. I'll take a look at it.

PR #10355, opened 2017-07-06
#10284 Fixed bug of vertical two handlers slider.

### LINKED ISSUE #10284: Vertical slider with two handles doesn't behave as expected

1. Create a "two handles" slider like discribed in [foundation 6 documentation](http://foundation.zurb.com/sites/docs/slider.html#two-handles)
2. Add `vertical` class and `data-vertical="true"` as describe in [foundation 6 vertical slider documentation](http://foundation.zurb.com/sites/docs/slider.html#vertical)

#### What should happen:

"Two handles" should behave the same way vertically and horizontally.

#### What happened instead:

Cannot drag first slider to lower value, also cannot drag second slider to higher value.

Tested on chrome 57.0.2987

Foundation 6.3.1

#### Test case link:

- [CodePen issue demo](https://codepen.io/WnP/pen/dRJNpq)

hikouki:
Fixed #10284 issue.

Before calculating the boundary values of the two handlers,
it was necessary to calculate the vertical position.

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -200,6 +200,12 @@ class Slider extends Plugin {
 
     var isDbl = this.options.doubleSided;
 
+    //this is for single-handled vertical sliders, it adjusts the value to account for the slider being "upside-down"
+    //for click and drag events, it's weird due to the scale(-1, 1) css property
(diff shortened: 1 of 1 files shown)


kball:
Awesome, thanks @hikouki!

PR #10360, opened 2017-07-07
Fixed #9766

### LINKED ISSUE #9766: Positioning of dropdown is not correct when two buttons open the same dropdown

1. Create two buttons that both point to the same dropdown code using example markup on http://foundation.zurb.com/sites/docs/dropdown.html
2. Click your first button
3. Click your second button

#### What should happen:
The dropdown pane should appear in the right place depending on which button was clicked. This is the expected behavior based on previous versions of Foundation.

#### What happened instead:
The dropdown only appears next to the first button, regardless of which button has been pressed

#### Test case:
https://codepen.io/anon/pen/ygQoVj

hikouki:
Hi! ✈️ 

Changed Dropdown panes so that it is arranged with the position of primary anchor as a reference.

It may not be a beautiful code... 😨 
What would you do if you were me?

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -49,14 +49,13 @@ class Dropdown extends Positionable {
   _init() {
     var $id = this.$element.attr('id');
 
-    this.$anchor = $(`[data-toggle="${$id}"]`).length ? $(`[data-toggle="${$id}"]`) : $(`[data-open="${$id}"]`);
-    this.$anchor.attr({
+    this.$anchors = $(`[data-toggle="${$id}"]`).length ? $(`[data-toggle="${$id}"]`) : $(`[data-open="${$id}"]`);
+    this.$anchors.attr({
       'aria-controls': $id,
       'data-is-focus': false,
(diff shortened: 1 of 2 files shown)


kball:
Hi @hikouki,
Thanks for the PR!  I think this is broadly the right approach, but what I'd love to see is if we can solve this without storing the data on the DOM object itself, but rather store a reference to the current/primary in the javascript itself... we're moving towards more of a state-driven approach to the javascript and planning to abstract the DOM interactions behind and interface, so the more we can keep this as pure JS state the better.

hikouki:
I'll try! 👍

hikouki:
I tried. :)
How does it?

kball:
Looks great @hikouki! Would you be able to add a visual test or an example in the docs that we can use to straightforwardly test the behavior?

hikouki:
Leave it to me!

hikouki:
I did it. 🍣

kball:
Merged! Thanks @hikouki, sorry for the slow turnaround

Issue #10353, opened 2017-07-05
6.4.1 – `src/assets/js/custom.js` not included anymore like in 6.3.1?


Moonbase59:
The docs still say that all custom javascript files in `src/assets/js/` automatically get included into `app.js` (6.3.1 did) but it seems not to? Anything changed due to webpack?

[F6.4.1/SASS on Ubuntu]

kball:
Hi @Moonbase59 - can you clarify, are you talking about in the ZURB stack? If so, the answer is yes it has shifted b/c it's moved to using true module dependency, and we need to update the docs. Can you highlight the ones in particular you're seeing the outdated info on? Thanks!

Moonbase59:
@kball http://foundation.zurb.com/sites/docs/starter-projects.html#javascript-compilation

Any easy way to do this correctly in 6.4.1, a link maybe?

kball:
@Moonbase59 essentially you'll want to simply import/require the files you want to include. I believe webpack allows you to do either node style require/exports (http://openmymind.net/2012/2/3/Node-Require-and-Exports/) or ES6 modules: http://exploringjs.com/es6/ch_modules.html

Will hopefully have time to update docs soon, but if you get things working well and want to submit a PR to update those, that would be AWESOME :)

kball:
Going to close this based on the initial docs improvement, but we'll be continuing to work on those docs as we go.

udf2457:
@kball 

A little nudge to see if you're getting any nearer improving those docs ?  I stumbled across this problem today trying to work with the "new" FontAwesome 5. 

Putting `fontawesome-all.js` into `src/assets/js/` doesn't get compiled, I also tried `src/assets/js/lib/` So getting the modern FA5 SVG w/JS is proving to be tedious.

scottshuffler:
Same issue as @udf2457

ncoden:
@scottshuffler @udf2457 @Moonbase59 I updated the docs. See https://github.com/zurb/foundation-sites/pull/10925.

Does it solves your problem ? Is the docs clear enough now ?

scottshuffler:
@ncoden LGTM

PR #10365, opened 2017-07-07
First pass on updating the JS docs to better describe module bundling

kball:
Attempts to address https://github.com/zurb/foundation-sites/issues/10353

@Moonbase59 and @rafibomb can you both take a look and see if this makes sense/helps?

### DIFF
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -34,7 +34,9 @@ When configuring Babel to work with Foundation, you only need to include the `es
 
 ### File Structure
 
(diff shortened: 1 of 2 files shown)


IamManchanda:
Not been asked but this makes sense to me :)

matthewandrianakos:
I've got some technical skill, but not enough to parse this.

Can the docs include an example of how to include a module? 

I can't even figure out which files would need to be modified to achieve this.

rafibomb:
Yes - needs a bit more of the basics and this is a start

PR #10367, opened 2017-07-07
Add Button cursor property and bind it into a variable

### LINKED ISSUE #10366: Default <button> cursor should be configurable via _settings.scss
Should [default `cursor`](https://github.com/zurb/foundation-sites/blob/develop/scss/_global.scss#L205-L217) be a pointer (or a value configurable by `_settings.scss`), indicating that the button is clickable?

Hover over a `<button>` element.

#### What should happen:
Default `cursor` should be a pointer (or a value configurable by `_settings.scss`), indicating that the button is clickable.

#### What happened instead:
`cursor` CSS is not set, defaulting to `auto` and displaying an arrow.

Chrome on OSX

6.4.1, XY Grid

#### Test case link:
https://codepen.io/brentdodell/pen/WOgaov

IamManchanda:
This fixes #10366

cc @kball for review/merge!

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -109,6 +109,10 @@ $global-flexbox: true !default;
 /// @type Boolean
 $global-prototype-breakpoints: false !default;
 
+/// Button cursor's value, `auto` by default
+/// @type Keyword
+$global-button-cursor: auto !default;
+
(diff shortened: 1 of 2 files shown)


kball:
LGTM

Issue #10363, opened 2017-07-07
6.4.1: Foundation.MediaQuery.current empty when including foundation.util.mediaQuery separately


kitsimons:
I'm importing `foundation.core` and `foundation.util.mediaQuery` separately. Subsequently using `Foundation.MediaQuery.current` returns an empty string.

If I import `foundation` (everything) rather than the components, using `Foundation.MediaQuery.current` returns the current breakpoint as expected.

So this;

```
import 'foundation-sites/dist/js/plugins/foundation.core'
import 'foundation-sites/dist/js/plugins/foundation.util.mediaQuery'
```
```
$(document).foundation()
console.log('Current: ' + Foundation.MediaQuery.current)
```
Returns an empty string.

Whereas this;

```
import 'foundation-sites/dist/js/foundation'
```
```
$(document).foundation()
console.log('Current: ' + Foundation.MediaQuery.current)
```
Returns the current breakpoint.

This appears to be an issue with 6.4.1 (everything works as expected in 6.3.1)

kball:
Hi @kitsimons 
Good catch - the distfile for MediaQuery is not running init right now... it should, as the distfiles are intended to give "drop in" functionality. For now you should be able to workaround by adding `Foundation.MediaQuery._init()`

kitsimons:
Awesome, thanks @kball

flyingL123:
I just ran into this same issue using 6.4.3. I'm using webpack and am trying to import MediaQuery into one of my files, like this:

```
import { MediaQuery } from 'foundation-sites/js/foundation.util.mediaQuery';
```

Later in the file, `MediaQuery.current` will return an empty string unless I call `MediaQuery._init();` first.

Should the call to `_init` actually be in the MediaQuery files so that it's already initiated when imported, or am I supposed to call that private `_init` method?

flyingL123:
Yes @DanielRuf it still occurs because the `_init` method is not called in the `foundation-sites/js/foundation.util.mediaQuery` file, so you end up having go call `_init` yourself after importing it. Can't we just add some code to the file to call `_init` before it's exported? Maybe a variable to check if it's already been `init`'ed so it doesn't get called multiple times?

DanielRuf:
cc @ncoden should we reopen this?

flyingL123:
[...]

PR #10369, opened 2017-07-07
Distributed version of media query should initialize

kball:
Fixes https://github.com/zurb/foundation-sites/issues/10363


IamManchanda (approved):
LGTM

PR #10370, opened 2017-07-08
Fix missing `.` with class names with xy grid

brettsmason:
Noticed a bug when reworking grid container that a lot of the base classes (without margin/padding) were missing `.` in their class names and as a result weren't working.


IamManchanda (approved):
:+1:

IamManchanda:
LGTM... cc @kball for merge/review!

PR #10371, opened 2017-07-08
Grid Container Rework

brettsmason:
First take on the reworked grid container. Updates docs to show usage as well as modified visual test to show this visually.

To confirm what this does:

- reworks `grid-container` to have padding by default
- removes `grid-container-padded` but use its variables for the container padding
- adds `grid-container-fluid` and `grid-container-full`
- Makes `grid-padding-x` have negative margins if within `grid-container/grid-container-fluid` to properly line up with margin grids

This should get some solid testing before merging just to make sure this is the behaviour we want, but from the visual tests it works great.

### DIFF
--- a/docs/pages/xy-grid.md
+++ b/docs/pages/xy-grid.md
@@ -90,11 +90,24 @@ To define a grid type, simple set `.grid-margin-x` or `.grid-padding-x` on the g
 
 ## Grid Container
 
-The grid defaults to the full width of its container. In order to contain the grid, use the `.grid-container` class.
+The grid defaults to the full width of the available space. To contain it use the `grid-container` class. The container will be centered and have a max-width equal to your `$grid-container` setting (1200px by default), along with padding on the left/right equal to half your `$grid-container-padding` setting.
 
 ```html
 <div class="grid-container">
-  <div class="grid-x">
+  <div class="grid-x grid-margin-x">
+    <div class="cell small-4">cell</div>
+    <div class="cell small-4">cell</div>
+    <div class="cell small-4">cell</div>
+  </div>
+</div>
+```
+### Grid Container Fluid
+
+To stretch the content to the full width of the available space, simply use the class `grid-container-fluid` instead.
+
+```html
+<div class="grid-container-fluid">
+  <div class="grid-x grid-margin-x">
     <div class="cell small-4">cell</div>
     <div class="cell small-4">cell</div>
     <div class="cell small-4">cell</div>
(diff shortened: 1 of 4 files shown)


IamManchanda:
@brettsmason In v6, its 
```html
<button type="button" class="button primary">Primary</button>
```

In v7 (ofcourse with optional namespace and healthchecks) its gonna be 
```html
<button type="button" class="button button--primary">Primary</button>
```

---

So why not follow the same pattern with `full` & `fluid` and lets go with this below in 6

```html
<div class="grid-container full"></div>
```

and change like this in 7

```html
<div class="grid-container grid-container--full"></div>
```

brettsmason:
@IamManchanda Yeah that's what I was thinking. I'll change that later.
My one question regarding the `full` class: It didn't make any sense to me to hide the left/right padding using `full` for a padding grid so at the moment this only applies to margin grids. If you take a look at the visual test you will see what I mean. Do you think this makes sense/is correct?

kball:
On a readthrough, this looks good to me.

IamManchanda (approved):
LGTM... Merging!

IamManchanda:
The docs copy is updated ... Awesome work again @brettsmason :wink:

brettsmason:
@OndrejVasicek Here's a CodePen to illustrate the differences: https://codepen.io/brettsmason/pen/RZjdKo

You are correct in that you don't really need a container if you are going for the `full` example, its just there for consistency really (and if we ever move to a left only gutter it will be useful to solve the overlap issue).

The padding is there as now the grid container adds padding to margin and padding grids. This now means you can use both gutter types in the same design as the columns line up correctly (see in the top 2 examples they don't). Prior to 6.4.2 if using padding gutters on one element and margin gutters on another, the columns wouldn't be in line with each other.

Hope that makes sense.

OndrejVasicek:
Thank you Brett. It makes sense but it raises new question – I thought that one of the reasons we have now padding and margin gutters is the different behaviour on the edges (margin goes to the edge, padding have some padding). But now if I use container, it behaves the same. What’s the difference between padding and margin gutters in this case? When I should use padding and when margin? It seems like the existence of one of them is useless in case of using container.

Or was there a reason why it aligns the edges on the same places when using container?

JarvisH:
@OndrejVasicek

Just to add my two cents: the difference in padding/margin inside a grid-container would allow for colored backgrounds on padded cells, with contents still aligned to margin cells. Like this: [Codepen](https://codepen.io/jarvisH/pen/oGgaRR)

PR #10377, opened 2017-07-08
Fixed: I made the option posible recursively merge.

hikouki:
I made the option posible recursively merge for Abide component.
Only Abide component.

## Before

`Foundation.Abide.defaults.validators` was overwritten.

```js
var options = {
  validators: {
    notEqualTo: function (el, required, parent) {
      return $(`#${el.attr('data-equalto')}`).val() !== el.val();
    }
  }
};

plugin = new Foundation.Abide($html, options);

Foundation.Abide.defaults.validators //=> { equalTo: function }
plugin.options.validators //=> { notEqualTo: function }
```

## After

```js
...

plugin = new Foundation.Abide($html, options);

Foundation.Abide.defaults.validators //=> { equalTo: function }
plugin.options.validators //=> {  equalTo: function, notEqualTo: function }
```

## Why did you correspond only to Abide?

There was only Abide component which included the object in the option.

```js
for (var name in Foundation._plugins) {
  var defaults = Foundation._plugins[name].defaults
  for (var key in defaults) {
    if (defaults[key] instanceof Object) {
      console.log(name);
    }
  }
}
```
=> `abide`

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -19,7 +19,7 @@ class Abide extends Plugin {
    */
   _setup(element, options = {}) {
     this.$element = element;
-    this.options  = $.extend({}, Abide.defaults, this.$element.data(), options);
+    this.options  = $.extend(true, {}, Abide.defaults, this.$element.data(), options);
 
     this.className = 'Abide'; // ie9 back compat
     this._init();
--- a/test/javascript/components/abide.js
+++ b/test/javascript/components/abide.js
@@ -18,6 +18,22 @@ describe('Abide', function() {
       plugin.$element.should.be.an('object');
       plugin.options.should.be.an('object');
     });
+
+    it('the options are recursively merged', function() {
+      $html = $('<form data-abide novalidate></form>').appendTo('body');
+
+      var options = {
+        validators: {
+          notEqualTo: function (el, required, parent) {
+            return $(`#${el.attr('data-equalto')}`).val() !== el.val();
+          }
+        }
+      };
+
+      plugin = new Foundation.Abide($html, options);
+
+      plugin.options.validators.should.includes.keys('equalTo', 'notEqualTo');
+    });
   });
 
   describe('validateInput()', function() {


kball:
Looks good to me

benjaminkohl:
I'm not sure if this is related, but attempting to add custom validators using [the example in the Abide documention](https://foundation.zurb.com/sites/docs/abide.html#adding-custom-pattern-and-validator) is no longer working for me in Foundation v6.4.3. Also, it appears the code sample got messed up somehow. The function is floating below the Foundation initialization.

ghsolutions:
I have the same problems after upgrading from 631 to 643. when i put my own patterns in the foundation.js-File  to the other default patterns it works well, 

but with 

Foundation.Abide.defaults.patterns['zeit'] = /^(0[0-9]|1[0-9]|2[0-3])(:[0-5][0-9]){1,2}$/,
Foundation.Abide.defaults.patterns['datumdeutsch'] = /^\d{2}[\/\-.]\d{2}[\/\-.](?:19|20)[0-9]{2}$/,

in app.js it does not work.

hikouki:
Plugins(Abide) are load only once.
That's when `$(document).foundation()` was called.
https://github.com/zurb/foundation-sites/blob/develop/js/foundation.core.js#L158

So, you need to set `Foundation.Abide.defaults.*` before `$(document).foundation()` is called.

BigBlockStudios:
I still cannot get custom validators to work at all: 

`
Foundation.Abide.defaults.validators['validation_fails']; // seems to get ignored

// Foundation.Abide.defaults.validators.*; //this fails with unexpected token

// Init Foundation
$(document).foundation();

Foundation.Abide.defaults.validators['custom_validator'] = function($el,required,parent) {
  return false;
};`

what is the syntax for "setting" our custom validators before foundation initializes???

Lazerproof:
@BigBlockStudios  
You have to use patterns array instead of validators 
```
Foundation.Abide.defaults.patterns['phone'] = /\([0-9]{3}\)\s[0-9]{3}-[0-9]{2}-[0-9]{2}/;
$(document).foundation();
```

ncoden:
@BigBlockStudios See this example:

```js
// Set a validator for all Abide forms
Foundation.Abide.defaults.validators['global_custom_validator'] = function($el,required,parent) { return false; };

// Initialize my own Abide form with a custom validator
var plugin = new Foundation.Abide($('#my-custom-data-abide'), {
    validators: { local_custom_validator: function (el, required, parent) { return false; } }
});

console.log(plugin.options.validators);
// Show both `global_custom_validator` and `local_custom_validator`
// Others abide forms that would be initialized with `$(document).foundation();` would only have `global_custom_validator`
```

PR #10381, opened 2017-07-09
Off-canvas with sticky / fixed elements [old]

SassNinja:
Guess I've come out as an off-canvas power user with my prior PR :wink: 
So it's probably no suprise I'm facing a new off-canvas problem (apart from the support of push transition for nested off-canvas elements).

I'm often using a sticky top bar for medium down to open an off-canvas navi.
What really bothers me then is the loss of the fixed state while the off-canvas with push transition is opened.
This is caused by the transform scope of the content container and as far as I know Brett, discarding the transform setup is not an option for him (I agree to this because performance is definitely important).

That's why I've created the possibility to preserve the fixed state anyway.
I'm pseudo fixing the sticky elements by setting inline CSS when opening the off-canvas (position absolute values) and restoring the original values after having closed the off-canvas again.

This PR adds the new attribute `data-off-canvas-sticky` that has to be added to each sticky / fixed element that is supposed to get pseudo fixed on opening an off-canvas with push transition.
Besides I've introduced the new event `closedEnd.zf.offcanvas` that gets triggered once the close transition has finished.

@brettsmason @kball **what do you think? Looking forward to getting you feedback on this!**

You can easily see my changes in action when you open the off-canvas docs page, scroll down a bit and open a push off-canvas.
Currently (live) the fixed navi disappears. In my fork it remains sticky (pseudo fixed).

### DIFF
--- a/docs/layout/default.html
+++ b/docs/layout/default.html
@@ -41,7 +41,7 @@
         </div>
       </a> -->
 
-      <div class="docs-sticky-top-bar">
+      <div class="docs-sticky-top-bar" data-off-canvas-sticky>
         {{> navigation}}
         {{> mobile-navigation}}
         {{> search-bar}}
@@ -54,7 +54,7 @@
 {{> body}}
         </div>
         <div class="medium-3 large-2 medium-pull-9 large-pull-10 hide-for-small-only columns" data-sticky-container>
-          <div class="docs-nav-container" data-sticky data-anchor="body-container" data-margin-top="6">
+          <div class="docs-nav-container" data-sticky data-anchor="body-container" data-margin-top="6" data-off-canvas-sticky>
             {{> component-list}}
           </div>
         </div>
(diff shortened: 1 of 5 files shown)


IamManchanda:
> SassNinja  wants to merge 1 commit into `zurb:develop` from `SassNinja:develop`

@SassNinja We are frankly not accepting the branch name with `develop` of the user from now on! (I just now made the rule :p)

Please it would be great if you can `git checkout -b my-custom-branch` and send us a PR ?? 

---

By the way ... thanks for this :+1:

IamManchanda:
@SassNinja Hope you wont be angry with me on this thread close... I just want you to create a branch with new name!

SassNinja:
@IamManchanda I don't understand why a my fork's branch name matters but sure, I'll copy & paste everything in a feature branch

doertedev:
Cross-linking #10408 - the new PR :)

PR #10395, opened 2017-07-12
use https as protocol

DanielRuf:
This PR adds https to the protocol relative URLs so https is enforced.

This solves different issues:
* testing locally works (does not use the file protocol anymore)
* enforces https and so can use http/2 (faster)
* prevents MitM attacks
* protocol relative URLs are an anti-pattern for a few years
* ...

https://jeremywagner.me/blog/stop-using-the-protocol-relative-url
https://www.paulirish.com/2010/the-protocol-relative-url/


kball:
👍

PR #10404, opened 2017-07-13
Update guidelines as per discussion!

IamManchanda:
cc @kball @brettsmason @ncoden


IamManchanda:
Gonna close as this has become a mess due to some history issues in the repo we have had. Will create new PR instead!

PR #10405, opened 2017-07-13
Resolve a RTL uncentered push/pull issue

Seanom:
Using RTL direction. Setting a content to be uncentered and also using push/pull will do so from the incorrect direction

Using $global-left resolves this, we are using the following classes, which produced the issue

```small-11 small-centered medium-8 large-uncentered large-6 large-push-3 columns```


kball:
Makes sense to me!

PR #10408, opened 2017-07-14
Off-canvas fixed / sticky

SassNinja:
Guess I've come out as an off-canvas power user with my prior PR :wink: 
So it's probably no suprise I'm facing a new off-canvas problem (apart from the support of push transition for nested off-canvas elements).

I'm often using a sticky top bar for medium down to open an off-canvas navi.
What really bothers me then is the loss of the fixed state while the off-canvas with push transition is opened.
This is caused by the transform scope of the content container and as far as I know Brett, discarding the transform setup is not an option for him (I agree to this because performance is definitely important).

That's why I've created the possibility to preserve the fixed state anyway.
I'm pseudo fixing the sticky elements by setting inline CSS when opening the off-canvas (position absolute values) and restoring the original values after having closed the off-canvas again.

This PR adds the new attribute `data-off-canvas-sticky` that has to be added to each sticky / fixed element that is supposed to get pseudo fixed on opening an off-canvas with push transition.
Besides I've introduced the new event `closedEnd.zf.offcanvas` that gets triggered once the close transition has finished.

@brettsmason @kball **what do you think? Looking forward to getting you feedback on this!**

You can easily see my changes in action when you open the off-canvas docs page, scroll down a bit and open a push off-canvas.
Currently (live) the fixed navi disappears. In my fork it remains sticky (pseudo fixed).


IamManchanda:
Thanks for this ... you didnt asked me though but i will look into it this weekend :smile:

SassNinja:
@IamManchanda your feedback is appreciated as well 😉 

I mentioned only @brettsmason and @kball because they had reviewed my previous off-canvas PR

SassNinja:
ping @brettsmason & @IamManchanda 

Did you already manage to have a look at this feature?

brettsmason:
@SassNinja Sorry been really busy.
I'll make time tonight to go over it, and thanks for putting a PR together!

brettsmason:
@SassNinja Managed to take a look at this last night.
First impressions are it seems to work really well for this use case. I still need to check it out in other browsers, but works solidly in Chrome at least.

My only query is the Sticky plugin was supposed to be rewritten for 6.4 (but slipped) to use the `position: sticky` behaviour, with a fallback. If/when this was implemented then this fix would no longer be required from what I understand.

@kball what do you think?

SassNinja:
@brettsmason thanks for your first feedback!

I haven't tested a `position: sticky` solution yet because browser support is still quite bad:
http://caniuse.com/#search=sticky

Imo it doesn't make sense to rely on a fallback if most of the browsers will have to use it.
Even mozilla has marked it as experimental although FF is partially supported:
https://developer.mozilla.org/de/docs/Web/CSS/position

Do you think it can be used for production anyway?

brettsmason:
@SassNinja sorry should been a bit clearer with my sticky comment :smile:

I wasn't really involved with it, but I think the plan was to use something like https://github.com/soenkekluth/sticky state and extend it, but I don't know the details. @kball will know more about the plans.

kball:
As far as I know @colin-marshall has not had time to do anything with sticky. 

For v7 I'm wondering if we should detach this in favor of a recommended plugin (e.g. sticky-state: https://github.com/soenkekluth/sticky-state)

SassNinja:
@brettsmason @kball I've tested your mentioned plugin / `position: sticky` for a simple off-canvas dummy.

I'm afraid it doesn't seem to solve the problem.
Here's a quick codepen where you can see what I mean and play around a bit with it
https://codepen.io/Sheldor/pen/gxaJKG

To sum it up:
1. as soon as I add `overflow: hidden` to the offcanvas wrapper (page) the sticky feature doesn't work any more. Without this overflow property a horizontal scrollbar appears on opening the offcanvas.
2. if I nest the offcanvas in the content container, the same issue this PR is fixing occurs: the offcanvas elem disappears after having scrolled down

So in my eyes the mentioned plugin is not the solution.
What do you think?

brettsmason:
@SassNinja Sorry I haven't got back to you sooner on this.

Looking at your CodePen if I add `overflow-x: hidden` to `body it appears to work well.

Could we therefore not add a class to body on open to hide the overflow on the correct axis and remove it when the off canvas has closed?

SassNinja:
@brettsmason you're right `overflow-x: hidden` on the body seems to fix issue 1 (although I'm worrying about the impact on other components / custom code due to this new body style...)

However, this doesn't fix issue 2 completely: the opened nested off-canvas disappears on scrolling down.
Imo there's no alternative to setting a height via JS (= viewport height) and disable content scrolling if nesting because 100% height doesn't work in this case.
Do you agree?

doertedev:
Any news on this?

patricklindsay:
Bump @IamManchanda @brettsmason @SassNinja 

Is there an available workaround until this is fixed?

DanielRuf:

DanielRuf:
Still relevant and we need a rebase here =)

doertedev:
Can I just pay someone for this :(

ncoden:
@SassNinja 👋 Hi. I resolved conflicts, fixed some bugs and added tests.

kball:
Fixing up existing conflicts, reviewed that it works, just going to merge this F'er.

PR #10421, opened 2017-07-16
Fix for #10407 reveal content classes

### LINKED ISSUE #10407: Off-canvas and reveal-for-large ... bug

There is a bug when using reveal-for-large and the off-canvas together. It displays as should for large screens and small screens. The column hides for medium screens as it should but the space does not shrink, leaving a blank space (extra margin-left). This issue is only in the new 6.4 xygrid version. Ive tried it on my own fresh installed projects and in codpen, foundation version 6 and 6.4 and it happens only in the 6.4.

#### What should happen:
the off canvas menu should shrink/hide on medium screens and lower and reveal on large screens only

#### What happened instead:
the off canvas shows for large, hides for small and medium, but in medium only it leaves a blank space (extra margin-left) and collapses only when in small screen size
chrome, Microsoft Edge
6.4
#### Test case link:
[CodePen with Foundation 6.4.1, XY Flexbox grid](https://codepen.io/NoreagoCan/pen/dRaBoZ?editors=1000)

SassNinja:
This PR fixes https://github.com/zurb/foundation-sites/issues/10407

Due to performance improvements the reveal related classes get only updated if the param `hasReveal` is set to true. Since `_addContentClasses()` is calling `_removeContentClasses()` it must loop through that param what was missing. I've fixed this now!

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -190,7 +190,7 @@ class OffCanvas extends Plugin {
    * @private
    */
   _addContentClasses(hasReveal) {
-    this._removeContentClasses();
+    this._removeContentClasses(hasReveal);
     this.$content.addClass(`has-transition-${this.options.transition} has-position-${this.position}`);
     if (hasReveal === true) {
       this.$content.addClass(`has-reveal-${this.position}`);
--- a/test/visual/offcanvas/offcanvas-reveal-modal-large.html
+++ b/test/visual/offcanvas/offcanvas-reveal-modal-large.html
@@ -1,0 +1,85 @@
+<!doctype html>
+<!--[if IE 9]><html class="lt-ie10" lang="en" > <![endif]-->
+<html class="no-js" lang="en" dir="ltr">
+  <head>
+    <meta charset="utf-8">
+    <meta http-equiv="X-UA-Compatible" content="IE=edge">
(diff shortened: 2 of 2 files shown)


rafibomb:
@SassNinja Thanks for this! @IamManchanda @NoreagoCan can you help test this PR for merge?

IamManchanda:
@rafibomb digging it in!

brettsmason:
Pulled this down and ran the `reveal` visual test - can confirm this works correctly.
Looks good to merge from my point of view, thanks @SassNinja 😄

IamManchanda:
Yup this works awesome! Merging it had a visual test myself too!

IamManchanda:
Hey @SassNinja Although this was merged 
But me and @brettsmason had a deep look into it just now 
and 
theres something still buggy going on with `position-right`. If you resize the window to smaller screen, refresh and then increase size, it works fine
If you resize from large to smaller, it doesn't work correctly!

Can you look into it 
Visit => http://localhost:3000/offcanvas/offcanvas-reveal-modal-large.html

IamManchanda:
The original bug reported with `position-left` works fine though!
But seems issue with `position-right`

brettsmason:
@SassNinja @IamManchanda Looking at it more - it looks like once `off-canvas-content` has the class `has-reveal-left/right` they aren't being correctly removed on change of breakpoint. You will know more than me though, I'm a bit out of the loop on off canvas at the moment 😄

PR #10428, opened 2017-07-18
Breakpoint specific sizes for off-canvas

SassNinja:
This PR add the possibility to define a different off-canvas size for small and medium by replacing the number with a map.

In a current project the off-canvas is supposed to have clearly more width for medium than for small.
But no matter what value I choose (even percentaged or viewport related), it doesn't look good for both breakpoints.
The bottom line is I need two different off-canvas sizes.

The new maps `$offcanvas-sizes` and `$offcanvas-vertical-sizes` replace the prior `$offcanvas-size` and `$offcanvas-vertical-size`. Since I don't consider any custom setting of these deprecated variables I've added warnings that appear during the gulp process.

Alternatively I could leave the variables name untouched and add several conditions to check if it's a number or a map. But imo a warning is better because it isn't much work for the user to update the settings and keeps the code cleaner.

@brettsmason @IamManchanda what do you think about this feature?

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -491,8 +491,14 @@ $meter-fill-bad: $alert-color;
 // 25. Off-canvas
 // --------------
 
-$offcanvas-size: 250px;
-$offcanvas-vertical-size: 250px;
+$offcanvas-sizes: (
+  small: 250px,
+  medium: 350px,
+);
+$offcanvas-vertical-sizes: (
+  small: 250px,
+  medium: 350px,
+);
 $offcanvas-background: $light-gray;
 $offcanvas-shadow: 0 0 10px rgba($black, 0.7);
 $offcanvas-inner-shadow-size: 20px;
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -6,14 +6,6 @@
 /// @group off-canvas
 ////
 
-/// Deprecated settings warning
-@if variable-exists(offcanvas-size) {
-  @warn '$offcanvas-size is deprecated and not used anymore! Please update your settings and use the map $offcanvas-sizes instead';
-}
-@if variable-exists(offcanvas-vertical-size) {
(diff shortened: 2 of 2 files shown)


kball:
For backwards compatibility with old settings files, can you have the old variables set up the new ones if they're defined? Can also print a deprecation warning, but it's best if we don't break projects that have the older settings and not the newer.

SassNinja:
@kball a deprecation warning was already included by me but you're right, still supporting the deprecated settings would not be amiss.

I've updated the PR and take over the values now.

kball:
Awesome... this is looking good as far as it goes - makes me wonder though if we should make the logic iterate through the sizes (to allow someone to do this for any arbitrary map of breakpoints) rather than explicitly call out small & medium... thoughts?

Also @brettsmason would love your thoughts & review

IamManchanda:
This is looking cool from outset... let me dig in...

scss/settings/_settings.scss
```diff
@@ -491,8 +491,14 @@ $meter-fill-bad: $alert-color;
 // 25. Off-canvas
 // --------------
 
-$offcanvas-size: 250px;
-$offcanvas-vertical-size: 250px;
+$offcanvas-sizes: (
+  small: 250px,
+  medium: 350px,
```
IamManchanda: Settings.scss is automatically generated! Remove this file changes

IamManchanda (changes requested):

IamManchanda (approved):
Requested one change above ... apart that looks awesome after digging in!

Have tested it out and can confirm that this works great!
Awesome @SassNinja

scss/settings/_settings.scss
kball: yeah though it's kind of find - it will get overridden anyway.

kball:
I'm going to merge this as is, because I'd like to get a point release out and want to land on testable set of changes... that said, would really love a new PR that iterates over the breakpoints to allow any set rather than hard coding small & medium

Issue #10429, opened 2017-07-18
Drilldown breaks focus trapping functionality of the off-canvas


dbombine:
#### How to reproduce this bug:

1. Set up a basic instance of the off-canvas.
2. Add a basic instance of the drilldown inside off the off-canvas instance.
3. Press the button to trigger the off-canvas.
4. Tab through the drilldown links using keyboard.
5. Notice that the focus moves to the inside the off-canvas-content element when it should stay trapped within the off-canvas.

#### What should happen:
Focus should stay inside the off-canvas.

#### What happened instead:
Focus moved outside.

Also notice that the drilldown submenus disappear when they are navigated to using keyboard (arrow keys).

#### Test case link:
[CodePen with Foundation](https://codepen.io/dbombine/pen/XgwOrZ)

kball:
@Owlbertz can you take a look at this?

Owlbertz:
Seems indeed like a bug as this is not the case without a Drilldown. Will check this out when I got the time.

Owlbertz:
Sorry, I totally forgot about this. I found out the issue lies in a `stopImmideatePropagation()` call that prevents the OffCanvas listener from trapping the focus.

I updated your pen with a fixed version here: https://codepen.io/Owlbertz/pen/rpBPdY and will submit a PR. Please pull the PR down and check if this solves the problem so it can be merged.

dbombine:
@Owlbertz, thank you very much! That did fix the issue. However, in the codepen you linked to, the focus goes elsewhere when going in the opposite direction (shift + tab). In my code it just stays trapped on the very first item of the drilldown.

Owlbertz:
It now leaves the Drilldown and focuses the element before the Drilldown (i.e. the OffCanvas' close button). Would you consider this new behavior as worse? I feel like it's more consistent this way.

dbombine:
It looks like it's behaving correctly when going down the tree (tab), but
doesn't keep the focus trapped in the off-canvas when going back up
(shift + tab) - it just focuses on the Codepen's JS area instead of the
off-canvas close button.

Owlbertz:
For me it jumps from the close button to the last Drilldown element as expected.

![drilldown-focus](https://user-images.githubusercontent.com/4280373/33852515-2dc15e70-debb-11e7-9466-58fd2fd488eb.gif)

dbombine:
Must be browser-specific then somehow. Getting this misbehaviour in Safari.

[...]

PR #10394, opened 2017-07-12
update lock file with all dependencies

DanielRuf:
This PR updates the lock file to and include more packages. This was the result of the first `npm install` running Node 8 which updated the lock file and added the missing locked dependencies.

PR #10431, opened 2017-07-19
Drilldown method to show any menu

SassNinja:
This PR adds the new method `_showMenu` to the drilldown component for opening any (sub)menu in it **without** clicking through every menu on the way to it.

I usually like the drilldown menu because it's a nice and intuitive navi which I often use in the off-canvas. But it has a huge downside: there's currently no method to jump into a specific submenu without clicking through several submenus.

I've added a visual test in this PR that shows the difference between the current `_show` and the new `_showMenu`. You'll see what I mean when you check the 'Use current show' checkbox and click the buttons C, B, X (in this order) and then use the back links.
http://localhost:3000/drilldown/drilldown-show-menu.html

The new method is probably a great feature for everyone who needs to e.g. auto open the current page in the drilldown if using it as page navi.

@IamManchanda @kball @brettsmason what do you think?


IamManchanda:
This is not awesome... this is super awesome !!! Hats off 
Its like deeplinking came into drilldown's :wink: 

cc @kball for merge/review!

IamManchanda (approved):
:+1:

js/foundation.drilldown.js
```diff
@@ -343,6 +343,86 @@ class Drilldown extends Plugin {
...
+   * @private
+   * @param {jQuery} $elem - the target submenu (`ul` tag)
+   * @param {boolean} trigger - trigger drilldown event
+   */
+  _setHideSubMenuClasses($elem, trigger) {
+    $elem.removeClass('is-active').addClass('invisible').attr('aria-hidden', true);
+    $elem.parent('li').attr('aria-expanded', false);
+    // if (trigger === true) {
+    //   $elem.trigger('hide.zf.drilldown', [$elem]);
+    // }
```
IamManchanda: Remove these comments!

IamManchanda (changes requested):

js/foundation.drilldown.js
SassNinja: The lines are commented in now. Instead of removing the possiblity to trigger a close event I'm  just not using it in _showMenu anymore.

SassNinja:
@IamManchanda many thanks for your kindly words! I'm glad to hear this feature is appreciated 😃 

Apart from your requested change (see my comment above) I've made another slight improvement:
The method `_showMenu` has now a second param called `autoFocus` which will prevent the auto focus of the first link in the target (sub)menu if false.
I think this might by helpful in cases where the auto focus is not desired (e.g. if the drilldown is hidden when calling that function). Default remains true.

IamManchanda (approved):
Seems very Good!

kball:
Testing through the visual test I see a few issues

1. When I click in using the 'showMenu' option I get some weird focus behavior (the focus ring which is supposed to only show on keyboard nav starts showing on elements within the drilldown.
2. When I keyboard nav around, I seem to occasionally trigger some sort of behavior where the entire drilldown becomes hidden.

Both of these were in Chrome latest (59) on mac

SassNinja:
@kball thanks for your feedback!

1. That's actually a feature: I've added this auto focus so the user can jump right into the drilldown via keyboard after having clicked the trigger (button). Would you prefer no auto focus by default?
Currently the auto focus may get disabled by explicitly setting the param to false
```javascript
$drilldown.foundation('_showMenu', $menu, false);
``` 
2. Would you describe more detailed how to reproduce this?
But I think I know what you mean. When I click 'Show X', then above the buttons (lose focus) and then press several times the TAB key the drilldown menus get mixed up.
However **this isn't caused by my changes / PR** because it also happens if you don't click any button but open "Item A3" the usual way (clicking on it), then click above the buttons (lose focus) and then press several times the TAB key. I think it's a general drilldown focus management issue and may be related to this issue https://github.com/zurb/foundation-sites/issues/10429

SassNinja:
ping @kball  – would be great to get your feedback (in particular on the 1st point)

Besides I've noticed this PR and some others are broken, precisely its history.
- https://github.com/zurb/foundation-sites/pull/10404
- https://github.com/zurb/foundation-sites/pull/10456
- https://github.com/zurb/foundation-sites/pull/10394
- https://github.com/zurb/foundation-sites/pull/10408

Surprisingly everathing was fine when I looked into it last time and there haven't been pushes in the meantime (at least in my PRs).
Any idea what might have broken them?

IamManchanda:
Poke @kball ... Now what needs to bee done ?

kball:
@SassNinja with regards to the PRs getting wonky, unfortunately we had some issues with the develop branch... we've fixed those (and added protection to prevent future occurrences) but old PRs seem to still be a little wonky... going to try closing and reopening to see if that fixes it.

kball:
With regards to your questions - I think we can separate #2 into a differnet issue as I've inconsistently seen it in other situations... for #1 we generally don't show the focus ring for non-keyboard nav, so I'd say lets make the default no focus, and then this should be good to go.

Since it doesn't seem that closing and reopening fixed the history, I'm going to close entirely, can you try creating a new PR with these changes and the update above? Thanks!

PR #10433, opened 2017-07-19
Rename duplicate $block-grid-max` to `$xy-block-grid-max` ...

### LINKED ISSUE #10432: $block-grid-max variable in settings twice
$block-grid-max is in settings.scss under both "3. The Grid" and "56. XY Grid". When using Float Grid whatever $block-grid-max is under The Grid will be overwritten by the XY Grid value (later in the file).

Suggestion:
$block-grid-max setting should occur only once and be moved to a separate heading until Float Grid is  removed. Or at least 1 or both commented out and notated.

brettsmason:
Allow for both types of block grids.
Closes #10432

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -185,7 +185,7 @@
 @mixin xy-block-grid-classes($margin-grid: true, $padding-grid: true) {
   @if $padding-grid {
     @include -zf-each-breakpoint {
-      @for $i from 1 through $block-grid-max {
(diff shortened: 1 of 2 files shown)


kball:
Migration note: `$block-grid-max` was not separated for xy-grid and regular grids. The xy-grid setting has been separated out to `$xy-block-grid-max`. If you have changed the value of `$block-grid-max` and are using the xy-grid, you should update `$xy-block-grid-max`.

Issue #10387, opened 2017-07-11
XY-Grid cell gutters not output at larger breakpoints


rbayliss:
#### How to reproduce this bug:

```scss
.SearchForm {
  @include xy-cell(shrink, $gutter-position: right left bottom, $vertical: true);
  @include breakpoint(xxlarge) {
    @include xy-cell(shrink, $gutter-position: right left bottom top, $vertical: true);
  }
}
```

#### What should happen:
```css
.SearchForm {
    flex: 0 0 auto;
    height: auto;
    margin-right: 0.625rem;
    margin-left: 0.625rem;
    margin-bottom: 0.625rem; }
    @media screen and (min-width: 90em) {
     .SearchForm {
        flex: 0 0 auto;
        height: auto;
        margin-right: 0.9375rem;
        margin-left: 0.9375rem;
        margin-bottom: 0.9375rem;
        margin-top: 0.9375rem; } }
```

#### What happened instead:
```css
.SearchForm {
    flex: 0 0 auto;
    height: auto;
    margin-right: 0.625rem;
    margin-left: 0.625rem;
    margin-bottom: 0.625rem; }
    @media screen and (min-width: 90em) {
     .SearchForm {
        flex: 0 0 auto;
        height: auto; } }
```
Note the lack of margin at the XL breakpoint.  
#### Browser(s) and Device(s) tested on:
SASS compiled with node-sass-chokidar
#### Foundation Version(s) you are using:
6.4.1

#### Test case link:
Sorry - not sure how to set up a CodePen with Foundation SCSS.

rbayliss:
Looking at the xy-cell mixin:

```scss
@if $gutter-output {
  // If gutters = map
  @if(type-of($gutters) == 'map') {
    // If $gutters map has a key = $breakpoint, output the value
    @if (map-has-key($gutters, $breakpoint)) {
      @include xy-gutters($gutter, $gutter-type, $gutter-position);
    }
  }
  // If not a map
  @else {
    @include xy-gutters($gutter, $gutter-type, $gutter-position);
  }
}
```
Unless overridden, $gutters = $grid-margin-gutters, which are:
```scss
$grid-margin-gutters: (
  small: 20px,
  medium: 30px
) !default;
```
Since the xxlarge breakpoint doesn't exist in the $gutters map, it doesn't get printed, even though xy-gutters() is capable of falling back to a smaller breakpoint.

svenraju:
Hello,

is there a solution for this?

[...]

PR #10437, opened 2017-07-19
Cell should always emit gutters if -output is enabled

kball:
Fixes https://github.com/zurb/foundation-sites/issues/10387 based on my assessment of the situation in https://github.com/zurb/foundation-sites/issues/10387#issuecomment-316537053

@brettsmason please review.


brettsmason:
@kball This looks good. Your comment about the mixin does indeed make sense and I think is the correct way to go. I've tested this locally and is fine so going to merge.

Issue #10444, opened 2017-07-21
Down arrow scrolls page on dropdown menu when on last menu item


shonin:
#### How to reproduce this bug:

This is visible on the documentation site

1. [Go to the docs](http://foundation.zurb.com/sites/docs/dropdown-menu.html)
2. Tab to a dropdown menu
3. Use the down arrow to move down through the items in the list
4. observe that the page starts scrolling

#### What should happen:

The page doesn't scroll

#### What happened instead:

The page scrolls

#### Browser(s) and Device(s) tested on:

OSX 10.12 - Chrome

#### Foundation Version(s) you are using:

6.4

PR #10445, opened 2017-07-21
Always prevent default on down arrow on the dropdown menu

rykener:
Fixes Issue #10444

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -173,10 +173,8 @@ class DropdownMenu extends Plugin {
       });
 
       var nextSibling = function() {
-        if (!$element.is(':last-child')) {
-          $nextElement.children('a:first').focus();
-          e.preventDefault();
-        }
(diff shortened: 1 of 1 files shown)


kball:
So if I understand this correctly, in the last-child case what will happen is $nextElement will be an empty jquery object, so that part of the code will do nothing, while the prevent default will still fire preventing the page from scrolling.

I think this makes sense, but would love to get @Owlbertz's opinion as our expert on keyboard nav and accessibility

Owlbertz:
Can't currently pull this down and check it out, but based on the diff I'd say this makes sense.

shonin:
That's correct @kball. I just decided it would be best to use the same logic that arrowing up uses. 

On arrow up, the `$prevElement` isn't an empty jquery object (it's the last item in the list), but that is a separate issue that is not as critical as this one.

kball:
Great. Merged! Thank you @shonin, and congrats on first merged PR

Issue #10305, opened 2017-06-30
Grid-margin doesn't work with grid-frame


OndrejVasicek:
#### How to reproduce this bug:
1. Set up grid-y with some children
2. Add class `grid-frame` to spread the size to 100% of the screen
3. Add class `grid-margin-y` to add margin gutters between the children

I used grid-y, because the class `grid-frame` has bigger meaning here. When using grid-x the width is 100% by default already. But the bug is present in grid-x as well.

#### What should happen:
The margin gutters should be between the children but not on the edges.

#### What happened instead:
There is some weird spacing in the bottom of the page. It seems like the negative margins from the grid-margin-y class doesn’t work, when the element I 100vh.
![grid-frame](https://user-images.githubusercontent.com/3864012/27740136-321e0b5c-5db1-11e7-97b9-c6e5aa8169c4.png)

#### Browser(s) and Device(s) tested on:
Current Chrome and FF.

#### Foundation Version(s) you are using:
6.4
#### Test case link:
https://codepen.io/OndrejVasicek/pen/weyOoZ

kball:
Thanks @brettsmason - this is not actually sufficient - the outcome is about the same as we only have 1 gutters worth of spacing but the height calc is taking into account 2 gutters worth...

I *think* this may actually be a related problem to https://github.com/zurb/foundation-sites/issues/10297

For the margin grid to work properly we may need to insist on having a container which containers the overflow, and have the frame in that case add the gutter height in.

kball:
I've started a branch to work on this with a simple visual test using the codepen code here: https://github.com/zurb/foundation-sites/tree/fix-grid-frame-with-margin

brettsmason:
@kball I think I've worked out a fix (atleast a CSS solution - not sure how to handle with mixins).
Take a look at this CodePen: https://codepen.io/brettsmason/pen/LLBoNB
Basically we can set the height using `calc()` at `100vh + gutter`.

What do you think?

IamManchanda:
@brettsmason That's dope!

kball:
Looks great to me, you're the man @brettsmason! Want to put together a PR?

brettsmason:
@kball sorry yes I'll put a pr together for this today or tomorrow depending on time.

PR #10448, opened 2017-07-21
Xy grid frame fix for margin grids

kball:
Fixes https://github.com/zurb/foundation-sites/issues/10305

@OndrejVasicek as the original filer of the issue, could you take a look?

cc @IamManchanda @brettsmason


IamManchanda:
Notes: Will look into it tomm!

OndrejVasicek:
Well I’m not sure the suggested solution (https://codepen.io/brettsmason/pen/LLBoNB) really works. I can still see some margin at the bottom.

brettsmason:
@OndrejVasicek That CodePen is slightly different to the PR. The PR adds a full amount of gutter to the height calculation. The only requirement for this to work is overflow on the body needs to be hidden, but I've added a note to the docs about this.

From my testing and review this works great so I'm going to merge, thanks for the help on this as always @kball 😄 👍

PR #10451, opened 2017-07-22
Iterate over off-canvas sizes map

SassNinja:
This PR iterates over the `$offcanvas-sizes` and `$offcanvas-vertical-$sizes` map instead of using hard coded small & medium.

It's an extension of my previous sizes map PR that is already closed / merged https://github.com/zurb/foundation-sites/pull/10428

@kball do you wanna review this one? (since it was your change request)

### DIFF
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -196,25 +196,24 @@ $maincontent-class: 'off-canvas-content' !default;
   @if $position == left {
     top: 0;
     left: 0;
-    width: map-get($sizes, small);
     height: 100%;
-
-    transform: translateX(-(map-get($sizes, small)));
     overflow-y: auto;
 
-    @include breakpoint(medium) {
-      width: map-get($sizes, medium);
-      transform: translateX(-(map-get($sizes, medium)));
+    @each $name, $size in $sizes {
+      @include breakpoint($name) {
+        width: $size;
+        transform: translateX(-$size);
+      }
     }
     
     // Sets the position for nested off-canvas element
     @at-root .#{$maincontent-class} .off-canvas.position-#{$position} {
-      transform: translateX(-(map-get($sizes, small)));
 
-      @include breakpoint(medium) {
-        transform: translateX(-(map-get($sizes, medium)));
+      @each $name, $size in $sizes {
+        @include breakpoint($name) {
+          transform: translateX(-$size);
+        }
       }
-
       &.is-transition-overlap.is-open {
         transform: translate(0, 0);
       }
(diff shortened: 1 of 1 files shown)


IamManchanda:
i have a done this a lot so i can tell you this looks good from outset ... let me dig in and let you know?

IamManchanda (approved):
LGTM ... tested it out with visual testing also ... have reviewed the code too and this looks good!

cc @kball for merge!

kball:
👍

kball:
So... @SassNinja and @brettsmason we've run into an issue with this and it's paired PR (#10428) that showed up during QA.

The problem is this - we're nesting calls to `@include breakpoint` in the reveal-for-x classes. You can see this as we call
```
 @include breakpoint($name) {
        .position-left.reveal-for-#{$name} {
          @include off-canvas-reveal(left);
        }
        //.. other positions
      }
```

But the `off-canvas-reveal` mixin itself runs include breakpoints... this results in CSS that looks like

```
    @media print, screen and (min-width: 64em) and (min-width: 40em) {
      .position-bottom.reveal-for-large ~ .off-canvas-content {
        margin-bottom: 350px; } }
```

This behaves inconsistently across browsers - in chrome it appears to behave as expected (min-width 64em), but on Edge it does the wrong thing (min-width 40em).

This goes away if there's only one off-canvas sizing (so https://github.com/zurb/foundation-sites/pull/10504 changing defaults actually hides this bug), but it's definitely broken in the responsive sizing case.

Unfortunately we didn't catch this until close to release, so I'm inclined to roll back these changes to get a clean release but if we can figure out a clean fix over the weekend I'd be happy with that too.

SassNinja:
@kball surprisingly I can't find your described CSS output when looking into the compiled foundation.css in `_build/assets/css/`
````
@media print, screen and (min-width: 64em) and (min-width: 40em)
````

However I agree it's problematic to include the breakpoint mixin in another breakpoint mixin.

My idea is to add a fourth param to the `off-canvas-reveal` mixin to pipe through the breakpoint name. 
Then we can check if this name is set in the off-canvas-sizes map  and if not take the closest down.

What do you think?
As far as I can see there isn't a function for this yet, is there?

SassNinja:
Ok, I found a suitable function called `-zf-get-bp-val()` 😄 
Besides I was able to reproduce the problematic CSS output (I just didn't have medium in the sizes map)

Good news: the idea I've mentioned above seems to work!
I've made PR to fix it: https://github.com/zurb/foundation-sites/pull/10510

brettsmason:
@SassNinja you are awesome! I'll try look over this tomorrow when I'm around and get this included for the release.

PR #10471, opened 2017-07-27
Remove text-decoration from link buttons

DaSchTour:
Remove text-decoration from link buttons. If text-decoration is set in global settings we don't want this to apply on button.

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -377,7 +377,8 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
     }
   }
   a.button {
-    &:hover {
+    &:hover,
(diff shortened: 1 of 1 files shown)


scss/components/_button.scss
```diff
@@ -376,4 +376,9 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
       margin-#{$global-left}: 0;
     }
   }
+  a.button {
+    &:hover {
```
IamManchanda: &:hover, &:focus ??

IamManchanda (approved):
LGTM

IamManchanda:
cc @kball for merge!

PR #10473, opened 2017-07-27
Update tabs.md to fix link code.

jessgusclark:
Remove `/tabs/` from the link to Tab 2. Code as is throws JavaScript error, removing this fixes the tabs.

### DIFF
--- a/docs/pages/tabs.md
+++ b/docs/pages/tabs.md
@@ -13,7 +13,7 @@ There are two pieces to a tabbed interface: the tabs themselves, and the content
 ```html
 <ul class="tabs" data-tabs id="example-tabs">
   <li class="tabs-title is-active"><a href="#panel1" aria-selected="true">Tab 1</a></li>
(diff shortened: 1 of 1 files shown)


IamManchanda (approved):

IamManchanda:
cc @kball for merge!

jeromelebleu:
Hi @jessgusclark , which JavaScript error is thrown please? With that change, `data-tabs-target` and `href` are now the same. It was set to something else to illustrate the use of the `data-tabs-target`attribute, and I'm wondering if this documentation part is still obvious now... Is a simple `href="#"` could be used instead?

jessgusclark:
@jeromelebleu it wasn't a javascript change it was an HTML change in the <li>. No JS error was thrown, there was no functionality executed. I would guess that `href="#"` wouldn't work since `#/tabs/panel2` didn't work. It would be something to test though.

jeromelebleu:
The `data-tabs-target` attribute has precedence over the `href` one to identify the target, so it should work in both cases...

PR #10491, opened 2017-08-01
Fix typo

### LINKED ISSUE #10472: Medium and up class prefixes being generated as small
Following on from [this comment](https://github.com/zurb/foundation-sites/issues/10302#issuecomment-318004528) I have discovered that classes prefixed with their breakpoint name eg `size-collapse`, `size-uncollapse` and `size-unstack` are all receiving the prefix `small`.

I've tracked this down to [this commit](https://github.com/zurb/foundation-sites/commit/2f84122213e318910ba6c8e2059da3014194397e) - if I revert this then classes are generated correctly.

I've done a diff for generated CSS with and without that PR and all it changes is the incorrect prefixes, so looks like thie is indeed the culprit.

@kball Can you confirm?

kball:
Should fix https://github.com/zurb/foundation-sites/issues/10472

@brettsmason please review

### DIFF
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -145,7 +145,9 @@ $breakpoint-classes: (small medium large) !default;
   $pbp: index($-zf-breakpoints-keys, $print-breakpoint);
 
   // Make breakpoint size available as a variable
-  $old-zf-size: $-zf-size;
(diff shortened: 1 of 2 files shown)


brettsmason:
Following our Slack discussion and testing I approve this PR 👍 
Thanks for sorting out my mess 😉

PR #10492, opened 2017-08-01
Collapse sizing bug

kball:
Fixes problem where responsive collapse classes were getting the wrong breakpoints for margin grids.


brettsmason:
@kball I've been through and reviewed this.
From the visual test you provided this does indeed seem to fix the issue.

Looks good to merge 👍

PR #10500, opened 2017-08-03
Modified Kitchen Sink Page

danclaytondev:
Added data-closable attribute to closable callout div so it closes properly

That's all!

### DIFF
--- a/docs/pages/kitchen-sink.md
+++ b/docs/pages/kitchen-sink.md
@@ -306,7 +306,7 @@ description: Everything but.
   <button class="close-button" aria-label="Close alert" type="button" data-close>
     <span aria-hidden="true">&times;</span>
   </button>
-  <p>This is a static close button example.</p>
+  <p>This is a close button example.</p>
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


kball:
Thanks for the PR! This makes the callout closable, but to get it to close, we need to make the close button also have a `data-close` attribute so it will trigger the appropriate event.

danclayton50:
Thanks, done that now and it works!

IamManchanda (approved):

docs/pages/kitchen-sink.md
```diff
@@ -302,8 +302,8 @@ description: Everything but.
 </div>
 
 ```html_example
-<div class="callout">
-  <button class="close-button" aria-label="Close alert" type="button">
+<div class="callout" data-closable>
+  <button class="close-button" aria-label="Close alert" type="button" data-close>
   </button>
   <p>This is a static close button example.</p>
```
IamManchanda: I think you need to remove the word `static` ? No ?

IamManchanda (changes requested):

danclayton50:
Removed the word static.

kball:
Perfect! Merged, thanks @danclayton50!

PR #10504, opened 2017-08-03
Make default sizing of off-canvas backwards compatible

kball:
We noticed in QA that the new responsive off-canvas sizing felt odd because it changed the defaults... we want to keep the capabilities, but leave the default behavior the way it was for now.


brettsmason:
Looks good to me - merging!

Issue #10474, opened 2017-07-28
[Off-canvas] Documentation : Example no more available in the documentation


AmaZili:
Paragraph: Responsive Off-Canvas (Putting it all together) URL : http://foundation.zurb.com/sites/docs/off-canvas.html#responsive-off-canvas-putting-it-all-together-

The text states that an example is available:

> For an example of off-canvas on small screens and Top Bar Menu with Dropdowns, check out this Building Block: http://zurb.com/building-blocks/top-bar-with-off-canvas

This url ( http://zurb.com/building-blocks/top-bar-with-off-canvas )does not exist anymore

kball:
@IamManchanda @rafibomb Do we have a better example in the new building blocks or should we port that old one?

IamManchanda:
@kball http://foundation.zurb.com/building-blocks/blocks/multilevel-offcanvas-menu.html

AmaZili:
This one : http://foundation.zurb.com/building-blocks/blocks/multilevel-offcanvas-menu.html does not implement the top-bar feature as an alternative to the large version menu on mobile

rafibomb:
Yeah - should be a new one! Can we get some help on this @AmaZili ?

AmaZili:
Hi Rafi,

 

I may give a hand on this.

For the moment I am working on having a 6.4.1 megamenu that work… https://codepen.io/AmaZili/pen/WEjRBo

If this can help

 

Philippe

 

De : Rafi [mailto:notifications@github.com] 
Envoyé : vendredi 11 août 2017 02:03
À : zurb/foundation-sites <foundation-sites@noreply.github.com>
Cc : Philippe <philippe@amazili.com>; Mention <mention@noreply.github.com>
Objet : Re: [zurb/foundation-sites] [Off-canvas] Documentation : Example no more available in the documentation (#10474)

 

Yeah - should be a new one! Can we get some help on this @AmaZili <https://github.com/amazili>  ?

AmaZili:
Hi Raffi,

 

I am working on a project that imply drilldown menu, but can’t get it to work with foundation 6.4.

I have wrote a codepen available here : https://codepen.io/AmaZili/pen/ZJvzdR

 

Click on item1, then on item1A and see the truncation of the menu….

 

Do you want me to put that in the forum on fill it as a bug in github ?

 

Thanks for your help.

 

And once again, Zurb is really doing a really good job

 

All the best

 

Philippe @AmaZIli

[...]

PR #10506, opened 2017-08-04
Update building block example from old no-longer-present to a newer example

kball:
Fixes https://github.com/zurb/foundation-sites/issues/10474#issuecomment-320054839


rafibomb:
not a good example - we'll need a new one

PR #10509, opened 2017-08-05
Drilldown method to show any menu II

SassNinja:
This is a recommit of my previous PR whose history got mixed up
https://github.com/zurb/foundation-sites/pull/10431

@kball as requested by you I've changed the auto focus feature now so it only gets triggered if explicitely setting the param to true

```javascript
$drilldown.foundation('_showMenu', $menu, true);
```


kball:
Merged! Excellent.

PR #10510, opened 2017-08-05
Fix off-canvas sizes iteration for reveal

SassNinja:
This is a bugfix of the off-canvas-sizes map iteration in the off-canvas-reveal mixin that showed up during QA and was reported in my closed PR:
https://github.com/zurb/foundation-sites/pull/10451

There have been a problematic CSS output because the iteration includes the breakpoint mixin while it gets included by another breakpoint mixin.
I've fixed this now by piping through the breakpoint name as 4th param what lets me get rid of the iteration within the reveal mixin.

@kball please let me know if it looks good from your side, too

### DIFF
--- a/docs/pages/off-canvas.md
+++ b/docs/pages/off-canvas.md
@@ -397,6 +397,28 @@ Advanced off-canvas users may use the new `contentId` option to bind an element 
 
 ---
 
+## Off-canvas Sizes
+
+In v6.4.2 the type of the off-canvas size variables has changed from number to map. This lets you define breakpoint specific sizes instead of one value for all.
+The map may contain every key that is defined in `$breakpoint-classes`.
+
+<div class="warning callout">
+  Please note the sizes maps do currently not work perfectly for the reveal classes. If sizes are defined for medium and large, `.reveal-for-medium` will only consider the medium value. This is going to get fixed in a future release.
+</div>
+
+```scss
+$offcanvas-sizes: (
+  small: 250px,
+  medium: 350px,
+);
+$offcanvas-vertical-sizes: (
+  small: 250px,
+  medium: 350px,
+);
+```
+
+---
+
 ## Migrating from versions prior to v6.4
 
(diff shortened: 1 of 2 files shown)


brettsmason:
@SassNinja I've ran through the visual tests and can confirm this fixes the issue I found in QA with Edge. I've also looked over the generated CSS and all looks good to me now.

@kball might want to give it a quick look but this looks good to merge 👍

kball:
hmm... so I don't think this quite gets it working 100%, because it doesn't handle the off-canvas responsive sizing properly...

If I have a responsive sizing map that looks like
```
$offcanvas-sizes: (           
  small: 250px,               
  large: 350px,
) !default;   
```

and have `reveal-for-medium` on an off-canvas, then load on a large screen I will see the off-canvas to have width 350px but the off-canvas-content to only have margin-right 250px.

SassNinja:
@kball you're right, that case won't work but is this a major issue in your eyes?
When is the next release scheduled and can this get fixed afterwards (if not much time left)?

I'd like to see this feature in the upcoming release since it doesn't break an existing project and provide more control regarding the off-canvas size.

I'm not sure how fast this **limitation** you've described can be fixed.
The problem here is that the reveal mixin is called within a breakpoint mixin so there's by design no possibility to consider more than the current breakpoint.
To get arround this I'd probably have to change this
````sass
@include breakpoint($name) {
    .position-left.reveal-for-#{$name} {
        @include off-canvas-reveal(left, $offcanvas-reveal-zindex, $maincontent-class, $name);
    }
}
````
into something like this
````sass
.position-left.reveal-for-#{$name} {
    @each $breakpoint in $breakpoint-classes {
        @if BREAKPOINT GTE NAME AND IS IN SIZES MAP {
            @include breakpoint($breakpoint) {
                @include off-canvas-reveal(left, $offcanvas-reveal-zindex, $maincontent-class, $breakpoint);
            }
        }
    }
}
````
But even with a working condition the result will be duplicate CSS and thus requires a rework of the reveal mixin. Probably nothing that can be done in a few minutes 😐

kball:
I suppose it's probably fine given the base case backwards doesn't break... lets just highlight in the documentation for the new responsive sizing feature that it doesn't currently play nicely with the `reveal-for-x` feature, and that will be fixed in a future release

SassNinja:
Ok, sizes chapter with warning note is added on the docs page now!

PR #10519, opened 2017-08-07
[docs] Fix title on deprecation notices

Odaeus:
It's a common mistake to use "depreciated" where "[deprecated](https://en.wikipedia.org/wiki/Deprecation)" is meant so I assume this is a mistake. If it was an intentional choice please close this PR. 😃


rafibomb:
Crazy that that slipped through - good catch and thanks!

PR #10522, opened 2017-08-07
Original dropdown position & alignment and CSS classes

SassNinja:
I've used the dropdown component to build a custom tooltip.
While working on it I've got two problems:

1. the original position & alignment gets changed if an overlap is detected but doesn't get restored once the overlap case isn't taking effect anymore
2. there are no CSS classes set that provide the current position & alignment what is a huge problem if you need to adjust the styling according to the current value (in my case a tooltip arrow whose direction is dependent on the dropdown pane's position)

This PR fixes both of these problems by saving the original values and restoring them before the overlap check. Same for the missing CSS classes that get set & updated every time the position is set.

@kball what do you think about this? Any objections?

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -105,7 +105,9 @@ class Dropdown extends Positionable {
    * @private
    */
   _setPosition() {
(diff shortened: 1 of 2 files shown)


kball:
looks good to me

PR #10533, opened 2017-08-12
Update /scss/components/_off-canvas.scss

macgraphic:
Correction to the left/right off-canvas panel & Height map of a top/bottom off-canvas panel

Error pinged up in CodeKit.

### DIFF
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -8,7 +8,7 @@
 
 /// Width map of a left/right off-canvas panel.
 /// @type Map
-$offcanvas-sizes: (
+$offcanvas-size: (
   small: 250px,
 ) !default;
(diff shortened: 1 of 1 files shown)


SassNinja:
@macgraphic would you please describe more detailed what is the itention of your PR?

You've changed `$offcanvas-sizes` into `$offcanvas-size` and removed the deprecation warnings.
Both is a bad idea and will result in a broken project.

1. `$offcanvas-size` was a number used prior to v6.4.2 and has been replaced by the map `$offcanvas-sizes`
2. the purpose of the deprecation warnings is to make the user update his settings (the warnings will disappear then)

IamManchanda:
I am closing this as I am assuming this is by mistake but please let me know if its not the case?

Issue #10543, opened 2017-08-16
Why does reveal scroll on mobile?


SassNinja:
I'm using a modified reveal on small screen size that doesn't have the default height of 100vh.
But then I've detected the plugin scrolls up on opening and restores the original scroll position after having closed the reveal.
This only occurs on mobile devices which are identified via user agent instead of media query.

I'm wondering why is this scrollTop reset & restoration necessary at all?
It isn't visible with the default height and looks weird if the reveal's height is less than 100vh.

**Can this 'feature' just be removed or is there any particular reason for this I'm missing?**

SassNinja:
Here's a codepen to show what I mean:
https://codepen.io/KaiTheGreat/pen/xLPQJw

If you open it on a mobile device (or in dev tools with mobile user agent), scroll down a bit, click the button and the close the reveal, you can see the mentioned behavior.

SassNinja:
I wasn't able to resist :smile: 
I've made a PR that adds a new option to disable this behavior: https://github.com/zurb/foundation-sites/pull/10546

But if this mobile behavior isn't desired at all / needed anymore, I would adjust my PR and remove it completely.

kball:
It's not a desirable behavior - I believe it was inserted to try to capture scroll state as at some point on mobile (particularly ios) that was being lost... if this bug is no longer present (ie opening & closing reveal after scrolling stays at the same place) across ios and android by all means let's just ditch the behavior.

patrick-huber:
I'm seeing the same behavior on my mobile reveals. I've added in a temporary workaround in my app js code (below), but would love to see pull request #10546 or a similar update make it in

```
// Fixing bug in reveal on mobile devices
var scrollPosition = 0;
var modalOpen = false;
var $reveal = $('.reveal');
$reveal.on('closeme.zf.reveal', function() {
  if(!modalOpen) scrollPosition = $(document).scrollTop();
});
$reveal.on('open.zf.reveal', function() {
  modalOpen = true;
});
$reveal.on('closed.zf.reveal', function() {
  $(document).scrollTop(scrollPosition);
  modalOpen = false;
});;
```

SassNinja:
I'm closing this as the current code in develop seems to be fixed: the scrollTop is preserved in my tests.

I only had a problem on iOS. But I've made a PR for this:
https://github.com/zurb/foundation-sites/pull/10786

PR #10546, opened 2017-08-16
Reveal mobileScrollTop option

SassNinja:
This PR adds a new option to the reveal plugin called 'mobileScrollTop'.
The purpose of this option is to provide a possibility for disabling the mobile scrollTop behavior that I've spotted in this issue: https://github.com/zurb/foundation-sites/issues/10543

As described in the issue report I still don't know why the scrollTop value gets set to 0 while the reveal is opened on a mobile device (identified by user agent instead of media query) at all.
But since this PR adds a new option to disable this behavior it definitely doesn't cause any issues at existing projects.

I'va also added a visual test so you can easily see the benefit of this option.
http://localhost:3000/reveal/mobile.html

@kball **would be great to get your feedback on this!**

If this behavior isn't necessary I would adjust my PR and simply remove it instead of adding a new option.

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -380,7 +380,7 @@ class Reveal extends Plugin {
           $('html, body').removeClass('is-reveal-open');
         }
         if(_this.originalScrollPos) {
-          $('body').scrollTop(_this.originalScrollPos);
+          $('html, body').scrollTop(_this.originalScrollPos);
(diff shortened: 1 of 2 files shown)


SassNinja:
As mentioned in #10543 the issue is only present for IOS with latest develop branch.
That has been fixed in #10637 so I close this PR.

PR #10550, opened 2017-08-17
Fix SASS Deprecations

### LINKED ISSUE #10534: xy-grid gutters deprecation warning when updating to 6.4.2
I installed foundation-sites 6.4.2 as a node module. 
I am using sass 3.5.1 and after uprading to 6.4.2 I get the following error:

    /usr/local/bin/sass` --no-cache --update base_foundation.scss:base_foundation.css

DEPRECATION WARNING on line 33 of ../../node_modules/foundation-sites/scss/xy-grid/_gutters.scss:
#{} interpolation near operators will be simplified in a future version of Sass.
To preserve the current behavior, use quotes:

  unquote("#{$operator}#{$gutter}")

You can use the sass-convert command to automatically fix most cases.

DEPRECATION WARNING on line 42 of ../../node_modules/foundation-sites/scss/xy-grid/_gutters.scss:
#{} interpolation near operators will be simplified in a future version of Sass.
To preserve the current behavior, use quotes:

  unquote("#{$operator}#{$gutter}")

You can use the sass-convert command to automatically fix most cases.

error ../../node_modules/foundation-sites/scss/xy-grid/_classes.scss (Line 415: Invalid CSS after "...grid-margin-y')": expected "}", was ".grid-frame.gri...")

Reverting to 6.4.1 makes the problem disappear.

jessedoyle:
Hi!

I recently noticed that `foundation-rails` was generating a deprecation warning with `sass` >= `3.4.20`.

Here's the warning I was noticing:

```
DEPRECATION WARNING on line 33 of /Users/me/src/app/vendor/bundle/ruby/2.4.0/gems/foundation-rails-6.4.1.2/vendor/assets/scss/xy-grid/_gutters.scss:
#{} interpolation near operators will be simplified in a future version of Sass.
To preserve the current behavior, use quotes:

  unquote("#{$operator}#{$gutter}")

You can use the sass-convert command to automatically fix most cases.
```

As suggested by the SASS compiler, I've squelched the deprecation by using the `unquote` function.

This fixes https://github.com/zurb/foundation-sites/issues/10534.

SEE: https://github.com/sass/sass/issues/1778

### DIFF
--- a/scss/xy-grid/_gutters.scss
+++ b/scss/xy-grid/_gutters.scss
@@ -30,7 +30,7 @@
 
       // Loop through each gutter position
       @each $value in $gutter-position {
-        #{$gutter-type}-#{$value}: #{$operator}$gutter;
+        #{$gutter-type}-#{$value}: unquote("#{$operator}#{$gutter}");
       }
     }
   }
@@ -39,7 +39,7 @@
 
     // Loop through each gutter position
     @each $value in $gutter-position {
-      #{$gutter-type}-#{$value}: #{$operator}$gutter;
+      #{$gutter-type}-#{$value}: unquote("#{$operator}#{$gutter}");
     }
   }
 }


kball:
@brettsmason can you review?

jessedoyle:
@brettsmason - Sorry - I don't intend to spam you! I just want to bump this so it doesn't get stale.

The deprecation warnings in our CI tests are becoming troublesome 😄

IamManchanda:
This looks good to me => http://sass-lang.com/documentation/Sass/Script/Functions.html#unquote-instance_method ... Poked up brett for his review... :) 
Dont worry @jessedoyle ;)

brettsmason:
@jessedoyle sorry for the delay! Haven't had much time for Foundation over the last few weeks.
So the Sass seems to compile fine still. I have no way of checking if it fixes your idea, but as you can confirm it does I'll go ahead and merge this.

jessedoyle:
Thanks @brettsmason!

ncoden:
For reference: 

> Sass will now emit a warning when a named color is used directly in interpolation (#{}) in a context where the hex representation would likely be invalid. This should help users avoid having their stylesheets break in compressed output mode.
> -- [Sass v3.4.0 changelog](https://github.com/sass/sass/blob/stable/doc-src/SASS_CHANGELOG.md#smaller-improvements)

ncoden:
For reference:

> Certain ways of using #{} without quotes in property and variable values have been deprecated in order to simplify the feature.
>
> Currently, #{} behaves unpredictably. If it's used near operators, it will cause those operators to become part of an unquoted string instead of having their normal meaning. This isn't an especially useful feature, and it makes it hard to reason about some code that includes #{}, so we're getting rid of it.
> 
> In the new world, #{} just returns an unquoted string that acts like any other unquoted string. For example, foo + #{$var} will now do the same thing as foo + $var, instead of doing the same thing as unquote("foo + #{$var}").
> 
> In order to ease the transition, Sass will now emit deprecation warnings for all uses of #{} that will change behavior in 4.0. We don't anticipate many warnings to appear in practice, and you can fix most of them automatically by running sass-convert --in-place on the affected files.
> 
> [Sass v3.4.20 changelog](https://github.com/sass/sass/blob/master/doc-src/SASS_CHANGELOG.md#3420-09-december-2015)

PR #10570, opened 2017-08-23
Reveal: Add the possibility to disable the vOffset

therufa:
I added the possibility to entirely disable the vOffset on the
component, so there is no need in the style layer to force-override it.


kball:
Looks good! Thanks @therufa and congrats on first merged PR!

PR #10575, opened 2017-08-24
Fix Progress Bar Example valuenow

wesinator:
Fix valuenow on 25% example to match the percent value.


kball:
Good catch, thanks!

PR #10582, opened 2017-08-26
Shouldnt we Exclude Node Modules folder

IamManchanda:
Taken from official babel docs 
https://github.com/babel/babel-loader#usage

> **babel-loader is slow!**
> 
> Make sure you are transforming as few files as possible. Because you are probably
> matching `/\.js$/`, you might be transforming the `node_modules` folder or other unwanted
> source.
> To exclude `node_modules`, see the `exclude` option in the `loaders` config as documented above.
> You can also speed up babel-loader by as much as 2x by using the `cacheDirectory` option.
> This will cache transformations to the filesystem.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -17,6 +17,7 @@ module.exports = {
       // Compiles ES6 and ES7 into ES5 code
       {
         test: /\.js$/,
+        exclude: /node_modules/,
         use: [
           {
             loader: 'babel-loader',


IamManchanda:
@kball @DaSchTour can anyone of you review this!

kball:
I don't think this actually will do anything, as the Foundation JS doesn't depend on anything from node modules directly within this build. Might make a difference in the template, but for the framework build I think it is irrelevant

PR #10589, opened 2017-08-28
[Feature Request] Update color regex pattern to support other values also

### LINKED ISSUE #10588: [Abide] Flags for color pattern?
I am regular expression these days and I am amazed why aren't we using flags 
See => [Regular Expressions - Advanced searching with flags](https://developer.mozilla.org/en/docs/Web/JavaScript/Guide/Regular_Expressions#Advanced_searching_with_flags) 

What i mean here is for example => [foundation.abide.js#L562](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.abide.js#L562)
The current regex is redundant => `/^#?([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$/`
The correct Regex would be => `/^#?([a-f0-9]{6}|[a-f0-9]{3})/i`

and secondly why aren't we using the `g` flag ... Is it intentional ?
cc @kball

IamManchanda:
Now its Hex (#F00 or #FF0000) || RGB || RGBA || HSL || HSLA
This also closes #10588 on merge

See => https://git.io/v5GkW
Credits where due: @lemnis | Awesome Lisa

cc @kball || @ncoden for review or merge!

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -559,7 +559,7 @@ Abide.defaults = {
     day_month_year : /^(0[1-9]|[12][0-9]|3[01])[- \/.](0[1-9]|1[012])[- \/.]\d{4}$/,
 
     // Hex (#FFF or #FFFFFF) || RGB || RGBA || HSL || HSLA
-    color : /^(#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$/i,
+    color : /^(#?([0-9a-f]{3}|[0-9a-f]{6})|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$/i,
 
     // Domain || URL
     website: {
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -558,8 +558,8 @@ Abide.defaults = {
     // DD/MM/YYYY
     day_month_year : /^(0[1-9]|[12][0-9]|3[01])[- \/.](0[1-9]|1[012])[- \/.]\d{4}$/,
 
-    // #FFF or #FFFFFF
-    color : /^#?([a-fA-F0-9]{6}|[a-fA-F0-9]{3})$/,
+    // Hex (#FFF or #FFFFFF) || RGB || RGBA || HSL || HSLA
+    color : /^(#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$/i,
 
     // Domain || URL
     website: {


IamManchanda:
That said @lemnis => Why #FF00FF00 ??

IamManchanda:
Ohk so FWIW ... I have updated @lemnis regex with my correction

The updated regex (updated code from lisa to mine) updates

- This add support for `f00`, `ff0000` 
- This remove support for `#ff00ff00` and `#ff00`

---

Here is the updated test (`g` and `m` flags in screenshot doesn't matter here)

regexp-color

IamManchanda:
Update: Dang there is still some corrections to be made... **DO NOT MERGE**

lemnis:
Added some extra explanations to the gist, but why would you check against hex values without the hashtag? I don't  see the reason, but I'm not that familiar with the source code of Foundation

IamManchanda:
@lemnis Yeah CSS4 Hex Notation is not supported right now ... correct me if I am wrong @kball 

> why would you check against hex values without the hashtag? I don't see the reason, but I'm not that familiar with the source code of Foundation

Funny but even i don't know ... if you see the old version its supporting non hash tag

----

But here is another question for you
Currently the tests are passing for all of these => https://git.io/v5ZH7
But is rgb(300, 0, 0) valid ... Is negative values valid ... Is percentages used in RGB
I am sorry I am a bit unsure here so please help me in the same ??

IamManchanda:
Oh Dang ... Just saw the w3 docs and they are supported (negative values) ... Stupid me never used it ... sorry for being so dump :p 

For reference check this comment by @lemnis => https://git.io/v5Z7k

@kball **This is good to merge**

lemnis:
Sometimes it's harder to start reading, then just code. Don't blame yourself 😉

IamManchanda:
@lemnis Yup sometimes ... FWIW thanks for the code! I think this is cool!

kball:
I'd love to see if we can get some of the tests into our own test suite - can you take a look at adding regex tests to `test/javascript/components/abide.js`? If you need help I suspect @Owlbertz can help, as he's done much of the unit test work.

IamManchanda:
Ohk kev will do it with @Owlbertz and get this in! Thanks 

That said @lemnis and @kball , what do you think of `i` flag 
 raised by @SassNinja here => https://git.io/v5nPR
Shall we use `i` flag or not in here ??

SassNinja:
@IamManchanda according to the CSS specs it's no problem to use the `i` flag here since the CSS syntax is case-insensitive.
So `RGB()` and `rGb()` work the same way `rgb()` does.

However, you've probably already noticed I'm a friend of being more strict regarding the case sensitivity :wink: (something as `rGb()` simply doesn't look right in my eyes).

https://www.w3.org/TR/CSS2/syndata.html#characters
https://stackoverflow.com/questions/17967371/are-property-values-in-css-case-sensitive

ncoden:
@SassNinja We're not talking about what we _promote_, but what we _support_. If we're expecting a CSS color, we should support any CSS color. Liking or not its syntax is not the question there.

SassNinja:
>  If we're expecting a CSS color, we should support any CSS color. Liking or not its syntax is not the question there.

So the rule is (as already mentioned by @IamManchanda [here](https://github.com/zurb/foundation-sites/pull/10593#issuecomment-325973760)):
If the browser supports it, we do it as well.

In this case the `i` flag is fine for me.

kball:
Going to remove the ready to merge until we get some unit tests on this

IamManchanda:
@kball Will do so this coming week!

---

@ncoden Thanks for support ;) 
@SassNinja Great!

kball:
@IamManchanda any update on unit tests? :D

IamManchanda:
@kball Sorry couldn't find much so only able to learn hello world testing on mocha and chai till now
https://github.com/IamManchanda/mocha-chai

Plans to do more learning this week or so, as enrolled in this free udacity course today too!
https://in.udacity.com/course/javascript-testing--ud549

@kball @Owlbertz Do you guys have any quick free articles on getting started on testing 
Any free video tutorials/articles etc?

DanielRuf:
So we still need some unit tests here. Who could help with providing some?

SassNinja:
I can take care of the tests as I'm familiar with mocha.

But apart from that I've checked the adjusted regex (again) an am wondering

1. why is `f00` considered as valid? `color:f00` is not valid CSS!
2. why is `#ff00FF00` considered as invalid or rather why is CSS4 hex notation excluded? browser support is quite good and it should be up to the user imo if he uses it https://caniuse.com/#search=hex
3. what about color name such as `red`? I mean `color:red` is definitely valid CSS but considered invalid by the regex

The original regex from @lemnis seems to be more suitable in me eyes than the modified one from @IamManchanda 

@ncoden @DanielRuf what do you think abt this?

ncoden:
IMHO all CSS colors should be supported and all invalid CSS colors should not be. The only problem is with CSS color names. There's like [150 colors](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value) and it seems to be a big waste to all support them manually. I would be okay with allowing any keyword (so `fff` too too) or no one.

SassNinja:
> I would be okay with allowing any keyword (so fff too) or no one (safer).

Imo we should go the safer way so any match is a valid color so sth as `xyz` is never true.

The original regex seems to fulfill this requirement

```regex
^(#[0-9a-f]{3}|#(?:[0-9a-f]{2}){2,4}|(rgb|hsl)a?\((-?\d+%?[,\s]+){2,3}\s*[\d\.]+%?\))$
```

abide-regex

ncoden:
@SassNinja What about:
```
rgb(0, 0, 0, 0%)
hsl(0, 0%, 0%, 0%)
```

SassNinja:
Good catch @ncoden !

(this would work in chrome & firefox btw but not in safari & IE)

The reason it matches is rgb, rgba, hsl and hsla are combined in one group to keep the regex more dry and simple.
To catch `rgb(0, 0, 0, 0%)` I would have to create 4 groups instead of one.

Apart from that have to even farer to catch mixed value (e.g. `rgb(255, 100%, 100%)` is invalid).
Precisely all of the following properties match the current regex but must be false:

```
rgb(0, 0, 0, 0%)
rgb(255, 100% 100%)
rgba(0, 0, 0, 0%)
hsl(0, 0%, 0%, 0%)
hsl(0%, 0%, 0%)
```

I'm in two minds whether to prioritize regex size or to prioritize a perfect validation (except of the color names).

What do you think?

DanielRuf:
> regex size or to prioritize a perfect validation

validation (more important) > size (less important) imo

ncoden:
> validation (more important) > size (less important) imo

And we're talking about something like 0.1ko in size.

SassNinja:
> And we're talking about something like 0.1ko in size. Maybe 0.5ko for a perfect RegExp.

I wasn't worried about the file size / execution time but the maintainability of the code.
However I've worked over the regex to (hopefully) catch all invalid values:

```regex
^(#[0-9a-f]{3}|#([0-9a-f]{2}){2,4}|rgb\((-?\d+\s*,\s*){2}(-?\d+\s*)\)|rgb\((-?\d+%\s*,\s*){2}(-?\d+%\s*)\)|rgba\((-?\d+\s*,\s*){3}(-?\d+(\.\d+)?\s*)\)|rgba\((-?\d+%\s*,\s*){3}(-?\d+(\.\d+)?\s*)\)|hsl\((-?\d+\s*,\s*)(-?\d+%\s*,\s*)(-?\d+%\s*)\)|hsla\((-?\d+\s*,\s*)(-?\d+%\s*,\s*){2}(-?\d+(\.\d+)?\s*)\))$
```

abide-regex-2

Review appreciated :)

One thing to be defined:
#rrggbbaa hex color notation is (of course) not supported by IE/Edge – should the regex it consider valid anyway? Or CSS3 only?

ncoden:
> #rrggbbaa hex color notation is (of course) not supported by IE/Edge – should the regex it consider valid anyway? Or CSS3 only?

We follow references and #rrggbbaa is a valid CSS(4) color. It's up to the end developer to care about the compatibility of the colors he can accept as input to display or transform.

PR #10593, opened 2017-08-28
Closes #10590: Refactor Regular expressions with `\w` and `i` flag!

### LINKED ISSUE #10590: [Abide] Most Regular expressions need to refactor
1. `i` flag for both cases .. so that code is not reduntant
2. `\w` instead of `a-zA-Z0-9` ... See => https://www.w3schools.com/JSREF/jsref_regexp_wordchar.asp ... as Less code helps shorten bugs, first rule of programming

Concerned Code => [foundation.abide.js#L533-L570](https://github.com/zurb/foundation-sites/blob/develop/js/foundation.abide.js#L533-L570)

IamManchanda:
This closes #10590 on merge, I have tested all of these changes i did at http://regexr.com/

### DIFF
--- a/js/foundation.abide.js
+++ b/js/foundation.abide.js
@@ -531,8 +531,8 @@ Abide.defaults = {
   validateOnBlur: false,
 
   patterns: {
-    alpha : /^[a-zA-Z]+$/,
-    alpha_numeric : /^[a-zA-Z0-9]+$/,
+    alpha : /^[a-z]+$/i,
+    alpha_numeric : /^\w+$/i,
     integer : /^[-+]?\d+$/,
     number : /^[-+]?\d*(?:[\.\,]\d+)?$/,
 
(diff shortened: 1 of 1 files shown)


IamManchanda:
@kball I have added Ready to merge coz i tested... but would request you to cross check and then merge :)

Deckluhm:
`\w` also matches `_` so you can't simply replace `[a-zA-Z0-9]` by `\w`.

SassNinja:
@Deckluhm good objection – the underscore of course prevents replacing `[a-zA-Z0-9]` with `\w`

@IamManchanda as I mentioned in https://github.com/zurb/foundation-sites/issues/10590 I don't recommend to overuse the `i` flag e.g. regarding the url pattern where users might not want `HTTP` or `hTTp` to pass the validation

IamManchanda:
> @IamManchanda as I mentioned in #10590 I don't recommend to overuse the i flag e.g. regarding the url pattern where users might not want HTTP or hTTp to pass the validation

@SassNinja I am not overusing it... If you see clearly ... I have not added `i` flag to every situation but only where uppercase and lowercase are being repeated.
That said on `hTTp` i would say that this line is only concerned with that matter => https://github.com/zurb/foundation-sites/pull/10593/files#diff-37998d120faf57b3e7580109c8e306e2L546

IamManchanda:
> > \w also matches _ so you can't simply replace [a-zA-Z0-9] by \w.
> 
> @Deckluhm good objection – the underscore of course prevents replacing [a-zA-Z0-9] with \w

Is there any `everything but not underscore` stuff ?

Deckluhm:
> Is there any everything but not underscore stuff ?
Not that I am aware of (but I'm definitely not a regex expert).

Deckluhm:
> Is there any everything but not underscore stuff ?

Not that I am aware of (but I'm definitely not a regex expert).

kball:
@IamManchanda Given you're already going to be looking into unit tests on Abide for #10589 lets add some unit tests here as well to make sure we're getting good testing on these as we do refactors. Thanks!

IamManchanda:
Ohk sure @kball I would love to 
... On `i` flag to be honest ... I don't agree with @SassNinja (Sorry Kai) because I think this is fine!

But I think @Deckluhm issue is valid ... we need to find a way to remove underscore from `\w` 

---

Will tag @lemnis , she might be able to help me on the regex issue as discussed :arrow_down:  

> `\w` also matches `_` so we can't simply replace [a-zA-Z0-9] by `\w`. So is there any way we can say `everything but not underscore` ??

SassNinja:
@IamManchanda my concerns regarding the `i` flag are based and the thought what happens with the submitted data after the validation (e.g. when saved in a database).
I'm just worrying some people expect the validation to fail if the user enters an url with 'hTTp' or a color with 'rGb()' – even though it works for the browser.

Talking about the `\w` issue:
You can say `everything but not underscore` because this would also include characters that are not included in `\w`.

What you can do is that `\w but not underscore` by replacing `[a-zA-Z0-9]` with `[^\W_]`.
But to be honest, I'm not sure if this improves the code readability... (may be a matter of taste)

IamManchanda:
> I'm just worrying some people expect the validation to fail if the user enters an url with 'hTTp' or a color with 'rGb()' – even though it works for the browser.

If browser allows it then should we!

IamManchanda:
> What you can do is say \w but not underscore by replacing `[a-zA-Z0-9]` with `[^\W_]`.
But to be honest, I'm not sure if this improves the code readability... (may be a matter of taste)

Maybe but I am quite sure that this minimizes the code redundancy!

kball:
FWIW I prefer the simple and explicit nature of [a-zA-Z0-9] over using fancy combos with negation.

IamManchanda:
@kball  [a-z0-9] with an `i` flag ?

SassNinja:
> If browser allows it then should we!

If that's the rule, I've no more objections to the `i` flag
(although I personally still prefer the more strict pattern 😉 )

JeremyEnglert:
@IamManchanda - what's the status of this?

DanielRuf:
It seems the last discussed changes are still missing (excluding underscore and so on).

SassNinja:
Seems there's no way to replace `a-zA-Z0-9` with `\w` without including `_` (except of using some negation which does not simply the code but does the opposite).

I think the only pattern change in this PR that is fine is 'alpha' (not sure if case insensitivity regarding 'url' matters).

I'm closing this PR as only changing 'alpha' does not justify the PR (consistency is more important in this case imo).

@IamManchanda feel free to reopen if you've got an good idea regarding the `\w` problem.

Issue #10579, opened 2017-08-25
Slider fails from float trailing zero on data-start


knabe:
#### How to reproduce this bug:

1. Add a trailing 0 to the data-start (6.0) 
2. Try to slide a slider
3. Slider breaks

#### What happened instead:
Slider breaks -- returns NaN

#### Browser(s) and Device(s) tested on:
Chrome & Safari desktop

#### Foundation Version(s) you are using:
6.3, 6.4

#### Test case link:
https://codepen.io/knabe/pen/MvBGQQ

#### Additional notes
I can only recreate with data-start

PR #10594, opened 2017-08-28
Fix slider data-start float bug 

knabe:
@IamManchanda  per your request

data-start float with a trailing 0 is typed as a string. This causes an undesired concatenation breaking the slider. 

[#10579](https://github.com/zurb/foundation-sites/issues/10579)

### DIFF
--- a/js/foundation.slider.js
+++ b/js/foundation.slider.js
@@ -153,7 +153,7 @@ class Slider extends Plugin {
       pctOfBar = this._logTransform(pctOfBar);
       break;
     }
-    var value = (this.options.end - this.options.start) * pctOfBar + this.options.start;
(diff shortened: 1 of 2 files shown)


IamManchanda (approved):

IamManchanda:
@kball This looks cool ( https://www.w3schools.com/jsref/jsref_parsefloat.asp ) 
Is this good to go and Can we merge this ?

kball:
OK, so from what I can tell the reason this shows up with start and not with end is because that '+' turns it into a string concatenation, while if you try to use '-' it will actually transform it to numbers. Innnteresting. Ok, good to go. Thanks for the PR!

knabe:
glad I could help!

PR #10597, opened 2017-08-29
Add missing param to scss doc for flex-grid-row

### LINKED ISSUE #10578: [Flex Grid] Documentation error for flex-grid-row
On the Sass reference for the flex grid there's a parameter missing on `flex-grid-row`: `$wrap` is the fifth parameter the mixin receives but it's not mentioned in the docs (which made me wonder why the gutter I was setting wasn't working 🙂)

kball:
Adds missing param to scss docs for flex-grid-row. Fixes #10578


IamManchanda (approved):
Thanks!

PR #10599, opened 2017-08-30
Add JSHint file and add support for esversion 6!

IamManchanda:
Have tested same on **sublime** but expected to work on all editors that are leveraging JS hint Plugin!


IamManchanda:
Other editor I am thinking that this is useful is the one and only GIthub's Atom https://github.com/AtomLinter/linter-jshint/issues/237#issuecomment-199614365

kball:
Seems reasonable

PR #10600, opened 2017-08-30
Set border-width variable on button-base mixin

marcelox:
When using a modified hollow button width (i.e. 2px instead of default 1px), 
non-hollow buttons could have the same border-width as set on $button-hollow-border-width

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -44,7 +44,8 @@ $button-radius: $global-radius !default;
 
 /// Border width for hollow outline buttons
 /// @type Number
-$button-hollow-border-width: 1px !default;
+$button-border-width: 1px !default;
+$button-hollow-border-width: $button-border-width;
 
 /// Sizes for buttons.
 /// @type Map
@@ -100,7 +101,7 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   }
 
   -webkit-appearance: none;
-  border: $button-hollow-border-width solid transparent;
+  border: $button-border-width solid transparent;
   border-radius: $button-radius;
   transition: $button-transition;
 
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -100,7 +100,7 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   }
 
   -webkit-appearance: none;
-  border: 1px solid transparent;
+  border: $button-hollow-border-width solid transparent;
   border-radius: $button-radius;
   transition: $button-transition;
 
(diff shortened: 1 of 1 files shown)


kball:
I think this makes sense. @rafibomb @iammanchanda any thoughts?

scss/components/_button.scss
```diff
@@ -100,7 +100,7 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   }
 
   -webkit-appearance: none;
-  border: 1px solid transparent;
+  border: $button-hollow-border-width solid transparent;
```
IamManchanda: Create a new variable `button-border-width` I guess ... This is not (only) for button hollow ??

IamManchanda (dismissed):

scss/components/_button.scss
IamManchanda: and also make `$button-hollow-border-width` equal to `$button-border-width`

marcelox: Do we need a button border width other than for styling hollow buttons?

IamManchanda: $button-hollow-border-width is not correct for non hollow buttons 

Create new variable `$button-border-width` and set to 1px
Edit `$button-hollow-border-width` to `$button-border-width` for avoiding code redundancy!

marcelox: okay!

scss/components/_button.scss
```diff
@@ -44,7 +44,8 @@ $button-radius: $global-radius !default;
 
 /// Border width for hollow outline buttons
 /// @type Number
-$button-hollow-border-width: 1px !default;
+$button-border-width: 1px !default;
+$button-hollow-border-width: $button-border-width;
```
kball: Can we make this a `!default` as well so that if someone has an older settings file that manipulated this, it will still work?

scss/components/_button.scss
```diff
@@ -100,7 +101,7 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
   }
 
   -webkit-appearance: none;
-  border: 1px solid transparent;
```
ncoden: We have enough problems with having to create a variable for every single property of each element, let's not begin to make a _part_ of each property customizable 😉.

Why not `border: $button-border;` ?

ncoden (changes requested):
Hello @marcelox. Sorry for the late review !
Poke @IamManchanda

ncoden (approved):
I made the required changes
@IamManchanda Can you review ?

ncoden:
Note: it would be even better to have a set of mixin to transform/add defaults/etc the `border` property and other shorthand property.
Something like:
```scss
$button-border: #fff !default;

//...
$-border-default: 1px solid transparent;
border: border-default($button-border, $-border-default); -> 1px solid #fff
```

joeworkman:

https://foundation.discourse.group/t/equalize-hollow-and-normal-buttons/2319/2

PR #10602, opened 2017-08-30
Fixes #9344

### LINKED ISSUE #9344: [Input Group] Flex-mode input group input height does not match non-input-group inputs
Foundation 6.2.4

I'm being pedantic...

An input-group input's height does not match up with non-input-group inputs when displayed side-by-side as in the demo. It's off by a few pixels.

**Demo:**

http://simshaun-static.s3-website-us-east-1.amazonaws.com/foundation/6.2.4/

**Example:**

image

Removing `height: auto` on `.input-group-field` solves the problem, but I don't know if that causes issues elsewhere or if it causes problems in non-flex mode.

simshaun:
Implements the newer input height formula from #9681 for input groups.

# Note 1:
I think it would be nice if the input height formula was moved to a single location so this issue can't reoccur in the future if the formula were to change again. The same formula is now in `_input-group.scss`, `_text.scss`, and `_select.scss`. However, I'm not sure of the best way to do this. Maybe a new file `scss/forms/_vars.scss` that `_input-group.scss`, `_text.scss`, and `_select.scss` can all use?

# Note 2:
I removed the hardcoded height on `.input-group-field` in non-flexbox-mode so that the input/select/textarea element would just inherit its normal height. There shouldn't be any reason to override the height on `.input-group-field` unless it's being used on a non input/select/textarea element. I guess this could be considered a breaking change in that scenario. Should I revert?

### DIFF
--- a/scss/forms/_input-group.scss
+++ b/scss/forms/_input-group.scss
@@ -22,7 +22,7 @@ $input-prefix-border: 1px solid $medium-gray !default;
 $input-prefix-padding: 1rem !default;
 
 @mixin foundation-form-prepostfix {
-  $height: ($input-font-size + $form-spacing * 1.5);
+  $height: ($input-font-size * $input-line-height) + (get-side($input-padding, 'top') + get-side($input-padding, 'bottom')) - rem-calc(1);;
 
   .input-group {
     display: if($global-flexbox, flex, table);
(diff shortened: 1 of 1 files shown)


kball:
@simshaun Thanks for putting this together! For note 1 I definitely agree, and would like to see that as part of this PR. Following the pattern of other components, this would probably be in `forms/_forms.scss` prior to the imports of the other pieces - that said, I think that is problematic because it makes it impossible to import just one piece. I like your idea of a `_vars.scss` file, which then could be imported wherever necessary... @ncoden @brettsmason @IamManchanda what do you think?

rafibomb:
Yes - Like the idea of a _vars.scss for the common things. Also like that you removed the height on the flex version as it's not needed (flex children inherit same height).

kball:
Going to merge this as a step forward, but would like to see us implement the `_vars.scss` approach going forward.

Issue #10470, opened 2017-07-27
BUGFIX for Magellan [ not working ]


h-e-r-b-e-r-t:
Even on [http://foundation.zurb.com/sites/docs/magellan.html](http://foundation.zurb.com/sites/docs/magellan.html) Magellan is not working.
That's because of an bug in .scss rsp. js: THE **is-active** CLASS IS ADDED TO **a-tag** BUT ALL SCSS IS FOR PARENT ELEMENT TO RECEIVE **is-active** CLASS!

https://github.com/zurb/foundation-sites
Foundation for Sites
Version 6.4.1
```
_bower_components/foundation-sites/scss$ grep -rni is-active .     
./components/_tabs.scss:141:  &.is-active {
./components/_dropdown-menu.scss:154:    .is-active > a {
./components/_menu.scss:41:/// Backward compatibility for menu state. If true, this duplicate `active` with `is-active`. 
./components/_menu.scss:430:    .is-active > a {
./components/_orbit.scss:72:    &.is-active {
./components/_orbit.scss:148:    &.is-active {
./components/_accordion.scss:93:  :last-child:not(.is-active) > & {
./components/_accordion.scss:112:    .is-active > &::before {
./components/_drilldown.scss:104:      &.is-active {
```

PLEASE FIX THIS BUG!

SassNinja:
@h-e-r-b-e-r-t thanks for reporting this issue! I've just checked it on the live docs page and can confirm it's not working as expected.

You've already spotted the reason: the menu component needs the is-active class on the li-tag and not on the a-tag. But since Magellan is not restricted to the menu component by design I think the best way would be to simply extend the menu so it considers an is-active on both, li-tag and a-tag.

@kball @IamManchanda what do you think?

IamManchanda:
Will dig it in tonight!

h-e-r-b-e-r-t:
@SassNinja, @kball, @IamManchanda

What's about adding '_is-active_' to a-tag and creating a new plugin-option '_data-parent_' for **jQuery .parent()**. If option is choosen, '_is-active_' is added to parent else not. 
If so, Magelan would be really universal, even for a-tags which are not wrapped.

IamManchanda:
That looks interesting !

kball:
@h-e-r-b-e-r-t Sounds like a reasonable compromise solution. Do you want to do a PR to add it?

h-e-r-b-e-r-t:
Hi,
[...]

PR #10614, opened 2017-09-05
Fixing magellan is-active issue  ||  adding traversing to magelan $active

h-e-r-b-e-r-t:
[https://github.com/zurb/foundation-sites/issues/10470#issuecomment-319499782](https://github.com/zurb/foundation-sites/issues/10470#issuecomment-319499782)

__menu.scss_ &&  __dropdown-menu.scss_:
Fixing: '_**is-active**_' class is added to a-tag but all scss is for parent element to receive  is-active class.
Making _**is-active**_ class general => see changes in magellan.js

_magellan.js_:
Adding option _**activeParents**_ for traversing $active element.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -165,8 +165,10 @@ class Magellan extends Plugin {
     }
 
     this.$active.removeClass(this.options.activeClass);
-    this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
-
+    this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').parents()[this.options.activeParents]
+  	               ? $( this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').parents()[this.options.activeParents] ).addClass(this.options.activeClass)
+  	               : this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
+    
     if(this.options.deepLinking){
       var hash = "";
       if(curIdx != undefined){
@@ -238,6 +240,13 @@ Magellan.defaults = {
    * @default 'is-active'
    */
   activeClass: 'is-active',
+    /**
+  * If >= 0 index of parents of active link, where activeClass is added/removed
+  * @option
+  * @type {number}
+  * @default -1
+  */
+ activeParents: -1,
   /**
    * Allows the script to manipulate the url of the current page, and if supported, alter the history.
    * @option
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -37,8 +37,17 @@ class Magellan extends Plugin {
   _init() {
     var id = this.$element[0].id || GetYoDigits(6, 'magellan');
     var _this = this;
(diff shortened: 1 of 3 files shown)


js/foundation.magellan.js
```diff
@@ -165,8 +165,10 @@ class Magellan extends Plugin {
     }
 
     this.$active.removeClass(this.options.activeClass);
-    this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
-
+    this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').parents()[this.options.activeParents]
+  	               ? $( this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').parents()[this.options.activeParents] ).addClass(this.options.activeClass)
+  	               : this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
+    
     if(this.options.deepLinking){
```
SassNinja: You shouldn't travel up the complete DOM tree twice!
I would rather write it this way and use jQuery's `eq()` so you don't have to recreate a jQuery object.

```javascript
this.$active = this.options.activeParents
                ? this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').parents().eq(this.options.activeParents).addClass(this.options.activeClass)
                : this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
```

js/foundation.magellan.js
```diff
@@ -238,6 +240,13 @@ Magellan.defaults = {
...
    */
   activeClass: 'is-active',
+    /**
+  * If >= 0 index of parents of active link, where activeClass is added/removed
+  * @option
+  * @type {number}
+  * @default -1
+  */
+ activeParents: -1,
   /**
```
SassNinja: If you use jQuery's `eq()` as suggested above, this has to be `null`default and e.g. `0` if you want the first parent.

scss/components/_dropdown-menu.scss
```diff
@@ -152,7 +152,7 @@ $dropdown-menu-item-background-active: transparent !default;
     }
 
     // Active state
-    .is-active > a {
+    .is-active {
       background: $dropdown-menu-item-background-active;
```
SassNinja: This breaks existing projects that rely on `is-active` to be on the links's parent instead of on the link itself.
Thus I wouldn't change this or extend it:
```sass
.is-active {
    &, > a {
```

SassNinja:
@h-e-r-b-e-r-t I've reviewed your code.

Apart from my comments I'm wondering if we really need an option travel up the DOM multiple levels?
I'd prefer to only offer one level, so rather a boolean (activeParent) than a number what lets you replace `parents().eq()` with `parent()`.

@kball @IamManchanda what do you think?

scss/components/_menu.scss
```diff
@@ -441,7 +441,7 @@ $menu-icons-back-compat: true !default;
     }
 
     // Active state
-    .is-active > a {
+    .is-active {
```
kball: This seems like a pretty big change... what is the thinking here?

DanielRuf:

kball:
Closing this PR as the feedback was never addressed or responded to.

PR #10629, opened 2017-09-08
issue-10628: Define variables before use in _breakpoint.scss

### LINKED ISSUE #10628: Breakpoint variables not scoped properly
In `scss/utils/_breakpoint.scss` ([see here](https://github.com/zurb/foundation-sites/blob/develop/scss/util/_breakpoint.scss#L326)) the variables like `$small-up` and `$large-down` are scoped only to the if statements in which they are defined. 

The solution for this problem is to define the variable before the if statement (as null or an empty string), and then allow the if statement to give it the value desired. 

This bug is blocking our organization from fully migrating off of Foundation 5

1. Try to use the `$small-up` variable

#### What should happen:

The code should compile

#### What happened instead:

`Undefined variable: "$small-up"`

6.4, the issue goes back at least to 6.3

gaulrobe:
Fixes #10628

When using the `$small-up`, etc variables we were getting `Undefined variable: "$small-up"` because those variables need to be defined outside the if-statements. Went with empty strings for parity with the data type that is expected for these variables.

This is rather urgent of a fix for our organization since we are in the process of upgrading from Foundation 5 to Foundation 6 and use these variables to create @media CSS.

Thanks for taking a look!

### DIFF
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -323,25 +323,39 @@ $breakpoint-classes: (small medium large) !default;
   }
 }
 
+$small-up: '';
+$small-only: '';
+
 @if map-has-key($breakpoints, small) {
   $small-up: screen;
   $small-only: unquote('screen and #{breakpoint(small only)}');
 }
+
+$medium-up: '';
+$medium-only: '';
 
 @if map-has-key($breakpoints, medium) {
   $medium-up: unquote('screen and #{breakpoint(medium)}');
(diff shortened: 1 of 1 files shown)


IamManchanda (approved):

IamManchanda:
I think this makes sense...

cc @kball for review...

kball:
Merged! Thanks @gaulrobe

shonin:
Thanks @kball! When can we hope for the next release?

PR #10630, opened 2017-09-08
Add xy styles to all docs

kball:
Since XY is orthogonal to the other grids, there should be no side effects (didn't see any in testing), and this lets us incrementally merge XY grid examples into every part of the docs.


kball:
oops should be for master

PR #10637, opened 2017-09-11
Linked hoverable dropdown toggle (forceFollow option)

SassNinja:
I'm sometimes using dropdown toggles that are hoverable and also clickable (direct link).
But this is a huge problem when it comes to touch devices (mostly mobile & tablet but not exclusive).

Thus I've made this PR that adds a new option called `forceFollow` that is inspired by the option from the dropdownMenu component.
If true (default) it supports using linked dropdown toggles that work as described above.

Precisely I've hooked into the click event listener and do a preventDefault if the dropdown is hoverable and not opened yet.
Without my PR the first click opens the dropdown but also immediately opens the link (href).

@kball @IamManchanda would you please review and let me know if it looks good for you?

I've also included a visual test but you can see it working in this codepen as well (make sure you're using a touch devices or touch emulation in dev tools):
https://codepen.io/KaiTheGreat/pen/PJYmXj

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -127,7 +127,9 @@ class Dropdown extends Positionable {
    * @private
    */
   _events() {
-    var _this = this;
+    var _this = this,
+        hasTouch = 'ontouchstart' in window || (typeof window.ontouchstart !== 'undefined');
+
     this.$element.on({
       'open.zf.trigger': this.open.bind(this),
       'close.zf.trigger': this.close.bind(this),
@@ -135,8 +137,16 @@ class Dropdown extends Positionable {
       'resizeme.zf.trigger': this._setPosition.bind(this)
     });
 
-    this.$anchors.off('click.zf.trigger')
-      .on('click.zf.trigger', function() { _this._setCurrentAnchor(this); });
+    this.$anchors.off('click.zf.trigger').on('click.zf.trigger', function(e) {
+        _this._setCurrentAnchor(this);
+
+        if (_this.options.forceFollow && hasTouch && _this.options.hover) {
+          var hasClicked = $(this).attr('data-is-click') === true;
+          if (hasClicked === false && _this.$element.attr('aria-hidden') === 'true') {
+            e.preventDefault();
+          }
+        }
+    });
 
     if(this.options.hover){
       this.$anchors.off('mouseenter.zf.dropdown mouseleave.zf.dropdown')
@@ -356,7 +366,6 @@ Dropdown.defaults = {
    * @default ''
    */
   positionClass: '',
-
   /**
    * Position of dropdown. Can be left, right, bottom, top, or auto.
    * @option
@@ -407,7 +416,14 @@ Dropdown.defaults = {
    * @type {boolean}
    * @default false
    */
-  closeOnClick: false
-}
+  closeOnClick: false,
+  /**
+   * Boolean to force overide the clicking of toggle link (href) to perform default action, on second touch event for mobile.
+   * @option
+   * @type {boolean}
+   * @default true
+   */
+  forceFollow: true
+};
 
 export {Dropdown};
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -137,7 +137,8 @@ class Dropdown extends Positionable {
       'resizeme.zf.trigger': this._setPosition.bind(this)
     });
 
-    this.$anchors.off('click.zf.trigger').on('click.zf.trigger', function(e) {
+    this.$anchors.off('click.zf.trigger')
+      .on('click.zf.trigger', function(e) {
         _this._setCurrentAnchor(this);
 
         if (_this.options.forceFollow && hasTouch && _this.options.hover) {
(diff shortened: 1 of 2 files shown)


IamManchanda:
Thanks digging it in!

IamManchanda (approved):
Superb @SassNinja

IamManchanda:
Kai this is wonderful! 
cc @kball for the merge!

IamManchanda:
> On desktop mouseover opens the dropdown pane and click opens the link's href.
On mobile the first touch opens the dropdown pane and the second touch (while still opened) opens the link's href.

I just loved it!

SassNinja:
Harry thanks for the kind words :blush: 

Would be great to see this merged so I can use it in several projects.

IamManchanda:
Sorry but I forgot, you need to update the docs too for the merge! @SassNinja

SassNinja:
@IamManchanda are you sure this is necessary?

This hover-touch feature is enabled by default so the user doesn't have to do anything after updating.
Some kind of documentation may be relevant if the user doesn't want this feature. But I think the option's description in the Plugin Options section already covers this.

So imo it's a little bit too much if I create a new section just for describing what's working out-of-the-box.
Or have I got you wrong?

vcarneiro:
This is awesome! Implemented it and it works like a charm!

SassNinja:
@IamManchanda what do you think? Ready for merge?

Would love to see this merged so I can use it in my projects without temporary adjusting the sources.

IamManchanda:
I think @ncoden will handle it ... I am bit busy and out of time these days!
Hope you can understand Kai? :)

js/foundation.dropdown.js
```diff
@@ -407,7 +417,14 @@ Dropdown.defaults = {
    * @type {boolean}
    * @default false
    */
-  closeOnClick: false
-}
+  closeOnClick: false,
+  /**
+   * Boolean to force overide the clicking of toggle link (href) to perform default action, on second touch event for mobile.
```
ncoden: I would rather say something like:
`If true, always follow the toggle link (or any default action) on click. Otherwise, wait for the second touch on touch devices`.

js/foundation.dropdown.js
```diff
@@ -136,7 +138,16 @@ class Dropdown extends Positionable {
     });
 
     this.$anchors.off('click.zf.trigger')
-      .on('click.zf.trigger', function() { _this._setCurrentAnchor(this); });
+      .on('click.zf.trigger', function(e) {
+        _this._setCurrentAnchor(this);
+
+        if (_this.options.forceFollow && hasTouch && _this.options.hover) {
+          var hasClicked = $(this).attr('data-is-click') === true;
+          if (hasClicked === false && _this.$element.attr('aria-hidden') === 'true') {
```
ncoden: `[aria-hidden]` is a side-effect. In the rest of the code we rely on `.is-open`.

js/foundation.dropdown.js
```diff
@@ -136,7 +138,16 @@ class Dropdown extends Positionable {
     });
 
     this.$anchors.off('click.zf.trigger')
-      .on('click.zf.trigger', function() { _this._setCurrentAnchor(this); });
+      .on('click.zf.trigger', function(e) {
+        _this._setCurrentAnchor(this);
+
+        if (_this.options.forceFollow && hasTouch && _this.options.hover) {
```
ncoden: Could you add some comments ?

ncoden (changes requested):
Hi. Here's my review.

@SassNinja What do you think ? And will you take care of the requested changes ?

ncoden:
I tested it. It works but there is one issue: once I opened the dropdown with a link... how do I close it ? I cannot toggle it (it would be a second tap and follow the link), and a "click away" is not listened.

SassNinja:
I'll adjust the code as requested in your review.
I took over some parts from the dropdownMenu component (comments, aria-hidden) but since it's a different component you're right, it could be clearer.

But @ncoden regarding the issue you've mentioned: I can't reproduce it in every browser / device...
Chrome dev tools (mobile emulation) & iOS work without problems (clicking outside the dropdown/pane closes it) but firefox dev tools have the problem you described – same for the usual hover dropdown (2nd button in my visual test `linked-dropdown-toggle.html`)

Would you tell me in which browser / OS you've had the issue in talk?

ncoden:
@SassNinja  I may have tested it on firefox dev tools. So I guess it's ok.

SassNinja:
As I said @ncoden the same issue has to occur for you when using a regular hover dropdown in 'firefox mobile'.
But although firefox mobile may not be really relevant I'm curious what causes this issue there and will try to find the reason.

However this PR is not related to this so from my side it's good.

I've just pushed another commit that considers your review (more / clearer comments).
Apart from that I've reworked the condition in the click event slightly so explicit `forceFollow:false` is also supported.

Would you please (re)review and merge if ok?

ncoden (approved):
👍

ncoden:
Since this PR introduce a new feature, I delay it to `v6.5.0`.

PR #10639, opened 2017-09-12
Fix bug in foundation.util.touch

taktes:
Hello! I found a bug with foundation Slider - it's works incorrectly on Firefox Mobile for Android - I can't move the slider and I'm getting an error in console that `event` variable is not defined.
After adding `event` param error was fixed and a slider works fine from my mobile browser finally.


kball:
Wow wonder how that snuck through. Nice catch!

PR #10648, opened 2017-09-14
Fix reset for y-grid to clear max-width as well

kball:
Fixes issue where grid-y cells are still constrained by the width set by grid-x cells.

Fixes 
xy grid foundation for sites 6 docs

To 
xy grid foundation for sites 6 docs-1

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -66,6 +66,7 @@
 @mixin xy-cell-reset($vertical: true) {
   $direction: if($vertical == true, width, height);
   #{$direction}: auto;
+  max-#{$direction}: none;
 }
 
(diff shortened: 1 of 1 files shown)


rafibomb:
I see the docs does that but not in a 6.4.3 codepen https://codepen.io/rafibomb/pen/yzgXZm
In the CSS it looks like width is reset properly
screen shot 2017-09-25 at 2 19 34 pm

kball:
Digging into this further, the situation occurs when you have *both* the xy grid *and* the traditional flex-grid in the page. The XY grid is much more careful about specificity than the traditional grid... that said, this *is* intended to be a supported case, so i'd like the reset to work.

rafibomb:
That's what it is - makes sense!

Issue #10457, opened 2017-07-24
XY grid - grid frame, full height issue


brgrz:
There's an issue with .grid-frame and .cell-block-y which don't get height: 100% (they only have the max-height: 100% set), which kinda defeats the point of a grid frame/app frame. The issue is that if the content within .cell-block-y doesn't overflow then it never hits the 100% height, which, if you have background applied results in columns that are too short and gives weird look.

The issue can be demonstrated if you delete some paragraphs from the official docs grid frame example. I'm attaching the photo for you to visualize what I mean:

![xy-grid-frame-height](https://user-images.githubusercontent.com/193816/28514399-21450b1c-705a-11e7-997b-89ab6d57df6d.png)

Imho most users will require those columns to take up 100%, especially when background is applied.

brgrz:
I don't know if it's just me but I'm having huge issues with nesting those XY grids and keeping everything in line, especially full height but not having overflows cut off

see https://github.com/zurb/foundation-sites/issues/10571, it has a lot in common with this issue

brgrz:
It's still required to manually put `height: 100%` to the `.cell-block-y` cells when you want them to be 100% height when there's not enough content in the cells. So this issue remains open even though I closed the related issue.

kball:
Hmm... I think you're probably right that cell blocks should naturally expand to get to 100%, but would love to get some more opinions. What do you think @oxyc @JeremyEnglert @brettsmason @rafibomb?

oxyc:
I actually subscribed to this issue as it hit me as well and I thought it was a bug.

rafibomb:
Yeah - that's right, first child nested .medium-cell-block-container > .grid-x - merged PR!

DanielRuf:
I guess https://github.com/zurb/foundation-sites/issues/10766 is about the same.

DanielRuf:
> Yeah - that's right, first child nested .medium-cell-block-container > .grid-x - merged PR!

This did not really fix anything as far as we can see.

joeworkman:
This issue has been mentioned on **Foundation Open Source Community**. There might be relevant details there:

https://foundation.discourse.group/t/challenge-getting-a-full-height-sidebar/2391/3

DanielRuf:
This is why I mentioned / linked it @davewsmith.

davewsmith:
Adding that a manual style of `height: 100%` does not achieve the desired result. `height: 100vw` does.

[...]

PR #10649, opened 2017-09-14
Add height pinning to cell-block-y

kball:
Fixes https://github.com/zurb/foundation-sites/issues/10457


IamManchanda (approved):
LGTM

kball:
This introduced a bug in Safari, turned up in QA. I believe using min-height:100% will have the same impact, and does not appear to introduce the same bug. Any reasons I'm missing why height pinning using min-height wouldn't work? (CC @oxyc @brgrz as the folks who were involved in #10457)

PR #10652, opened 2017-09-15
Fix for .grid-y.align-justify -> .cell .cell.auto .cell configuration

DaSchTour:
Given the following DOM
```lang:html
<div class="grid-y align-justify" style="min-height: 100vh">
  <div class="cell"></div>
  <div class="cell auto"></div>
  <div class="cell"></div>
</div>
```
The current setup with leads to overlapping of the middle cell.

Seams to be a Chrome Bug. Codepen Example (shrink the window so that the middle content doesn't fit completely): https://s.codepen.io/DaSch/debug/OxVxaP/RBMOJexjGzgk

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -77,6 +77,9 @@
   }
   @elseif ($size == 'auto') {
     #{$direction}: auto;
+    @if ($vertical) {
+      flex-basis: auto;
+    }
     $val: if($margin-gutter == 0, 100%, calc(100% - #{rem-calc($margin-gutter)}));
   }
   @elseif ($size == 'shrink') {


rafibomb:
I can't get to your codepen anymore to see the issue. I set one up here: https://codepen.io/rafibomb/pen/rGjwLg

What is the use-case of a .cell without .shrink or .auto? What is the expected behavior?

DaSchTour:
@rafibomb Well I didn't think that much about the use-case but just saw that this combination doesn't work and could simply be fixed

rafibomb:
Ok - so a sensible default makes sense to me. In the old flex grid, a .column was set to auto unless a size class is used. 

Let's make the case of a cell without an .auto class do flex basis auto for all orientations.

@brettsmason Can you buy into this?

kball:
Expected behavior for an XY grid cell without auto or shrink for width is 100%... changing it to flex basis auto would break a lot of things.

That said, it seems a little weird to do that for Y. Wrapping behavior in Y doesn't feel intuitive at all...
@brettsmason would love your thoughts on this

rafibomb:
@kball Yes - flex: 1 0 0; on a .cell in the X direction, right?

kball:
No - `1 0 0` is auto grow... we use the equivalent of `0 0 100%` to be full width, but to get things working across IE etc we actually use `0 0 auto` and `width: 100%`

brettsmason:
@kball I agree with this:
> Expected behavior for an XY grid cell without auto or shrink for width is 100%

We spent a lot of time discussing this when working on the grid, and IMO the default should definately stay this way - it made no sense to me having `auto` behaviour cells in small screen when generally people will likely have 100% width in more cases.

rafibomb:
Thanks for the explanation! So for this PR - are we aligned?

kball:
So... I set up another pen to show this, and applied what I think this fix is doing, and I still see the bug: https://codepen.io/kball/pen/xXXmBK

I think there are 2 core questions here.

1. What height should a `cell` with no modifiers be in a `grid-y`? If it were to be consistent with `grid-x`, it would be full height; as it stands it is minimal height. This stems from the difference in the ways `x` and `y` are treated by browser, but I'm not convinced it is correct.

2. What should wrapping behavior be for a `grid-y`? As it stands, `grid-y` is `column no-wrap`, but that inconsistency creates confusing behavior when you start talking about full cells and the like.

rafibomb:
I thought the questions is if _no cells have auto or shrink_. That's where things get weird. If a vertical grid has only .cells, the expected behavior seems like it should be like this https://codepen.io/rafibomb/pen/EwwrPx

rafibomb:
To make a vertical grid wrap, perhaps that should be an explicit class.

DaSchTour:
Guess this isn't relevant anymore

PR #10653, opened 2017-09-15
Add button expanded class for each breakpoint

DaSchTour:
Follow up to #7575 closing #7570

### DIFF
--- a/docs/pages/button.md
+++ b/docs/pages/button.md
@@ -39,7 +39,7 @@ Additional classes can be added to your button to change its size and shape.
 <a class="button" href="#">So Basic</a>
 <a class="large button" href="#">So Large</a>
 <a class="expanded button" href="#">Such Expand</a>
-<a class="small expanded button" href="#">Wow, Small Expand</a>
+<a class="small expanded-for-large button" href="#">Wow, Small + Expand for Large</a>
 ```
 
 ---
--- a/docs/pages/button.md
+++ b/docs/pages/button.md
@@ -59,7 +59,9 @@ Additional classes can be added to your button to change its size and shape.
 <a class="button large" href="#">So Large</a>
 <a class="button expanded" href="#">Such Expand</a>
 <a class="button small expanded" href="#">Wow, Small Expand</a>
-<a class="button small expanded-for-large" href="#">Wow, Small + Expand for Large</a>
+<a class="button small expanded-small" href="#">Wow, Expand only on small viewport</a>
+<a class="button small expanded-medium-down" href="#">Expand on medium and smaller</a>
+<a class="button small expanded-medium-up" href="#">Expand on medium and larger</a>
 ```
 
 ---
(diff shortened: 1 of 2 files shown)


IamManchanda:
@DaSchTour  as I said before

```
$responsive-expanded-button: true; // maybe `false` by default
```

Will also poke @kball

DaSchTour:
@IamManchanda well, good we talked about this. I checked the generated CSS and saw that my usecase (expanded buttons on small viewports) wasn't covered. So I improved the and added -up and -down modifiers.

IamManchanda:
@kball What do you say ... True or false?

scss/components/_button.scss
```diff
@@ -271,6 +275,29 @@ $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !defau
...
+        }
+        @if $size != $-zf-zero-breakpoint {
+          @include breakpoint(#{$size} down) {
+            &.expanded-for-#{$size}-down {
+              @include button-expand;
+            }
+          }
+
+          @include breakpoint(#{$size}) {
+            &.expanded-fpr-#{$size}-up {
```
IamManchanda: fpr

IamManchanda (changes requested):

scss/components/_button.scss
```diff
@@ -80,6 +80,10 @@ $button-hollow-hover-lightness: -50% !default;
 /// @type List
 $button-transition: background-color 0.25s ease-out, color 0.25s ease-out !default;
 
+/// Additional responsive classes for .expanded
+/// @type Boolean
+$responsive-expanded-button: true !default;
```
IamManchanda: @kball true or false

docs/pages/button.md
```diff
@@ -59,6 +59,7 @@ Additional classes can be added to your button to change its size and shape.
 <a class="button large" href="#">So Large</a>
 <a class="button expanded" href="#">Such Expand</a>
 <a class="button small expanded" href="#">Wow, Small Expand</a>
+<a class="button small expanded-for-large" href="#">Wow, Small + Expand for Large</a>
```
rafibomb: In terms of class naming - let's match our current naming scheme for the grid. 
large-expanded, medium-expanded etc https://github.com/zurb/foundation-sites/blob/develop/scss/grid/_flex-grid.scss#L216

scss/components/_button.scss
rafibomb: I say false

Also - let's rename the variable so it matches the others
$button-responsive-expanded

docs/pages/button.md
IamManchanda: Yup `large-expanded`

rafibomb:
After these 2 changes - looks good to merge! @DaSchTour are you able to make the updates so we can merge today?

DaSchTour:
@rafibomb done 😄

rafibomb:
One question came up while testing this. If we set `$button-responsive-expanded: false !default;` to false then how do CSS download users get this feature?

rafibomb:
In the meantime I updated the docs for this.

DaSchTour:
I guess CSS download users are and will be more and more limited in the number of features the get. In the long term I guess there will be more such "hidden" features. The only way so solve this would be to create CSS for every component with different settings. This way it would also be possible to only include certain components, just like in JavaScript.

Issue #10544, opened 2017-08-16
Switch to Popper.js for tooltips/dropdowns positioning?


FezVrasta:
I see you have a lot of logic to handle the tooltips and popovers position, but still the result is far from optimal.  
For instance, I see the tooltips won't flip to stay into viewport area if they overflow it.

Have you considered the idea to drop all your custom logic and rely on something like Popper.js?

Bootstrap v4 makes use of it and everything works nicely.

Project page: https://popper.js.org
Repo: https://github.com/FezVrasta/popper.js

IamManchanda:
Ah we like to use our own code for this type of stuff 

See 
Thread => https://github.com/zurb/foundation-sites/issues/9989
Inspiration => https://github.com/zurb/foundation-sites/issues/9799
PR => https://github.com/zurb/foundation-sites/pull/10053

IamManchanda:
> Have you considered the idea to drop all your custom logic and rely on something like Popper.js?

cc @kball but if you ask me I dont think so...

IamManchanda:
Btw just saw that you are the creator of the popper framework... So what are the benefits of Popper over current [JS](http://foundation.zurb.com/sites/docs/dropdown.html#explicit-positioning) code of ours @FezVrasta ??

FezVrasta:
I already provided a little explanation on the initial post. I'm going to write a comprensive list tomorrow is it's not a problem for you. It's a bit late here right now

IamManchanda:
No issye i will be waiting!

FezVrasta:
So, one of the problems I see is that the popovers/dropdowns/tooltips (poppers) don't attempt to appear (and stay) in the viewport, as you can see from this gif, they just appear in a non visible portion of the page.

![1](https://user-images.githubusercontent.com/5382443/29409053-dd4e00b8-834a-11e7-8dce-ef165365263f.gif)

I don't exactly know what's happening here, but looks like your algorithm have some problems detecting the correct place where to place the dropdowns:

![2](https://user-images.githubusercontent.com/5382443/29409151-411a191a-834b-11e7-9de0-1889dc34ffd8.gif)

Form my tests, I see that your algorithm doesn't support scrolling containers:
![3](https://user-images.githubusercontent.com/5382443/29409236-9f829914-834b-11e7-8c5b-d84faf6769c3.gif)

And last but not least, the PR that introduced the smart positioning in foundation added ~1500 rows of code you have to test and maintain.

[...]

PR #10672, opened 2017-09-22
[WIP] Transition to Typescript

DaSchTour:
Just to track progress here. This will be a lot of work, but just from starting I see a lot of possible problems in the current JS codebase.

1. There are internal functions not described correctly.
2. The Plugin class isn't used correctly by all plugins

But that said it's worth the effort to move to TS to find all this hidden possible problems.


IamManchanda:
Nice... Happy to contribute 
But cc @kball fot this feature request 

FWIW I support Daniel on this PR

IamManchanda:
FWIW, I will repeat my comment here

> Thanks to the type system, TypeScript helps a lot in minimizing bugs and unit tests.
> and, yeah this is not just an angular thing
> 
> - https://angular.io/guide/quickstart
> - https://www.typescriptlang.org/docs/handbook/react-&-webpack.html
> - https://vuejs.org/v2/guide/typescript.html
> 
> { PS: I haven't tried angular yet, and I am not learning TypeScript for learning angular but for being better in JavaScript }
> 

---

Also, Vue.js is also giving [preference](https://medium.com/the-vue-point/upcoming-typescript-changes-in-vue-2-5-e9bd7e2ecf08) to TypeScript now.
Also, in their roadmap

> **On the Roadmap: TypeScript Support in vue-cli**
After 2.5, we are planning to introduce official support for TypeScript in the next version of vue-cli in order to make it easier for TS+Vue users to kick off new projects. Stay tuned!

kball:
Woah! There's a ton of work in here @DaSchTour, nice work!

I think this is a very interesting direction, and as you point out opens up all sorts of better static analysis and will help us find more things. 

So long as we are able to maintain a straightforward migration path for the various ways folks are using older versions of the framework, I'm game.

IamManchanda:
@kball @DaSchTour ... what's the roadmap on this?
6.5? 7.0?

DaSchTour:
@IamManchanda Depends on how fast I'll get the work done and if there will be a complete rewrite for 7.0

IamManchanda:
Moving to a different language shouldn't be done in the same version series.
Yes typescript is better undoubtedly and should be used but not before **v7**

FWIW... I am adding a label @ v7

IamManchanda:
Also, for v7 rewrite, you should be considering this PR too #10544

js/entries/foundation-plugins.d.ts
```diff
@@ -0,0 +1,21 @@
+export { Abide } from '../foundation.abide';
```
IamManchanda: > js/entries/foundation-plugins.d.ts

Shouldn't the new folder be ts

DaSchTour: for compatibility I think we should not rename everything

Owlbertz:
@DaSchTour I would consider myself rather experienced with TS. If you want to delegate something, feel free to get in touch with me.

DaSchTour:
@Owlbertz I'll possibly won't have time to do anything on this topic till and and of this year. So you could try to go on with this as you like. Fix the TS errors that are currently poping up. Add the typescript builder to webpack/gulp. And I guess backporting of JS changes from other PRs also needs to be done.

IamManchanda:
One fun (little unrelated) stuff,
JavaScript wrapped with **Typescript** not only let you do the web
but also make you at good place to learn ahead cross platform **modern** mobile development!

Here are quick links I found out:

- [Swift is like TypeScript](https://alhazmy13.github.io/swift-is-like-typescript/
- )
- [Kotlin is like TypeScript](https://gi-no.github.io/kotlin-is-like-typescript/)
- [Swift is like Kotlin](http://nilhcem.com/swift-is-like-kotlin/)

So it's a good language that should be learn today if you are doing JavaScript

js/entries/foundation-plugins.d.ts
IamManchanda: Yeah same like how kotlin does with java
Gotcha ... Agree!

IamManchanda:
V7 requests are being closed for the time being. We have them labeled so we can readdress them in the future.

DaSchTour:
Well. I don't know why this was labeled V7? But that was never meant to be only for V7.
But as I don't have time for this maybe it's better to not do TypeScript at Foundation. We already stoped using Foundation JS for new projects.

Issue #10125, opened 2017-06-03
In RTL mode align-right is not set correctly


mozaffar:
When you generate a customized css in RTL mode the value for class ".align-right" is not set correctly:

Currently:
`.align-right {
     -ms-flex-pack: end;
      justify-content: flex-end;
}`

and It should be:
`.align-right {
     -ms-flex-pack: start;
      justify-content: flex-start;
}`

Because the content should be started from right and be stick to the right side of the container ;)

IamManchanda:
cc @abdullahsalem

abdullahsalem:
Thanks @mozaffar for reporting this issue, and @IamManchanda for the mention.

rafibomb:
Closing in favor of #10676

PR #10676, opened 2017-09-25
Fixed flex horizontal alignment. Issue #10125

imabdulsal:
This PR fixes this issue #10125

The variable `$-zf-flex-justify` in `scss/util/_flex.scss` contains static values for the two properties **left** and **right**, but they should be set depending on the current direction context that is stored in `$global-text-direction`. 
Technically, we cannot use the variable `$global-text-direction` in any Scss utility because `$global-text-direction` lives in `_global.scss` which is called after `_util.scss`, therefore I changed that variable -`$-zf-flex-justify`- to a function (`zf-flex-justify($text-direction)`) that takes the direction as an argument, as well as I defined an internal global variable with the origin name `$-zf-flex-justify`  in `_global.scss`to void changing in any Scss component that was using that variable name.

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -121,6 +121,9 @@ $global-button-cursor: auto !default;
 $global-left: if($global-text-direction == rtl, right, left);
 $global-right: if($global-text-direction == rtl, left, right);
 
+// Internal variable that contains the flex justifying options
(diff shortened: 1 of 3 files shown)


IamManchanda (approved):
LGTM
cc @kball

PR #10678, opened 2017-09-25
update with minor fix for manual installation docs

asecondwill:


rafibomb:
I think this doesn't apply anymore. Bower is still supported but it's not a dependency of the build system and not explicitely used in it. I don't see a bower.json file in the ZURB Stack that it would point to doing a bower i https://github.com/zurb/foundation-zurb-template/blob/master/package.json#L12

rafibomb:
@kball confirmed? No need for bower install?

kball:
ZURB template no longer uses bower but I didn't make the same switchover in the basic template...  probably should do that instead of this change.

rafibomb:
PR open!

Issue #9038, opened 2016-07-20
[Media Object] Mixins not functioning in flex mode


marcusmoore:
**How can we reproduce this bug?**

_Assuming we are compiling our assets_
1. Enable flex mode via instructions [here](http://foundation.zurb.com/sites/docs/flexbox.html)
2. Create a new component using the SASS mixins (`media-object-container` and `media-object-section()`)
3. Create page with an example code from [docs](http://foundation.zurb.com/sites/docs/media-object.html) and created component

**What did you expect to happen?**
The output from the example code and the component code would be the same.

**What happened instead?**
The required `.main-section` is overridden by the compiled media object mixin.

The 

**Test case:**
Here is a [gist](https://gist.github.com/marcusmoore/afca0fa78ee1c6482b1ae2854c1d90da) with my code and some screenshots demonstrating the issue:
## Desktop

![screenshot 2016-07-20 13 05 26](https://cloud.githubusercontent.com/assets/1141514/17001391/acc311ae-4e7b-11e6-80c7-81a6c1f77e9e.png)
## Mobile

![screenshot 2016-07-20 13 05 59](https://cloud.githubusercontent.com/assets/1141514/17001392/acc4a33e-4e7b-11e6-957f-bd3a4d808fce.png)

marcusmoore:
I may have fixed part of the problem by adding `flex: 1 1 0px;` manually:

```
.MediaObject {
  @include media-object-container;


  &__section{
    @include media-object-section();

    &--main{
      // Pulled from _media_object.scss
      @if $global-flexbox {
          flex: 1 1 0px;
      }
    }
  }
}
```

rafibomb:
I don't think there is any flex mode set up for this component. Should be straight forward to add it though.

kball:
@rafibomb I see you added help wanted so going to assume assigning it to yourself doesn't mean you don't want someone to take it... @marcusmoore would you like to take a crack at a PR to address this?

marcusmoore:
@kball Unfortunately I don't have the time right now. I can take a look in a few weeks if it's still open but I can't commit to it right now.

Martskin:
I took a swing at refactoring the Media Object mixins here: https://github.com/zurb/foundation-sites/pull/10679

brettsmason:
This looks like the PR addressing this was successfully merged into core so closing this.

PR #10679, opened 2017-09-25
Addresses Media Object mixin issue in Flex mode. #9038.

Martskin:
I refactored the Media Object mixins to try and address #9038.

### DIFF
--- a/scss/components/_media-object.scss
+++ b/scss/components/_media-object.scss
@@ -26,6 +26,18 @@ $mediaobject-image-width-stacked: 100% !default;
   @if $global-flexbox {
     flex-wrap: nowrap;
   }
+
+  img {
+    max-width: none;
+  }
+
+  @if $global-flexbox {
+    &.stack-for-#{$-zf-zero-breakpoint} {
+      @include breakpoint($-zf-zero-breakpoint only) {
+        flex-wrap: wrap;
+      }
+    }
+  }
 }
 
 /// Adds styles for sections within a media object.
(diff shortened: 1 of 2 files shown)


kball:
This makes sense to me, thanks @Martskin!

Would you be able to add a visual test case for this to confirm the behavior & make sure we don't create any regressions? There's information on how to create a visual test here: https://github.com/zurb/foundation-sites/wiki/Testing-Guide#visual-regression-tests

Martskin:
@kball 

No problem! I added a visual test page named `media-object-mixins-flex-mode.html`. Lemme know if you want anything else changed or added.

Also, I think this line in the Visual Regression Tests guide may not be necessary anymore?
>If you need to test flexbox mode, change the stylesheet in the <link> tag to foundation-flex.css

scss/components/_media-object.scss
```diff
@@ -50,6 +62,27 @@ $mediaobject-image-width-stacked: 100% !default;
   > :last-child {
     margin-bottom: 0;
   }
+
+  .stack-for-#{$-zf-zero-breakpoint} & {
```
ncoden: I agree the code would be more cleaner with this in `media-object-selection`. But sadly, `.stack-for-medium .media-object-section` is not the same as `.media-object.stack-for-medium .media-object-section`, bugs can be introduced because `.stack-for-*` is used for other components.

With this selector, if any parent use `stack-for-` for its own purpose, this component will be affected. So i'm moving this to `@mixin media-object-container` to ensure backward compatibility.

(I don't like it neither, it prevent to reduce the specificity 😢  ).

ncoden (approved):

PR #10681, opened 2017-09-26
Fixed javascript source folder

grantcarthew:
Just fixing a doc issue.

### DIFF
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -34,7 +34,7 @@ When configuring Babel to work with Foundation, you only need to include the `es
 
 ### File Structure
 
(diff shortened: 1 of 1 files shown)


rafibomb:
Should be src/assets/js I think. Can you update?

grantcarthew:
The module directory I have in my `node_modules` does not have src at all. The top level directories are `_vendor`, `assets`, `customizer`, `dist`, `js`, and `scss`. This is v6.4.3 of foundation-sites.

kball:
👍

kball:
Merged! Thanks @grantcarthew

Issue #10226, opened 2017-06-21
Incorrect usage for the Cite element


glennphilp:
#### How to reproduce this bug:

When using a blockquote or q element the developer should include a cite attribute or cite element. However, the current usage for [cite](http://foundation.zurb.com/sites/docs/typography-base.html#blockquotes) by Foundation is incorrect.

#### What should happen:

According to: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/cite) and [HTML 5 Doctor](http://html5doctor.com/element-index/#cite) the cite should be:

> A creative work may include a book, a paper, an essay, a poem, a score, a song, a script, a film, a TV show, a game, a sculpture, a painting, a theater production, a play, an opera, a musical, an exhibition, a legal case report, a computer program, a web site, a web page, a blog post or comment, a forum post or comment, a tweet, a written or oral statement, etc.

#### What happened instead:

Foundation is currently using a person's name.

#### Recommendation for fixing issue:

Person name:
`<span class="cite">Person Name</span>`
Citing a piece of work:
`<cite>[Book Title|Movie Title...]</cite>`

AndrewKvalheim:
Seems consistent with the [specification][]; its first example is:

```html
<p>In the words of <cite>Charles Bukowski</cite> -  
<q>An intellectual says a simple thing in a hard way. An artist says a hard thing in a simple way.</q></p>
```

Presumably the person's name serves as a proxy to the "written or oral statement" being cited.

---

That example highlights another issue, though. Foundation naively styles all citations as blocks, interfering with simple inline usage. That behavior should probably be scoped to block quotes.

[specification]: https://www.w3.org/TR/html5/text-level-semantics.html#the-cite-element

kball:
I think @AndrewKvalheim you are probably correct that cite should probably not be a block element. Looking to see when that was introduced, it appears to have been pulled over from Foundation 5.  Wonder if any of @brettsmason @gakimball @andycochran or other old guard folks know why that was the case, or have opinions on this subject?

plweil:
+1!

"That example highlights another issue, though. Foundation naively styles all citations as blocks, interfering with simple inline usage. That behavior should probably be scoped to block quotes."

glennphilp:
Although the W3C shows examples of using a person's name, they contradict themselves in the opening two statements.

[...]

PR #10687, opened 2017-09-29
Allow inline `<cite>` elements

jmsfwk:
Move the block styling and reduced font size on `<cite>` to only
apply when it is a child of `<blockquote>`.

Fixes issue #10226

### DIFF
--- a/docs/pages/typography-base.md
+++ b/docs/pages/typography-base.md
@@ -261,6 +261,10 @@ A definition list (`<dl>`) is used to display name-value pairs, like metadata or
 
 Sometimes other people say smart things, and you may want to mention those things with a quote. We've got you covered.
 
+<div class="callout">
+  By default, `<cite>` takes the look of the `.cite-block` component. In Sass, you can customize it with <a href="#sass-variables">`$cite-*` variables</a> or disable it by setting `$enable-cite-block` to false.
+</div>
+
 <div class="docs-codepen-container">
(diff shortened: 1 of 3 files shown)


kball:
This makes sense to me, but is a breaking change... can we wrap up the block form of citation into a mixin so that if someone wants to re-instate the old behavior it is as simple as 
```
cite {
  @include block-citation();
}
```

ncoden:
Hey @jmsfwk. Thank you, I see you made the requested changes, and sorry for the late reply. I think this PR goes to the right direction, which is trying to not impose any design and simply providing a set of easily customizable tools.

It's sad I was not active when you opened this PR, I would recommand a more "component" approach rather than "allowing a old-behavior reset". Anyway, I'll take care of the changes and merge this right after.

Sorry again for the late reply.

ncoden:
@jmsfwk For your future pull requests, you can create a dedicated branch (like feat/optional-block-cite) and create a pull request from it. So you can continue working on your branch develop while this PR is open.

ncoden:
I pushed.
Now we have a `.cite-block` component, generated from its `cite-block()` mixin, and used as default (or not) for the `<cite>` tag according to the `$enable-cite-block` property.

So no migration is needed anymore, and we're slowly moving to a "component" approach for typography.

ncoden (approved):

PR #10689, opened 2017-10-01
Add grow to xy-cell-base

flatline-studios:
Add the ability to use the xy-cell-base mixin to grow the flex item to take up all available space.

This is effectively the same as: `.flex-child-grow`, found @ https://foundation.zurb.com/sites/docs/flexbox-utilities.html#vanilla-flexbox-helper-classes - But there isn't a way to do this through the current SASS mixins.

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -44,7 +44,7 @@
 
 /// Sets base flex properties for cells.
 ///
-/// @param {Keyword} $size [full] - The size of your cell. Accepts `full`, `auto` or `shrink`.
+/// @param {Keyword} $size [full] - The size of your cell. Accepts `full`, `auto`, `shrink` or `grow`.
 @mixin xy-cell-base($size: full) {
   @if($size == 'full') {
     // This is the base style, all others inherit from it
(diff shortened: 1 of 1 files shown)


kball:
This seems reasonable to me. @brettsmason any objections?

brettsmason:
@kball No problem with it, though isn't `grow` the same as our `auto` behaviour?
In hindsight I probably would of called it `grow` rather than `auto` but oh well 😄

flatline-studios:
@brettsmason Unfortunately not...

AUTO:
`flex: 1 1 0px;`
GROW (proposed):
`flex: 1 0 auto;`

Because `auto` has the flex-basis set to 0px, as well as being told it can shrink, then it won't grow, it will just fit in to whatever is left over.

GROW would mean, "Take up AT LEAST your minimum space, and ALWAYS take up some amount of the remaining space (depending on other flex-grow: n elements), but NEVER take up less".

It's basically a more aggressive than the `flex: 1 1 0px` that's set up for `auto` currently.

brettsmason:
@flatline-studios Thanks for confirming and apologies for my slight lack of understanding there 😄 
Sounds a decent idea to me in that case.

PR #10691, opened 2017-10-02
Option to change font-weight of buttons

DaSchTour:
Adding option to set the font-weight of buttons. This is useful if the body font weight is set to a lighter font which is to light for buttons with solid background.

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -11,8 +11,9 @@
 $button-font-family: inherit !default;
 
 /// Font weight for button elements.
+/// Ignored if null (default)
 /// @type Font-Weight
-$button-font-weight: inherit !default;
+$button-font-weight: null !default;
 
 /// Padding inside buttons.
 /// @type List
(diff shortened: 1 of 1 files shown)


scss/components/_button.scss
```diff
@@ -91,6 +95,9 @@ $button-responsive-expanded: false !default;
   vertical-align: middle;
   margin: $button-margin;
   font-family: $button-font-family;
+  @if ($button-font-weight != inherit && $button-font-weight != $global-weight-normal) {
+    font-weight: $button-font-weight;
+  }
```
ncoden: One could want to enforce the `inherit` value on button font-weight (because it was previously set otherwise, by an other package for example). Same for $global-weight-normal.

For all default value that should not generate any code, we should use `null`. Any property with `null` as value will not be ignored in the CSS compilation.

So you can simply set `button-font-weight` to `null`, with maybe a comment like `If different than null, set the button font-weight to the given value. Otherwise, no property is set and the browser will inherit the property.`

ncoden (changes requested):
@DaSchTour I'll take care of the requested changes.

ncoden (approved):

PR #10698, opened 2017-10-05
Fix autoprefixer documentation for Sass setup

pdeffendol:
The documentation for configuring gulp-autoprefixer does not match the actual configuration used in the build scripts.  This updates the documentation to match.


rafibomb:
Thanks!

Issue #10552, opened 2017-08-18
Accessibility for Abide (form validation) please


udf2457:
Hi,

It does not appear that you have implemented any WCAG compliant accessibility for Abide (e.g. setting _aria-invalid_).

ncoden:
See #10699. It should be released in `v6.4.4`.

PR #10699, opened 2017-10-05
Abide accessibility

Owlbertz:
Improved Abide accessibility, based on issue #10552.
As this PR will not create a full WCAG compliance, it is at least a step towards it.

Maybe @udf2457 could give some feedback.

### DIFF
--- a/docs/pages/abide.md
+++ b/docs/pages/abide.md
@@ -29,30 +29,30 @@ These input types create a text field: `text`, `date`, `datetime`, `datetime-loc
   <div class="row">
     <div class="small-12 columns">
       <label>Number Required
-        <input type="text" placeholder="1234" aria-describedby="exampleHelpText1" aria-errormessage="exampleErrorText1" required pattern="number">
-        <span class="form-error" id="exampleErrorText1">
+        <input type="text" placeholder="1234" aria-describedby="example1Hint1" aria-errormessage="example1Error1" required pattern="number">
+        <span class="form-error" id="example1Error1">
           Yo, you had better fill this out, it's required.
         </span>
       </label>
-      <p class="help-text" id="exampleHelpText1">Here's how you use this input field!</p>
+      <p class="help-text" id="example1Hint1">Here's how you use this input field!</p>
     </div>
     <div class="small-12 columns">
       <label>Password Required
-        <input type="password" id="password" placeholder="yeti4preZ" aria-describedby="exampleHelpText2" aria-errormessage="exampleErrorText2" required >
-        <span class="form-error" id="exampleErrorText2">
+        <input type="password" id="password" placeholder="yeti4preZ" aria-describedby="example1Hint2" aria-errormessage="example1Error2" required >
+        <span class="form-error" id="example1Error2">
           I'm required!
         </span>
       </label>
-      <p class="help-text" id="exampleHelpText2">Enter a password please.</p>
+      <p class="help-text" id="example1Hint2">Enter a password please.</p>
     </div>
     <div class="small-12 columns">
       <label>Re-enter Password
-        <input type="password" placeholder="yeti4preZ" aria-describedby="exampleHelpText3" aria-errormessage="exampleErrorText3" required pattern="alpha_numeric" data-equalto="password">
-        <span class="form-error" id="exampleErrorText3">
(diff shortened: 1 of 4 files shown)


udf2457:
@Owlbertz Normally I'd love to, but I'm drowning in work at the moment, so it might be 2–3 weeks (at least) before I can get round to looking properly at it.  So for the moment the best I can do is thank you for your hard work. ;-)

atainton:
@Owlbertz any update?

atainton:
@udf2457 any update, would be great for a disabled user to know when there are validation errors

Owlbertz:
@atainton Don't hesitate to check this out and give feedback on your own for this.

udf2457:
@atainton et al.

I'm afraid I'm currently a bit backlogged at work - more backlogged than I was in October, which apparently is somehow humanly possible !  So realistically I'm not going to be able to do much testing of the above commit until much before the end of the year.

Sorry to disappoint, but such is life !

ncoden:
I'll check this. I want to learn about accessibility and this is a good opportunity for that ;)

Owlbertz:
@ncoden glad to hear that.
I don't want to advert for myself too much, but if you want to get started with accessibility maybe [my collection of related ressources](https://github.com/Owlbertz/a11y-collection) might be helpful.

ncoden:
@Owlbertz I added some aria attributes to the docs and tested it. It works well ;). I am particulary impressed by the `aria-live="assertive"` to "pop" an error message to the user.

The next step is to add these attributes when we can. I'm thinking to `aria-live` for all `[data-abide-error]`, and `aria-describedby` for all `.form-error` and `[data-form-error-for]`.

What do you think ?

ncoden:
@Owlbertz 

* [x] Set `[aria-live]` attribute on `[data-abide-error]`, according to the `a11yErrorLevel` option.
* [x] Set `[aria-describedby]` attribute on field, targetting the first form error.
* [x] Set `[for]` attribute on form errors, targetting their field.
* [x] `a11yAttributes` option to disable all this. For now, it does not disable `aria-invalid`.
* [ ] Full documentation.
* [ ] Unit tests.

ncoden:
Poke @Owlbertz. Would you have some time to review this ? :)

ahukkanen:
I don't know if @ncoden is still around but I would be interested in why you decided to add the `aria-describedby` to point to the error field if the `aria-describedby` attribute is not defined for the field.

This causes the screen reader to announce the error element event before the user has done anything wrong. I created a quick sample here what happens (try it out with a screen reader):
https://jsfiddle.net/f3g2rx1j/show

The screen reader is announcing here the following:
- "Text Required" field
- Editable
- Required
- Yo, you had better fill this out, it's required.
- (On some screenreaders it reads also the placeholder after the `aria-describedby` element)

We received feedback from accessibility auditors that this can be confusing for users using the assistive technologies. The user would not expect the error text to **describe** the contents of the input field. Instead, the screen reader should announce the error only when it becomes visible on the page, i.e. when the user leaves the field without entering anything to it (or when they return back to the field after first leaving it empty).

If there is a valid reason why you added the `aria-describedby` attribute to point to the first error message, we would be interested to hear the reasoning to better understand this and also to justify its use for the accessibility experts.

The described current behavior would be completely fine in case it happened when the error was already visible on the page, i.e. when the user returns back to the field after they first left it empty. But now that they hear the error description when it is still invisible on the screen, it creates confusion.

Could you share some insight on this?

---

Here's the example code if it disappears for some reason from JSFiddle:
```html
<!DOCTYPE html>
<html>
<head>
<title>Foundation test</title>
<!-- Compressed CSS -->
<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/foundation-sites@6.7.4/dist/css/foundation.min.css" crossorigin="anonymous">

<!-- Compressed JavaScript -->
<script src="https://code.jquery.com/jquery-3.6.0.min.js" crossorigin="anonymous"></script>
<script src="https://cdn.jsdelivr.net/npm/foundation-sites@6.7.4/dist/js/foundation.min.js" crossorigin="anonymous"></script>
</head>
<body>
  <form data-abide novalidate data-abide="true" data-validate-on-blur="true" data-live-validate="true">
    <label>
      Text Required
      <input type="text" placeholder="Go to this field with a screenreader (don't type anything)" required>
      <span class="form-error">
        Yo, you had better fill this out, it's required.
      </span>
    </label>
  </form>

<script>
$(document).ready(() => {
  $("body").foundation();
});
</script>
</body>
</html>
```

Issue #10701, opened 2017-10-05
Push, Pull, Offset aren't optional 


DusanMilko:
Using latest Foundation with SASS

@include foundation-grid(
  $pull: false,
  $push: false,
  $offset: false
);

pull, push, and offset aren't optional with foundation-grid. They should be... instead setting these variables renames the attribute...

By default the grid creates for example:

```
  .medium-1 {
    width: 8.33333%; }
  .medium-push-1 {
    position: relative;
    left: 8.33333%; }
  .medium-pull-1 {
    position: relative;
    left: -8.33333%; }
  .medium-offset-0 {
    margin-left: 0%; }
```

I only need, thus the other options should be optional

```
  .medium-1 {
    width: 8.33333%; }
```

colin-marshall:
Looks like this got merged so I'm closing out this issue.

PR #10702, opened 2017-10-05
Push, pull, & offset optional for foundation-grid

DusanMilko:
Allowing push, pull, offset, center, and uncenter to be optional for foundation-grid

Fixes issue #10701

### DIFF
--- a/scss/grid/_classes.scss
+++ b/scss/grid/_classes.scss
@@ -163,20 +163,13 @@
     }
 
     // Gutter adjustment
-    @if $uncenter {
-      .#{$-zf-size}-#{$uncenter} {
-        @include grid-col-unpos;
-      }
-    }
-
-    @if $push {
-      .#{$-zf-size}-#{$push}-0 {
-        @include grid-col-unpos;
-      }
-    }
-
-    @if $pull {
-      .#{$-zf-size}-#{$pull}-0 {
+    $-gutter-unpos-selector: (
+      if($uncenter, '.#{$-zf-size}-#{$uncenter}', null),
+      if($push, '.#{$-zf-size}-#{$push}-0', null),
+      if($pull, '.#{$-zf-size}-#{$pull}-0', null),
+    );
+    @if ($uncenter or $push or $pull) {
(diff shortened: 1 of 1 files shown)


scss/grid/_classes.scss
```diff
@@ -150,15 +156,29 @@
...
+
+    @if $push {
+      .#{$-zf-size}-#{$push}-0 {
+        @include grid-col-unpos;
+      }
+    }
+
+    @if $pull {
+      .#{$-zf-size}-#{$pull}-0 {
+        @include grid-col-unpos;
```
ncoden: `grid-col-unpos` will generate the same following 4-lines each time it is called:
```css
  position: static;
  float: float;
  margin-right: 0;
  margin-left: 0;
```

So we should rather try to generate a selector with or without uncenter/push/pull classes, instead of duplicating the CSS.

Something like: 
```scss
$selector: null;
$selector: if($uncenter, '#{$selector}.#{$-zf-size}-#{$uncenter} ,', $selector);
$selector: if($push, '#{$selector}.#{$-zf-size}-#{$push} ,', $selector);
$selector: if($pull, '#{$selector}.#{$-zf-size}-#{$pull} ,', $selector);

#{$selector} {
  @include grid-col-unpos;
}
```
But with some utilities to be more readable.
See: https://www.sassmeister.com/gist/dde2d7e79198b0cfab25e9ec54340b95

ncoden (changes requested):

ncoden:
@DusanMilko Sorry for the late reply. Would you have time to work on this ?
However, i'll take care of it (tomorrow or the next day).

ncoden:
@DusanMilko For your future pull requests, you can create a dedicated branch (like `feat/optional-push-pull-grid`) and create a pull request from it. So you can continue working on your branch `develop` while this PR is open.

ncoden (approved):
I pushed to requested changes

DusanMilko:
@ncoden ok cool, thanks. Sorry I was on vacation off grid.

PR #10704, opened 2017-10-05
Fixes issue with cell block height pinning in Safari

kball:
Fixes an issue introduced in https://github.com/zurb/foundation-sites/pull/10649 and turned up while doing QA on 6.4.4-rc1, which results in cell blocks not functioning properly in Safari

### DIFF
--- a/scss/xy-grid/_frame.scss
+++ b/scss/xy-grid/_frame.scss
@@ -61,7 +61,7 @@
   @if $vertical == true {
     overflow-y: auto;
     max-height: 100%;
-    height: 100%;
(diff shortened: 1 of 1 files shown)


ncoden:
@oxyc @brgrz Could you give your opinion on this ? Does this solve your problem in https://github.com/zurb/foundation-sites/issues/10457 without introducing a new bug in Safari ?

colin-marshall:
I believe this also fixes #10757.

DanielRuf:
Already fixed by https://github.com/zurb/foundation-sites/pull/11085

PR #10715, opened 2017-10-10
fix: prevent "space" on input to open dropdown

Owlbertz:
I had the usecase where I used a Dropdown as a auto-completion/suggestion feature. The Dropdown was attached to an input element and shows when the input has focus. Issue was then, that everytime a space was entered into the input, the Dropdown would receive focus and prevent the typing within the input.

I am not sure, whether this is too specific or not to become part of the code base.

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -178,7 +178,7 @@ class Dropdown extends Positionable {
 
       Keyboard.handleKey(e, 'Dropdown', {
         open: function() {
-          if ($target.is(_this.$anchors) && !$target.is('input')) {
+          if ($target.is(_this.$anchors) && !$target.is('input, textarea')) {
             _this.open();
             _this.$element.attr('tabindex', -1).focus();
             e.preventDefault();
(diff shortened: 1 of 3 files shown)


js/foundation.dropdown.js
```diff
@@ -178,7 +178,7 @@ class Dropdown extends Positionable {
 
       Keyboard.handleKey(e, 'Dropdown', {
         open: function() {
-          if ($target.is(_this.$anchors)) {
+          if ($target.is(_this.$anchors) && !$target.is('input')) {
```
ncoden: Not only `input` should be able to get a `space` without losing the focus. I think we have two solutions:
* determine all the native elements that should be able to receive `space` and make a big filter
* add an option to the `dropdown` element to prevent the keyboard catching

ncoden (approved):

Owlbertz:
Good finding with the `textarea`s @ncoden, we should check for `button` and `select` as well I guess.

ncoden:
Why `<button>` ? We do expect  `space` or `enter` on a button to open the dropdown. In the other case, I would simply used `.is(':focus')` to ignore all focusable anchors.

Owlbertz:
Sorry, I just realized now that I was in a different context. you are correct.

ncoden:
See: https://github.com/zurb/foundation-sites/pull/10899

PR #10720, opened 2017-10-12
Add jsDelivr hits badge

LukasDrgon:
Hey, I noticed that foundation-sites gets 39M hits per month on jsDelivr, so I thought you might like this badge.


ncoden (approved):
Hey @LukasDrgon. Thank you, we love new badges ;)

PR #10744, opened 2017-10-22
Update integrity value of foundation.min.{css,js} 

GreenRecycleBin:


ncoden:
Hello @GreenRecycleBin, sorry for the late reply.
I will close this PR, since it is now recommended to use `sha384` instead of `sha256`

PR #10752, opened 2017-10-25
Just to keep things standard!

IamManchanda:
cc @kball for merge!


ncoden:
We always like standards ;)
_We also like standard pull request names title ;)_

Issue #10606, opened 2017-08-31
Wrong version for meteor + wrong dependancy for fourseven:scss


mackgorski:
Foundation wont work properly with meteorjs. In latest version of the package from atmosphere there is not latest version of foundation - what I get is zurb:foundation-sites@6.4.1 instead of zurb:foundation-sites@6.4.3 and what's more it is tied to very old fourseven:scss@3.4.1 instead of fourseven:scss@4.5.4 - this prevents the meteor server from auto refreshing on scss file change and many more if even anyone can be successful in in installation.
#### How to reproduce this bug:
1. `$ meteor add fourseven:scss@4.5.4`
2. `$ meteor add zurb:foundation-sites`

#### What should happen:
Should install packages without any problems. And properly refresh browser automatically on every scss file change.

#### What happened instead:
```
meteor add zurb:foundation-sites
 => Errors while adding packages:

While selecting package versions:
error: Conflict: Constraint fourseven:scss@3.4.1 is not satisfied by fourseven:scss 4.5.4.
Constraints on package "fourseven:scss":
* fourseven:scss@4.5.4 <- top level
* fourseven:scss@3.4.1 <- zurb:foundation-sites 6.4.1
```

#### Foundation Version(s) you are using:
I want the latest one 6.4.3 but meteor shows 6.4.1 :/ no option of change.

Please make it work! Also please provide a path where the foundation files are stored in meteor after downloading.

danwild:
This is still a major issue, the atmosphere pkg is quite outdated at this point, to the point of being unusable. 

However; thankfully by following @juliancwirko's [great example](https://github.com/juliancwirko/meteor-npm-foundation-test) I have been able to use foundation `v6.4.3` + `fourseven:scss@4.9.0`.

I'm guessing npm might be preferred approach going forward to eliminate overhead of maintaining atmosphere pkgs...

juliancwirko:
@danwild yes, in newest versions of Meteor there is no need to use package from atmosphere. It was built before it was possible to simply use npm packages and imports.

ncoden:
@danwild @juliancwirko See https://github.com/zurb/foundation-sites/pull/10825#issuecomment-399265805

> Actually, this introduced an issue: Foundation is considered as compatible anymore with older Meteor versions (but it should be).
>
> I think we can use the following version range instead:
> ```
> fourseven:scss@3.4.1 || 4.0.0'
> ```
> 
[...]

PR #10773, opened 2017-11-16
Update fourseven:scss version for Meteor 1.6 #10606

user-222:
Fixes the bug in #10606 where foundation-sites cannot be used under Meteor 1.6 as fourseven:scss@3.4.1 is no longer supported (due to Node upgrade).  This PR allows using the newer versions of fourseven:scss that are supported by Meteor 1.6 and Node 8.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,13 +1,3 @@
 {
-	"plugins": [
-  	"transform-es2015-arrow-functions",
-  	"transform-es2015-block-scoped-functions",
-  	"transform-es2015-block-scoping",
-  	"transform-es2015-classes",
-  	"transform-es2015-destructuring",
-  	"transform-es2015-template-literals",
-  	"transform-es2015-parameters",
-  	"transform-es2015-shorthand-properties",
-  	"transform-es2015-spread"
-  ]
+
 }
--- a/.babelrc
+++ b/.babelrc
@@ -7,7 +7,7 @@
   	"transform-es2015-destructuring",
   	"transform-es2015-template-literals",
   	"transform-es2015-parameters",
-		"transform-es2015-shorthand-properties",
-  	"transform-es2015-spread",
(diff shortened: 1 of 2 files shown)


blakedut2:
Checks fail because the tests are run by Node 6 which doesn't natively support es2015, whereas Node 8 behind Meteor 1.6 does.  Is this a matter of upgrading foundation to Node 8?  Or just not supporting Meteor 1.6?  Either way my changes are 100% working on Node 8 as-is.

DanielRuf:
Please do not remove the babel plugins which are needed for compiling Foundation Sites for browsers.

The upgrade of node-sass should be the only thing that is done in this PR.

ncoden:
@blakedut2 We need the transpilation for browser compatibility. Could you check if an update of `fourseven:scss` only (without removing the Babel plugins) is enough to run Foundation on Meteor ?

raym:
I removed the hard-coded `fourseven:scss` version number on my cloned fork. Now the package tries to use `fourseven:scss@0.9.6`, which is probably why the version was hard-coded in the first place. I'll look into why it uses `0.9.6` and report back. I'll open my own PR if this one remains abandoned.

```
$ meteor add zurb:foundation-sites
 => Errors while adding packages:             
                                              
While loading package fourseven:scss@0.9.6:
error: Command failed: /Users/raym/.meteor/packages/meteor-tool/.1.6.0.12o3jcm++os.osx.x86_64+web.browser+web.cordova/mt-os.osx.x86_64/dev_bundle/bin/npm rebuild
--update-binary
...
```

raym:
My PR is here:
https://github.com/zurb/foundation-sites/pull/10825

DanielRuf:
Then close this PR as it is replaced by the other. You could have also rebased and force-pushed here without creating a new branch and PR, just as sidenote.

PR #10778, opened 2017-11-19
Update xy-grid.md

rlhawk:
Add $frame-grid to list of arguments for foundation-xy-grid-classes mixin.


ncoden (approved):
Thank you @rlhawk for your first contribution 🥇

Issue #10774, opened 2017-11-16
Precompile the library to CommonJS.


ShadyXV:
Trying to run tests with Jest doesn't work after importing foundation-sites on the component.
[#2937](https://github.com/facebookincubator/create-react-app/issues/2937)

Pharserror:
👍  on this.

ncoden:
Fixed in https://github.com/zurb/foundation-sites/pull/10864

bobwalker99:
The change made in #10864 doesn't fix the referenced create-react-app issue (facebook/create-react-app#2937) - The problem is with the jQuery import (`import $ from jQuery';`).

To recreate:

1. Create a react app with the standard commands:
```
npx create-react-app my-app
cd my-app
npm start
```
2. Add a reference to the develop branch of foundation-sites in package.json:
```
yarn add zurb/foundation-sites#develop
```
3. Edit `src\App.js` and add the following:
```
import { Foundation } from "foundation-sites/js/foundation.core";
```
4. run `yarn test`. output is as follows:
```
 FAIL  src\App.test.js
  ● Test suite failed to run

    C:\Users\bob.walker\dev\my-app\node_modules\foundation-sites\js\foundation.core.js:3
    import $ from 'jquery';
    ^^^^^^

    SyntaxError: Unexpected token import

      at new Script (vm.js:51:7)
      at Object.<anonymous> (src/App.js:4:19)

Test Suites: 1 failed, 1 total
Tests:       0 total
Snapshots:   0 total
Time:        5.685s
Ran all test suites related to changed files.
```

Changing all the imports in the foundation source (under my app's `node_modules`) to `import {$} from jQuery;` fixes it for me, but I tried forking and making this change in the foundation-sites project, but I get test failures that I don't get before the change. Happy to investigate further if someone can provide guidance?

ncoden:
@bobwalker99 `foundation-sites/js/foundation.core` is the source file and will always be in ES6+. Please use the following instead to import Foundation from the precompiled dist file:

```js
import { Foundation } from 'foundation-sites';
```
Or to get the core only:
```js
import { Foundation } from 'foundation-sites/dist/js/plugins/foundation.core';
```

I don't see how using `import {$} from jQuery;` would fix anything.

Issue #10336, opened 2017-07-04
jQuery.foundation() is not a function


voogieJames:
I am using foundation-sites plugins in my existing project, and after updating foundation to 6.4.1 `$(document).foundation();` returns an 'is not a function error'. 
But issue was solved by calling` Foundation.addToJquery($);` beforehand. 
So question: is it a bug or a feature?

klloe:
I've the same issue, your fix doesn't work either @voogieJames.

voogieJames:
@klloe, before that I've updated jQuery to make sure its version is **>= 3**.. and also 'npm update -g'

klloe:
@voogieJames Already use the CDN version

kball:
Hey @voogieJames @klloe I'd like to help track this down - how are you downloading and installing Foundation? What does your build look like? And are you using the built dist files or importing from the sourcefiles?

klloe:
@kball I'm using the basic template and load all the necessary scripts via CDN

```
  <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.2.1/jquery.min.js" integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4=" crossorigin="anonymous"></script>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/what-input/4.2.0/what-input.min.js" integrity="sha256-7wGy9IEPmo9Jk/I86siWbZFPeclfjFitAAfmvxBUc7k=" crossorigin="anonymous"></script>
  <script src="https://cdnjs.cloudflare.com/ajax/libs/foundation/6.4.1/js/foundation.min.js" integrity="sha256-Nd2xznOkrE9HkrAMi4xWy/hXkQraXioBg9iYsBrcFrs=" crossorigin="anonymous"></script>
    
  <script>
    $(document).foundation();
  </script>
```

kball:
Hi @klloe,

Using that exact pasted script, I see the following error:
```
Failed to find a valid digest in the 'integrity' attribute for resource 'https://cdnjs.cloudflare.com/ajax/libs/foundation/6.4.1/js/foundation.min.js' with computed SHA-256 integrity 'ssuvZ7clQBE8C6fdKygrQfRqIx5bFb/F/Abeqq1dr/w='. The resource has been blocked.
```

[...]

PR #10782, opened 2017-11-21
Add docs to new npm entry-point for CommonJS users.

Pharserror:
Related to #10774 and #10336.
The goal here is to add some instructions for CommonJS users who may struggle with understanding what the intention of this file is and how it should be used. This is my best guess based on my own experience. These instructions are similar to those given by @kball [here](https://github.com/zurb/foundation-sites/issues/10336#issuecomment-315955426) but, which worked for me.

### DIFF
--- a/dist/js/npm.js
+++ b/dist/js/npm.js
@@ -1,3 +1,25 @@
+/* TO USE:
+ *
+ * This entry-point assumes that you are doing 1 of 2 things:
+ * 1. Your webpack (or other transpilation library) config is setup to transform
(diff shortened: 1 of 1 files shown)


ncoden:
Thank you @Pharserror.
@kball Any thoughts on this ?

ncoden:
I'm closing this since Foundation now exports ES5/ES6-compatible modules. See https://github.com/zurb/foundation-sites/pull/10864. This sould be released in `v6.4.4`.

Pharserror:
👏

PR #10786, opened 2017-11-23
Fix reveal enableScroll calculation of scrollTop

SassNinja:
This PR fixes an issue with the scrollTop calculation of the `_enableScroll` method in the reveal component.
Precisely the issue occurred for me on iOS / iPhone.

After some debugging I found out the line `parseInt($("html").css("top"))` doesn't work as expected in this scenario because the class `.is-reveal-open` got removed before.
Therefore the html element got `position:static` again an returns `auto` as top-value on iOS.

To fix this I've moved the scrollTop calculation from the disabledScroll/enableScroll methods to the place they get called and pipe them through as param.
So now the top value gets calculated before `.is-reveal-open` is removed.

@IamManchanda @kball would you review?

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -186,7 +186,8 @@ class Reveal extends Plugin {
   * Disables the scroll when Reveal is shown to prevent the background from shifting
   * @param {number} scrollTop
   */
-  _disableScroll(scrollTop){
+  _disableScroll(scrollTop) {
+    scrollTop = scrollTop || $(window).scrollTop();
     if ($(document).height() > $(window).height()) {
       $("html")
         .css("top", -scrollTop);
@@ -197,7 +198,8 @@ class Reveal extends Plugin {
   * Reenables the scroll when Reveal closes
   * @param {number} scrollTop
   */
-  _enableScroll(scrollTop){
+  _enableScroll(scrollTop) {
+    scrollTop = scrollTop || parseInt($("html").css("top"));
     if ($(document).height() > $(window).height()) {
       $("html")
         .css("top", "");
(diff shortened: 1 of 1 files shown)


js/foundation.reveal.js
```diff
@@ -406,7 +410,7 @@ class Reveal extends Plugin {
 
       _this.$element.attr('aria-hidden', true);
 
-      _this._enableScroll();
+      _this._enableScroll(scrollTop);
```
ncoden: Could we move `enableScroll` before removing `is-reveal-open`, instead of passing the scrollTop ? I think this would be cleaner, but I don't know the codebase enough to be sure this would not cause others bugs.

SassNinja: I first tried this way and it fixed the iOS issue (original scrollTop restored after having closed the reveal).
But unfortunately non-iOS / desktop has got the issue then!

The reason is that the class `is-reveal-open` adds fixed position to the html and thus the scrollTop restoration don't work before removing that class.

My solution was the only way I've found to get rid of the issue for all cases.

SassNinja: @ncoden what do you think?

Would be great to see this iOS issue be gone in my projects (without adjusting the sources temporary)

ncoden: Sorry for the late reply. I'll check if it can be cleaner, then merge this.

ncoden:
@SassNinja I think it is clean enough now. You you check if this works in your project ? I have some trouble to reproduce the issue.

SassNinja:
Yup @ncoden it's still working for me.
While testing I've noticed the issue occurs on iOS 9.3 for me but not on iOS 11.2 anymore.
Nevertheless I think it's a good idea to fix it with this PR.

PR #10814, opened 2017-12-21
.load() -> .on('load') for jQuery3

ahebrank:
These are the ones I could still find. See e.g., #8975


ahebrank:
Oh, oops, wrong upstream.

ahebrank:
And never mind, dupe of 10729

Issue #10818, opened 2017-12-23
Responsive gutters for xy-cell() doesn't work as expected.


Lazerproof:
My settings:
```
$breakpoints: (
  small: 0,
  medium: 640px,
  large: 1024px,
  xlarge: 1200px,
  xxlarge: 1440px,
);
$print-breakpoint: large;
$breakpoint-classes: (small medium large);
```
```
$xy-grid: true;
$grid-container: $global-width;
$grid-columns: 12;
$grid-margin-gutters: (
  small: 20px,
  medium: 30px,
);
$grid-padding-gutters: $grid-margin-gutters;
$grid-container-padding: $grid-padding-gutters;
$grid-container-max: $global-width;
$xy-block-grid-max: 8;
```

This code
```
@include xy-cell(12);
	
@include breakpoint(medium) {
	@include xy-cell(8);
}
```
outputs
```
.contacts {
  width: calc(100% - 1.875rem);
  margin-right: 0.9375rem;
  margin-left: 0.9375rem;
}

@media print, screen and (min-width: 40em) {
  .contacts {
    width: calc(66.66667% - 1.875rem);
    margin-right: 0.9375rem;
    margin-left: 0.9375rem;
```
I'm not sure is it bug or designed behavior, but based on old float grid behavior, behavior of xy-gutters mixin and some other opened issues, I expected that gutters for xy-cell mixin outside of breakpoint mixin should fallback to the smallest breakpoint instead of the largest.
So I expected to get 

```
.contacts {
  width: calc(100% - 1.25rem);
  margin-right: 0.625rem;
  margin-left: 0.625rem;
}

@media print, screen and (min-width: 40em) {
  .contacts {
    width: calc(66.66667% - 1.875rem);
    margin-right: 0.9375rem;
    margin-left: 0.9375rem;
```
So currently it leads to more coding to get desired result 
```
@include xy-cell(12, $breakpoint: small)
```
or 
```
@include breakpoint(small) {
    @include xy-cell(12)
}
```

Also falling back to the largest breakpoint gutters conflicts with xy-gutters($negative: true) which outputs gutters for every breakpoint.
Version 6.4.4-rc1

ncoden:
Hello @Lazerproof.

For the following test code:
```scss
$breakpoints: (
  small: 0,
  medium: 640px,
  large: 1024px,
  xlarge: 1200px,
  xxlarge: 1440px,
);
$grid-margin-gutters: (
  small: 20px,
[...]

PR #10819, opened 2017-12-23
fix: prevent leaking of `-zf-size` in `-zf-each-breakpoint` mixin

ncoden:
Closes https://github.com/zurb/foundation-sites/issues/10818
Note: this also remove some useless logic in `@mixin breakpoint`


ncoden:
>  @IamManchanda self-assigned this 2 days ago

poke @IamManchanda

PR #10827, opened 2017-12-29
fix: remove tab outline when not using keyboard 10798

### LINKED ISSUE #10798: [Tabs] remove outline from click
The outline that is automatically added when a tab is clicked is very confusing.

So far we have to remove the outlines from the elements.

It would be good to have an option to remove the blue focus ring using a setting.

ncoden:
> No `outline` should be show when it is not for accessibility reasons, since it is not part of the Foundation design. So when `what-input` detect a mouse, we should have `outline: 0;` on the tab links

Closes #10798

### DIFF
--- a/scss/components/_tabs.scss
+++ b/scss/components/_tabs.scss
@@ -91,6 +91,7 @@ $tab-content-padding: 1rem !default;
   float: #{$global-left};
 
   > a {
+    @include disable-mouse-outline;
     display: block;
     padding: $padding;
     font-size: $font-size;


Owlbertz (changes requested):
@ncoden I guess your modifications of the Reveal SCSS-files in this PR are a mistake?

ncoden:
@Owlbertz Yes it is

ncoden:
@Owlbertz I fixed this.

Owlbertz (approved):
LGTM now

PR #10829, opened 2017-12-30
Drilldown parent fix

### LINKED ISSUE #10803: Drilldown with parent links doesn't allow for proper use of keyboard 'Enter' key

1. Create a basic drilldown component with data-parent-link set to true.
2. Add a valid href attribute to a submenu parent item.
3. Navigate to that parent item using keyboard. Press 'Enter'.
4. Navigate to the newly opened submenu parent item with a valid href. Press 'Enter' again.
 
#### What should happen:
1. When the submenu opens, the focus should be set on the submenu parent.
2. Pressing 'Enter' on the submenu parent item should take you to the location in its href attribute.

#### What happened instead:
1. When the submenu opens, focus is set incorrectly on the first submenu child.
2. Pressing 'Enter' on the submenu parent item makes it act like the 'Back' button.

All browsers.

Any version. Codepen example is using 6.3.1.

#### Test case link:

- [CodePen with Foundation 6.3.1](https://codepen.io/dbombine/pen/XgwOrZ)

Owlbertz:
Implemented suggestion of @dbombine in #10803.

Closes #10803

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -226,7 +226,7 @@ class Drilldown extends Plugin {
           if ($element.is(_this.$submenuAnchors)) {
             _this._show($element.parent('li'));
             $element.parent('li').one(transitionend($element), function(){
-              $element.parent('li').find('ul li a').filter(_this.$menuItems).first().focus();
+              $element.parent('li').find('ul li a').not('.js-drilldown-back a').first().focus();
             });
             return true;
           }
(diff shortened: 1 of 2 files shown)


ncoden (approved):
I don't like the way it is implemented (with selectors and a lot of assumptions about the HTML architecture), but the whole framework is like that, this is standard jQuery and I don't see how it could be improved.

I may have this impression because I dev only in Ember for 2 months and we don't deal with the DOM in that framework.

Anyway. LGTM 👍

Owlbertz:
Yeah, I am not really happy with the way I done it, too. But as you said the whole framework works like that. This wasn't the case when we started Foundation 6 I guess, but became more and more this way as more collaborators opened PRs. However, let's hope we can keep it cleaner with F7.

DaSchTour:
@ncoden I feel your pain. We need to stop using jQuery 😉 while currently working on an angular project I hate looking into old jQuery code more and more.

IamManchanda:
Same here guys ;)

Issue #10832, opened 2018-01-02
Add data-update-history option for magellan deep linking


jfurler:
The current behaviour for magellan deep linking is to push the browser state, which updates the history. I see that tabs and accordions have an additional `data-update-history` option which defaults to false, which allows you to configure whether or not you want this behaviour. It would be good if this option was also available for magellan.

PR #10833, opened 2018-01-02
Add updateHistory option to magellan

jfurler:
https://github.com/zurb/foundation-sites/issues/10832

### DIFF
--- a/docs/pages/magellan.md
+++ b/docs/pages/magellan.md
@@ -86,3 +86,10 @@ This below example is a simplified version of the table of contents on the right
 </div>
 ```
 
+---
+
+## Browser history
+
+When the `data-deep-link` option is set to `true`, the active section of the magellan is recorded by adding a hash with the active magellan section ID to the browser URL. By default, magellan *replace* the browser history (using `history.replaceState()`).
+
+Modify this behavior by using attribute `data-update-history="true"` to *append* to the browser history (using `history.pushState()`). In the latter case the browser back button will track each section the magellan gone through (in most case, this is not recommended).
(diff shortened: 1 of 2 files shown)


ncoden:
Thank you @jfurler for the PR.

It makes me think about _why_ we currently use `pushState` instead of `replaceState`. I found no reasons to support this. Actually, I found a lot of reasons why we should **not** use `pushState` and even discourage.
* Scrolling on the page is a __passive__ interaction. We do not expect it to be recorded or to have side-effects (others than loading things).
* Taking a quick look to the entire page may result in being locked in 10/15/20 history states.

So I disabled `updateHistory` by default, and added some documentation.

Owlbertz:
I think it is good that we have an option to switch between `replaceState()` and `pushState()`.
While you are right that `replaceState()` would most likely be the preferred option, I can imagine huge one-paged websites that tend to have different section with sub-page character, so removing this option might be wrong. However, I would be ok with making `updateState()` the default option.

ncoden:
@Owlbertz You mean `replaceState()` instead of `updateState()`, right ?

Owlbertz:
Uhm, yes!

ncoden:
Ok. So we can merge this ?
There will be a warning in the release notes.

Owlbertz:
I haven't pulled this down, but codewise it looks ok. Only thing I have to mention is that feature detection is only done for `window.history.pushState()` and never for `window.history.replaceState()` - but I assume that the existence of the former also gives information of the latter.

jfurler:
@Owlbertz I copied the logic from the "reveal" module, so yeah I just assumed the check was safe.

I only made `updateHistory` true by default to avoid a change in behaviour, but if you guys are happy with default false then that's fine with me :)

PR #10835, opened 2018-01-03
typo

gpaumier:


ncoden (commented):
Thx @gpaumier. I'll check if there is others mistakes like this one.

ncoden:
I fixed 2 others mistakes.

PR #10837, opened 2018-01-03
Fix for #10836 (clear hash in Magellan w/ deep linking)

### LINKED ISSUE #10836: Magellan w/ deep linking doesn't clear the hash when there's no longer an active link

On a page with with Magellan and `data-deep-linking` enabled and a top-margin greater than `data-threshold` (defaults to 50px) such that there is no active link until you scroll past that top-margin:

1. Load the page (without a hash).
2. Scroll down to the first Magellan-tracked element such that the window's location updates to that element's ID (e.g., `#first-heading`).
3. Scroll back to the very top (e.g., by pressing HOME on your keyboard).

#### What should happen:

What should happen after each of the steps to reproduce:

1. The window's location should not have a hash.
2. The window's location should have a hash of `#first-heading`.
3. The window's location should not have a hash.

#### What happened instead:
What actually happens after each of the steps to reproduce:

1. The window's location does not have a hash.
2. The window's location has a hash of `#first-heading`.
3. The window's location has a hash of `#first-heading`.

Chrome 63.0.3239.84 on macOS Sierra (10.12.6)

6.4.3

nvenegas:
Fixes #10836.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -169,7 +169,7 @@ class Magellan extends Plugin {
 
     if(this.options.deepLinking){
       var hash = "";
-      if(curIdx != undefined){
+      if(curIdx !== undefined){
         hash = this.$active[0].getAttribute('href');
       }
       if(hash !== window.location.hash) {
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -165,7 +165,11 @@ class Magellan extends Plugin {
     }
 
     this.$active.removeClass(this.options.activeClass);
-    this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
+    if(curIdx !== undefined){
+      this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
+    }else{
+      this.$active = $();
+    }
 
     if(this.options.deepLinking){
       var hash = "";
@@ -174,7 +178,8 @@ class Magellan extends Plugin {
       }
       if(hash !== window.location.hash) {
         if(window.history.pushState){
-          window.history.pushState(null, null, hash);
+          var url = curIdx !== undefined ? hash : window.location.pathname;
+          window.history.pushState(null, null, url);
         }else{
           window.location.hash = hash;
         }
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -178,7 +178,10 @@ class Magellan extends Plugin {
       }
       if(hash !== window.location.hash) {
         if(window.history.pushState){
-          var url = curIdx !== undefined ? hash : window.location.pathname;
+          // If there is no active idx, move to the same url without hash
+          // https://stackoverflow.com/a/5298684/4317384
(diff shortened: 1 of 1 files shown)


nicolasv:
I can't run the tests locally: I run into [#217: phantomjs terminated with signal SIGSEGV](https://github.com/nathanboktae/mocha-phantomjs/issues/217).

js/foundation.magellan.js
```diff
@@ -165,16 +165,21 @@ class Magellan extends Plugin {
...
 
     if(this.options.deepLinking){
       var hash = "";
-      if(curIdx != undefined){
+      if(curIdx !== undefined){
         hash = this.$active[0].getAttribute('href');
       }
       if(hash !== window.location.hash) {
         if(window.history.pushState){
-          window.history.pushState(null, null, hash);
```
nicolasv: Previously, when there was no active element (`curIdx === undefined`) we would `window.history.pushState(null, null, '')` which is effectively a no-op.

nicolasv:
> I can't run the tests locally: I run into #217: phantomjs terminated with signal SIGSEGV.

Pointing mocha-phantomjs to my homebrew installed phantomjs binary works makes them run:

```diff
diff --git a/package.json b/package.json
index 5abc54fc2..03efc2726 100644
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "test:ci": "npm run test:sass && npm run test:javascript:ci",
     "test:sass": "mocha test/sass/test_sass.js",
     "test:javascript:transpile": "gulp sass:foundation && gulp test:transpile-js",
-    "test:javascript:phantomjs": "npm run test:javascript:transpile && mocha-phantomjs --ignore-resource-errors test/javascript/index.html",
+    "test:javascript:phantomjs": "npm run test:javascript:transpile && mocha-phantomjs -p /usr/local/bin/phantomjs --ignore-resource-errors test/javascript/index.html",
     "test:javascript:browserstack": "npm run test:javascript:transpile && browserstack-runner",
     "test:javascript:ci": "npm run test:javascript:transpile && mocha-phantomjs --ignore-resource-errors test/javascript/index.html && browserstack-runner",
     "test:visual": "gulp test",
```

nicolasv:
Manually testing the `magellan/active-item.html` visual test (via `npm run test:visual`) looks good: scrolling back to the _Pre-anchor Section_ clears the location hash.

js/foundation.magellan.js
```diff
@@ -165,21 +165,26 @@ class Magellan extends Plugin {
...
         hash = this.$active[0].getAttribute('href');
       }
       if(hash !== window.location.hash) {
-        if (window.history.pushState) {
-          if (this.options.updateHistory) {
-            window.history.pushState({}, '', hash);
-          } else {
-            window.history.replaceState({}, '', hash);
+        if(window.history.pushState){
+          var url = curIdx !== undefined ? hash : window.location.pathname;
```
ncoden: `window.location.pathname` is only `/path`, but an url can contains: `/path?query#fragment`. So the query params will be lost

ncoden (commented):
Found a bug, but everything else looks good. 👍

ncoden (approved):
Fixed 👍

ncoden:
Merged. Thank you @nicolasv for you first PR 🎉

nicolasv:
Thanks, @ncoden!

PR #10845, opened 2018-01-09
Use calc in button group with single child

TheBnl:
Without calc the single button in a button group would overlap the grid because of set negative margin. This becomes noticeable when the spacing is set to, for example, the grid gutter.


ncoden (approved):
I generally don't like the use of `calc`... but this code is a mess with all the styles applied automatically, and the flex/non-flex compatibility.

I checked a lot of use cases. It seems to works.

Thank you @TheBnl for your first contribution 🎉

PR #10848, opened 2018-01-10
Added solution to compile problem for npm users

jnystromdesign:


docs/pages/sass.md
```diff
@@ -77,7 +77,8 @@ Finally, add an `@import` statement to the top of your primary Sass file. Refer
 @import 'foundation';
 ```
 
-You're also going to want a settings file for your project, which will allow you to modify the default styles of Foundation. **[Download the latest settings file here](https://raw.githubusercontent.com/zurb/foundation-sites/master/scss/settings/_settings.scss)**, add it to your project as `_settings.scss`, then import it *before* Foundation itself.
+You're also going to want a settings file for your project, which will allow you to modify the default styles of Foundation. **[Download the latest settings file here](https://raw.githubusercontent.com/zurb/foundation-sites/master/scss/settings/_settings.scss)**, add it to your project as `_settings.scss`, then import it *before* Foundation itself. 
+For users of npm, you need to change `@import util/util` to a full path to the util (node_modules/foundation-sites/scss/util/util) in the settings file in order to compile it.
```
ncoden: I don't think this concerns only to NPM users. All users with a package manager or a custom installation of Foundation may have the Foundation missing from the Sass `includePaths` and should edit `utils/utils`.

Maybe we could reformulate this to:

```
<div class="callout>
    The settings file needs to import `util/util` from Foundation. Please ensure that the Foundation folder is included in Sass or change `@import util/util` for it to points to the full path of the file. For example, NPM users may need to change the import to `node_modules/foundation-sites/scss/util/util`.
</div>
```

ncoden: We could also add a message inside `settings.scss`

ncoden (changes requested):

docs/pages/sass.md
DanielRuf: Or in webpack where we can use `~packagename`

ncoden:
@jnystromdesign Would you have some time to discuss/provide the requested changes ? Or should be take care of this ?

jnystromdesign:
@ncoden Uhm. Etither callouts aren't showing in the markdown preview at GitHub or I've done something stupid. Sry. Just want to contribute, but I'm somewhat of a newb when it comes to collaboration on GitHub.

ncoden:
I think it should be explained in the documentation _and_ in the codebase. For the callout, it's because the GitHub markdown preview doesn't have our "callout" class. This doc is not pure markdown and will be visible only in https://foundation.zurb.com/sites/docs.

ncoden:
I just remembered that `foundation.scss` is automatically generated by `octophant`. So I'm merging this. 👍 Thank you @jnystromdesign

ncoden (approved):

PR #10851, opened 2018-01-10
fix: enforce `show-for-sr` and others visibility classes with important #10850

### LINKED ISSUE #10850: [Feature Request] !important rules on 'show-for-sr'
The only CSS property on `.show-for-sr` with an '!important' rule is `position`. However, there are several cases where it would be more fail-proof to add `!important` to most (all?) properties: if you use this class you definitely want to hide the text, no matter if another class on the same element changes any of these properties. 

It also shouldn't be dependent on the cascade – on one of my projects I have some CSS rules as part of the critical path (and the rest of the CSS loaded after that), which means, for example, that the `forms` CSS rules are loaded after the visibility classes, causing some issues on a `<input type="file">` with the `show-for-sr` class.

ncoden:
> Most of our classes should depends on cascade. Our components must be customizable (in Sass AND in CSS) so their properties must be overridable without having to increase the specificity.
>
> However, I agree that some classes can have a "deterministic and uncustomizable behavior", and so can use !important on their properties to ensure that. But we should be careful about this, explicit in the documentation, and not begin to use !important everywhere to make things "work" quickly.
>
> In a perfect CSS architecture, this would be an entirely separated type of classes. A sort of "stand-alone state" with a particular prefix.
-- https://github.com/zurb/foundation-sites/issues/10850#issuecomment-356741458

See: https://hugogiraudel.com/2016/10/13/css-hide-and-seek/ (cc @HugoGiraudel)

Closes: https://github.com/zurb/foundation-sites/issues/10850

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -207,27 +207,38 @@
 /// Makes an element visually hidden, but still accessible to keyboards and assistive devices.
 /// @link http://snook.ca/archives/html_and_css/hiding-content-for-accessibility Hiding Content for Accessibility
 /// @link http://hugogiraudel.com/2016/10/13/css-hide-and-seek/
-@mixin element-invisible {
-  position: absolute !important;
-  width: 1px;
-  height: 1px;
-  padding: 0;
-  overflow: hidden;
-  clip: rect(0,0,0,0);
-  white-space: nowrap;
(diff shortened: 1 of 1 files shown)


ncoden:
Note: This introduce breaking changes ⚠️.

Issue #10854, opened 2018-01-11
[Accordion Menu] $accordionmenu-padding not working properly


goerba:
**Foundation version 6.4.3**

The easiest way to check this out, is by going to the [Accordion Menu](https://foundation.zurb.com/sites/docs/accordion-menu.html) page on the Foundation Docs site. Scroll down to an example accordion menu and inspect 'Item 1' with your developer toolbar.

You will see the following:

`.menu a, .menu .button {`
    `line-height: 1;`
    `text-decoration: none;`
    `display: block;`
    `padding: 0.7rem 1rem;`
`}`

`.accordion-menu a {`
   `padding: 0.7rem 1rem;`
`}`

The `.accordion-menu a` style is being overruled by the `.menu a` style, because of the order in which the SCSS mixins are loaded in `foundation.scss`.

So even when you change the `$accordionmenu-padding` variable in the settings, it will never be used on this level, because of that.

ncoden:
Currently, import in `foundation.scss` are sorted alphabetically.

**This is a very bad idea**.

Import order in Sass matters a lot as the last CSS properties declared will override the first ones. So components should be sorted in a more "generic" to more "specific" components. When a component relies on others (like `accordion-menu` with `menu`), these dependencies should be imported before.

ncoden:
@goerba See https://github.com/zurb/foundation-sites/pull/10856

Issue #9260, opened 2016-10-15
Dropdown submenu not working in flex mode.


ycrepeau:
The top bar menu (responsive) has a drop down submenu to allow user to pick the language in her/his choice. When Foundation 6 is in Flexmode, the dropdown does not work (the submenu remains open).

![screen 1](https://cloud.githubusercontent.com/assets/6821466/19408461/894c64ba-928b-11e6-9d2d-6a236de9f767.png)

In foundation_and_overide.scss, I had the following:

``` scss
$global-flexbox: true;
//@include foundation-grid;
@include foundation-flex-grid;
@include foundation-flex-classes;
```

The workaround was:

``` scss
//$global-flexbox: true;
@include foundation-grid;
//@include foundation-flex-grid;
//@include foundation-flex-classes;
```

The haml code is:

``` haml
-# /app/views/layout/_header.html.haml
/ Start Top Bar
.title-bar{"data-hide-for" => "medium", "data-responsive-toggle" => "example-menu"}
  %span{"data-toggle" => "example-menu.top-bar", :type => "button"}
    %button.menu-icon
  .title-bar-title Yeti

#example-menu.top-bar
  .top-bar-left
    %ul.dropdown.vertical.medium-horizontal.menu
      %li.menu-text Yeti
      %li= link_to "Home", root_path

  .top-bar-right
    %ul.dropdown.menu.vertical.medium-horizontal(data-dropdown-menu)
      - if user_signed_in?
        %li= link_to "Edit profile",  edit_user_registration_path
        %li= link_to "Logout",  destroy_user_session_path, :method => :delete
      - else
        %li=link_to "Log in!", new_user_session_path

      %li.has-submenu= render 'layouts/locale_selector', controller: controller_name, action: action_name
```

and

``` haml
-# /app/views/layouts/_locale_selector.html.haml
%a{:href=>"#"}
  %span.fa.fa-lg.fa-globe
%ul.menu.submenu.vertical(data-submenu)
  - previous_enabled = true
  - locale_languages.each do |language|
    - if language[:disabled]
      - if previous_enabled
        - previous_enabled = false
        %li
          %hr

      %li
        %span{:style=>'font-size:75%; color:#888; margin-left:15px;'}
          =language[:label]
    - else
      %li
        = link_to language[:label],
                  controller: controller,
                  action: action,
                  set_locale: language[:locale],
                  page: params[:page]
```

masonhale:
+1 I'm having the same issue.

[...]

PR #10856, opened 2018-01-11
chore: sort Sass components in categories

ncoden:
Closes:
- https://github.com/zurb/foundation-sites/issues/10854
- https://github.com/zurb/foundation-sites/issues/9260

> Currently, import/includes in foundation.scss are sorted alphabetically. This is a very bad idea.
> 
> Import/includes order in Sass matters a lot as the last CSS properties declared will override the first ones. So components should be sorted in a more "generic" to more "specific" components. When a component relies on others (like accordion-menu with menu), these dependencies should be imported and included before.

**Changes**: Sort CSS includes in the following categories:
1. **Global styles**
2. **Generic components** (controls and HTML-oriented components)
3. **Basic components** (components with enclosed layout styles)
4. **Containers** (components with side-effects on internal layout)
5. **Layout** (components with side-effects on external layout)
6. **Helpers** (highly specific and non-inheritable classes)

⚠️  Breaking changes MAY HAVE been introduced. Tests are required to ensure it did not created bugs.


ncoden:
@colin-marshall @brettsmason Could you review this ?
(also poke @SassNinja)

ncoden:
I checked all component and everything looks fine.
@SassNinja Could you make a review ?

colin-marshall:
@ncoden I like this better, but it seems the categories themselves would be more organized if it could be sorted alphabetically when dependencies don't matter. For example, I think `drilldown-menu` should come before `dropdown-menu`. Another example is the sorting of generic components category looks to be completely random, but maybe there are dependencies I'm not seeing.

ncoden:
@colin-marshall hmm you are right. I'll change that.

ncoden:
I resolved conflicts with `develop`, moved `typography` and `forms` to the "Global styles" category, and sorted components inside categories alphabetically.

DanielRuf (approved):

PR #10868, opened 2018-01-17
fix: fix off-canvas option `revealOn` when used without classes #10867 

### LINKED ISSUE #10867: [Off Canvas] Off canvas not visible with revealOn option passed without .reveal-on classes
The Off-canvas `revealOn` option is infered from its class (`.reveal-on-medium` == `revealOn: 'medium',`), but the `.reveal-on` class is not set when  `revealOn` alone is given, while it is required for the off-canvas to be visible.

* Create an Off-canvas with `[data-revealed="true"]` and `[data-reveal-on=...]`
* Look, and see......

#### What should happen:

The Off Canvas is visible.

#### What happened instead:

The Off Canvas takes the place but is not visible.

Foundation v6.4.3

#### Test case link:

https://codepen.io/ncoden/pen/Rxevga

ncoden:
Closes https://github.com/zurb/foundation-sites/issues/10867

Other changes:
* fix Off-canvas `revealOn` option when used with `revealClass` option
* add a visual test for off-canvas `revealOn` option

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -3,7 +3,7 @@
 import $ from 'jquery';
 import { Keyboard } from './foundation.util.keyboard';
 import { MediaQuery } from './foundation.util.mediaQuery';
-import { transitionend } from './foundation.util.core';
+import { transitionend, RegExpEscape } from './foundation.util.core';
 import { Plugin } from './foundation.plugin';
 
 import { Triggers } from './foundation.util.triggers';
@@ -118,12 +118,16 @@ class OffCanvas extends Plugin {
       }
     }
 
-    this.options.isRevealed = this.options.isRevealed || new RegExp(this.options.revealClass, 'g').test(this.$element[0].className);
+    // Get the revealOn option from the class.
+    var revealOnClass = this.$element[0].className.match(RegExpEscape(this.options.revealClass) + '([^\s])+', 'g');
+    if (revealOnClass) {
+      this.options.isRevealed = true;
+      this.options.revealOn = this.options.revealOn || revealOnClass[1];
+    }
 
-    // Get the revealOn option from the class / ensure the `reveal-on-*` class is set.
-    if (this.options.isRevealed === true) {
-      this.options.revealOn = this.options.revealOn || this.$element[0].className.match(/(reveal-for-medium|reveal-for-large)/g)[0].split('-')[2];
-      if (this.options.revealOn) this.$element.first().addClass(`reveal-for-${this.options.revealOn}`);
+    // Ensure the `reveal-on-*` class is set.
+    if (this.options.isRevealed === true && this.options.revealOn) {
+      this.$element.first().addClass(`${this.options.revealClass}${this.options.revealOn}`);
       this._setMQChecker();
     }
 
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -122,9 +122,8 @@ class OffCanvas extends Plugin {
 
     // Get the revealOn option from the class / ensure the `reveal-on-*` class is set.
     if (this.options.isRevealed === true) {
-      if (this.options.revealOn)
       this.options.revealOn = this.options.revealOn || this.$element[0].className.match(/(reveal-for-medium|reveal-for-large)/g)[0].split('-')[2];
-      this.$element.first().addClass(`reveal-for-${this.options.revealOn}`);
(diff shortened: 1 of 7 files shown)


SassNinja (changes requested):
I think the condition has to placed one line below so it affects `this.$element.first().addClass(`reveal-for-${this.options.revealOn}`);`
Otherwise you will end up with `reveal-for-null` if there's neither a reveal-for-x class nor a revealOn attribute.

SassNinja:
Another thing I've noticed:
This line is bad imo because it only works with the default breakpoint classes
`this.options.revealOn = this.options.revealOn || this.$element[0].className.match(/(reveal-for-medium|reveal-for-large)/g)[0].split('-')[2];``

Once you've got custom breakpoints (e.g. xmedium) it won't work correctly.
Think it should be changed into
`this.options.revealOn = this.options.revealOn || this.$element[0].className.match(/(reveal-for-\w+)/g)[0].split('-')[2];``

What do you think @ncoden ?

ncoden:
> I think the condition has to placed one line below

Yes. I don't know what happened. It's fixed.

ncoden:
> This line is bad imo because it only works with the default breakpoint classes

I fixed this, and an other one below, when using both `revealOn` and `revealClass` options. I would like to add a visual test for this one, but we cannot generate custom Sass for visual tests for now.

SassNinja:
>  would like to add a visual test for this one, but we cannot generate custom Sass for visual tests for now.
Does `for now` mean there are plans to support this in future?
Just curious – how would you do this? Individual SCSS files in the subfolders (e.g. `/test/visual/offcanvas/nested.scss`) that get compiled?

ncoden:
Yes. Or all files in a same folder. But the point is we should have a way to get custom sass in visual tests.

ncoden:
@SassNinja Could you review the last changes ?

js/foundation.offcanvas.js
```diff
@@ -118,10 +118,16 @@ class OffCanvas extends Plugin {
       }
     }
 
-    this.options.isRevealed = this.options.isRevealed || new RegExp(this.options.revealClass, 'g').test(this.$element[0].className);
+    // Get the revealOn option from the class.
+    var revealOnClass = this.$element[0].className.match(RegExpEscape(this.options.revealClass) + '([^\s])+', 'g');
```
SassNinja: match() does not work this way – you must either use a hard coded pattern or a RegExp pattern if you need variables in it.
So it has to look like this
```javascript
var revealOnClass = this.$element[0].className.match(new RegExp(RegExpEscape(this.options.revealClass) + '([^\s])+', 'g'));
```

js/foundation.offcanvas.js
```diff
@@ -118,10 +118,16 @@ class OffCanvas extends Plugin {
...
+    if (revealOnClass) {
+      this.options.isRevealed = true;
+      this.options.revealOn = this.options.revealOn || revealOnClass[1];
```
SassNinja: It has to be `revealOnClass[0]` – but this only works if a match has been found and otherwise throw an error...

SassNinja (commented):
I've reviewed your last changes @ncoden 

Although I like getting away from the split I think `revealOn` is supposed to be the breakpoint name – not the class name

js/foundation.offcanvas.js
ncoden: Fixed

js/foundation.offcanvas.js
ncoden: Not anymore since I switched to `regexp.exec(str)`

js/foundation.util.core.js
```diff
@@ -24,6 +24,18 @@ function GetYoDigits(length, namespace){
...
+/**
+ * Escape a string so it can be used as a regexp pattern
+ * @function
+ * @see https://stackoverflow.com/a/9310752/4317384
+ *
+ * @param {String} str - string to escape.
+ * @returns {String} - escaped string
+ */
+function RegExpEscape(str){
+  return str.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
```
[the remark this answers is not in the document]
ncoden: https://stackoverflow.com/a/6076269/4317384

`/` is the pattern delimiter, but has no sens in regexp, it does not need to be escaped.

ncoden:
@SassNinja I tested on my side. Do you think it is ready to merge ?

SassNinja:
> @SassNinja I tested on my side. Do you think it is ready to merge ?

Yup, looks good to me and ready to merge

PR #10877, opened 2018-01-22
Adds ability to fire Trigger listeners immediately. Fixes #10869

### LINKED ISSUE #10869: For sites with ads (or long window.onload delay), util.triggers init happens far too late

1. Place a garden-variety modal (and accompanying button) components on a page that can have a  delayed `window.load` event.  In our particular situation, this delay is caused by the presence of standard display ads (served from Google DFP) on the page, but any delay in `window.load` will yield the same result.
2. View the page
3. Note that the button to trigger the modal won't be respond to user interaction until that delayed `window.load` event finally fires.

#### What should happen:
Button should be able to trigger the modal as soon as DOM is ready. 

#### What happened instead:
Buttons aren't reactive to input until `window.load` fires, giving the appearance of broken UI

Any typical browser

6.4.3

#### Test case link:
https://jsfiddle.net/q7ged6so/4/

#### Questions
1. Is there a particular reason `window.load` is waited for?  
1. Could this be changed to DOM ready instead without interfering with other behaviors?    
a. Or perhaps made configurable to be one or the other?

kentontaylor:


ncoden:
Hi @kentontaylor, Thank you for your PR, I'll review it tonigh.

ncoden (commented):
Hi @kentontaylor !

Your PR works, but seems a bit hacky. Foundation should just try to initialize himself as soon as possible. A `iDontCareInitializeNow()` function would only fails in most of its use cases.

As far as I know, there is no reason to wait for all scripts to be loaded. When html is dynamically added we expect it to be initialized manually anyway.

I replaced `$(window).on('load')` by `$(document).ready()` and it seems to work. Unit tests runs and all components are fine. However, I'm not totally confident with this.

@colin-marshall @Owlbertz @brettsmason Do any of you know `Trigger` enough to ensure that using `$(document).ready()` does not introduce weird bugs ?

Owlbertz:
I really don't want to sound rude, but the broken unittests indicate an issue to me.

ncoden:
@Owlbertz The broken unittests comes from elsewhere. I did not figured out from where, it works in local.

DanielRuf:
> @Owlbertz The broken unittests comes from elsewhere. I did not figured out from where, they works in local.

> I really don't want to sound rude, but the broken unittests indicate an issue to me.

https://github.com/zurb/foundation-sites/issues/10886

DanielRuf:
Closes https://github.com/zurb/foundation-sites/issues/10869

DanielRuf:
@ncoden I guess this is another candidate for the new helper utility.

ncoden:
`onLoad` is already used for this (see https://github.com/zurb/foundation-sites/pull/11077). The issue here is that `$(window).on('load')` is triggered once external ressources are loaded. Problem is that some external ressources are not useful at all for the page to be usable and can take a long time to load.

So we _may_ have to use `$(document).ready()` instead. Problem is: it's hard to figure out what this could break in the current users' projects.

Issue #10883, opened 2018-01-25
[Interchange] Support for background images without file extension


diniscorreia:
I know this is an edge case, but is there any way to have Interchange support background images without a file extension? 

On one of my projects I have to use the company's image server, which returns the images with a URL without an actual file extension. I basically modified the regex on Interchange to match the image server's subdomain.

ncoden:
@diniscorreia See: https://github.com/zurb/foundation-sites/pull/10884

JeremyEnglert:
@diniscorreia, see @ncoden's message above. It looks like it may meet your needs.

In the meantime, I'm going to close out this ticket. Expanding upon this idea will be considered for the next version of Foundation - thanks for the idea!

PR #10884, opened 2018-01-25
feat: add "type" option to force interchange replacement method #10883

ncoden:
Add `type` option to Interchange:

> type {string} ['auto'] - Type of the responsive ressource to remplace. It can takes the following options:
> - `auto` (default): choose the type according to the element tag or the ressource extention
> - `src`: replace the `[src]` attribute, recommanded for images ``.
> - `background`: replace the `background-image` CSS property.
> - `html`: replace the element content.

Others changes:
* Clean some parts using ES6 features in Interchange

Closes https://github.com/zurb/foundation-sites/issues/10883


js/foundation.interchange.js
```diff
@@ -83,6 +84,22 @@ class Interchange extends Plugin {
     }
   }
 
+  /**
+   * Check options valifity and set defaults for:
+   * - `data-interchange-type`: if set, enforce the type of remplacement (auto, src, background or html)
```
DanielRuf: Souldn't it be "replacement"?

js/foundation.interchange.js
```diff
@@ -146,29 +163,39 @@ class Interchange extends Plugin {
...
 
-    var _this = this,
-        trigger = 'replaced.zf.interchange';
+    var trigger = 'replaced.zf.interchange';
+
+    var type = this.options.type;
+    if (type === 'auto') {
+      if (this.$element[0].nodeName === 'IMG')
+        type = 'src';
+      else if (path.match(/\.(gif|jpg|jpeg|png|svg|tiff)([?#].*)?/i))
```
DanielRuf: `jpe?g` could be also used. But what about webp and other formats?

js/foundation.interchange.js
```diff
@@ -198,7 +225,19 @@ Interchange.defaults = {
    * @type {?array}
    * @default null
    */
-  rules: null
+  rules: null,
+
+  /**
+   * Type of the responsive ressource to remplace. It can takes the following options:
```
DanielRuf: remplace => replace
can takes => can take

js/foundation.interchange.js
```diff
@@ -198,7 +225,19 @@ Interchange.defaults = {
...
+   * - `auto` (default): choose the type according to the element tag or the ressource extention
```
DanielRuf: extention => extension

js/foundation.interchange.js
```diff
@@ -198,7 +225,19 @@ Interchange.defaults = {
...
+   * - `src`: replace the `[src]` attribute, recommanded for images `<img>`.
```
DanielRuf: recommanded => recommended

test/javascript/components/interchange.js
```diff
@@ -95,6 +98,16 @@ describe('Interchange', function() {
       spy.restore();
     });
 
+    it('replaces following the `type` option', function() {
```
DanielRuf: replaces following the `type` option => replaces the following `type` option

DanielRuf (commented):
Looks good so far but needs some changes in the texts.

js/foundation.interchange.js
ncoden: It should. `remplacement` is french ;)

js/foundation.interchange.js
ncoden: I kept the same regex as before. But you're right, any supported `` format should be added here.

ncoden:
Thanks for the review. I shoud definitely pay more attention to typo.

PR #10885, opened 2018-01-26
test some build related stuff

DanielRuf:


DanielRuf:
Interesting. It still fails. I'll use this PR here for testing and finding a solution.

Issue #10901, opened 2018-02-01
Responsive Accordion Tabs: .foundation('destroy') doesn't remove events handler attached to the window element


mbellet:
I'm working with Responsive Accordion Tabs, if you initialize a element manually and after destroy the initialization with .foundation('destroy') and after that you initialize again the element you will have 2 event handlers attached to the window.
This will execute 2 functions that will cause malfunctions of the plugin.

I think that destroy method should remove the handlers that is attached from the initialization in the window element.

I don't know if it's only in "Responsive Accordion Tabs" plugin maybe this behavior could also happen in all hybrid Foundation plugins like "Responsive Navigation".

ncoden:
Hi @mbellet. See https://github.com/zurb/foundation-sites/pull/10902. It should be released in `v6.4.4`.

PR #10902, opened 2018-02-01
fix: remove events handler when ResponsiveAccordionTabs is destroyed #10901

ncoden:
Closes https://github.com/zurb/foundation-sites/issues/10901


kball:
This looks good to me.

PR #10864, opened 2018-01-14
chore: fix ES5/ES6 Foundation imports

ncoden:
### Changes:
* Build Foundation as an UMD module
* Make `js/entry/foundation.js` an ESM entry point.
* Delete unused `dist/js/npm.js`

See commits for more details.

### Entry points:
* `dist/js/foundation.js`: [UMD](https://github.com/umdjs/umd) (browser-compatible) entry point [**default**]
* `js/entries/foundation.js`: [ESM](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) (es6) entry point

### ✅ Closes:
* https://github.com/zurb/foundation-sites/issues/10852 Using foundation 6 js with brunch and npm instead of bower and webpack (@fmj02)
* https://github.com/zurb/foundation-sites/issues/10774  Precompile the library to CommonJS (@ShadyXV)
* https://github.com/zurb/foundation-sites/issues/10800 Webpack LibraryTarget - export UMD (@tomm1996)

### ➡️ Related to:
* https://github.com/zurb/foundation-sites/pull/10849 pkg.main should be an es5 module not es6 (@iainbeeston)
* https://github.com/zurb/foundation-sites/pull/10782 Precompile the library to CommonJS (@ShadyXV)
* https://github.com/zurb/foundation-sites/issues/10336 jQuery.foundation() is not a function (@voogieJames)
* https://github.com/zurb/foundation-sites/issues/10645 ReferenceError: Can't find variable: exports (@melcarthus)
* https://github.com/zurb/foundation-sites/issues/10723 zurb foundation unknown word error (@dhisharma)
* https://github.com/zurb/foundation-sites/issues/10910 Couldn't find preset "es2015" relative to directory (@bonovski)

PR #10903, opened 2018-02-01
[WIP] fix: fix ES6 plugins webpack build

ncoden:
Bug introduced in https://github.com/zurb/foundation-sites/pull/10864

Changes:
* Export the mutated `Foundation` object and the plugin in every plugin entry

### DIFF
--- a/js/entries/plugins/foundation.abide.js
+++ b/js/entries/plugins/foundation.abide.js
@@ -2,3 +2,5 @@ import { Foundation } from './foundation.core';
 
 import { Abide } from '../../foundation.abide';
 Foundation.plugin(Abide, 'Abide');
+
+export { Foundation, Abide };
--- a/js/entries/plugins/foundation.accordion.js
+++ b/js/entries/plugins/foundation.accordion.js
@@ -2,3 +2,5 @@ import { Foundation } from './foundation.core';
 
 import { Accordion } from '../../foundation.accordion';
 Foundation.plugin(Accordion, 'Accordion');
+
+export { Foundation, Accordion };
--- a/js/entries/plugins/foundation.accordionMenu.js
+++ b/js/entries/plugins/foundation.accordionMenu.js
@@ -3,3 +3,4 @@ import { Foundation } from './foundation.core';
 import { AccordionMenu } from '../../foundation.accordionMenu';
 Foundation.plugin(AccordionMenu, 'AccordionMenu');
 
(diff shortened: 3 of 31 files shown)


colin-marshall:
@ncoden this fixed all but 1 of the `TypeErrors`. I am left with this one:
```
Uncaught TypeError: Foundation.Magellan is not a constructor
```

Because of this, the table of contents menu still doesn't work on the docs when I build them. The error traces back to this line:
https://github.com/zurb/foundation-docs/blob/master/js/docs.tableOfContents.js#L23

ncoden:
Related to https://github.com/webpack/webpack/issues/4824

ncoden:
So `develop` is broken because I am too stupid to test correctly a PR before merging it.

The problem in this issue is:
1. Webpack transpilation to UMD require a particular format for `externals` (see https://webpack.js.org/configuration/output/#module-definition-systems).
2. Exports in Foundation plugins are "named" (and not defaults), so importing them as  `external` require to make external targets objects (in which we can pick up the named plugin). (see https://github.com/zurb/foundation-sites/pull/9965/commits/abfd426baaeb510b9e2d06a82b902036402b1d39)
3. External format for UMD does not support the "object target".

ncoden:
To do:
[ ]: temporary disable UMD bundling (#10911)

Then, there is multiple paths we can take to resolve this issue:
* switch to Rollup (but I'm not sure if it will resolve this problem)
* patch webpack (difficult!)
* create a plugin for webpack to add the missing feature (even more difficult!)

ncoden:
So I was writting an issue to webpack, and explaining the problem helped me (as often) to understand what we're doing wrong.

On a side, we want to use **named export**, so we see each plugin file not as the whole plugin, but as a container for the plugin.
On the other side, we want to import externals as the plugin itself (stocked in `window.Foundation.*`). This was resolved by the `externals: { './foundation.core`: { Foundation: window.Foundation }, ...}` trick  because we were creating the "container" on the fly from (for this example) `window.Foundation`.

So: the solution is to bundle each generated file in its own webpack "Library": webpack embed all the file exports in a dedicated object. This is the container we want. So we use that "Library" as external instead of `window.Foundation.*`.

I have a solution that seems to work. I need to make some more tests and optimizations before commiting.

ncoden:
Closing this in favor of https://github.com/zurb/foundation-sites/pull/10930

Issue #10384, opened 2017-07-11
[Magellan] Update event fires constantly


mattbag:
I wasn't able to find a solution online, and I'm not confident in fiddling with this event, but most of all I'm sure with the rebuild solving this issue might be easy enough...
I imported in my js triggers,smoothscroll and magellan. It works great, but since I'm attaching a few functions it seems a bit of a waste.
Cheers!

#### How to reproduce this bug:

1. Step one: any magellan instance
2. Step two: show the magellan event in console
3. Step three: ponder a solution

#### What should happen:
magellan event should fire after the active magellan section as been updated

#### What happened instead:
the magellan event fires on every scroll motion

#### Browser(s) and Device(s) tested on:
Chrome

#### Foundation Version(s) you are using:
6.4.1

#### Test case link:
https://codepen.io/mattbag/pen/LLXJVP?editors=1111

#### Older version possibile solution
https://github.com/zurb/foundation-sites/issues/9068

kball:
@IamManchanda can you take a look at this one as part of your entre into the JS world?

IamManchanda:
Will look into it ... cool!

mattbag:
I don't know if it relates at all, while studying vanilla js I came up with this solution
https://codepen.io/mattbag/pen/WZvoPo?editors=0101

Basically gets the index of the section to check the trigger. Hope it helps somehow. Thanks!

Issue #9068, opened 2016-07-30
Magellan update event fires too often.


Owlbertz:
I feel like the Magellan `update.zf.magellan` fires too often.
The docs state that it "Fires when magellan is finished updating to the new active element.", but as seen in the image it fires more often making it not usable to reliable hook into **after** the update process.

![image](https://cloud.githubusercontent.com/assets/4280373/17270383/a7558452-5661-11e6-9c38-5e0f6c079e07.png)

kball:
Definitely the case. It appears to fire continuously on scrolling updates, and I'm seeing it anywhere from 2 to 6 times on clicking an update in the `active-item.html` visual test.

It looks like there's a lot of functionality running each time (including manipulating classes) that really shouldn't be... I think the fix is to either check the class to find current active target, or probably better keep a JS reference to current active index, and bail early around https://github.com/zurb/foundation-sites/blob/develop/js/foundation.magellan.js#L156 if index hasn't changed.

Owlbertz:
Bump @kball

kball:
@Owlbertz thanks for prodding - been slammed with some other stuff, do you have time to take a look?

coreysyms:
Hey fellas, just ran into this issue and solved it for me by doing this. Not sure if it is the best solution for the rest of the world, but wanted to share to see if it perked anyone else's mind. I basically just use the logic that is in place to my advantage, which in _updateActive states. Take all hrefs and removed the active class, then apply active to the one that is currently calculated as active. I slipped this "if" in between the remove and update active class to grab the current and see if the href is the same as the one being applied the active class.

```
this.$active.removeClass(this.options.activeClass);

if ($(this.$active[0]).attr("href") !== $(this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]')[0]).attr("href")) {
  this.$element.trigger('update.zf.magellan', [this.$active]);
}

this.$active = this.$links.filter('[href="#' + this.$targets.eq(curIdx).data('magellan-target') + '"]').addClass(this.options.activeClass);
```

The trigger only fires when the href is actually different from the previous one. Running 6.3.0

brettsmason:
@Owlbertz @coreysyms This looks like something that should be fixed but my JS knowledge is limited.
Any suggestion on what to do with this?

coreysyms:
this is not fixed. just tested. I can push a PR something to fix it.

coreysyms:
If someone would like to double check this, please do.

PR #10906, opened 2018-02-02
fix: fire Magellan update event only where there is a change

coreysyms:
Fix for #9068 and #10384. Fires update only where there is a change.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -151,6 +151,7 @@ class Magellan extends Plugin {
   _updateActive(/*evt, elem, scrollPos*/) {
     if(this._inTransition) {return;}
     var winPos = /*scrollPos ||*/ parseInt(window.pageYOffset, 10),
+        changed = false,
         curIdx;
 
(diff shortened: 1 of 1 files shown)


ncoden (commented):
Looks good 👍, but I think the code readibility could be improved. The Magellan code is already difficult to read, let's try to not make it worst ;)

ncoden:
@coreysyms Also, could you submit PR from a branch on your own fork, instead of the main Foundation repo ? ;)

ncoden:
@coreysyms I fixed a bug (the event was not fired anymore when the active link were reset), and refactored the function.

@colin-marshall or @JeremyEnglert  Could you take a look at this please ?

ncoden:
There should be no breaking changes or new features, so I plan this for `v6.4.4`.

DanielRuf:
lgtm

DanielRuf:
Do we still need this branch here?

ncoden:
@DanielRuf No. Nice catch

PR #10907, opened 2018-02-03
feat: add `code-inline` and `code-block` typography helpers #9614

### LINKED ISSUE #9614: <code> tag vs long named codes (mainly on mobile)

1. write a long word in a `<code>` tag
2. place it in a container with less width than the word
3. the `<code>` element will overflow the width of the container

#### What should happen:
The word should break and add another line - not overflow the container

This could be fixed with:
```
.fix {
ms-word-wrap: break-word;
word-wrap: break-word;
}
```
#### What happened instead:
The `<code>` block will overflow the container width. This should not happen, on mobile will break the viewport width and add a vertical scrollbar
#### Test case:

-http://codepen.io/wecodeart/pen/GrJyQY

ncoden:
Changes
* add `code-inline` and `code-block` mixins and classes in ` typography/_helpers`
* remove `<code>` defaults properties
* add `$enable-code-inline` option to control wether `<code>` is based on `.code-inline
* add docs for `code-inline` and `code-block` in typography-base

Closes #9614

### DIFF
--- a/docs/pages/typography-base.md
+++ b/docs/pages/typography-base.md
@@ -306,6 +306,29 @@ Remember to escape angle brackets when printing HTML: <code>&lt;div&gt;</code>
 
 ---
 
+Use `.code-inline` component to apply the code style when you want.
+
+```html_example
+<span class="code-inline">I am not code, but I am displayed as if.</span>
+```
+
+---
+
+Use the `.code-block` component to create a block of code.
+
+```html_example
+<code class="code-block">{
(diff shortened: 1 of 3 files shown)


JeremyEnglert:
At first glance, this looks solid! I'll pull it down later today and test a few different scenarios.

ncoden:
@JeremyEnglert Did you have the time to make your tests ?

DanielRuf (approved):

PR #10915, opened 2018-02-05
Fix accessibility of pagination examples.

### LINKED ISSUE #9334: [Pagination] Documentation suggests invalid HTML role="navigation"
The documentation for Pagination includes this example:

`<ul class="pagination" role="navigation" aria-label="Pagination">`

with this comment:

> Note that the container has the attributes `role="navigation"` and `aria-label="Pagination"`. These explain the purpose of the component to assistive technologies.

I noticed my IDE warning that `"navigation"` is not an acceptable value for the `role` element in this context. At first I thought this was a bug in the IDE, but more research confirmed this:

http://stackoverflow.com/a/14181478/29470 says:

> `<ul>` has a default implied ARIA semantic of `list`, and may only validly be set to either `directory`, `list`, `listbox`, `menu`, `menubar`, `presentation`, `tablist`, `toolbar` or `tree`, so setting it to `navigation` is invalid and breaks the `list` semantic that the `<ul>` element has in the first two cases.

This can be confirmed in the [HTML5 spec](https://www.w3.org/TR/html5/dom.html#sec-implicit-aria-semantics)

Maybe it would be better to suggest wrapping the `<ul>` in a `<nav>` element instead.

Owlbertz:
According to
- http://www.a11ymatters.com/pattern/pagination/
- http://stackoverflow.com/a/14181478/29470

Closes #9334

### DIFF
--- a/docs/pages/pagination.md
+++ b/docs/pages/pagination.md
@@ -9,7 +9,7 @@ sass: scss/components/_pagination.scss
 
 A pagination list is just a `<ul>` with the class `.pagination`, and a series of `<li>`/`<a>` pairs. Add the class `.current` to an `<li>` to mark the current page, or `.disabled` to add disabled styles to a link.
 
-Note that the container has the attributes `role="navigation"` and `aria-label="Pagination"`. These explain the purpose of the component to assistive technologies.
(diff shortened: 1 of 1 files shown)


ncoden (approved):
👍 Thanks @Owlbertz

PR #10921, opened 2018-02-06
fix: fix DropdownMenu closing when touching to scroll #9823

### LINKED ISSUE #9823: Dropdown submenus close on touch scroll gesture

1. Trigger Item 1 menu dropdown
2. Move down / touch gesture down the page

Bug occurs in both 6.3.0 and 6.3.1 Foundation versions 

#### What should happen:
The submenu should stay open, otherwise if the submenu is larger than the viewport the submenu items aren’t reachable.

#### What happened instead:
The submenu closes after moving down / touch gesture down the page.
#### Test case:

- [Test Dropdown Menu with Foundation 6.3.0]( http://documentation.knowledgekube.co.uk/test-foundation/dropdown-6-3-0/)
- [Test Dropdown Menu with Foundation 6.3.1]( http://documentation.knowledgekube.co.uk/test-foundation/dropdown-6-3-1/)

imabdulsal:
This solution fixes this issue #9823.

**Issue:**
On a touchable screen, when you open up a long dropdown menu that goes beyond the viewport boundary -i.e., #9823-, then attempts to **move down/up** to keep interacting with the menu, the menu will disappear! That happens because the current attached events `mouseup.zf.dropdownmenu touchend.zf.dropdownmenu` on the body object cannot figure out if that `touchend` comes after the user **moved down/up** or the user just tapped on the screen, once.

**Solution:**
I added the flag `this.isTouchMove` to help `touchend`'s callback function to figure out if the user was **moving down/up** or just tapped once.

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -74,6 +74,7 @@ class DropdownMenu extends Plugin {
       }
     }
     this.changed = false;
+    this.isTouchMove = false;
     this._events();
   };
 
@@ -270,13 +271,20 @@ class DropdownMenu extends Plugin {
   _addBodyHandler() {
     var $body = $(document.body),
         _this = this;
-    $body.off('mouseup.zf.dropdownmenu touchend.zf.dropdownmenu')
+    $body.off('touchmove.zf.dropdownmenu')
+         .on('touchmove.zf.dropdownmenu', function(){
+           _this.isTouchMove = true;
+         })
+         .off('mouseup.zf.dropdownmenu touchend.zf.dropdownmenu')
          .on('mouseup.zf.dropdownmenu touchend.zf.dropdownmenu', function(e) {
            var $link = _this.$element.find(e.target);
-           if ($link.length) { return; }
+           if ($link.length || _this.isTouchMove){
+             _this.isTouchMove = false;
+             return;
+           }
 
            _this._hide();
-           $body.off('mouseup.zf.dropdownmenu touchend.zf.dropdownmenu');
+           $body.off('mouseup.zf.dropdownmenu touchend.zf.dropdownmenu touchmove.zf.dropdownmenu');
          });
   }
 


ncoden (commented):
The `this.isTouchMove` gesture seems pretty generic. Could we make it more generic, handled by the `touchend` trigger ?

Also, is there any case where a `touchmove` is trigger, then a `touchend` unrelated to the `touchend`, then  the `touchmove` `mouseup` ?

And finally, some comments would be useful.

abdullahsalem:
You are right @ncoden, the instance variable name `this.isTouchMove` is too generic. In fact, I tried to make it more specific, but I wasn't lucky to find more specific name. The plugin does not have another instance variable that shares this variable's functionality, and it has no other event that needs for such that flag, therefore I thought it might be ok if I leave it as it is.
Any other suggested names for that instance variable?

ncoden:
What I meant is: the logics there to try to remember whether or not the detected `touchend` comes from a `touchmove` is pretty generic. I'm asking if we could just have a way to say in the DropdownMenu "i look for touchend that did not had a touchmove before", but it seems only possible with custom jQuery events.

Also, I wonder why we use `mouseup` and `touchend` here instead of click.

> With a mouseup event, you can click somewhere else on the screen, hold down the click button, and move the pointer to your mouseup element, and then release the mouse pointer. A click event requires the mousedown and mouseup event to happen on that element.
>
> The normal expectation is that a click requires both the mousedown and mouseup event, so I'd recommend the click event.
>
> From the possible duplicate, it appears that mouseup and mousedown events can also be caused by mouse buttons other than the left click button. Which is very different from what a generic user would expect.
> -- https://stackoverflow.com/q/14805225/4317384

I would personally expect elements to lost focus (and maybe close or anything) on mousedown _or_ on click, but not mouseup or right-click/middle-click.

See for example:
feb-08-2018 23-54-18

@abdullahsalem  What do you think ? Is there any reasons here to use `mouseup touchend` ?
Also, I'll take a look at the git history, maybe tomorrow.

ncoden:
poke @abdullahsalem ;)

ncoden:
So `touchend` was introduced in https://github.com/zurb/foundation-sites/pull/7344/files. I have troubles to understand why as nothing is commented...

DanielRuf:
> So touchend was introduced in https://github.com/zurb/foundation-sites/pull/7344/files. I have troubles to understand why as nothing is commented...

So was this the cause for the issue that is fixed by this PR?
The code suggests that it closes the dropdown when something around it is clicked / touched to no keep it open then wen a click on the body occurs.

ncoden:
@DanielRuf Yes it is. But my question was: why `mouseup`+`touchend` and not `click` ?

I got it: `click` is often not triggered on mobile, same for `touch`. I'm looking for a better event...

DanielRuf:
> I got it: click is often not triggered on mobile, same for touch. I'm looking for a better event...

Yes, `touchstart` + `touchend` is the native `tap` / `click` on mobile and touch devices.

ncoden:
Actually `touchstart` end `touchend` are triggered when scrolling on mobile, so it's not the same. This is the issue here.

I made some research and it's always the same "solutions" coming back:
* using `touchstart` (we don't want to do that here)
* `touchstart`+`touchend` and cancel if `touchmove` got triggered (like the current implementation)
* using jQuery mobile's `tap` event

So... I guess all I can do it's cleaning this PR a bit with a dedicated utility for "tap/click" events

ncoden:
Closing in favor of #11091

PR #10928, opened 2018-02-11
fix typo

wolfbeisz:
replace 'msot' with 'most'


ncoden (approved):
👍 Thanks for your first contribution @wolfbeisz

Issue #10563, opened 2017-08-21
[Dropdown Menu] Incorrect documentation


Karl456:
Just been racking my brains out as to why the events for Dropdown menu weren't working, seems like this is because the documentation is incorrect.

The documentation (found here: http://foundation.zurb.com/sites/docs/dropdown-menu.html#Item-2) says:
```
show.zf.dropdownMenu
hide.zf.dropdownMenu
```

When this should be (notice the lower-case M):
```
show.zf.dropdownmenu
hide.zf.dropdownmenu
```

DanielRuf:
This will be finally fixed by https://github.com/zurb/foundation-docs/pull/24

PR #10929, opened 2018-02-11
Fix: dropdown menu event names in docs

JeremyEnglert:
Fixes: https://github.com/zurb/foundation-sites/issues/10563

I'll admit, I'm a bit confused on exactly how to submit this PR. Because to fix the event name in the docs, it looks like we have to edit it in the code. 

@ncoden - gonna need your help here. :)

### DIFF
--- a/js/foundation.dropdownMenu.js
+++ b/js/foundation.dropdownMenu.js
@@ -310,7 +310,7 @@ class DropdownMenu extends Plugin {
     if (this.options.closeOnClick) { this._addBodyHandler(); }
     /**
      * Fires when the new dropdown pane is visible.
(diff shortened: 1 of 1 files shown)


ncoden (approved):
This is our Git workflow for now. Even if the change is "in the code", this is considered as a Documentation change, so it's good for master.

But I agree with you that this is confusing. As said in private:
> this is not a good practice at all. `master` should be safe and  in theory contains only new releases.

LGTM 👍 Good job.

DanielRuf:
Do we still need this branch?

ncoden:
No ;)

PR #10555, opened 2017-08-18
Fixes #9151 - Drilldown event issue

IamManchanda:
Updated by mistake ... Correcting now!
Closes #9151

Issue #9151, opened 2016-09-01
[Drilldown Menu] Wrong JS events listed in Documentation


callumisitt:
The Javascript events in the documentation for the Drilldown Menu appear to be wrong. I tried the open event, which is listed in the docs as 'open.zf.drilldownMenu', but nothing was happening when I set up a listener for it. Using 'open.zf.drilldown' however, worked as intended. This was with version 6.2.3.

ahdigital:
This is still an issue in version 6.4.

ozeone:
How come this has not been fixed in a release?
Googled for 1h before finding this, and solving my problems.

The event name is still open.zf.drilldownMenu in docs, and needs to be open.zf.drilldown

shready:
Looks like it was fixed in f7e63fd by changing the docs title, but then the title was changed back 72844c8 for consistency. That probably reintroduced the issue.

IamManchanda:
Appologies folks... Have to accept my mistake... I just updated it for the sake of docs search but that was a mistake...

Fixed it now ... Added a PR!

PR #10932, opened 2018-02-12
fix generated event name

DanielRuf:
As a sidenote, using the title for event names is not a great idea. We should use some variable `fn-name` or similar.

### DIFF
--- a/docs/pages/drilldown-menu.md
+++ b/docs/pages/drilldown-menu.md
@@ -1,5 +1,5 @@
 ---
-title: Drilldown Menu
+title: Drilldown
 description: Drilldown is one of Foundation's three menu patterns, which converts a series of nested lists into a vertical drilldown menu.
 video: 8qPQRXl52hI
 scss: scss/components/_drilldown.scss


ncoden (commented):
I don't get it. What is the link between the doc title and the event names ?

DanielRuf:
> I don't get it. What is the link between the doc title and the event names ?

Well, run `npm run deploy:docs` and take a look at the build/drilldown-menu.html file.

DanielRuf:
See `formatJsEventName` at https://github.com/zurb/foundation-docs/blob/master/lib/handlebars/javascript.js

DanielRuf:
The change is already done in develop but not master where the current docs live and so the docs are atill wrong.

DanielRuf:
Also see

https://github.com/zurb/foundation-sites/issues/9151

https://github.com/zurb/foundation-sites/pull/10555

DanielRuf:
I hereby propose to add another frontmatter variable which is used if it is set. But probably in another PR.

JeremyEnglert:
I thought the names were generated from the inline names/comments. Such as here: https://github.com/zurb/foundation-sites/blob/1c6ce2d69e976b8d0ead214c2b2619548466c858/js/foundation.drilldown.js#L302

If this isn't the case, the recently merged PR (https://github.com/zurb/foundation-sites/pull/10929) isn't correct.

DanielRuf:
> I thought the names were generated from the inline names/comments. Such as here:

Unfortunately no.

DanielRuf:
screenshot_20180212-192608 is the current status.

DanielRuf:
And https://github.com/zurb/foundation-docs/blob/master/templates/partials/javascript-reference.hbs#L107

JeremyEnglert:
Eeeek. Definitely something we need to address in F7. Contributing to the docs should be one of the easier things to do. 

Waiting for feedback from @ncoden, as we may have to address previous PRs as well.

ncoden:
I'm investigating for a solution a bit cleaner.

ncoden:
See https://github.com/zurb/foundation-docs/pull/24/files

ncoden:
See also: https://github.com/zurb/foundation-sites/pull/10933

PR #10936, opened 2018-02-13
Revert "fix: fix side-effect of jQuery .filter() in Abide.addA11yAttr…

DanielRuf:
…ibutes", fixes #10935

This reverts commit c6a69067701b24ceaccd2bad625c37898ad0b445.

PR #10938, opened 2018-02-13
switch to Node.js 8 to use package-lock.json

DanielRuf:
This is not the best solution but still recommended as the only issue and cause is phantomjs 1.x and it just affects the unit tests in this specific environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "6"
+  - "8"
 install:
   - npm install -g bower browserstack-runner
   - npm install


ncoden:
If I understand correctly, since #10936 is merged we can close this ?

ncoden:
@DanielRuf If I understand correctly, since #10936 is merged we can close this ?

DanielRuf:
It was meant as a fast solution instead of https://github.com/zurb/foundation-sites/pull/10944 as the failing tests are produced by phantomjs with jQuery 3.3.

Chrome headless does not have these issues afaik.

DanielRuf:
The lock file locked the build to the versions that are pinned in the lock file. So far Travis and npm do not use the lockfile and instead use the latest releases.

ncoden:
Also changing the node version in tests to just hide errors while we are supposed support node >= 4 is not a long-term solution ;)

DanielRuf:
> we are supposed support node >= 4

Was not aware of this. I guess you mean https://github.com/zurb/foundation-sites/blob/develop/package.json#L110 and the other dependencies.

But we should keepn an eye on Node 8 (and 9) in the future as npm 5, npx and so on are much faster now =)

PR #10939, opened 2018-02-14
Allow breakpoint mixin to accept a list

homerjam:
This fixes the following issue whereby gutters aren't set as expected for larger sizes:
```
.my-cell {
      @include xy-cell(12);

      @include breakpoint(medium) {
        @include xy-cell(6/12);
      }
}
```

by allowing the following:
```
.my-cell {
      @include xy-cell(12);

      @include breakpoint(medium, large, xlarge, xxlarge) {
        @include xy-cell(6/12);
      }
}
```

[Changes without white space](https://github.com/zurb/foundation-sites/pull/10939/files?w=1)

### DIFF
--- a/docs/pages/xy-grid.md
+++ b/docs/pages/xy-grid.md
@@ -449,6 +449,10 @@ We also have a shorthand option for the above which outputs the same CSS:
 }
 ```
 
+<div class="callout warning">
+  If you pass multiple breakpoints to the <code>breakpoint</code> mixin, it will duplicate its content for each of them. Be careful to only use <code>breakpoint</code> with properties that should change across breakpoints.
+</div>
+
 ### Custom Block Grid
 
 Use the `xy-grid-layout()` mixin to create your own block grid.
--- a/docs/pages/xy-grid.md
+++ b/docs/pages/xy-grid.md
@@ -411,17 +411,34 @@ The cell size calculator can also be accessed as a function. This gives you the 
 
 ### Responsive Grids
 
-Pair `xy-cell` with the `breakpoint()` mixin to make your grid responsive.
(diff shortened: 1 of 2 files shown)


DanielRuf:
Isn't this more like a feature or really a bugfix?

homerjam:
I'm not sure to be honest, I couldn't figure out from #10141 and #10387 whether this was intentional behaviour or not.

ncoden (changes requested):
Hi @homerjam,

Basically, this change allow to duplicate a `breakpoint()` content across several breakpoints.

For now `breakpoint()` only takes one size as parameter, since before `xy-grid` there was no reason to duplicate the same content across breakpoints. But now `xy-grid` automatically detect the breakpoint it is in, so this would be a legit use case.

However, I am afraid that not everyone will use it correctly and that many will end up duplicating a lot of code without knowing.

We could add that feature as is with a very good documentation. But I would like us to also look for others solutions.

Note: "request change" because the mixin documentation is not updated.

homerjam:
Hi @ncoden 

Glad to see this is being considered and you understand the use case.

I've briefly updated the inline documentation as requested.

For others possibly arriving here this is how I'm working around the issue for now:
```
.my-cell {
     @include xy-cell(12);

      @each $size in (medium, large, xlarge, xxlarge) {
        @include breakpoint($size) {
          @include xy-cell(6/12);
        }
      }
}
```

Thanks!

ncoden (approved):
Hi @homerjam. I made the following changes:

* [x] Use the arglist annotation instead of list
* [x] Add usage cases in the XY Grids docs
* [x] Add warning about code duplication

DanielRuf (approved):
lgtm.

Definitely helpful feature and the implementation seems is ok.

homerjam:
Awesome! Thanks @ncoden

PR #10943, opened 2018-02-14
tests: migrate to mocha-headless-chrome

DanielRuf:
This PR tries to fix the unit test suite. Switching away from phantomjs to Chrome headless.


package.json
```diff
@@ -80,6 +80,7 @@
     "is-empty-object": "^1.1.1",
     "js-yaml": "^3.8.4",
     "mocha": "^3.4.2",
+    "mocha-headless-chrome": "^1.8.2",
     "mocha-phantomjs": "^4.0.2",
```
ncoden: `mocha-phantomjs` is not required anymore, right ?

package.json
```diff
@@ -15,9 +15,9 @@
     "test:ci": "npm run test:sass && npm run test:javascript:ci",
     "test:sass": "mocha test/sass/test_sass.js",
     "test:javascript:transpile": "gulp sass:foundation && gulp test:transpile-js",
-    "test:javascript:phantomjs": "npm run test:javascript:transpile && mocha-phantomjs --ignore-resource-errors test/javascript/index.html",
+    "test:javascript:phantomjs": "npm run test:javascript:transpile && mocha-headless-chrome -a ignore-resource-errors -f test/javascript/index.html",
```
ncoden: We can rename that script since phantomjs is not used anymore. Maybe to `test:javascript:units` ?
Also, is there any downside or incompatibility with switching to `mochheadless-chrome` ?

test/javascript/components/offcanvas.js
```diff
@@ -104,7 +104,7 @@ describe('Off Canvas', function() {
           plugin.$element.should.have.class('is-open');
           $('body').should.have.class('is-off-canvas-open');
           done();
-        }, 1);
+        }, 80);
```
ncoden: Why this ?

package.json
DanielRuf: Exactly. I'm doing the cleanup later this day.

package.json
DanielRuf: Right. I'll do this then or rewrite the commit.

DanielRuf: > Also, is there any downside or incompatibility with switching to mochheadless-chrome

Not that I knew of.

test/javascript/components/offcanvas.js
DanielRuf: Because it is too short for CI and produced a failing test.

ncoden:
@DanielRuf  Also, please try to use clearer commit and PR titles ;)

test/javascript/components/offcanvas.js
DanielRuf: I general there was a race condition which gave `null`.

DanielRuf: Mh, testing. Might be another issue there =/

DanielRuf: We might have to change this there as both run in parallel as it seems (the event and the other code part).

ncoden: > In general there was a race condition which gave null.

I don't understand what you're talking about

> Because it is too short for CI and produced a failing test.

I think this is caused by the `opened.zf.offcanvas` event being triggered _before_ the class and everything is set, which is not a good thing (because the offcanvas is not fully `opened`).

We should resolve that instead of putting a workarounds in the tests.

DanielRuf:
> @DanielRuf Also, please try to use clearer commit and PR titles ;), with a namespace and a short resume of what is change and why.

Understood. Will do this then when I rewrite the commit message.

ncoden:
Please avoid to force push on a public PR. This break the PR history and prevent others to understand the context of each comment.

Once someone commented/reviewed your PR, consider opening a new clear PR on an other branch if you want to rebase.

DanielRuf:
> Please avoid to force push on a public PR. This break the PR history and prevent others to understand the context of each comment.

Yup, saw it, took too much for the force push =/ Will redo it =)

PR #10941, opened 2018-02-14
getFnName unit tests ADDED

dragthor:
`Foundation.getFnName()` unit tests added.

PR #10944, opened 2018-02-15
tests: migrate to mocha-headless-chrome

DanielRuf:
* migrate to Chrome headless for unit tests
* remove wrongly used `end()` from abide label filtering, fixes a failing unit test
* update npm script name to make clear that we do not use phantojs anymore
* trigger opened.zf.offcanvas at the end to prevent race conditions


test/javascript/util/box.js
```diff
@@ -33,8 +33,8 @@ describe('Foundation box', function () {
 
       var dims = Foundation.Box.GetDimensions($("#rect-test"));
 
-      dims.width.should.equal(200);
-      dims.height.should.equal(100);
+      Math.round(dims.width).should.equal(200); // Math.round fix for IE 11 (probably because of rem)
+      Math.round(dims.height).should.equal(100); // Math.round fix for IE 11 (probably because of rem)
```
ncoden: Dimensions are defined in pixels, so why `probably because of rem` ?

test/javascript/util/core.js
```diff
@@ -97,7 +97,7 @@ describe('Foundation core', function() {
       var name = Foundation.getFnName(B);
 
       name.should.be.a('string');
-      name.should.be.equal('');
+      name.should.be.equal('B');
```
ncoden: I seen that failed in the tests. These tests were added today in https://github.com/zurb/foundation-sites/pull/10941. It worked well with `phantomjs`.

I'm not sure of what we should expect in this case. @dragthor do you have an opinion on this ?

test/javascript/components/offcanvas.js
```diff
@@ -104,7 +104,7 @@ describe('Off Canvas', function() {
           plugin.$element.should.have.class('is-open');
           $('body').should.have.class('is-off-canvas-open');
           done();
-        }, 1);
+        }, 30);
```
ncoden: See https://github.com/zurb/foundation-sites/pull/10943#discussion_r168351520

I think we need a timeout here because the opened.zf.offcanvas event is triggered before the class and everything is set, which is not a good thing (because the offcanvas is not fully opened).

We should resolve that instead of putting a workarounds in the tests.

ncoden:
Please provide a full list of the changes and their reasons. In each commit (give explainaitions, references, links to related issues/PR/commits...) _and_ in the PR description.

test/javascript/components/offcanvas.js
DanielRuf: Definitely.

DanielRuf:
> Please provide a full list of the changes and their reasons. In each commit (give explainaitions, references, links to related issues/PR/commits...) and in the PR description.

Ok, will do so late this day. It's currently 2 am here ;-)

ncoden:
> Ok, will do so late this day. It's currently 2 am here ;-)

Same ;-)

DanielRuf:
Resolve https://travis-ci.org/zurb/foundation-sites/builds/341697796#L2882 in another PR with the creator of the unit test?

test/javascript/util/core.js
DanielRuf: phantomjs 1.9 or 2? Not sure as Chrome stable is used for the new test suite and this is more stable than the old phantomjs and should behave correctly. Not sure about the other browsers from Browserstack.

ncoden:
> Resolve https://travis-ci.org/zurb/foundation-sites/builds/341697796#L2882 in another PR with the creator of the unit test?

Agreed

test/javascript/util/box.js
DanielRuf: Take a look at the previous Travis builds. IE11 glitches when we use rem and uses 99.99994... instead of 100. These are rounding bugs that we also see in production on normal websites when we use rem units.

ncoden: When we use rem units on what ? Dimensions here are `height:100px; width:200px;`.

test/javascript/util/core.js
ncoden: phantomjs `v2.1.1`

test/javascript/util/box.js
DanielRuf: I can revert this but still Browserstack reports 99.9999… px instead of 100 so it fails there.

DanielRuf: See https://travis-ci.org/zurb/foundation-sites/builds/341688519#L3786

DanielRuf: https://github.com/DURAARK/wiregen/blob/91e9642bc63417d32c9adcdfa5b9e664035cc556/extract_floorplan/EXTRACT.html

Should we really ignore this bug in IE11 or handle it somehow? Either we need a change in the unit test to catch it or find another solution.

DanielRuf: I'll revert it in about an hour. It seems the unit test was so far not looked at in detail as it was always like this.

https://travis-ci.org/zurb/foundation-sites/builds/326776466#L2881

DanielRuf: I'll try to reproduce it in our own Browserstack instance with a minimal test. I still gues that it is related to some specific css property as we often see this in production when using Foundation Sites.

js/foundation.offcanvas.js
```diff
@@ -328,6 +324,12 @@ class OffCanvas extends Plugin {
       this.$overlay.addClass('is-closable');
     }
 
+    /**
+     * Fires when the off-canvas menu opens.
+     * @event OffCanvas#opened
+     */
+    this.$element.trigger('opened.zf.offcanvas');
+
```
ncoden: I think this should be at the very end, after every class/attributes has been set, so after `this._addContentClasses`. Look at the others components for reference.

ncoden (commented):
Also does IE 11 works with https://github.com/zurb/foundation-sites/pull/10944/commits/3e86d6e396c9534a0da25841e3c5cf758a89302b ? If yes, what was the problem ?

js/foundation.offcanvas.js
DanielRuf: Also after the return statement which closes it early?

DanielRuf: > Also does IE 11 works with 3e86d6e ? If yes, what was the problem ?

No, I just reverted because it clearly makes not much sense when it is a browser bug ;-)
Not reproducible with IE11 in our own browserstack instance

DanielRuf: https://codepen.io/DanielRuf/pen/XZeqjO was used for testing

DanielRuf: > Also after the return statement which closes it early?

Well, I guess yes.

ncoden: > Also after the return statement which closes it early?

The return statement is inside the `transitionend` event. It has no impact on the `opaned.zf.offcanvas` and runs way after it.

DanielRuf: Right, saw it also. The best and cleanest solution is to trigger the event always after adding all needed classes as triggering events and adding classes (when the event is triggered before this) produces race conditions as far as I can see.

DanielRuf:
> Also does IE 11 works with 3e86d6e ? If yes, what was the problem ?

I guess the problem was always there but not handled / solved
https://github.com/zurb/foundation-sites/pull/10944#discussion_r168386781

https://travis-ci.org/zurb/foundation-sites/builds/326776466#L2881
https://travis-ci.org/zurb/foundation-sites/builds/267290437#L2623

Not sure which exact IE11 release and Windows KB patches are used on the used browserstack instance.

DanielRuf:
Added some information in the initial comment. Please let me know if this is ok. Or should I split the commits into several branches and PRs?

ncoden:
There's still race conditions with dynamically added component (https://github.com/zurb/foundation-sites/pull/10947) but I think it's good to go.

DanielRuf:
> trigger opened.zf.offcanvas at the end to prevent race conditions

Probably I should move this to another PR. Here we have just a change for the offcanvas component itself to prevent race conditions in it directly. I did not tackle the other components (which is the plan for the other single PRs, I have to check all components and split it into work packages for the PRs).

ncoden:
Agreed

DanielRuf:
The initial comment was updated. Please let me know if something is missing @ncoden.

DanielRuf:
Ah, right =)

DanielRuf:
Done and also created https://github.com/zurb/foundation-sites/pull/10967

ncoden:
I updated `package-lock` for `mocha-headless-chrome` :)

Issue #9047, opened 2016-07-25
[Sticky] Bug does not play nicely with ajax content


phifa:
I think I tried almost all foundation plugins and got them all working with ajax content, except for sticky. You can see it on this page: 
http://p352632.mittwaldserver.info/ 

You can see sticky is working. But when you first visit this site and go via the menu to the sticky plugin, the plugin does not get initialized. 
http://p352632.mittwaldserver.info/page2.html

This is how I initialze the plugins: 

```

// Init on document ready
$(document).ready(function(){
    pageInit(document);
});

// Init with Ajax
$(function(){
    'use strict';
    var $page = $('#page'),
        options = {
            debug: false,
            prefetch: true,
            blacklist: '.no-smoothState',
            cacheLength: 7,
            onStart: {
                duration: 200, // Duration of our animation
                render: function ($container) {
                    // Add your CSS animation reversing class
                    $container.addClass('is-exiting is-inprogress');
                    // Restart your animation
                    smoothState.restartCSSAnimations();
                }
            },
            onProgress: {
                duration: 0,
                render: function ($container) {
                }
            },
            onReady: {
                duration: 0,
                render: function ($container, $newContent) {
                    // Remove your CSS animation reversing class
                    $container.removeClass('is-exiting');
                    // Inject the new content
                    $container.html($newContent);
                }
            },
            onAfter: function($container, $newContent) {
                pageInit($newContent);
            }
        },
        smoothState = $page.smoothState(options).data('smoothState');
});


function pageInit(scope) {
    // Init Foundation Plugins
    $(scope).foundation();
}


```

alanontheweb:
I second that. Same exact issue here.

phifa:
anybody from the ZURB / Sticky team can reproduce this issue?

phifa:
@rafibomb or anybody else? Nothing?

kball:
Hi @phifa - thank you for the example, I see the bug happening - I'd like to dig in and solve it.  Is this example in a repo I can pull down and do some debugging on with a local copy of Foundation?

phifa:
[smoothfoundation.zip](https://github.com/zurb/foundation-sites/files/502252/smoothfoundation.zip)
@kball No repo thus far. I attached the files as a zip for you. Thanks for checking in!

kball:
This is due to the same thing as https://github.com/zurb/foundation-sites/issues/8394

[...]

PR #10947, opened 2018-02-15
fix: support dynamically added components #8016

### LINKED ISSUE #8016: [Magellan] active class not being applied when content is loaded dynamically
**How can we reproduce this bug**
1. Create a magellan nav with sticky
2. Populate content dynamically (angular)
3. Initialize foundation js after content is loaded

**What did you expect to happen** 
Magellan should apply an 'active' class when the element is in view

**What happened instead** 
No active classes are applied to the menu. 

I'm using F6 in an angular application and i'm trying to use ng-repeat with the requisite magellan markup. I'm initializing the foundation js after content is loaded. plugins (sticky, accordion, dropdown) seem to work, just not magellan.

Codepen: http://codepen.io/npapazian/pen/MKVGrr

DanielRuf:
We have to check for `document.readyState === "complete"` when the window and content is already loaded and more content is loaded afterwards that is needed for the magellan and sticky solution.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -83,7 +83,7 @@ class Magellan extends Plugin {
           duration: _this.options.animationDuration,
           easing:   _this.options.animationEasing
         };
-    $(window).one('load', function(){
+    if (document.readyState === 'complete') {
       if(_this.options.deepLinking){
         if(location.hash){
           _this.scrollToLoc(location.hash);
@@ -91,7 +91,17 @@ class Magellan extends Plugin {
       }
       _this.calcPoints();
       _this._updateActive();
-    });
+    } else {
+      $(window).one('load', function(){
+        if(_this.options.deepLinking){
+          if(location.hash){
+            _this.scrollToLoc(location.hash);
+          }
+        }
+        _this.calcPoints();
+        _this._updateActive();
+      });
+    }
 
     this.$element.on({
       'resizeme.zf.trigger': this.reflow.bind(this),
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -159,6 +159,14 @@ class OffCanvas extends Plugin {
   _setMQChecker() {
     var _this = this;
 
+    if (document.readyState === 'complete') {
+      if (MediaQuery.atLeast(_this.options.revealOn)) {
+        _this.reveal(true);
+      } else {
+        _this.reveal(false);
+      }
+    }
+
     $(window).on('changed.zf.mediaquery', function() {
       if (MediaQuery.atLeast(_this.options.revealOn)) {
         _this.reveal(true);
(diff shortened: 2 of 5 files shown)


DanielRuf:
Working codepen: https://codepen.io/DanielRuf/pen/LQOWmE

ncoden:
I just found the same issue with `Magellan`. There is probably others issues like this across all components. :/

DanielRuf:
Right, we may have to add this to some more components.

DanielRuf:
I'll check the other components, prepare an issue and collect them there and link the PRs then (I would rather go step for step or component for component to have better and more atomar changes which is easier for bisecting and regression testing imo).

ncoden:
I was going to take care of this and open a PR but ok, have fun ;)

DanielRuf:
> keeping everything in one PR is easier to manage.
Agreed. I'll prepare the other commits then. If you want you can also push to my branch. Or do we already have some PRs for the other components?

ncoden:
>  Or do we already have some PRs for the other components?

I don't think. There was PRs before that resolved some cases manually (and sometimes badly), but everything is merged by now.

DanielRuf:
Also see https://github.com/zurb/foundation-sites/issues/9047

ncoden:
Thanks.

DanielRuf:
Pushed a new commit. Did not find more occurences where we wait for the load event.

ncoden:
@DanielRuf Seen. LGTM but require some cleaning as said at https://github.com/zurb/foundation-sites/pull/10988#discussion_r171405136, maybe in an other PR (that or in that PR but with a squash after). Do you want to take care of this or should I ?

DanielRuf:
> LGTM but require some cleaning as said at #10988 (comment), maybe in an other PR (that or in that PR but with a squash after). Do you want to take care of this or should I ?

Would be glad if I couls get some assistance here so feel free to prepare a PR with the needed changes to make them all consistent then.

js/foundation.magellan.js
```diff
@@ -83,15 +83,25 @@ class Magellan extends Plugin {
...
+      $(window).one('load', function(){
+        if(_this.options.deepLinking){
+          if(location.hash){
+            _this.scrollToLoc(location.hash);
+          }
+        }
+        _this.calcPoints();
+        _this._updateActive();
+      });
+    }
```
ncoden: > When  a Magellan component is dynamically created, the window DO NOT scroll  to the target corresponding to the location hash as the magellan did not  created the target and should appear without effects on it after the  page already loaded.
>
> -- https://github.com/zurb/foundation-sites/pull/11077#issue-175772567

ncoden (changes requested):

ncoden:
Closing in favor of https://github.com/zurb/foundation-sites/pull/11077

Issue #10892, opened 2018-01-29
Foundation.toggler.js is causing: `TypeError: Cannot read property '0' of undefined` when using ES6 modules


aaronjpitts:
When importing foundation plugins using ES6 modules, foundation.toggler.js is causing: `TypeError: Cannot read property '0' of undefined` if input is undefined on line 55 in js/foundation.toggler.js. This is the simple fix:

```
  if (input) {
    // Allow for a . at the beginning of the string
    this.className = input[0] === '.' ? input.slice(1) : input;
  }
```

ncoden:
Hi @aaronjpitts.

Thanks for the issue report and the fix. However I would like to make sure that this fix a correct usage of `Toggler`. Could you please describe your usage case ?

aaronjpitts:
Sure, so I am creating a Vue plugin which simple creates directives for all Foundation javascript widgets (https://github.com/aaronjpitts/vf-directives), I am implementing the toggler Vue directive below.

```
import $ from 'jquery'
import { Foundation } from 'foundation-sites/js/foundation.core'
import { Toggler } from 'foundation-sites/js/foundation.toggler'

Foundation.plugin(Toggler, 'Toggler')

export default {
  install (Vue) {
    Vue.directive('toggler', {
      inserted (el) {
        el.vfToggler = new Foundation.Toggler($(el))
      },
      unbind (el) {
        el.vfToggler.destroy()
      }
    })
  }
}
```

kball:
I think the fix is not correct, because the use case here is not quite right - toggler as designed is supposed to listen for the toggle event and simply toggle a class. Calling it without a class as an argument should not be valid!

However, I *do* think it should accept that class as an option and not only as a data attribute, which would allow us to wrap it up more nicely in a vue directive

DanielRuf:
@ncoden Is this resolved by https://github.com/zurb/foundation-sites/pull/10952 which is merged?

ncoden:
Yes, thanks for the bumb.

PR #10952, opened 2018-02-17
fix: make Toggler target an option #10892

ncoden:
Related to: https://github.com/zurb/foundation-sites/issues/10892

> ...toggler as designed is supposed to listen for the toggle event and simply toggle a class. Calling it without a class as an argument should not be valid!
>
> However, I do think it should accept that class as an option and not only as a data attribute, which would allow us to wrap it up more nicely in a vue directive
>
> -- @kball https://github.com/zurb/foundation-sites/issues/10892#issuecomment-362864469

### DIFF
--- a/js/foundation.toggler.js
+++ b/js/foundation.toggler.js
@@ -50,7 +50,10 @@ class Toggler extends Plugin {
     }
     // Otherwise, parse toggle class
     else {
-      input = this.$element.data('toggler');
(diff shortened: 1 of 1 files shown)


ncoden:
@colin-marshall Would you have some time to review this ? :)

DanielRuf (approved):
LGTM

PR #10957, opened 2018-02-19
Add check for changes to title attribute before showing tooltip

### LINKED ISSUE #10898: Changing the 'title' property of a tooltip element does not change the tooltip

1. Create an element with a tooltip.
2. Change the `title` property of the element.
3. Hover over the element for the tooltip.

#### What should happen:

The new title should be used for the tooltip.

#### What happened instead:

The old title is used for the tooltip.

Primarily Chrome, although this seems like the defined behavior rather than a browser specific bug.

Using version 6. This is a new project so it is a fresh download.

#### Test case link:

https://codepen.io/tedivm/pen/KQpxLd

colin-marshall:
Resolves #10898.

Adds a check to see if the title attribute is empty before showing the tooltip. If it's not empty, the `tipText` option is updated with the new title text, the tooltip element is updated with the contents of `tipText`, and the title attribute is reset.

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -135,6 +135,14 @@ class Tooltip extends Positionable {
     }
 
     var _this = this;
+
+    // Update Tooltip text if title attribute has changed since init 
+    if (this.$element.attr('title') !== '') {
+      this.options.tipText = this.$element.attr('title');
+      this.template.text(this.options.tipText);
+      this.$element.attr('title', '');
+    }
+
     this.template.css('visibility', 'hidden').show();
     this._setPosition();
     this.template.removeClass('top bottom left right').addClass(this.position)
@@ -145,7 +153,6 @@ class Tooltip extends Positionable {
      * @event Closeme#tooltip
      */
     this.$element.trigger('closeme.zf.tooltip', this.template.attr('id'));
-
 
     this.template.attr({
       'data-is-active': true,


DanielRuf:
Is this for dynamically added title attributes?

js/foundation.tooltip.js
```diff
@@ -135,6 +135,14 @@ class Tooltip extends Positionable {
     }
 
     var _this = this;
+
+    // Update Tooltip text if title attribute has changed since init 
+    if (this.$element.attr('title') !== '') {
+      this.options.tipText = this.$element.attr('title');
+      this.template.text(this.options.tipText);
+      this.$element.attr('title', '');
+    }
```
ncoden: This seems a bit duplicated with the initialization. Also, this does not support HTML.
I think we could factorize this and

https://github.com/zurb/foundation-sites/blob/53f0ee83aa0963f0dde56811277f03cfaa29c130/js/foundation.tooltip.js#L52-L60

.... as a `_reflowTitle` function, and maybe add a `_reflow` also.

ncoden (changes requested):

ncoden:
@DanielRuf Dynamically added and updated title attributes.

ncoden:
Not a bugfix but a feature as we never explicitly said that we support dynamically added/updated properties.

js/foundation.tooltip.js
colin-marshall: @ncoden I agree. If `allowHTML` is true, can the HTML be set through the title attribute or does that only work by setting it through the `tipText` option?

ncoden: For a simple API, I would say that `title` and `tipText` should be considered as aliases. But using `title` for html content is not a good practice and we may not allow that....

ncoden:
> For a simple API, I would say that title and tipText should be considered as aliases. But using title for html content is not a good practice and we may not allow that....

@colin-marshall Any thoughts on this ?

PR #10960, opened 2018-02-20
Added 'disabled' styling to switches.

jellisii:
I've run up on some situations where I need to fix a switch into a position.  This adds CSS to reflect the disabled attribute on the input.


DanielRuf:
id, for and name should be different for the second example.

Please use the master branch for changes in the docs.

ncoden:
@DanielRuf We do not want to change the docs on `master` as it describe a new feature on `develop`.

scss/components/_switch.scss
```diff
@@ -208,6 +208,10 @@ $switch-paddle-transition: all 0.25s ease-out !default;
   // <input> element
   .switch-input {
     @include switch-input;
+      // indicate a disabled switch
+    &:disabled ~ label {
```
ncoden: I don't like this selector (non BEM-like) but this is the CSS architecture of the component, so good enough.

scss/components/_switch.scss
```diff
@@ -208,6 +208,10 @@ $switch-paddle-transition: all 0.25s ease-out !default;
   // <input> element
   .switch-input {
     @include switch-input;
+      // indicate a disabled switch
```
ncoden: This is not aligned ;)

scss/components/_switch.scss
```diff
@@ -208,6 +208,10 @@ $switch-paddle-transition: all 0.25s ease-out !default;
...
+    &:disabled ~ label {
+      cursor: not-allowed;
```
ncoden: You can use `$switch-cursor-disabled` instead of a hard-coded property (many people don't like the `not-allowed` cursor.

Also, you could add `$switch-opacity-disabled` (with `.5` or `.66`) as default so we can recognize a disabled switch without having to move the cursor on it.

ncoden (changes requested):

DanielRuf:
> @DanielRuf We do not want to change the docs on master as it describes a new feature on develop.

Ah ok =) So when it comes with new changes, it should be in the same branch (develop).

ncoden:
If we want to improve the doc of the last version, we push on master so it can be deployed quickly. If we improve/add some docs for features/components of the next version, we keep it on develop and it will made be public when the next version is released.

scss/components/_switch.scss
```diff
@@ -208,6 +216,11 @@ $switch-paddle-transition: all 0.25s ease-out !default;
   // <input> element
   .switch-input {
     @include switch-input;
+    // indicate a disabled switch
+    &:disabled ~ .switch-paddle {
+      cursor: $switch-cursor-disabled;
+      opacity: $switch-opacity-disabled;
+    }
```
ncoden: Sorry for not putting that in my first review, but could you move that to `@mixin switch-paddle`, just after the `input:checked` styles ?

ncoden (commented):
Thanks for the changes :)

ncoden (approved):
👍  Thanks @jellisii. I'll resolve conflicts.

jellisii:
Thank you @ncoden and @DanielRuf .

ncoden:
@jellisii  For your next pull request, you can push your work to a dedicated branch (like `feat/add-switch-disabled-state` and create a pull request from it. So you can continue working on your develop branch without polluting your existing pull requests.

ncoden:
Preview:
screen shot 2018-02-26 at 23 44 11

ncoden:
Everything looks good so I'm merging this.
Thanks again @jellisii 👍 .

Issue #10742, opened 2017-10-20
[6.4.3] foundation.scss - "$xy-grid: false" parameter missing in foundation-everything mixin


franklang:
Current source :
@mixin foundation-everything(
  $flex: true,
  $prototype: false
) {

Corrected source :
@mixin foundation-everything(
  $flex: true,
  $prototype: false,
  $xy-grid: false
) {

DanielRuf:
Thanks for reporting. Will check that soon.

DanielRuf:
`$xy-grid: true` is the default in Foundation Sites 6.4.

PR #10962, opened 2018-02-20
docs: the flexgrid is used if $xy-grid is set to false

### LINKED ISSUE #10425: [Flex Grid] .row and .column classes aren't receiving default styles

Simply create a div.row with a nested div.column. Notice that there is no longer a max-width set by default.

#### What should happen:

The div.row should have the max-width equal to whatever is defined as $global-width.

#### What happened instead:

Neither the div.row nor div.column are receiving any styles like they were in v6.3.

Chrome on Macbook Pro.

#### Test case link:

- [CodePen with Foundation 6.4.1, XY Flexbox grid and MotionUI](https://codepen.io/joshdean/pen/QgRKJK)

DanielRuf:
This PR closes #10425.

The change makes clear that if `xy-grid` is set to default the `flex-grid` is used as fallback.

Currently we have the following in scss/foundation.scss
```scss
 @include foundation-global-styles;
  @if not $flex {
    @include foundation-grid;
  }
  @else {
    @if $xy-grid {
      @include foundation-xy-grid-classes;
    }
    @else {
      @include foundation-flex-grid;
    }
  }
```

So it also seems to be the case that the float grid is used if `$flex: false` is set.


docs/pages/xy-grid.md
```diff
@@ -26,6 +26,8 @@ The XY grid is supported in Chrome, Firefox, Safari 6+, IE10+, iOS 7+, and Andro
 
 **XY grid is the default Foundation grid**. It is present by default in `foundation.css` of CDN link or package managers. In Sass, it will be generated by default by `@include foundation-everything` (unless `$global-flex` or `$xy-grid` are set to false).
 
+If `$xy-grid: false` is set, the flex grid is used instead.
```
ncoden: I think we should rather reformulate the whole sentence. Like:

> ... (unless `$xy-grid` or `$global-flex` is set to false, it will then generate the Flex Grid or the Float Grid depending on `$global-flex`).

Or

> ... (unless `$xy-grid` or `$global-flex` is set to false, it will then generate the Flex Grid if `$global-flex` is true, or the Float Grid otherwise).

DanielRuf: I'll think about it and push a new commit tomorrow. Thanks for the feedback.

DanielRuf:
Also saw https://github.com/zurb/foundation-sites/issues/10742

ncoden:
@DanielRuf You'll create a PR i guess ?

ncoden (approved):
👍

DanielRuf:
> @DanielRuf You'll create a PR i guess ?

In a separate PR I guess as this would be an actual code change.

PR #10964, opened 2018-02-21
Update off-canvas docs regarding nested limitation

### LINKED ISSUE #10328: [Off-canvas] Docs - open nested off-canvas push does overlap instead.
Click example here: 
http://foundation.zurb.com/sites/docs/off-canvas.html#nested-off-canvas
open nested off-canvas push

Overlaps instead of pushes

SassNinja:
This PR fixes https://github.com/zurb/foundation-sites/issues/10328

To make it more clear only overlap transition is possible when using a nested off-canvas, I've added a further callout (apart from the one in the nested section) right in the transition section.


ncoden:
@SassNinja You are targeting `develop` but this behavior is present in `6.4.3`, right ?

SassNinja:
@ncoden I thought it will be part of 6.4.4 since it's no critical issue. But sure I may target master instead.

Is it safe to simply change the base here on github or is it better to recreate the branch starting from master?

ncoden:
So I tested and, as I expected, this only change the branch your PR is compared to. So I would advice you to rebase your changes on `master` (with an other branch) and open an other PR.

PR #10966, opened 2018-02-21
Update off-canvas docs

SassNinja:
Fixes https://github.com/zurb/foundation-sites/issues/10328

To make it more clear only overlap transition is possible when using a nested off-canvas, I've added a further callout (apart from the one in the nested section) right in the transition section.

@ncoden this time base is master branch ;) (in contrast to https://github.com/zurb/foundation-sites/pull/10964)


ncoden (approved):
👍 LGTM, thanks

PR #10967, opened 2018-02-21
fix: trigger opened.zf.offcanvas at the end to prevent race conditions

DanielRuf:
This prevents a race condition and triggers the event after all other things are done.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -280,6 +280,7 @@ class OffCanvas extends Plugin {
    * @param {Object} event - Event object passed from listener.
    * @param {jQuery} trigger - element that triggered the off-canvas to open.
    * @fires OffCanvas#opened
+   * @todo also trigger 'open' event?
    */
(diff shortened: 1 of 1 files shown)


js/foundation.offcanvas.js
```diff
@@ -280,6 +280,7 @@ class OffCanvas extends Plugin {
    * @param {Object} event - Event object passed from listener.
    * @param {jQuery} trigger - element that triggered the off-canvas to open.
    * @fires OffCanvas#opened
+   * @todo also trigger 'open' event?
```
ncoden: Part of https://github.com/zurb/foundation-sites/pull/10949

ncoden (approved):
👍  Thanks

js/foundation.offcanvas.js
DanielRuf: Right. It's a todo annotation so we don't forget to revisit this (I hope so). =)

PR #10969, opened 2018-02-21
Fix Off-canvas In-canvas feature

### LINKED ISSUE #10585: off-canvas in-canvas: on resize move from off-canvas to in-canvas without adding class .is-closed

1. Step one: go to http://foundation.zurb.com/sites/docs/off-canvas.html, section In-Canvas to Off-Canvas

2. Step two: look at classes for the item that has 
class="off-canvas in-canvas-for-large position-right is-transition-overlap is-closed"

3. Step three: resize viewport to trigger the off-canvas button, open this off-canvas element with button so it looks like in code: 
class="off-canvas in-canvas-for-large position-right is-transition-overlap is-open"

4. Step four: resize the viewport with still open off-canvas menu window to large again, css code should now read:
class="off-canvas in-canvas-for-large position-right is-transition-overlap is-closed"
instead it reads:
class="off-canvas in-canvas-for-large position-right is-transition-overlap is-open"
while the in-canvas-for-large is already in effect, effectively breaking anything that relies on proper is-open/ is-closed state

#### What should happen:
on viewport resize it has to recalculate if it is still off-canvas open or closed

#### What happened instead:
see above

chrome; IOS; 

6.4.3

Main problem I see here that on viewdirection (rotate) change on mobile it breaks layouts;

SassNinja:
Fixes https://github.com/zurb/foundation-sites/issues/10585

To sum up the problem:
Most of the In-canvas feature is done via CSS only. But when opening the Off-canvas on e.g. small viewport (as regular offcanvas) and then doing a window resize to large viewport where it becomes in-canvas, some issues occur (overlay still visible etc.)

This PR fixes this by adding an `inCanvasOn` option. 
To fix existing projects automatically (easy upgrading) I'm extracting the breakpoint name from the `.is-in-canvas-for-[BREAKPOINT]` class (if available) and use it as option if none set.

The docs have been updated accordingly with a callout.


ncoden:
I took a quick look at it in the subway and (obviously) did not understood everything. I'll take an other look tomorrow.

docs/pages/off-canvas.md
```diff
@@ -346,13 +346,17 @@ For an example of off-canvas, checkout this top bar with off-canvas navigation a
 
 ## In-Canvas to Off-Canvas
 
-With this feature you can have a standard page element move off-canvas at a particular breakpoint. Use the new class <code>.in-canvas-for-[BREAKPOINT]</code> for this. This differs from the <a href="#reveal-on-larger-screens">Reveal on Larger Screens</a> feature it doesn't actually open the off-canvas for specific screen sizes but overrides the off-canvas styles so it behaves as a regular page element. This way you can place an element anywhere on the page and move it into off-canvas for e.g. small screens.
+With this feature you can have a standard page element move off-canvas at a particular breakpoint. Use the new class <code>.in-canvas-for-[BREAKPOINT]</code> and the <code>inCanvasOn</code> option for this. This differs from the <a href="#reveal-on-larger-screens">Reveal on Larger Screens</a> feature as it doesn't actually open the off-canvas for specific screen sizes but overrides the off-canvas styles so it behaves as a regular page element. This way you can place an element anywhere on the page and move it into off-canvas for e.g. small screens only.
```
ncoden: You can replace `Use _the new_ class...` btw. It's 10-months old and already in `v6.4.0`. Also, options are the standard and recommanded "way to go" to customize components, so I think you could put it first.

Something like:
> You can make a standard page element move off-canvas at a particular breakpoint with the `inCanvasOn` option (or the `.in-canvas-for-*` classname).

js/foundation.offcanvas.js
```diff
@@ -136,6 +137,20 @@ class OffCanvas extends Plugin {
...
 
+    let inCanvasFor = this.$element.attr('class').match(/\bin-canvas-for-(\w+)/);
+
+    if (!this.options.inCanvasOn && inCanvasFor && inCanvasFor.length === 2) {
+      // Extract breakpoint name from CSS class if no inCanvasOn option set but using .in-canvas-for-[BREAKPOINT]
+      this.options.inCanvasOn = inCanvasFor[1];
+    } else if (this.options.inCanvasOn && !inCanvasFor) {
+      // Add CSS class if only inCanvasOn option set
+      this.$element.addClass(`in-canvas-for-${this.options.inCanvasOn}`);
+    }
```
ncoden: What if:
```html
<div data-off-canvas data-in-canvas-for="medium" class="off-canvas in-canvas-for-large">
```

With that case, the breakpoint is considered as "medium" in Js but "large" in CSS. Here how I see things to ensure a consistent behavior:
* Classname is a (backward-compatible) way to set the option.
* The option also require require the class to be set to work properly.

And about priorities: component class (less common = more specific = prioritized) > component options > default options.

So according to me, the check should be the following (pseudo-code):
```js
// Set the `inCanvasOn` option if a class is found
if (inCanvasSplit.length === 2) {
  this.options.inCanvasOn = inCanvasSplit[1]
}

// Ensure the required class is set
if (this.options.inCanvasOn) {
  this.$element.addClass(`in-canvas-for-$(this.options.inCanvasOn}`);
}
```

js/foundation.offcanvas.js
```diff
@@ -136,6 +137,20 @@ class OffCanvas extends Plugin {
...
+
+    if (this.options.inCanvasOn) {
+      this.isInCanvas = MediaQuery.atLeast(this.options.inCanvasOn);
+    }
```
ncoden: Duplicated with `:178`. You can move this to its own function like `_reflowBreakpoint`.

js/foundation.offcanvas.js
```diff
@@ -157,6 +172,18 @@ class OffCanvas extends Plugin {
...
     }
+
+    if (this.options.inCanvasOn) {
+      $(window).on('changed.zf.mediaquery', () => {
+        if(MediaQuery.atLeast(this.options.inCanvasOn)) {
+          this.close();
+          this.isInCanvas = true;
+        } else {
+          this.isInCanvas = false;
+        }
```
ncoden: Basically:
```js
if (true) {
  true;
} else {
  false;
}
```
Can be factorized ;)

js/foundation.offcanvas.js
```diff
@@ -282,7 +309,7 @@ class OffCanvas extends Plugin {
    * @fires OffCanvas#opened
    */
   open(event, trigger) {
-    if (this.$element.hasClass('is-open') || this.isRevealed) { return; }
+    if (this.$element.hasClass('is-open') || this.isRevealed || this.isInCanvas) { return; }
```
ncoden: hmm Why do we rely on `isInCanvas` or responsive classes instead of calling `.open()` and relying on `.is-open`

ncoden (changes requested):
I finally had the time to review it ;)

SassNinja:
Thank @ncoden for the review and sorry for the late response: have been really busy at work...

I've adjusted the code now according to your feedback.
I really like to priority order you suggested and have implemented it this way:
CSS class > option > default
Besides I've adjusted the docs accordingly and prevented duplicate code by using a `_checkInCanvas()` method.

But regarding your question abt my changes in the `open()` method I haven't understood you.
My intention was to prevent any open action if inCanvas is currently true. So I'm just triggering a return in such a case.
Do you have another solution in mind?

docs/pages/off-canvas.md
SassNinja: Yup the 'new' is definitely not up-to-date anymore – I've adjusted the text. 
I've removed the usage of the `.in-canvas-for-[BREAKPOINT]` class from the text/example and mention it only in a callout anymore because usually only the  JS option should be used.

js/foundation.offcanvas.js
SassNinja: Right, code should be DRY. I've now moved the code into a new function called `_checkInCanvas` because this is most clear imo.

js/foundation.offcanvas.js
SassNinja: Yup didn't notice this while writing it. It's refactored in `_checkInCanvas` now.

js/foundation.offcanvas.js
ncoden: I had some trouble to understand the role of `isInCanvas` here, so 3 weeks later it's even worst :D. I'll try to take a fresh look at this tomorrow.

ncoden:
@SassNinja Could you resolve conflicts ? (merge dev)

SassNinja:
ok, I've merged `develop` into the in-canvas feature branch now

js/foundation.offcanvas.js
SassNinja: @ncoden have you been able to take a look at this last requested change?
Would like to see this PR merged for 6.5.0 :-)

ncoden:
Ok I understand how incanvas works now. So to make the element going "in-canvas" we have `off-canvas.in-canvas-for-#{$name}` that take care of everything. But this is CSS-only so when opening the off-canvas and going back to in-canvas, the state of the Js offCanvas component doesn't change (and the overlay is still visible).

I think we could go further, making the whole inCanvas state controlled in Js (instead of having both CSS and Js media query). Instead of duplicating the `@mixin in-canvas` code across breakpoint, we would have a `.is-in-canvas` modifier added and removed by the Js, like `.is-open`. Also, we do not _close_ the offCanvas when going "in canvas" but _hide it_, so when going back to offCanvas the offCanvas stay open.

@SassNinja What do you think ? Just an idea, this does not have be in this PR.

SassNinja:
Right @ncoden that's exactly how the incanvas is supposed to work!

I like your idea to reduce the code and thought about it.
But I think moving the breakpoint logic from CSS to JS will also cause FOUC.
The incanvas feature is supposed to 'tranform' an off-canvas elem into regular page content and if this styling is dependent on JS there will be delay until the content appears.

So I think it's best way here to not move the breakpoint logic into JS and leave as it is.

ncoden:
>  But I think moving the breakpoint logic from CSS to JS will also cause FOUC.

You're right. I'll try to think of an other solution. If I got nothing before Friday I'll merge it.

SassNinja:
@ncoden regarding you haven't posted an other solution I assume it's good to merge? :-)

ncoden:
@SassNinja Friday is not finished yet ;)

ncoden (approved):
👍 Ok, LGTM as I found nothing better. Thank you @SassNinja ;)

ncoden:
Delaying this to `v6.6.0` because this add a new feature.

PR #10977, opened 2018-02-22
Fix broken link to XY grid

NeonArray:
There was a forward slash in the path which was causing a 404.


ncoden (approved):
👍 Thank you @ocularrhythm

Issue #10820, opened 2017-12-23
show-for() and hide-for() gap at exact width of breakpoint


Lazerproof:
#### How to reproduce this bug:
```
$breakpoints: (
      small: 0,
      medium: 568px,
      xxmedium: 940px,
      large: 1024px,
      xlarge: 1200px,
      xxlarge: 1440px,
);

.a {
  @include hide-for(xxmedium);
}

.b {
  @include show-for(xxmedium);
}
```
At 940px both element are visible. So implementation of hide-for() should be changed to $breakpoint-value - 1.

ncoden:
Thank you @Lazerproof for your bug reports ;) .
I'm working on it.

ncoden:
See https://codepen.io/ncoden/pen/PEbGYR
This happens when the browser font-size is smaller than 16px. The gap of `1/16em` (so `1px`) between the two breakpoints become `0.625px` and the max limit of the lower breakpoint is round to the upper pixel.

I'll search why the breakpoint sizes are set in `em`, and what unit we could use to avoid this.

JeremyEnglert:
@ncoden - think we can squeeze this into the upcoming release?

ncoden:
Yes. Fixing this issue would require to use pixels instead of em/rem for grids. This is a major change.

Issue #10808, opened 2017-12-15
XY-Grid Container don't fit next to each other in Edge


Baedda:
I have got a problem in Microsofts Edge when using the XY Grid with a 13px base font size.

#### How to reproduce this bug:

Use the following Sass settings for the XY grid. It's important to use exactly the same values because with other cases the bug doesn't show up. Settings that I haven't listed below to use their defaults.

1. $global-font-size: 81.25%;
2. $global-width: rem-calc(990);
3. $grid-margin-gutters: rem-calc(30);

Then place two containers next to each other by using  @include xy-cell(6);

#### What should happen:

Now you should see two containers in one line. This is what happens in every browser. But not in Edge.

#### What happened instead:

In Edge there seems to be not enough space so the containers are linearised like on a mobile view. Maybe this is a calculation error in Foundation? Maybe you should always round down the calculated results of the calculated container widths? But maybe its an calculation error in Edge and there won't be any fix? Please help.

#### Foundation Version(s) you are using:

6.4.3

#### Test case link:

https://codepen.io/artursopelnik/pen/YYyXKp

In the code pen I wasn't able to use the Sass version of Foundation. So I use the CSS version and overwrite the Foundation CSS with the CSS that comes from my site to simulate the settings I have mentioned above.

ncoden:
Tested with BrowserStack:
- Firefox 49: works.
- IE11: works.
- Edge 15: works.
- Edge 16: **don't works** (also tested on my own pc).

This comes from Edge 16 that _round_ numbers to 2 decimal places. So for the width, `calc(50% - 2.3075...rem)` is round to `calc(50% - 2.3)`, same for the margins, and _unluckily_ the cell has a too big width.

It may appears only with your configuration because the margins are round more than the width, so they can fit the row. **This works with the default configuration because we are lucky**.

I can't believe we have a bug _that stupid_ in Edge 16 (and not Edge 15 or IE). This bug is so stupid that at first I did not trusted the Edge developer tool and checked the width row in Js (see https://codepen.io/ncoden/pen/baqmgZ).

That being said... The only thing we can do is allowing to disable to conversion to `rem` for grid, so the generated CSS may be pixel-perfect. But I am not sure it worth it for a _custom config_ on a browser _nobody uses_ (0.24% on Edge 16, 1.34% on Edge 15).

ncoden:
Edit: all version of Edge and IE seem to truncate values:
[...]

Issue #10779, opened 2017-11-20
rounding error in grid column width calculation in Firefox?


beda42:
#### How to reproduce this bug:

1. Create a .grid-x with grid-margin-x with 4 columns
2. Open it in Firefox (tested with FF 56 and 57 under Linux)
3. Set default font size to 14.

#### What should happen:
4 columns should be displayed side by side

#### What happened instead:
3 columns are in one row, the last one wraps under the first one

#### Browser(s) and Device(s) tested on:
FF 56 and 57, Linux, two different Linux PCs.

#### Foundation Version(s) you are using:
6.4.3

#### Test case link:
[simple demo](https://codepen.io/beda42/pen/zPRvQJ)

#### Notes:
* Changing the default browser font size to 13 or 15 or zooming in and out leads to the columns realigning properly.
* In Chrome, it works OK all the time.
* When looking at the CSS in the Inspector, the width is calculated like this: `width: calc(25% - 1.875rem);`. If I change the 1.875rem to 1.876rem, the problem goes away. This is why I think this might be a rounding error.
* The same occurs with 5 columns, not with 3.
* Screenshot attached.
![screenshot at 2017-11-20 10-04-29](https://user-images.githubusercontent.com/1674433/33010276-3b28f60a-cdda-11e7-9539-052660ac970e.png)

DanielRuf:
Related https://github.com/zurb/foundation-sites/issues/10808

DanielRuf:
Did you mean https://github.com/zurb/foundation-sites/pull/10978?

ncoden:
Duplicate of https://github.com/zurb/foundation-sites/issues/10808 yes.

Issue #10695, opened 2017-10-04
'show-for-small-only' and 'hide-for-small-only' elements both displaying at a certain width


hofuchi:
only happens for 1px with font-size and minimum font-size set to larger than default

rafibomb:
Can you provide more details on what screen width you are seeing this? Can you link a codepen demo so it can be replicated?

hofuchi:
![untitled](https://user-images.githubusercontent.com/9373942/32677862-bd2f5142-c625-11e7-84c1-694e85460e0c.png)

```
@media screen and (max-width: 39.9375em) {
  .hide-for-small-only {
    display: none !important; } }

@media screen and (max-width: 0em), screen and (min-width: 40em) {
  .show-for-small-only {
    display: none !important; } }
```

Sorry for the late reply

The picture above shows why they both show up at **large font-sizes only**
At exactly 959px screen width with a 24px font-zize, there's a gap where none of the above media queries are active, I'll see about creating a codepen but it might be hard because it's dependent on screen width and font-size

colin-marshall:
@hofuchi would appreciate a codepen if you have a minute to reproduce it for us. Thanks!

hofuchi:
@colin-marshall 
Here you go:
https://codepen.io/theGrahamScanlon/pen/PRqbjK?css-preprocessor=sass

Since I couldn't get `$global-font-size` working, go to your browser settings and set the font-size to 16px and the minimum-font-size to 16px.

In the codepen, resize the width to between 639px and 640px and both elements show up.

![screen shot 2018-03-11 at 10 03 47](https://user-images.githubusercontent.com/9373942/37254952-963092c0-2513-11e8-8898-0905fd358f13.png)

DanielRuf:
> set the font-size to 16px and the minimum-font-size to 16px

This is the default in most browsers.

I guess you mean https://codepen.io/DanielRuf/pen/eMNgyz?css-preprocessor=sass

DanielRuf:
> between 639px and 640px and both elements show up.

Not sure but I can not do this in Chrome.

DanielRuf:
https://github.com/zurb/foundation-sites/blob/4e21634356da296edd2f1c3e65d91140f991008b/dist/css/foundation.css#L5360
https://github.com/zurb/foundation-sites/blob/4e21634356da296edd2f1c3e65d91140f991008b/dist/css/foundation.css#L5364

DanielRuf:
[...]

PR #10978, opened 2018-02-22
Breakpoint Mixin `max-width` Values

millllllz:
# Why?
Browser windows can be fractions of pixels wide, under certain circumstances (e.g. zoom). In these cases, it is not enough to set the `max-width` to the `em` equivalent of one pixel less.

# What changed?
* updated `max-width` calculation
* updated tests to match new output

# Notes
I recognize this is nearly moving the goalposts and doesn't solve the core problem, but in lieu of a exclusive max-width, 4 places of accuracy seems to be the standard and should work for most.

### DIFF
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -77,7 +77,7 @@ $breakpoint-classes: (small medium large) !default;
   // Convert any pixel, rem, or unitless value to em
   $bp: -zf-bp-to-em($bp);
   @if $bp-max {
-    $bp-max: -zf-bp-to-em($bp-max) - .000001;
+    $bp-max: -zf-bp-to-em($bp-max) - .00001;
   }
 
   // Conditions to skip media query creation
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -77,7 +77,7 @@ $breakpoint-classes: (small medium large) !default;
   // Convert any pixel, rem, or unitless value to em
   $bp: -zf-bp-to-em($bp);
   @if $bp-max {
-    $bp-max: -zf-bp-to-em($bp-max) - 0.0001;
+    $bp-max: -zf-bp-to-em($bp-max) - .0001;
   }
 
   // Conditions to skip media query creation
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -77,7 +77,7 @@ $breakpoint-classes: (small medium large) !default;
   // Convert any pixel, rem, or unitless value to em
   $bp: -zf-bp-to-em($bp);
   @if $bp-max {
-    $bp-max: -zf-bp-to-em($bp-max) - (1/16);
+    $bp-max: -zf-bp-to-em($bp-max) - 0.0001;
   }
 
   // Conditions to skip media query creation
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -77,7 +77,7 @@ $breakpoint-classes: (small medium large) !default;
   // Convert any pixel, rem, or unitless value to em
   $bp: -zf-bp-to-em($bp);
   @if $bp-max {
-    $bp-max: -zf-bp-to-em($bp-max) - .0001;
+    $bp-max: -zf-bp-to-em($bp-max) - .000001;
   }
 
   // Conditions to skip media query creation
--- a/test/sass/_breakpoint.scss
+++ b/test/sass/_breakpoint.scss
@@ -24,10 +24,10 @@
 
   @include test('Breakpoint (Only Range) [function]') {
     $test: breakpoint(medium only);
-    $expect: '(min-width: 40em) and (max-width: 63.9999em)';
+    $expect: '(min-width: 40em) and (max-width: 63.99999em)';
 
     $test-lowest: breakpoint(small only);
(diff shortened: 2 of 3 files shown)


ncoden:
This should fix:
* https://github.com/zurb/foundation-sites/issues/10820 show-for() and hide-for() gap at exact width of breakpoint

Related to:
* https://github.com/zurb/foundation-sites/issues/10808 XY-Grid Container doesn't fit next to each other in Edge

scss/util/_breakpoint.scss
```diff
@@ -77,7 +77,7 @@ $breakpoint-classes: (small medium large) !default;
   // Convert any pixel, rem, or unitless value to em
   $bp: -zf-bp-to-em($bp);
   @if $bp-max {
-    $bp-max: -zf-bp-to-em($bp-max) - (1/16);
+    $bp-max: -zf-bp-to-em($bp-max) - 0.0001;
```
ncoden: Why `0.0001` ?

Side note, numbers < 1 are usually written without the first `0` in CSS/Sass. Like `.0001`

test/sass/_breakpoint.scss
```diff
@@ -24,10 +24,10 @@
 
   @include test('Breakpoint (Only Range) [function]') {
     $test: breakpoint(medium only);
-    $expect: '(min-width: 40em) and (max-width: 63.9375em)';
+    $expect: '(min-width: 40em) and (max-width: 63.9999em)';
```
ncoden: There is some tests to do in IE/Edge to ensure this works properly (or at least better than now). See #10808

ncoden (commented):
Thank you @JasonMiller. There is some tests and investigations to do (like why did we used `1px` in the first place and not lower), but it's a nice catch.

See also the remarks below.

scss/util/_breakpoint.scss
JasonMiller: Since all the `em` math works in 16ths, values end up at 4 decimals. I can remove the leading zero. Do you have a suggestion other than `.0001`?

test/sass/_breakpoint.scss
JasonMiller: Looking at #10808, wouldn't `.9375` be just as much a problem as `.9999`? I don't see this as introducing new issues. Thoughts?

ncoden: True but I prefer testing. Also, is there any reason for `.9999` and not `.999999` ?

JasonMiller: This is from my [previous comment](https://github.com/zurb/foundation-sites/pull/10978#discussion_r170259770), but since all the em math works in 16ths, values end up at 4 decimals. Do you have a suggestion other than `.0001` which would result in `.9999`?

ncoden: According to the W3C CSS3 reference:

> CSS theoretically supports infinite precision and infinite ranges for all value types; however in reality implementations have finite capacity. UAs should support reasonably useful ranges and precisions.
> -- https://www.w3.org/TR/css-values-3/#numeric-types

It is common is CSS to write values with 6 decimals as it is precise enough to ensure a sub-pixel difference. But when dealing with breakpoints I'm not sure it's what we want.

JasonMiller: I understand your concern. What can I do to move this along?

JasonMiller:
@ncoden, I'm seeing TravisCI failures that do not pertain to my deltas. Please advise.

ncoden:
@JasonMiller We know about the unit test fails. We are resolving it.

test/sass/_breakpoint.scss
JasonMiller: @ncoden is there a testing procedure you'd like me to follow here?

ncoden:
@JasonMiller I just have some tests to do across browsers, when I have the time for.

test/sass/_breakpoint.scss
ncoden: @JasonMiller I just have some tests to do across browsers, when I have the time for.

DanielRuf:
What is the current state? The change would solve calculation bugs in Firefox and other browsers.

ncoden:
The current state was "waiting for @ncoden to remember this PR". Thanks for the recall.

DanielRuf:
related issues
https://github.com/zurb/foundation-sites/issues/10779
https://github.com/zurb/foundation-sites/issues/10695

ncoden:
I tested it across browsers (Edge 16 and IE 9), it works fine and does not introduce new bugs. I also changed the precision from 4 to 6 decimals like discussed before, and applied the same precision to visibility classes.

ncoden:
> I was wrong, Sass actually has a default precision of 5 digits. Bootstrap-sass uses 8 digits but require to update the Sass compiler options.
> See: https://github.com/twbs/bootstrap-sass#sass-number-precision

I updated tests too.

ncoden:
Seems good so far. Thank you @JasonMiller for the PR 👍

JasonMiller:
@ncoden, sorry to re-open this, but I'm seeing some issues with Chrome at exact breakpoint values, potentially related to rounding. If you take a look at [this codepen](https://codepen.io/millllllz/pen/rKavaM), and resize your screen to `1024px`, you'll see the 4-digit precision rule is still a match. This would be a problem for rules that use the `down` keyword, e.g. `medium down`. Did you see any of this in your testing? If you can confirm it's an issue, I can open a new pull request.

Issue #10976, opened 2018-02-22
Breakpoint mixin omits print media type when using a named breakpoint and direction


rossta:
When using the breakpoint mixin with a named breakpoint *and direction*, print is omitted from the media query when it should be included.
 
#### How to reproduce this bug:

Use the breakpoint for a named media query less-than-or-equal to the print breakpoint and a direction:

```sass
.test {
  @include breakpoint(large) {
    display: none;
  }
}
```

#### What should happen:

When using a named media query (no direction specified) less-than-or-equal to the print breakpoint, the print media type is included in the media query:

```sass
.test {
  @include breakpoint(large) {
    display: none;
  }
}
```
Results in: 
```css
@media print, screen and (min-width: 1024px) {
  .test {
     display: none;
  }
}
```
So, I would expect the following sass
```sass
.test {
  @include breakpoint(large up) {
    display: none;
  }
}
```
to produce the same CSS:
```css
@media print, screen and (min-width: 1024px) {
  .test {
     display: none;
  }
}
```
#### What happened instead:

Though the arguments `large` and `large up` produce the same output for screen, the print media type is omitted for `large up`:
```css
@media screen and (min-width: 1024px) {
  .test {
     display: none;
  }
}
```
The same behavior exists for `small up`, `small down`, `medium up`, etc.

I believe this is because the breakpoint mixin does a key-value lookup based on the [whole string of the $value argument](https://github.com/zurb/foundation-sites/blob/f7aab3d77f3c951f72378aeadd9883a8323adfb6/scss/util/_breakpoint.scss#L150) unlike the breakpoint *function*, which [splits the string on the first word of its $value argument](https://github.com/zurb/foundation-sites/blob/f7aab3d77f3c951f72378aeadd9883a8323adfb6/scss/util/_breakpoint.scss#L43). The result is a named query is not found in the key lookup (since the expected key is `'large up'`, for example, so the [logic to include the print media type here](https://github.com/zurb/foundation-sites/blob/f7aab3d77f3c951f72378aeadd9883a8323adfb6/scss/util/_breakpoint.scss#L169) fails.

I believe the fix is to determine the named media query in the breakpoint mixin based on the first word only of the `$value` argument and possibly to extract the direction for use in the logic to determine whether to include the print media type.

#### Foundation Version(s) you are using:

Foundation 6.4.0

DanielRuf:
Not reproducible with Foudnation Sites 6.4.3

It generates

```css
[...]

PR #10979, opened 2018-02-22
fix: add the print media query for the right media query combinations

DanielRuf:
This covers more `down`, `up` and `only` cases.

Reference: #10976

### DIFF
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -147,8 +147,10 @@ $breakpoint-classes: (small medium large) !default;
   @for $i from 1 through length($values) {
     $value: nth($values, $i);
     $str: breakpoint($value);
-    $bp: index($-zf-breakpoints-keys, $value);
+    $bp: index($-zf-breakpoints-keys, nth($value, 1));
     $pbp: index($-zf-breakpoints-keys, $print-breakpoint);
+    // Direction of media query (up, down, or only)
+    $dir: if(length($value) > 1, nth($value, 2), up);
 
     $old-zf-size: null;
 
@@ -166,7 +168,7 @@ $breakpoint-classes: (small medium large) !default;
     // Otherwise, wrap the content in a media query
     @else {
       // For named breakpoints less than or equal to $print-breakpoint, add print to the media types
-      @if $bp != null and $bp <= $pbp {
+      @if ($bp != null and $bp == $pbp and ($dir == '' or $dir == 'up' or $dir == 'only')) or ($bp != null and $bp < $pbp and ($dir == '' or $dir == 'up'))  or ($bp != null and $bp >= $pbp and ($dir == 'down')) {
         @media print, screen and #{$str} {
           @content;
         }
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -168,7 +168,9 @@ $breakpoint-classes: (small medium large) !default;
     // Otherwise, wrap the content in a media query
     @else {
       // For named breakpoints less than or equal to $print-breakpoint, add print to the media types
-      @if ($bp != null and $bp == $pbp and ($dir == '' or $dir == 'up' or $dir == 'only')) or ($bp != null and $bp < $pbp and ($dir == '' or $dir == 'up'))  or ($bp != null and $bp >= $pbp and ($dir == 'down')) {
+      // generate print if the breakpoint affects the print-breakpoint (or smaller).
+      // This means the current condition only needs to be extended so 'down' always generates print.
+      @if $bp != null and ($bp <= $pbp or $dir == down) {
         @media print, screen and #{$str} {
           @content;
         }


scss/util/_breakpoint.scss
```diff
@@ -166,7 +168,7 @@ $breakpoint-classes: (small medium large) !default;
     // Otherwise, wrap the content in a media query
     @else {
       // For named breakpoints less than or equal to $print-breakpoint, add print to the media types
-      @if $bp != null and $bp <= $pbp {
+      @if ($bp != null and $bp == $pbp and ($dir == '' or $dir == 'up' or $dir == 'only')) or ($bp != null and $bp < $pbp and ($dir == '' or $dir == 'up'))  or ($bp != null and $bp >= $pbp and ($dir == 'down')) {
```
ncoden: I think we should move this to a cleaner "breakpoint compairaison" function. Putting all the cases here in a `if` is really confusing.

DanielRuf: Definitely. Could need some assistance here as I've not often written mixins and functions.

ncoden:
Also, there is a lot of cleaning to do in the `breakpoint` function and mixin. See https://github.com/zurb/foundation-sites/pull/9661/files ;)

scss/util/_breakpoint.scss
```diff
@@ -147,8 +147,10 @@ $breakpoint-classes: (small medium large) !default;
   @for $i from 1 through length($values) {
     $value: nth($values, $i);
     $str: breakpoint($value);
-    $bp: index($-zf-breakpoints-keys, $value);
+    $bp: index($-zf-breakpoints-keys, nth($value, 1));
```
rossta: 👍 Yes, this is consistent with the fix I had in mind

scss/util/_breakpoint.scss
ncoden: I'll take care of this.

DanielRuf: Thanks =)

ncoden: ... this weekend :)

DanielRuf: No problem.

scss/util/_breakpoint.scss
SassNinja: that's the same fix I was thinking abt ;)

scss/util/_breakpoint.scss
SassNinja: @DanielRuf @ncoden I'm not sure if this line has to be changed at all... what is the goal?

The comment says 'for named breakpoints less than or equal to $print-breakpoint' what I interpreted as starting point . The original `@include breakpoint(large)` was actually a `@include breakpoint(large up)` and thus also affects breakpoints greater than large.

So is the goal that `breakpoint(large only)` and `breakpoint(large down)` do affect print but `breakpoint(large)` does not? (although this would be a breaking change)
If not the condition `$bp != null and $bp <= $pbp` doesn't need to be changed imo

ncoden: I had some trouble to retrieve the problem here and the purpose of this change, but I think I get it. 

If I understand `$print-breakpoint` correctly, it must be used to prevent `print` to consider highest breakpoints because despite the high resolution of the paper (which is good for reading), this is not a good support for small _interfaces_ and interactive elements. Paper should almost be considered as a _tablet_ giving a fake resolution (like iPads, but _we_ are faking the resolution).

According to this, the objective is to make print ignoring all styles applied to _highest breakpoints only and not below_. This means that, for `$print-breakpoint: large;` on print:
* `large up/only/down`, `medium up/only/down`, `xlarge down` should be printable because their styles are applied to a _large_ (or smaller) screen so we should consider them as displayable on a _large_ (or smaller) print support.
* `xlarge only/up` should be not printable like before.

SassNinja: Ok @ncoden so the rule would be to generate print if the breakpoint affects the print-breakpoint (or smaller).
This means the current condition only needs to be extended so 'down' always generates print.

Since @DanielRuf has already added `$dir` I would simply change the condition to
```sass
@if $bp != null and ($bp <= $pbp or $dir == down) {
```

ncoden: Yes, and we can add an explantation comment and a link to this. @DanielRuf What do you think ?

DanielRuf: Sounds good. I might have thought to complex. Not sure what I covered that would not be covered by the simple `or ` condition. 🤔

DanielRuf:
> Also, there is a lot of cleaning to do in the breakpoint function and mixin. See https://github.com/zurb/foundation-sites/pull/9661/files ;)

Do we need any further changes in this PR here as we also need https://github.com/zurb/foundation-sites/pull/9661?

ncoden:
> Do we need any further changes in this PR here as we also need #9661?

See https://github.com/zurb/foundation-sites/pull/10979#discussion_r171845415

DanielRuf:
But doesn't this exclude / miss some cases?

Think about print breakpoint large.
I have an element which uses `medium down`. Because it should include `$bp >= $pbp and ($dir == 'down')`

DanielRuf:
> Think about print breakpoint large.
> I have an element which uses medium down. But this would be printed with the code of the comment I guess. But it should not because it should include $bp >= $pbp and ($dir == 'down')

/cc @SassNinja

ncoden:
@DanielRuf 
I'm not sure to understand the case. Do you mean an element using `breakpoint(medium down)` with `$print-breakpoint: large` ? If so, "print" will be in the generated breakpoint because `medium` < `large`.

DanielRuf:
But the element will not be visible on large so there should not be a print media query.

`@media print, screen and (max-width: medium) {`

ncoden:
hmm... There is a problem with the current implementation:

```scss
@media print, screen and (max-width: medium) {
```
...means "print OR (screen and max-width: medium)".

So for the following code:

```scss
$print-breakpoint: medium;
```

```scss
.only-for-small,
.only-for-medium {
    display: none;
}

@include breakpoint(small only) {
    .only-for-small { display: block }
}

@include breakpoint(medium only) {
    .only-for-medium { display: block }
}
```

```html
<div class="only-for-small">SMALL</div>
<div class="only-for-medium">MEDIUM</div>
```

Both `SMALL` and `MEDIUM` would be visible on print...
See: https://codepen.io/ncoden/pen/RMGOdx

So we may have to change that to https://codepen.io/ncoden/pen/BrLEeO OR https://codepen.io/ncoden/pen/wmzZLQ

DanielRuf:
> Both SMALL and MEDIUM would be visible on print... I don't think this is expected.
> See: https://codepen.io/ncoden/pen/RMGOdx

Huh? I don't think so. At least with my proposed version it should be on only and down afaik.

DanielRuf:
https://github.com/zurb/foundation-sites/pull/10979/files#diff-bf1c44b23206fb646211490bd0d02fffR171

ncoden:
> Huh? I don't think so. At least with my proposed version it should be on only and down afaik.

Your proposed version does not resolve the case but hide some of its effects. The bug has nothing to do with "when do we add `print`" but is caused by the way we generate the mediaquery, with "print OR [breakpoint]".

DanielRuf:
Ah okay.

DanielRuf:
What are the next steps here?

ncoden:
I took a look at the whole discussion and I think that this is the todo-list:
* [ ] Add support for `down` breakpoints https://github.com/zurb/foundation-sites/pull/10979/files#r171845415
* [ ] Ensure that only the right breakpoint is shown on print https://github.com/zurb/foundation-sites/pull/10979#issuecomment-373959270

kball:
@DanielRuf do you have a sense of current status of this PR?

DanielRuf:
> @DanielRuf do you have a sense of current status of this PR?

I'll check it this week.

DanielRuf:
> @DanielRuf do you have a sense of current status of this PR?

Probably just needs the proposed change from https://github.com/zurb/foundation-sites/pull/10979/files#r171845415

joeworkman (commented):
LGTM

joeworkman (approved):

joeworkman:

Issue #10685, opened 2017-09-28
Treat empty data attribute options as truthy


mindfullsilence:
#### How to reproduce this bug:

1. Step one
Create a dropdown panel, or any type of js-triggered component that can receive a true/false `data-[option]` attribute. 
2. Step two
Set the option to not contain a value (e.g. `<div data-dropdown data-hover></div>`)

#### What should happen:
I expected this to work similar to how the browser interprets boolean attributes when there is no value, defaulting to true-ish behavior. Similar to how the `checked` attribute works, where the following are equivalant and evaluate to "checked"
```
<input type="checkbox" checked />
<input type="checkbox" checked="checked" />
<input type="checkbox" checked="true" /> <!-- technically not valid html, but still evaluates to "checked" -->
```

#### What happened instead:
Hover did not work on the dropdown, the data attribute did not evaluate to true.

#### Browser(s) and Device(s) tested on:
Chrome

#### Foundation Version(s) you are using:
6.4.3

This is a somewhat opinionated ticket, but it did end up costing me a fair amount of time to debug (no errors/warnings, difficult to separate style issue from js issue from html issue). This could be something as simple as a documentation change, but wanted to propose that if a boolean data attribute contains no value, but is defined, that it default to "true". Food for thought.

PR #10980, opened 2018-02-23
fix: explicitely check if data-hover validates to false

DanielRuf:
Reference: https://github.com/zurb/foundation-sites/issues/10685

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -150,14 +150,14 @@ class Dropdown extends Positionable {
         if (_this.options.forceFollow === false) {
           // if forceFollow false, always prevent default action
           e.preventDefault();
-        } else if (hasTouch && _this.options.hover && _this.$element.hasClass('is-open') === false) {
+        } else if (hasTouch && _this.options.hover !== false && _this.$element.hasClass('is-open') === false) {
           // if forceFollow true and hover option true, only prevent default action on 1st click
           // on 2nd click (dropown opened) the default action (e.g. follow a href) gets executed
           e.preventDefault();
         }
     });
 
-    if(this.options.hover){
+    if(this.options.hover !== false){
       this.$anchors.off('mouseenter.zf.dropdown mouseleave.zf.dropdown')
       .on('mouseenter.zf.dropdown', function(){
         _this._setCurrentAnchor(this);
@@ -306,7 +306,7 @@ class Dropdown extends Positionable {
    */
   toggle() {
     if(this.$element.hasClass('is-open')){
-      if(this.$anchors.data('hover')) return;
+      if(this.$anchors.data('hover') !== false) return;
       this.close();
     }else{
       this.open();
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -150,14 +150,14 @@ class Dropdown extends Positionable {
         if (_this.options.forceFollow === false) {
           // if forceFollow false, always prevent default action
           e.preventDefault();
-        } else if (hasTouch && _this.options.hover !== false && _this.$element.hasClass('is-open') === false) {
+        } else if (hasTouch && (_this.options.hover === true || _this.options.hover === '') && _this.$element.hasClass('is-open') === false) {
           // if forceFollow true and hover option true, only prevent default action on 1st click
           // on 2nd click (dropown opened) the default action (e.g. follow a href) gets executed
           e.preventDefault();
         }
     });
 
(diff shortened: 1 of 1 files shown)


DanielRuf:
Todo: strict check for more attributes.

js/foundation.dropdown.js
```diff
@@ -150,14 +150,14 @@ class Dropdown extends Positionable {
...
-        } else if (hasTouch && _this.options.hover && _this.$element.hasClass('is-open') === false) {
+        } else if (hasTouch && _this.options.hover !== false && _this.$element.hasClass('is-open') === false) {
           // if forceFollow true and hover option true, only prevent default action on 1st click
           // on 2nd click (dropown opened) the default action (e.g. follow a href) gets executed
           e.preventDefault();
         }
     });
 
-    if(this.options.hover){
+    if(this.options.hover !== false){
```
ncoden: With `<div data-dropdown data-hover></div>`, `this.options.hover` is equal to an empty string here. So I wonder, do we need to:
- consider all non-false values as positives, which is a classic HTML approach
- ignore all invalid properties (expect a boolean but got a string), which is more sustainable approach.

The problem here is that we work with both HTML and Js values, and that in HTML `attribute` equals to `attribute=""`. Also, HTML boolean attibutes as considered as `true` as soon as they are set, even with a `"false"` as value...

With the current PR, `Foundation.Dropdown.defaults.hover = null` will make it considered as `true` even if its default is `false`.

DanielRuf:
So check better for strict true or any type of string?

js/foundation.dropdown.js
DanielRuf: So check better for strict true or any type of string?

ncoden:
> So check better for strict true or any type of string?

I don't know what solution to choose.

So I wonder, do we need to:
* consider all non-false values as positives
  ```
  if (this.options.hover !== false) ...
  ```
* ignore invalid properties (and use defaults instead)
  ```
  const validOptions = filterOptions({  }, Dropdown.defaults, this.$element.data(), options);
  this.options = filterOptions({ hover: Boolean, ... }, Dropdown.defaults, this.$element.data(), options);
  ```
* ignore invalid properties, but transform to true for Boolean
  (same code as above)

DanielRuf:
Did some changes to cover empty attr values and boolean true.

ncoden:
It's not only about this component. It's about how the whole framework should deal with Boolean/Empty attributes. We cannot make a manual check in each component for `true || ''`.

DanielRuf:
Right. Another approach should be better (the internal plugin system).

Where and how could we implement this in the best way?

DanielRuf:
@colin-marshall what do you think?

ncoden:
The simplest would be to use a small utility like
```js
// foundation.util.core.js
function boolAttr(value) {
  return value === true || (typeof value === 'string' && !value.length);
}
```
```js
// foundation.dropdown.js
if (boolAttr(this.option.hover)) ...
```

But it would be a bit dirty.

An other solution is (like given in a previous comment) a filter when initializing options. But it would be the part of a way bigger architecture change to efficiently control typing (typescript would not be a bad idea btw).

DanielRuf:
Should we plan this for v7 or how do we proceed with this PR?

ncoden:
I think it's not too much work to go with a little helpers in all components like:

```js
    _init() {
        this.options = $.extend({}, Dropdown.defaults, this.$element.data(), options);
        this.options.hover = boolAttr(this.options.hover, 'hover');
    }
```

Or 

```js
    _init() {
        this.options = $.extend({}, Dropdown.defaults, this.$element.data(), options);
        this.options = parseBooleanAttributes(this.options, ['hover', ...]);
    }
```

DanielRuf:
What should we choose? @colin-marshall @kball @Owlbertz @brettsmason

ncoden:
@DanielRuf Personally I would go for the 3rd solution.

I can expect this to be `true`:
```html
<div data-dropdown data-hover>...<div>
```

But not this:
```js
Foundation.Dropdown.defaults.hover = '';
```

DanielRuf:
@ncoden I guess you may know best how the helper should look and work like. Do you want to and can you take care of this?

ncoden:
I'll take care of this, probably Saturday.

DanielRuf:
I guess we have to rebase here. On which branch should I rebase?

ncoden:
`develop`. But still, it's not a priority (v6.6.0).

Issue #10548, opened 2017-08-16
Code Snippet Missing 'left' for Off-Canvas


mandimey:
Yo Rafi! 

In the code snippet for the off-canvas panel and button, the IDs don't match. As a result, the button doesn't trigger the panel to slide. The panel ID is offCanvas and the button toggle is offCanvasLeft. See http://foundation.zurb.com/sites/docs/off-canvas.html#complete-example + Click Trigger section. 

<3

JeremyEnglert:
@mandimey - would you like to submit a Pull Request to fix this?

DanielRuf:
Hi @mandimey would you have the time to provide a PR and need any assistance or can I provide a PR?

DanielRuf:
Closimg as this was fixed with https://github.com/zurb/foundation-sites/pull/10981

PR #10981, opened 2018-02-23
docs: fix the id of the example code for the click trigger

DanielRuf:
This PR fixes the code example to use the default off-canvas ID in the docs.

See https://github.com/zurb/foundation-sites/issues/10548

### DIFF
--- a/docs/pages/off-canvas.md
+++ b/docs/pages/off-canvas.md
@@ -105,7 +105,7 @@ To create a click trigger that opens the panel, add the attribute `data-open` or
 </div>
 
 ```html_example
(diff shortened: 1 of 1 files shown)


ncoden (approved):
👍 Thanks @DanielRuf 

Did you check the others buttons ?

DanielRuf:
Not yet. Will do so later today. Do you mean only the off-canvas page or all components?

ncoden:
I meant only the Off canvas.

DanielRuf:
Ok, will check that.

DanielRuf:
The others seem to be right. The codepen already uses the right identifier.

PR #10988, opened 2018-02-26
fix: fire updateActive after calcPoints

DanielRuf:
@ncoden this seems to be the case for the other randomly failing unit tests.

Please verify and test it with a few test iterations.

### DIFF
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -100,11 +100,15 @@ class Magellan extends Plugin {
         e.preventDefault();
         var arrival   = this.getAttribute('href');
         _this.scrollToLoc(arrival);
+        _this.calcPoints();
+        _this._updateActive();
       });
 
     this._deepLinkScroll = function(e) {
       if(_this.options.deepLinking) {
         _this.scrollToLoc(window.location.hash);
+        _this.calcPoints();
+        _this._updateActive();
       }
     };
 
@@ -129,7 +133,7 @@ class Magellan extends Plugin {
 
     SmoothScroll.scrollToLoc(loc, options, function() {
       _this._inTransition = false;
-      _this._updateActive();
+      //_this._updateActive();
     })
   }
 
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -93,22 +93,31 @@ class Magellan extends Plugin {
       _this._updateActive();
     });
 
-    this.$element.on({
-      'resizeme.zf.trigger': this.reflow.bind(this),
-      'scrollme.zf.trigger': this._updateActive.bind(this)
-    }).on('click.zf.magellan', 'a[href^="#"]', function(e) {
+    if (document.readyState === "complete") {
+      _this.$element.on({
+        'resizeme.zf.trigger': _this.reflow.bind(_this),
+        'scrollme.zf.trigger': _this._updateActive.bind(_this)
+      }).on('click.zf.magellan', 'a[href^="#"]', function(e) {
         e.preventDefault();
-        var arrival   = this.getAttribute('href');
+        var arrival   = _this.getAttribute('href');
         _this.scrollToLoc(arrival);
-        _this.calcPoints();
-        _this._updateActive();
       });
-
+    } else {
+      $(window).one('load', function(){
+        _this.$element.on({
+          'resizeme.zf.trigger': _this.reflow.bind(_this),
+          'scrollme.zf.trigger': _this._updateActive.bind(_this)
+        }).on('click.zf.magellan', 'a[href^="#"]', function(e) {
+          e.preventDefault();
+          var arrival   = _this.getAttribute('href');
+          _this.scrollToLoc(arrival);
+        });
+      });
+    }
+    
     this._deepLinkScroll = function(e) {
       if(_this.options.deepLinking) {
         _this.scrollToLoc(window.location.hash);
-        _this.calcPoints();
-        _this._updateActive();
       }
     };
 
--- a/js/foundation.magellan.js
+++ b/js/foundation.magellan.js
@@ -133,7 +133,6 @@ class Magellan extends Plugin {
 
     SmoothScroll.scrollToLoc(loc, options, function() {
       _this._inTransition = false;
-      //_this._updateActive();
     })
   }
 
--- a/test/javascript/components/magellan.js
+++ b/test/javascript/components/magellan.js
@@ -1,6 +1,6 @@
 describe('Magellan', function() {
 	var plugin;
-	var $html, $content; 	
+	var $html, $content;
 
   var generateUl = function(count) {
   	var html = '';
@@ -68,7 +68,7 @@ describe('Magellan', function() {
     });
 
 
-    it('fails gracefully when target does not exist', function(done) {
+    it('fails gracefully when target does not exist', function() {
     	var count = 5, duration = 200;
       $html = $(generateUl(count)).appendTo('body');
       $content = $(generateContent(count - 1)).appendTo('body');
@@ -78,17 +78,14 @@ describe('Magellan', function() {
 
       var hasError = false;
       var targets = $html.find('a');
-      if(targets.length){
+      try {
         var target = $(targets).eq(-1).attr('href');
         plugin.scrollToLoc(target);
-        hasError.should.equal(false);
-        done();
-			} else {
+      }
+      catch (err) {
         hasError = true;
-        hasError.should.equal(false);
-        done();
-			}
-      
+      }
+      hasError.should.equal(false);
     });
 
   });
--- a/test/javascript/components/magellan.js
+++ b/test/javascript/components/magellan.js
@@ -52,28 +52,23 @@ describe('Magellan', function() {
       	animationDuration: duration
       });
 
-
-
       // Jump to last section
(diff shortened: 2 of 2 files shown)


js/foundation.magellan.js
```diff
@@ -129,7 +133,7 @@ class Magellan extends Plugin {
 
     SmoothScroll.scrollToLoc(loc, options, function() {
       _this._inTransition = false;
-      _this._updateActive();
+      //_this._updateActive();
```
ncoden: No commented code please, never ;)

js/foundation.magellan.js
```diff
@@ -100,11 +100,15 @@ class Magellan extends Plugin {
         e.preventDefault();
         var arrival   = this.getAttribute('href');
         _this.scrollToLoc(arrival);
+        _this.calcPoints();
+        _this._updateActive();
```
ncoden: I investigated on this issue before. The easy solution is to put `this.calcPoints();` everywhere like @kball did 1 year ago (https://github.com/zurb/foundation-sites/pull/9491), and according to me this should work.

But `calcPoints()` is a huge operation that will get the position of all targets on the page. When the plugin was implemented, `calcPoints()` was only called on init/reflow so `_updateActive` on the scroll event works on pre-calculated positions. Calling `calcPoints` everywhere is against this optimized pattern.

So I think we should rather do the following:
* Consider a "page is loaded" when we receive the `onload` event OR the page is already loaded when the plugin is instantiated (related to https://github.com/zurb/foundation-sites/pull/10947).
* Wait for the page to load to listen for scroll events (or others events that trigger `_updateActive`).
  If the user begin to scroll before, we cannot update the magellan anchors anyway.
* Wait for the page to load before scrolling in tests.

ncoden (changes requested):
Nice catch but I don't think that's the best solution.

js/foundation.magellan.js
DanielRuf: Mh, sounds like a small refactor for this. Should we handle it in this PR or another one?

js/foundation.magellan.js
DanielRuf: Ok.

js/foundation.magellan.js
ncoden: I think we should clean that in this PR, as this is "anti-pattern" and can introduce some performance issues.

DanielRuf: Could need some advice. Do you have some specific solution in your mind how this should look like?

ncoden: hmm... For example
1. An an utility to check for `readyState === "complete"`, and will create a listener on `$(window).once('load', ...)` otherwise.
2. Use it in the `_event` function and start listening all others events in it (so we listen nothing before the page is loaded).
3. For the tests, I guess we can still use something like `$html.once('load',  ...)` and call `plugin.scrollToLoc(...)` within.

DanielRuf:
Pushed a new commit. Please review and test on your side (~80-100 test runs).

js/foundation.magellan.js
```diff
@@ -93,15 +93,28 @@ class Magellan extends Plugin {
...
+        _this.$element.on({
+          'resizeme.zf.trigger': _this.reflow.bind(_this),
+          'scrollme.zf.trigger': _this._updateActive.bind(_this)
+        }).on('click.zf.magellan', 'a[href^="#"]', function(e) {
+          e.preventDefault();
+          var arrival   = _this.getAttribute('href');
+          _this.scrollToLoc(arrival);
+        });
+      });
+    }
```
ncoden: Duplicated content.

We could do something like:
```js
// core.util.js
function onLoad(cb) {
  (document.readyState === "complete")
    ? cb()
    : $(window).one(cb);
 }
```
```js
// foundation.magellan.js
onLoad(() => {
  // code here
});
```

test/javascript/components/magellan.js
```diff
@@ -56,7 +56,9 @@ describe('Magellan', function() {
 
       // Jump to last section
       var target = $html.find('a').eq(-1).attr('href');
-      plugin.scrollToLoc(target);      
+      $html.one('load', function(){
+        plugin.scrollToLoc(target);
+			});
```
ncoden: The function here is executed dynamically, far after the test function exited. So the test happends before the scroll begin and therefore does not test it.

test/javascript/components/magellan.js
```diff
@@ -82,7 +84,9 @@ describe('Magellan', function() {
       var hasError = false;
       try {
 				var target = $html.find('a').eq(-1).attr('href');
-	      plugin.scrollToLoc(target); 
+        $html.one('load', function(){
+          plugin.scrollToLoc(target);
+        });
```
ncoden: Same here. No error triggered inside `scrollToLoc` will be triggered here.

See for example:
```js
// Numbers indicate the order in which each line is executed
...                               /* 1 */
try {                            
  setTimeout(/* 2 */ () => {
    throw new Error('err');       /* 7 */
  }, 0);                          /* 3 */
} catch(err) {                    /* [not executed] would be 4 */
  console.log('catched');         /* [not executed] would be 5 */
}
...                               /* 6 */
```

The code above will trigger an error. It will not be catched.

test/javascript/components/magellan.js
ncoden: A solution would be to wait for `$html.one('load'` to execute the test (and everything with it). I wonder if we can use `async await` in tests (for a simpler code)...

ncoden (changes requested):

test/javascript/components/magellan.js
DanielRuf: Ah you mean the `done`callback.

test/javascript/components/magellan.js
DanielRuf: True. Can you show or provide the solution? Currently not 100% clear to me how I can solve these things here as this component feels so tricky =/

test/javascript/components/magellan.js
ncoden: I forgot the `done` callback but yes.

```js
// .... initialize plugin

// wait for page to load
$html.one('load', function(){
  // scroll
  plugin.scrollToLoc(target);
  
  // wait for the scroll to takes effect
  setTimeout(function() {
    // ... test everything ...
    // and say we are done
    done();
  });
});
```

But as you see, this is a bit nested. It would be simpler with async/await

DanielRuf: `async await` is not supported in all browsers, for example IE11

ncoden: `async await` is supported almost nowhere. My point is that test files are transpiled (see [gulp/tasks/test.js:35](https://github.com/zurb/foundation-sites/blob/develop/gulp/tasks/test.js#L35)). But our current global babel config ([.babelrc](https://github.com/zurb/foundation-sites/blob/DanielRuf/fix/develop-dropdown-eventnames/.babelrc)) does not allow `async/await` (see https://medium.com/@dandv/using-es2016-es7-async-await-with-babel-e758bab1c4d5).

DanielRuf:
It seems the `calls reflow on viewport size change once` test also fails sometimes.

js/foundation.magellan.js
DanielRuf: Some util method for all components? Because we have this duplicated part probably in more components and tests where we do the same check.

DanielRuf: Sounds good. A candidate for another PR for the core utils?

ncoden: Ok, let's merge this as this, I will refactor all that in https://github.com/zurb/foundation-sites/pull/10947.

DanielRuf: Makes sense 👍 That should ensure that each PR produces consistent behavior and better overview. 

Personally I am not a big fan of mixing too much and big changes in one PR and separate them. But this is just my opinion.

ncoden:
> tests: make sure that the Magellan tests are properly running

hmm It seems to be a rollback. Tests may running but:
* the race condition should be back on `it scrolls the selected element into viewport`
* we do not actually test the plugin behavior on `fails gracefully when target does not exist` as we do not call the method when target does not exist.

DanielRuf:
> the race condition should be back on it scrolls the selected element into viewport

Could not verify it at least. Did around 70 - 80 runs and did not see them again.

js/foundation.magellan.js
ncoden: One PR = one action, with little or big changes. We can do a PR across the whole framework to refactor events on initialization, it's not a problem this we do more or less the same thing in all components and we're not lost in changes.

ncoden:
> All checks have passed

🍾

DanielRuf:
🎉 ✨

ncoden:
Note: there is probably a race condition in Interchange, I got this in 1 test on 100:
> * Interchange events() calls reflow on viewport size change once
>     Uncaught Error: Script error. (:0)

DanielRuf:
> Note: there is probably a race condition in Interchange, I got this in 1 test on 100:

https://github.com/zurb/foundation-sites/pull/10988#issuecomment-370040655

DanielRuf:
Not yet 100% sure what the cause is.

ncoden (approved):

PR #10993, opened 2018-02-28
Remove `.info` class name

gormus:
The example suggest there is a `.info` class available for labels. However there is no definition for `.info` class in v6.x, the demo is simply rendering the default `.label`.


DanielRuf:
Right, it seems it was one of the deprecated settings and was still in the docs.

https://github.com/zurb/foundation-sites/commit/834be750a798f18c3245ceb6cbf98f02671736db#diff-03112641c36e40dca2d56b484a8e6079

ncoden (commented):
👍 LGTM, I did not found other `info` in the code. Thanks @gormus

PR #10995, opened 2018-02-28
fix: change default width of the reveal to 80%

### LINKED ISSUE #8707: "Standard" Reveal width doesn't match docs (Foundation 6)

1. Build a Reveal modal using Foundation 6
2. View on a medium or larger screen width
3. See fixed 600px dialog

According to http://foundation.zurb.com/sites/docs/reveal.html#sizing the width of a medium+ Reveal dialog should be 80%.  This is what I expected coming from Foundation 5 as well.

**What happened instead?**

The width is set to 600px.

Either the documentation or the code is wrong.  I'm inclined to say it's the code that needs to be fixed.

DanielRuf:
This changes the default width for the reveal to the documented value of `80%`.

Closes #8707

### DIFF
--- a/scss/components/_reveal.scss
+++ b/scss/components/_reveal.scss
@@ -12,7 +12,7 @@ $reveal-background: $white !default;
 
 /// Default width of a modal, with no class applied.
 /// @type Number
-$reveal-width: 600px !default;
+$reveal-width: 80% !default;
 
 /// Default maximum width of a modal.
 /// @type Number


ncoden:
It does not seems to be a "fix" as this is a big change (from `600px` to `1092px` for 1366-wide screens). The "expected" value here seems far more closer to the `600px` visible in demo and all usages than the `80%` written somewhere in the doc.

DanielRuf:
Then are the docs unclear and the `80%` are wrong in general?

DanielRuf:
How where `80%` defined?
https://github.com/zurb/foundation-sites/blob/develop/scss/settings/_settings.scss#L109

DanielRuf:
> On medium-sized screens and larger, the width changes to 80%.

ncoden:
What I mean is: we can be somewhere between "wrong docs" and "wrong feature". The doc may be wrong (so `docs: ...` to fix the doc), but if you want to make the feature according to the doc, this is a far too breaking change too call this a "fix".

PR #10998, opened 2018-03-01
fix: copy the sourcemap files

### LINKED ISSUE #10997: CSS unnecessary sourcemap declaration

1. Download latest release from https://github.com/zurb/foundation-sites/archive/v6.4.3.zip
2. Go to dist/css directory
3. Verify that foundation.css last line has a sourcemap declaration, but sourcemap file doesn't exist.

#### What should happen:
It depends. From my point of view the Github release is used by other packages, like webjars, so it would be better if the result CSS has no sourcemap ? I'm not a CSS expert, so I might be wrong.

#### What happened instead:
CSS file has a sourcemap declaration, which results in a 404 error in browser.

Firefox, Linux

6.4.3

DanielRuf:
This copies the sourcemap files for the unminified development files.

Closes #10997

### DIFF
--- a/gulp/tasks/deploy.js
+++ b/gulp/tasks/deploy.js
@@ -60,7 +60,7 @@ gulp.task('deploy:dist', ['sass:foundation', 'javascript:foundation'], function(
       .pipe(gulp.dest('./dist/css'))
     .pipe(cssFilter.restore)
     .pipe(sourcemapFilter)
-    .pipe(gulp.dest('./dist/css'))
+      .pipe(gulp.dest('./dist/css'))
     .pipe(sourcemapFilter.restore)
     .pipe(jsFilter)
       .pipe(gulp.dest('./dist/js'))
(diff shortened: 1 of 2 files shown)


gulp/tasks/deploy.js
```diff
@@ -58,6 +59,9 @@ gulp.task('deploy:dist', ['sass:foundation', 'javascript:foundation'], function(
       .pipe(rename({ suffix: '.min' }))
       .pipe(gulp.dest('./dist/css'))
     .pipe(cssFilter.restore)
+    .pipe(sourcemapFilter)
+    .pipe(gulp.dest('./dist/css'))
```
ncoden: Missing tabulation here.

ncoden (approved):
I made some research and there seems to be no problem with publishing sourcemap to production. So LGTM 👍 .

gulp/tasks/deploy.js
DanielRuf: Good catch, will change it tomorrow morning if this is ok.

DanielRuf: Changed.

Issue #7916, opened 2016-01-19
Minor nit with styling of selects (possibly a firefox only issue)


r3wt:
A bit of background:
1. Only modification to Foundation is replace Arial Helvetica with Open Sans. 

*\* Issues **
1. the default padding of selects looks incorrect compared to text areas and text inputs(only tested in firefox). A modification i have found that looks correct is to change select's padding to `padding: 0.190rem;` .
   - before
     ![img](http://puu.sh/mB2hr/df6986d4da.png)
   - after
     ![img](http://puu.sh/mB2jn/a6b015bf36.png)
2. Selects allow text to run behind dropdown arrow of select box which is just _horrible_. The fix here is simple. 
   ![img](http://puu.sh/mB2uh/114931447d.png)

``` css
padding: 0.19rem 1.666rem 0.19rem 0.19rem;
text-overflow: ellipsis;
```

andycochran:
Ack! You're right; changing the fonts can cause this cropping issue. It's the padding. 

The fix for this is to not use top/bottom padding on `<select>` elements: 
```css
select {
  padding-top: 0;
  padding-bottom: 0;
}
```
The `line-height` could be set to mach `height`, but I'm not certain that's necessary.

whatcould:
I was just writing up a forum post about the left-alignment issue! Which is separate from the cropping / text-under-arrow issue.

The select box text left-alignment issue happens even on the Zurb Foundation site; in my testing, it doesn't depend at all on the font, seems contrary to @r3wt's experience (?)

Here is a direct screenshot clipped from the Foundation docs:

![select-menu-text-alignment-firefox](https://cloud.githubusercontent.com/assets/8288/12427328/90be7278-beac-11e5-9941-3c2c3fd8191b.png)

The Select elements have mis-aligned text in Firefox, and in Chrome on Windows (and slightly less mis-aligned on IE11/Win8). Graphic is Firefox 43 Mac, but it's a similar mis-alignment on FF & Chrome Windows. Chrome and Safari on Mac align the text in the Select elements the same as the input elements, with no extra padding.

If you take off the default 0.5rem padding, Chrome/Safari Mac both push the text right against the border, and these other browsers add. So it's some browser issue with the select boxes.

Interestingly, FF Mac has a double-text-indent bug on select boxes (http://stackoverflow.com/questions/28108434/firefox-select-element-doubles-text-indent-value), that doesn't seem to be what causes it but is maybe related.

[...]

PR #11000, opened 2018-03-02
fix: remove top and bottom padding from select tags

DanielRuf:
This removes the unneeded paddings which conflicts with other fonts.

Closes https://github.com/zurb/foundation-sites/issues/7916

### DIFF
--- a/scss/forms/_select.scss
+++ b/scss/forms/_select.scss
@@ -23,7 +23,7 @@ $select-radius: $global-radius !default;
 
   height: $height;
   margin: 0 0 $form-spacing;
-  padding: $input-padding;
+  padding: 0 $input-padding;
 
   appearance: none;
   border: $input-border;
--- a/scss/forms/_select.scss
+++ b/scss/forms/_select.scss
@@ -23,7 +23,7 @@ $select-radius: $global-radius !default;
 
   height: $height;
   margin: 0 0 $form-spacing;
-  padding: 0 $input-padding;
+  padding: $input-padding;
 
   appearance: none;
   border: $input-border;
@@ -76,6 +76,10 @@ $select-radius: $global-radius !default;
   &[multiple] {
     height: auto;
     background-image: none;
+  }
+  &:not([multiple]) {
+    padding-top: 0;
+    padding-bottom: 0;
   }
 }
 


ncoden:
Related to: https://github.com/TryGhost/Ghost/issues/1109

There's a lot of reports on Bugzilla about problems with line-height/padding in `<select>`, but none with this specific case.

ncoden (commented):
The problem comes from a too-big line heigh (not sure if this is related an internal `line-height` property) conflicting with `padding-top`/`padding-bottom` and the `height` calculated for it.

Computed `line-height` + `padding-bottom` +  `padding-top` is too big for the calculated height, and the text overflow.

Removing `padding-top`/`padding-bottom` resolves this issue but is inconsistent with the `input`/`textarea` (`@mixin form-text`) properties (that we cannot change because `textarea` need  paddings). However I don't know if this is an issue. We can also consider top/bottom paddings as specific textarea/form elements with auto height.

TODO: report this to Mozilla.

DanielRuf:
> TODO: report this to Mozilla.

Can you create a report at https://bugzilla.mozilla.org/enter_bug.cgi?

So should we not merge this and report this to Mozilla? Other browsers do not have this exact issue?

DanielRuf:
This also breaks multiple select in Chrome as it changed the padding.

https://codepen.io/DanielRuf/pen/qxVOJQ

DanielRuf:
Changed it a bit.

ncoden (commented):
I don't like using `:not()` but do not have a better solution for now. I'll make some more tests across browsers before merging.

ncoden:
The "padding" seems to work to fix the text alignement for all browsers, but it introduce a bug with the focus indicator height:

### Expected (IE 11)
screen shot 2018-04-11 at 23 34 32

### Actual

* **IE 11**
  screen shot 2018-04-11 at 23 33 08
* **IE 9**
  screen shot 2018-04-11 at 23 36 31
* **Firefox 56**
screen shot 2018-04-11 at 23 46 05

DanielRuf:
How do you mean this? I see no wrong behavior as focus is like active / selected.
The line-height  / box height of the option has changed which is correct.

DanielRuf:
Try native select without our styles and you see the same (as far as I remember).

ncoden (approved):

Issue #8484, opened 2016-03-30
unused variables in _tabs.scss, why ?


mDesignHub:
Within Foundation 6.2.0 _tabs.scss there are variables, which are never been used ?
row 34: $tab-expand-max: 6 !default;
row 46: $tab-content-color: foreground($tab-background, $primary-color) !default;

is this for later version ?

brettsmason:
@kball the first variable looks similar to what's used with the `.expanded` class in menus which works because of `display: table-cell` styling. I'd say this could be removed. However not sure on the 2nd variable, could this be turned into a usable variable or should it be removed?

I'm happy to tidy it up if someone can confirm that's the case.

kball:
It seems like the second one maybe should do something in the tabs-content mixin?  Maybe play around with using it there and see if the results make sense.  The first one looks like maybe it was intended for something similar to the way the button-group-expand mixin is used, but I don't see any evidence of code in that direction.

@andycochran I know you were looking at menus & button groups a lot, do you think there's a need for similar expand logic for tabs?

andycochran:
I like the idea of all expand classes using the same method. But I wonder about us going in circles with methods. We left `table-cell` for `inline-block` in one component, stopped using `float` in another, now they've all diverged. I wonder if one expand mixin could rule them all.

brettsmason:
Its maybe worth exploring, though I think you might get into specificity issues, and you have borders on some things.

DanielRuf:
`$tab-content-color` is actually used. `$tab-expand-max` is currently still not used and may be safe to remove (I guess so).

ncoden:
`$tab-expand-max` is introduced in the initial Tabs commit (https://github.com/zurb/foundation-sites/commit/acb151af2e7f9d1c80be5c0986cad586b2692eb6, @zurbrandon) and was never used. It is safe to remove.

PR #11001, opened 2018-03-02
fix: remove the unused tab-expand-max variable

DanielRuf:
This removes the unused `$tab-expand-max` variable.

Closes https://github.com/zurb/foundation-sites/issues/8484

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -800,7 +800,6 @@ $tab-active-color: $primary-color;
 $tab-item-font-size: rem-calc(12);
 $tab-item-background-hover: $white;
 $tab-item-padding: 1.25rem 1.5rem;
-$tab-expand-max: 6;
 $tab-content-background: $white;
 $tab-content-border: $light-gray;
 $tab-content-color: $body-font-color;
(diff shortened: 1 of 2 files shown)


scss/settings/_settings.scss
```diff
@@ -800,7 +800,6 @@ $tab-active-color: $primary-color;
 $tab-item-font-size: rem-calc(12);
 $tab-item-background-hover: $white;
 $tab-item-padding: 1.25rem 1.5rem;
-$tab-expand-max: 6;
```
ncoden: No need to remove variables in `settings/_settings.scss` as it is automatically generated ;)

ncoden (approved):
> `$tab-expand-max` is introduced in the initial Tabs commit (https://github.com/zurb/foundation-sites/commit/acb151af2e7f9d1c80be5c0986cad586b2692eb6, @zurbrandon) and was never used. It is safe to remove.
> -- https://github.com/zurb/foundation-sites/issues/8484#issuecomment-370075094

LGTM 👍

scss/settings/_settings.scss
DanielRuf: Mh, in the src / scss directory? Should I revert this file change then?

PR #11002, opened 2018-03-02
fix: add the xy-grid parameter to foundation-everything

DanielRuf:
This adds `$xy-grid` to `foundation-everything`.

Closes https://github.com/zurb/foundation-sites/issues/10742

### DIFF
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -72,12 +72,12 @@
   $prototype: false,
   $xy-grid: true
 ) {
-  @if $xy-grid {
-    $xy-grid: true !global;
-  }
-
   @if $flex {
     $global-flexbox: true !global;
+  }
+
+  @if $xy-grid {
+    $xy-grid: true !global;
   }
 
   @include foundation-global-styles;
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -69,7 +69,8 @@
 
 @mixin foundation-everything(
   $flex: true,
-  $prototype: false
+  $prototype: false,
+  $xy-grid: true
 ) {
   @if $flex {
     $global-flexbox: true !global;
--- a/scss/foundation.scss
+++ b/scss/foundation.scss
@@ -70,7 +70,7 @@
 @mixin foundation-everything(
   $flex: true,
   $prototype: false,
-  $xy-grid: true
+  $xy-grid: $xy-grid
 ) {
   @if $flex {
     $global-flexbox: true !global;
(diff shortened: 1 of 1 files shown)


ncoden (changes requested):
This makes the global `$xy-grid` config ([scss/xy-grid/_xy-grid.scss:11](https://github.com/zurb/foundation-sites/blob/f29e2d2ecf818844e62e7355846991f8b2989145/scss/xy-grid/_xy-grid.scss#L11)) useless.

We have the following solutions:
* use the global `$xy-grid` by default as parameter (better)
* set the global `$xy-grid` with the parameter (consistent with the `$flex` parameter)

I would personnally go for the _consistent_ option for now, but I'd like to get your opinion.

DanielRuf:
We have no `$global-xy-grid`, so how should we name the variable here in this file?

ncoden:
IMHO `$xy-grid`should work here as a param _and_ a global config.

ncoden:
@if $xy-grid { // test the local parameter
    $xy-grid: true !global; // set the global config
}

DanielRuf:
> @if $xy-grid { // test the local parameter
>     $xy-grid: true !global; // set the global config
> }

Changed it according to your suggestion.

scss/foundation.scss
```diff
@@ -69,8 +69,13 @@
 
 @mixin foundation-everything(
   $flex: true,
-  $prototype: false
+  $prototype: false,
+  $xy-grid: true
 ) {
+  @if $xy-grid {
+    $xy-grid: true !global;
+  }
```
ncoden:
```scss
$prototype: false,
$xy-grid: true
```
So:
```scss
@if $flex { 
  $global-flexbox: true !global; 
} 
@if $xy-grid { 
  $xy-grid: true !global; 
} 
```

Does it works as expected ? (changing the global variable).

scss/foundation.scss
```diff
@@ -69,8 +69,13 @@
 
 @mixin foundation-everything(
   $flex: true,
-  $prototype: false
+  $prototype: false,
+  $xy-grid: true
```
ncoden: In this case I think we should use:
```scss
$xy-grid: $xy-grid
```
for backward compatibility. The global `$xy-grid` setting should still have an effect.
Please check it works as expected.

ncoden (changes requested):

scss/foundation.scss
DanielRuf: Ok, will check this later today before pushing this proposed change.

DanielRuf:
Is something left or did we oversee something?

DanielRuf:
> style: change the code formatting without effects on the execution

Would be this then.

PR #11006, opened 2018-03-02
fix: IE 11 and older do not trigger popstate, use hashchange for chec…

### LINKED ISSUE #10771: deep linking to tabs seems not to work in MSIE 11
In f6 it should be possible to set an anchor link to a tabpanel. Clicking on it should open the referenced tab. This works fine in chrome and ff, but not in msie 11 (see jsfiddle-link, click on the "go to panel"-links)

#### What should happen:
Linked tab should open

#### What happened instead:
nothing

Internet Explorer 11 (11.0.9600.17843), Windows 7 Professional Service Pack 1

6.4.3

#### Test case link:
https://jsfiddle.net/1zptbLbm/

ADD:
I found a workaround, that seem to work at least in regular tabs: when I select the target tab onclick via "selectTab", everything is fine, smth like:

    $(function() {
        $('.deeplinks').on('click',function(ev){
            $("#example-tabs").foundation('selectTab',ev.currentTarget.hash.substr(1),true);
        });
        $(document).foundation();
    });

Unfortunately we use a responsive accordion tab, where "selectTab" is not implemented:
https://jsfiddle.net/3u172fdb/

DanielRuf:
IE 11 and older do not trigger the `popstate` event when the hash changes.
We have to use `hashchange` in this case.

It is a bug that will not be fixed. See https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3740423/

Closes https://github.com/zurb/foundation-sites/issues/10771

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -143,7 +143,16 @@ class Tabs extends Plugin {
     }
 
     if(this.options.deepLink) {
-      $(window).on('popstate', this._checkDeepLink);
+      // bug in IE
+      // @see https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/3740423/
+      // checks from http://browserhacks.com/ for IE11 and older which does not fire popstate on hashchange
+      var isIE_3_10 = /*@cc_on!@*/false;
+      var isIE_11 = '-ms-scroll-limit' in document.documentElement.style && '-ms-ime-align' in document.documentElement.style;
+      if(isIE_3_10 || isIE_11){
+        $(window).on('hashchange', this._checkDeepLink);
+      } else {
+        $(window).on('popstate', this._checkDeepLink);
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


ncoden (changes requested):
Same as for others PRs: `popstate` is used a lot within the framework (Tabs, Accordion, Magellan, Reveal).

Please add an utility in `foundation.util.core` (or an other utility file if you think it is more appropriate) and avoid code duplication.

DanielRuf:
> Same as for others PRs: popstate is used a lot within the framework (Tabs, Accordion, Magellan, Reveal).
> 
> Please add an utility in foundation.util.core (or an other utility file if you think it is more appropriate) and avoid code duplication.

Good suggestion. Similar to the `load` utility method that you suggested?

ncoden:
@DanielRuf Exactly (with a better name, specific enough to Foundation)

ncoden:
@DanielRuf  I started implemented a generic `popstateEvent` and made some research:

* [hashchange](https://developer.mozilla.org/en-US/docs/Web/Events/hashchange) is triggered _only when the hash change_. It is not similar to `popstate` as it will not be triggered when the url change. __IE 8+__.
* [popstate](https://developer.mozilla.org/en-US/docs/Web/Events/popstate) is triggered when the history state is changed. __Edge 14+__.

But none of these events will not being triggerd by `history.pushState()` (see https://codepen.io/ncoden/pen/ddLQmN). Only backward/forward history manipulations will (like clicking on a link or called `history.back()`). There is a good polyfill [here](http://felix-kling.de/blog/2011/01/06/how-to-detect-history-pushstate/) but it works by overriding `history.pushState`.

The thing is: we do not even need to listen on all history change. We currently use `popstate` (checked for Accordion, Magellan, Reveal and Tabs) only for `window.location.hash`, and could rely on `hashchange` instead without any problems. No need for a polyfill.

ncoden:
Closing this in favor of https://github.com/zurb/foundation-sites/pull/11011

PR #11011, opened 2018-03-04
fix: use `hashchange` event instead of `popstate` for IE/Edge compatibility

ncoden:
> We do not even need to listen on all history change. We currently use popstate (checked for Accordion, Magellan, Reveal and Tabs) only for window.location.hash, and could rely on hashchange instead without any problems. No need for a polyfill.

See: https://github.com/zurb/foundation-sites/pull/11006#issuecomment-370193395

#### Events:
* [popstate](https://developer.mozilla.org/en-US/docs/Web/Events/popstate): Edge 14+.
* [hashchange](https://developer.mozilla.org/en-US/docs/Web/Events/hashchange): IE 8+.

#### Updated components:
* Accordion
* Magellan
* Reveal
* Tabs

Previous PR: https://github.com/zurb/foundation-sites/pull/11006 (@DanielRuf)
Closes https://github.com/zurb/foundation-sites/issues/10771 (@hubert-ried)

### DIFF
--- a/js/foundation.accordion.js
+++ b/js/foundation.accordion.js
@@ -146,7 +146,7 @@ class Accordion extends Plugin {
       }
     });
     if(this.options.deepLink) {
-      $(window).on('popstate', this._checkDeepLink);
+      $(window).on('hashchange', this._checkDeepLink);
     }
   }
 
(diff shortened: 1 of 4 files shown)


DanielRuf (approved):
Should be sufficient as we do not have to listen for any history changes, just to hash changes.

ncoden:
@DanielRuf Yes this is the reason of this PR. We currently only need to listen on hash changes.

Issue #11015, opened 2018-03-05
Should the search cancel buttons in Chrome and Safari on macOS really be removed by default? If so, why?


gburning:
I was troubleshooting why the WebKit cancel buttons for search inputs don't show up and found that they've been explicitly removed in the CSS.

https://github.com/zurb/foundation-sites/blob/26475ef192727333d8ad8c0e5cbff4be06ea15e4/dist/css/foundation.css#L332-L335

Should this really be the case?

DanielRuf:
See https://github.com/zurb/foundation-sites/blob/1a7af341bccd571b65aa2a96b10bd0884576fb5a/_vendor/normalize-scss/sass/normalize/_normalize-mixin.scss#L529-L535

DanielRuf:
https://github.com/necolas/normalize.css/blob/7369f566898b8cefc2649b6a66e112857de4c46f/CHANGELOG.md

```
8.0.0 (February 2, 2018)
Remove support for older browsers Android 4, lte IE 9, lte Safari 7.
Don't remove search input cancel button in Chrome/Safari.
Form inputs inherit font-family.
Fix text decoration in Safari 8+.
```

DanielRuf:
https://github.com/zurb/foundation-sites/blob/develop/package.json#L86

DanielRuf:
https://github.com/necolas/normalize.css/issues/685

DanielRuf:
@ncoden should we file an issue at normalize-scss or overwrite this in our global styles?

ncoden:
Why not both ?

Also I'll take care of a cleaner version of https://github.com/zurb/foundation-sites/pull/11016 tomorrow.

PR #11016, opened 2018-03-05
fix: use normalize 8.0.0

DanielRuf:
normalize-scss has diverged from normalize.css and does not contain the latest upstream changes.

This also seems to correct the order of the rules in the foundation.min.css file.

Closes: #11015


scss/settings/_settings.scss
```diff
@@ -95,12 +95,10 @@ $global-menu-nested-margin: 1rem;
 $global-text-direction: ltr;
 $global-flexbox: true;
 $global-prototype-breakpoints: false;
-$global-button-cursor: pointer;
+$global-button-cursor: auto;
```
ncoden: `_settings.scss` should not be in this PR.

scss/foundation.scss
```diff
@@ -6,7 +6,7 @@
  */
 
 // Dependencies
-@import '../_vendor/normalize-scss/sass/normalize';
+@import '../_vendor/normalize.css/normalize';
```
ncoden: As you include a CSS file, CSS will be generated even when Foundation mixins are not generated. Including a CSS file directly is not a good approach as it goes against the following (very recommanded) architecture:
1. Import
2. Configure
3. Execute (In Sass, generate CSS)

Before relying on `normalize-scss`, we had normalize included in our own mixin. This is a perfect solution but if there nothing better to get normalize updated we'll go for it.

ncoden (changes requested):
Could you motivate the migration to normalize `v8.0.0` (benefits, downsides, minor and major/breaking changes...) ?

Please open an issue to discuss about that (and the best way to implement it). I'm closing this PR as the proposed solution is not compatible with the framework SCSS architecture (see the comments below).

scss/settings/_settings.scss
DanielRuf: Right. But there are also outdated / wrong settings ;-)

scss/foundation.scss
DanielRuf: You can import css files in Scss without any issues. It will not generate extra files. So far normalize-scss is far behind normalize and can not be directly compared.

scss/settings/_settings.scss
DanielRuf: Also check the min css files and how the order of rules is correct then.

DanielRuf:
> Could you motivate the migration to normalize v8.0.0 (benefits, downsides, minor and major/breaking changes...) ?

Ok, but sooner or later we will also go back to CSS probably (PostCSS) ;-)

scss/foundation.scss
ncoden: I am not talking about _extra file generation_ but _systematic code generation_. This means that if I do in my code:

```scss
import 'foundation-sites`;
```

Without running `@include foundation-everything`, I'll still have normalize code generated.

DanielRuf: Ah, I guess I'll do some research to find a way to wrap this in a mixin. Probably by using some etract like plugin to store the content for reuse.

DanielRuf:
> Ok, but sooner or later we will also probably go back to CSS probably (PostCSS) ;-)

In general PostCSS is faster (~3 times), has better performance, less problematic dependencies (node-gyp is not needed), provides better integration with other tools and the whole toolchain (webpack, linting, ...).

DanielRuf:
Can we not copy and rename the css file to scss? Any css file is a valid scss file.

ncoden:
@DanielRuf We can, but the problem is still there: an imported CSS or SCSS file with code in it will make the code directly imported into the user Sass project even with no mixin called. This is not what we want.

ncoden:
A solution _could be_ to do like we did before: put the _normalize_ code into its own mixin. But this would require to update it manually and this is why we switched to `normalize-scss`.

DanielRuf:
> @DanielRuf We can, but the problem is still there: an imported CSS or SCSS file with code in it will make the code directly imported into the user Sass project even with no mixin called. This is not what we want.

There have to be solutions to wrap this in mixins. At least in PostCSS which is just JS / ES6 plugins then =)

PR #11017, opened 2018-03-05
fix: mirror the previous and next button content of Orbit in rtl mode

DanielRuf:
Closes https://github.com/zurb/foundation-sites/issues/11008

### DIFF
--- a/scss/components/_orbit.scss
+++ b/scss/components/_orbit.scss
@@ -190,6 +190,14 @@ $orbit-control-zindex: 10 !default;
     @include orbit-next;
   }
 
+  // RTL support
+  @if $global-text-direction == rtl {
+    .orbit-previous,
+    .orbit-next {
+      transform: scale(-1, 1);
+    }
+  }
+
   .orbit-bullets {
     @include orbit-bullets;
   }


scss/components/_orbit.scss
```diff
@@ -190,6 +190,14 @@ $orbit-control-zindex: 10 !default;
     @include orbit-next;
   }
 
+  // RTL support
+  @if $global-text-direction == rtl {
+    .orbit-previous,
+    .orbit-next {
+      transform: scale(-1, 1);
```
ncoden: The HTML (so the used icons) are at the discretion of the user. We cannot assume that the `.orbit-previous` icon will be the "mirror" of the `.orbit-next` icon.

I see 3 solutions:
1. Fix nothing as this is not our responsability to show the right icon. Only provide some CSS bidi (lrt/rtl) visibility helpers and update the doc to show an example.
2. Do not reverse icons in Sass and reverse the action of the arrows (this may be confusing as arrows are named `previous` and `next`. We can rename the arrows to `left` and  `right` tho)
3. Keep `previous` and `next` arrows and add the `left` and `right` arrows aside. This may make the JS (and the component usage) a bit complex.

2) and 3) prevent the user to show an icon dedicated to _previous_ or _next_. 1) seems a way more customizable, but a bit complex to use.

Do you have an other solution in mind ?

ncoden (changes requested):

scss/components/_orbit.scss
DanielRuf: > provide some CSS bidi (lrt/rtl) visibility helper

That was also in my mind. But what about the defaults and examples?

ncoden: We would update them. It's not a problem.
But I'm not sure this is the best solution.

DanielRuf: Me neither.

cc @SassNinja @colin-marshall

DanielRuf: > Fix nothing as this is not our responsability to show the right icon. Only provide some CSS bidi (lrt/rtl) visibility helpers and update the doc to show an example.

Thsi is the best solution at the moment, or am I wrong?

DanielRuf: > Do not reverse icons in Sass and reverse the action of the arrows (this may be confusing as arrows are named previous and next. We can rename the arrows to left and right tho)

also not good for a11y

DanielRuf:
I guess we can close this PR and should provide LTR/RTL CSS helpers.

ncoden:
See https://github.com/zurb/foundation-sites/issues/11070

Issue #9123, opened 2016-08-18
Safari bug?


SimonWpt:
This is a mysterious Expand-Collapse-bug with Safari. Because it is a customer site I use a short url. All actions are done with mouse clicks.
1. Open https://agnt.eu/ghwmm
2. Click left menu icon
3. Expand **Warenkunde**
4. Click on **Das Messerhandwerk**
5. Now go back via **BROWSER BACK**
6. Collapse  **Warenkunde** 
7. Expand  **Warenkunde** 

Now sub menu expands and collapses automatic infinite without any user action.

SimonWpt:
It seems to me, that safari is the only browser, which saves the view state of a dropdown. When going back in Safari via *browser back* the state is expanded. In all other browers, the state is reseted to collapse.

ralts00:
Its true :( 
![giphy](https://cloud.githubusercontent.com/assets/5034252/17819584/31d2658a-660e-11e6-906f-f0401b7b0b52.gif)

coreysyms:
I have seen this too in the wild, I have overridden this in the past with jQuery to fire a page reload if event.persisted exists, but this is not a universal solution as most people would prefer not to reload the page. I had to do this because even targeting the menu to close via classes did not work. Only a hard reload worked.

Safari's handling of bfcache could be preferable for most things, but in this case we would rather not save the state of the DOM. A case specific solution for just this plugin would be the way to go, we just need to be careful not to effect other DOM elements that are preferably cashed. The team should discuss a solution

stuartstarrs:
This also appears to be just the latest versions of Safari. 
I can't replicate this on Version 8.0.8 (10600.8.9) for example.

coreysyms:
Here is my code, I have had good luck with doing this. Of course watch your div IDs to match.
```
       $("#accordion-nav-menu ul li a").click(function(e) {
			e.preventDefault();
			
			if ($(this).next().length === 0) {
				//close all submenus
				
				$('#accordion-nav-menu li.has-submenu').attr("aria-selected","false").attr("aria-expanded","false");
				$('#accordion-nav-menu li.has-submenu > ul').attr("aria-hidden", "true").removeClass("is-active").removeAttr("style").css("display","none");
				$('#accordion-nav-menu').foundation('hideAll');
				
				//close main menu
				$('#main-nav').removeClass("is-active");
				$("#nav-icon").removeClass("open");
				
				//wait a tick and fire url
				setTimeout(function() { 
[...]

PR #11021, opened 2018-03-07
chore: update the browserslist config according to the compability table

DanielRuf:
The browserslist config was not consistent and did not set supported browsers versions. This removed the support for Safari 9 and older as these are not supported anymore and have many bugs that can not be easily solved with a few workarounds.

https://foundation.zurb.com/sites/docs/compatibility.html

Reference: #9123 

Also Apple does not support older versions (just 11?).


ncoden (commented):
This update the compatibility according to the documentation, but I am not sure this is what we should do. The compatibility list is inconsistent with Android Browser `v2.3` (0% usage, same for `v3`/`v4` and 8 year old) and without Safari 9 (0.5% usage, iOS 9 so iPhone 4S/5/5C/6/6S/SE, 2.5 year old).

We can have some bugs in older browsers, this is always better than unsupporting it.

So... do we drop iOS 9 or do we update the documentation.

DanielRuf:
> iPhone 4S/5/5C/6/6S/SE

Attention, these devices still get OS upgrades / updates. Even for Safari ;-)

DanielRuf:
> do we update the documentation.

This would introduce more work than before. Safari 9 is quite old and the Foundation Sites 6 docs were like this the whole time I guess.

Deprecating old Safari versions should be ok.

We might either increase the Android version or check for android_chrome in general instead of Android. But this would be another step and for 6.5 probably useful.

ncoden:
> Attention, these devices still get OS upgrades / updates. Even for Safari ;-)

Yes, expect for 4s.

> Deprecating old Safari versions should be ok.

Yes, but I am not sure about Safari 9. And again, I would prefer a _partial support_ (with autoprefixer doing its job even if we didn't check if everything works correctly) and add it to the doc rather than dropping the CSS support and going for "it cost us nothing but we'll drop it anyway".

DanielRuf:
How do we proceed with this one?

ncoden:
@DanielRuf I would go for keeping iOS 7 in the CSS compatibility table (we can drop Android 2.3 and use Android 4.4 instead) and add a "partial support" section in the documentation with something like: "Actively supported but with some known bugs."

We _want_ to support these browsers as they are not too old, but sadly sometimes there is a bug that we cannot resolve, too bad. It's always better than dropping the support (and creating new bugs).

DanielRuf:
Pushed a new commit with a few changes. Please check if this is better.

docs/pages/compatibility.md
```diff
@@ -16,18 +16,20 @@ tags:
...
     <td class="works">Versions 9+</td>
   </tr>
   <tr>
     <td>Android Browser</td>
-    <td class="works">Versions 2.3+</td>
+    <td class="works">Versions 4.4+</td>
   </tr>
 </table>
 
+<sup>1</sup>iOS 7+ is actively supported but with some known bugs.
```
ncoden: Does _all_ iOS 7+ versions have bugs ? ;)
Please add the concerned versions and links to the issues.

DanielRuf: > Please add the concerned versions and links to the issues.

Well, this would be many. iOS Safari <= 10 has flexbugs issues and there are also numerous issues with Safari 9. And both are officially not supported by Apple anymore. That's why I wanted to make a cut. We can not support all these old and problematic Safari releases (which also did not support required form fields and so on).

So "some known bugs" is not right in my opinion.

https://www.reddit.com/r/web_design/comments/3la04p/psa_safari_on_ios9_has_a_media_query_bug/

Safari is the new IE for us in the frontend ;-)

ncoden: > Any time a web browser is tied to the release cycle of an OS it's going to be bad. IE suffered from this problem in the past and Safari both on iOS and the main OS suffers from not being on an independent release cycle. It's really stupid.

ncoden: What do you think @colin-marshall @SassNinja @JeremyEnglert ?
Foundation is known for its compatibility with older browsers (even if it's a bit less true now with flex & XY grid).

See the discussion above https://github.com/zurb/foundation-sites/pull/11021#issuecomment-373955647

ncoden:
@DanielRuf I think we can keep this PR as is for now, and start to drop olders browsers supports (like at least Safari 7 & 8) at the next minor version, as we try to make `v6.5.0` keeping maximum compatibility with `v6.4.4-rc1`. What do you think @DanielRuf ? (poke @SassNinja)

ncoden:
Poke @DanielRuf. https://github.com/zurb/foundation-sites/pull/11021#issuecomment-380598523

DanielRuf:
> What do you think ? (poke @SassNinja)

This?

The other is clear (keep it open for now?) to me.

ncoden:
> The other is clear (keep it open for now?) to me.

I don't understand. I meant merging it like this with support for android >= 4.4, ios >= 7 and ie >= 9

ncoden (approved):
👍

Issue #10413, opened 2017-07-14
Make $accordion-plusminus a parameter on accordion-title


brentdodell:
#### How to reproduce this bug:
Use accordion mixins (`accordion-item` in particular)

#### What should happen:
It would be nice to be able to override [$accordion-plusminus](https://github.com/zurb/foundation-sites/blob/develop/scss/components/_accordion.scss#L103) as a parameter on [@mixin accordion-title](https://github.com/zurb/foundation-sites/blob/develop/scss/components/_accordion.scss#L75-L81).

It would also be nice if you could specify what the *plus* and *minus* content is, so that different characters could be used

#### What happened instead:
Cannot override default setting of `$accordion-plusminus` when including `accordion-title` mixin, because no parameter is available for that value.

#### Foundation Version(s) you are using:
 6.4.1

#### Test case link:
- [accordion-item mixin without $plusminus parameter generates SASS error](https://codepen.io/brentdodell/pen/VWRMdP)
- [accordion-item mixin with $plusminus parameter](https://codepen.io/brentdodell/pen/YQgrjZ)

IamManchanda:
This looks good @brentdodell ... Would you like to do a PR for it ??

brentdodell:
> This looks good @brentdodell ... Would you like to do a PR for it ??

Sure! I will when I have a chance. I didn't want to jump into it too quickly because I hadn't looked over Zurb's contribution guidelines.

Are you thinking just parameterizing `$accordion-plusminus`, or also allowing for specifying custom `content` in place of `+`/`-` like I also suggested?

IamManchanda:
Hey @brentdodell thanks for considering contributing guidelines 
For the core,
i would think that just parameterizing `$accordion-plusminus` would be fine  

But then we have this too (http://foundation.zurb.com/sites/docs/prototyping-utilities.html) 
So for my thought adding a class where an accordion will have arrow instead of plusminus will be great too for optional prototypes!

Again, this is just very personal opinion and maybe other yetinauts may consider different
So i will cc @kball for his opionion ??

kball:
I would actually prefer the ability to set arbitrary content for open & closed states

IamManchanda:
@kball You mean to choose b/w plusminus and arrows ?? 
Yup that was on my mind too!

IamManchanda:
I think that will be good one choice also!

IamManchanda:
Hey @brentdodell ... Any updates on this ??

[...]

PR #11025, opened 2018-03-08
Allow for setting of custom plus/minus indicators for Accordion.

jellisii:
Addresses #10413


scss/components/_accordion.scss
```diff
@@ -14,6 +14,14 @@ $accordion-background: $white !default;
 /// @type Boolean
 $accordion-plusminus: true !default;
 
+/// Default indicator for plus when $accordion-plusminus is true
+/// @type String
+$accordion-plus-character: '\002B' !default;
```
ncoden: Could you rename it to `$accordion-plus-content` as this is a String that can contain several characters ?

scss/components/_accordion.scss
```diff
@@ -14,6 +14,14 @@ $accordion-background: $white !default;
...
+
+/// Default indicator for minus when $accordion-plusminus is true
+/// @type String
+$accordion-minus-character: '\2013' !default;
```
ncoden: `$accordion-minus-content`

scss/components/_accordion.scss
```diff
@@ -14,6 +14,14 @@ $accordion-background: $white !default;
 /// @type Boolean
 $accordion-plusminus: true !default;
 
+/// Default indicator for plus when $accordion-plusminus is true
```
ncoden: You can add "`" around variable names and values like elsewhere in the docs

ncoden (changes requested):
👍 Thanks @jellisii 

It make me think that we should have these options for every component where it could be useful to customize them, for lrt/rtl support and design customization (with a custom webfont for example).

ncoden (approved):
@jellisii I took care of the requested changes. Seems good to merge for `v6.5.0`

ncoden:
I added the same options for Pagination. I did not found others content where customization would be useful.

Issue #9727, opened 2017-01-30
[Reveal] btmOffsetPct being ignored


diniscorreia:
No matter what value I use in `btmOffsetPct`, the reveal element always stays in the same position (slightly closer to the top of the window).

Looking at the code for the plugin, `btmOffsetPct` doesn't seem to be used anywhere, except when setting the default value.

IamManchanda:
Please provide Codepen ??

diniscorreia:
@IamManchanda oops, sorry. Here you go: http://codepen.io/diniscorreia/pen/KaZeeY

IamManchanda:
Tested it out myself => http://codepen.io/IamManchanda/pen/YNYjKe
Yup i can confirm this is a ~~bug~~ Missing `data-options` setting !!

@kball Kevin sir can you look into it !

DanielRuf:
It is still not used in the component but defined as variable.

DanielRuf:
I guess it was removed in https://github.com/zurb/foundation-sites/commit/e9e3bb9d33ebd7808c47ffcd69a94c867483047f so it might be safe to completely delete the unused variable / option.

DanielRuf:
Closing as the PR is merged.

PR #11030, opened 2018-03-09
docs: remove the unused data-btm-offset-pct in reveal

DanielRuf:
`data-btm-offset-pct` is unused but stil referenced in the docs.

Closes https://github.com/zurb/foundation-sites/issues/9727


ncoden (approved):
👍

Issue #11029, opened 2018-03-09
Thumnail class anchor elements don't fill thumbnail "frame" when image is smaller than its container


colin-marshall:
#### How to reproduce this bug:

1. Go to https://foundation.zurb.com/sites/docs/thumbnail.html
2. Enlarge browser window 1350px or wider
3. Observe middle thumbnail example of `.thumbnail` class on anchor tag

#### What should happen:
Middle thumbnail image should appear like the first and third/last thumbnail

#### What happened instead:
<img width="869" alt="screen shot 2018-03-08 at 8 13 30 pm" src="https://user-images.githubusercontent.com/4453733/37188851-95712a04-230e-11e8-8aad-23ea7fc01ab8.png">
--

This happens because the `max-width: 100%` on the anchor makes the thumbnail "frame" take up the width of the parent container. When the image, at its largest width, is smaller than the container, you get the result above. 

This is easily fixable with some extra CSS, but I am not sure if there's a good reason to support having the thumbnail class on the anchor tag because it defeats the purpose if you have to add custom CSS rules to make it work.

DanielRuf:
See https://github.com/zurb/foundation-docs/pull/30

This is no bug in Foundation Sites, th styles are only in the docs.

PR #11032, opened 2018-03-09
docs: center the thumbnails

DanielRuf:
https://github.com/zurb/foundation-docs/pull/30

The styles of the thumbnails in the docs use `display: block`
 with auto margins on both sides for exemplary centering them.

But this does not work well with `display: block` on an a tag.
~~`display: table`~~ `text-center` on the parent works much better in this example which is just for the docs and demonstrating the thumbnails.

Closes https://github.com/zurb/foundation-sites/issues/11029


ncoden (commented):
👍 Please update `foundation-docs` too

ncoden:
Thank you.

PR #11036, opened 2018-03-10
chore: update eslint/gulp-eslint to v4

colin-marshall:
I updated my global eslint to v4 and started seeing this error from SublimeLinter:
`Unexpected top-level property "ecmaVersion"`

Caused by this line:
https://github.com/zurb/foundation-sites/blob/develop/.eslintrc#L2

That setting is already present and in the correct place on this line:
https://github.com/zurb/foundation-sites/blob/develop/.eslintrc#L20

Figured I might as well update eslint/gulp-eslint for foundation-sites while I was at it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,4 @@
 {
-  "ecmaVersion": 6,
   "env": {
     "browser": true,
     "builtin": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -57,53 +57,6 @@
         }
       }
     },
-    "CSSselect": {
-      "version": "0.4.1",
-      "resolved": "https://registry.npmjs.org/CSSselect/-/CSSselect-0.4.1.tgz",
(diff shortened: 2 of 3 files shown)


ncoden:
I'm testing it.
@colin-marshall Did you checked for breaking changes between v3 and v4 ?

ncoden (approved):
See https://www.diffchecker.com/SLxCxCEC

So LGTM. There's still a lot of work to do (but I don't think it's a good idea to get into it now), but not more.

colin-marshall:
@ncoden Are you talking about all the warnings and the errors that we need to take care of (Or adjust the lint rules to ignore)?

ncoden:
Yes. In theory we should take care of them, but right now (at the end of v6) it is too much work for too small benefits.

ncoden:
On v7 we should use Prettier with the Github integrations :)

Issue #10178, opened 2017-06-13
PositionClass option is ignored by Dropdown 6.3.1


Lazerproof:
'this.options.positionClass' is set to the value of _getPositionClass without checking for the prior existance of the option during init of Dropdown

#### How can we reproduce this bug?

1. Set the positionClass option
2. It has no effect

#### What did you expect to happen?
The position should respect the positionClass option

#### What happened instead?
The position is regexed from the classname instead

andycochran:
Is this still an issue in 6.4.3?

DanielRuf:
> DEPRECATED: Class applied to adjust open position.

https://github.com/zurb/foundation-sites/blob/477563de08ae39c69ebf40b87f10f871696956c4/js/foundation.dropdown.js#L371-376

DanielRuf:
Closing as https://github.com/zurb/foundation-sites/pull/11038 is merged.

PR #11038, opened 2018-03-10
docs: remove the deprecated positionClass of dropdown

DanielRuf:
Closes https://github.com/zurb/foundation-sites/issues/10178

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -348,14 +348,6 @@ Dropdown.defaults = {
    */
   hOffset: 0,
   /**
-   * DEPRECATED: Class applied to adjust open position.
-   * @option
-   * @type {string}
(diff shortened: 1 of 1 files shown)


ncoden (approved):
hmm... I was surprised that `positionClass` has no effect (as "deprecated" does NOT mean "removed").

`positionClass` what removed at https://github.com/zurb/foundation-sites/pull/10053 (https://github.com/DanielRuf/foundation-sites/commit/f224b0c8da35e50f06deba114dcfb28d4bea6e8a, @kball). So it was actually _never_ deprecated but immediately replaced by `position` and `alignment`. That's not great for backward compatibility and ease of migration 😕 

Anyway, LGTM 👍. Good for master as it fixes the `v6.4.3` documentation.

Issue #10626, opened 2017-09-07
Destroying visible reveal does not remove is-reveal-open class from body.


GCheung55:
#### How to reproduce this bug:

1. Click on the button to open the modal.

2. With JS, destroy the open reveal element.
```javascript
$('#exampleModal11').foundation('destroy');
```

3. Confirm modal is destroyed.
4. Check `body` element for `is-reveal-open` class.

#### What should happen:
Destroying the reveal element should remove the `is-reveal-open` class from the `body` element.

#### What happened instead:
`body` element retains `is-reveal-open` class after the reveal element is destroyed.

#### Browser(s) and Device(s) tested on:
Chrome 60.0.3112.113

#### Foundation Version(s) you are using:
6.4.3

#### Test case link:
https://codepen.io/anon/pen/WEqOrB - inside the modal is a link with a click event that destroys the reveal element. It will also print out a boolean determined by whether the `body` element has the `is-reveal-open` class or not.

Basically the `_destroy` method does not cleanup the `is-reveal-open` class, where as the `finishUp` method does.

Arredatore:
Could it be linked to my topic? I also experience trouble with reveal https://github.com/zurb/foundation-sites/issues/10617

iapassas:
You may consider the following 
    var alrt = new Foundation.Reveal($('#myReveal'));
    alrt.close();  
    if ($('body').hasClass('is-reveal-open')) { $('body').removeClass('is-reveal-open');  }
    var bl = $('.reveal-overlay').filter(function() {  return $(this).css('display') == 'block';  });
    $(bl).remove();

drball:
I was having this problem and solved it by removing these animation effects: 
`data-animation-in="fade-in-fast" data-animation-out="fade-out"`

PR #11041, opened 2018-03-11
fix: remove the open classes if no visible reveal exists after destro…

DanielRuf:
…ying one

Closes https://github.com/zurb/foundation-sites/issues/10626

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -278,6 +278,11 @@ class Reveal extends Plugin {
       $('html').addClass('is-reveal-open');
     }
 
+    function removeRevealOpenClasses() {
+
+      $('html').removeClass('is-reveal-open');
+    }
+
     // Motion UI method of reveal
     if (this.options.animationIn) {
       function afterAnimation(){
@@ -409,7 +414,7 @@ class Reveal extends Plugin {
       var scrollTop = parseInt($("html").css("top"));
 
       if ($('.reveal:visible').length  === 0) {
-        $('html').removeClass('is-reveal-open');
+        removeRevealOpenClasses();
       }
 
       Keyboard.releaseFocus(_this.$element);
@@ -469,6 +474,10 @@ class Reveal extends Plugin {
     this.$element.hide().off();
     this.$anchor.off('.zf');
     $(window).off(`.zf.reveal:${this.id}`);
+
+    if ($('.reveal:visible').length  === 0) {
+      removeRevealOpenClasses();
+    }
   };
 }
 
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -273,14 +273,6 @@ class Reveal extends Plugin {
 
     var _this = this;
 
-    function addRevealOpenClasses() {
-      $('html').addClass('is-reveal-open');
-    }
-
-    function removeRevealOpenClasses() {
-      $('html').removeClass('is-reveal-open');
-    }
-
     // Motion UI method of reveal
     if (this.options.animationIn) {
       function afterAnimation(){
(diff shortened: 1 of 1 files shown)


ncoden (approved):
~~LGTM 👍~~

js/foundation.reveal.js
```diff
@@ -409,7 +414,7 @@ class Reveal extends Plugin {
       var scrollTop = parseInt($("html").css("top"));
 
       if ($('.reveal:visible').length  === 0) {
-        $('html').removeClass('is-reveal-open');
+        removeRevealOpenClasses();
```
ncoden: Please use `this.removeRevealOpenClasses();` (and run the docs and tests before creating a PR)

js/foundation.reveal.js
```diff
@@ -469,6 +474,10 @@ class Reveal extends Plugin {
     this.$element.hide().off();
     this.$anchor.off('.zf');
     $(window).off(`.zf.reveal:${this.id}`);
+
+    if ($('.reveal:visible').length  === 0) {
```
ncoden: A comment would be great. Also, do you think the test should be duplicated or moved to `removeRevealOpenClasses` ?

js/foundation.reveal.js
```diff
@@ -278,6 +278,11 @@ class Reveal extends Plugin {
       $('html').addClass('is-reveal-open');
     }
 
+    function removeRevealOpenClasses() {
+
```
ncoden: empty line

ncoden (changes requested):
Actually...

DanielRuf:
I guess I oversaw a small bit, will change it later.

DanielRuf:
Just forgot the `this.`, will add later =)

js/foundation.reveal.js
DanielRuf: Will remove it then.

js/foundation.reveal.js
DanielRuf: Right. Will do so in 7-8 hours, currently not on one of my machines.

js/foundation.reveal.js
DanielRuf: Not sure, the other function also just adds the class and does no prior check if is is already added.

ncoden:
@DanielRuf Please for each PR:
* run tests
* start the docs
* ensure that no similar change can be made elsewhere
* ensure thet the documentation is up-to-date
* add comments and references, there's never too much comments ;)

DanielRuf:
> run tests
Might be a good idea to create a precommit hook in general. I'll add it to my todo list.

Will do the rest later today.

ncoden:
@DanielRuf Can you share your Todo list (as issues) ?

js/foundation.reveal.js
DanielRuf: `addRevealOpenClasses` also has an empty line after the function declaration.

js/foundation.reveal.js
ncoden: If I remember correctly `.addClass()` already do that check.

DanielRuf: Meant the check for `:visible`.

DanielRuf: Meant the `:visible` check around the calls.

DanielRuf:
We have to check the other failing unit tests on Windows (IE 11 and Firefox 58) as these are different issues.

Who could take a look at them?

js/foundation.reveal.js
ncoden: We may not have understood each other.

What I'm trying to say is that:
* For now, `addRevealOpenClasses` can be called on all Reveal initializations and will only add the class once (thanks to jQuery `.addClass`). So it takes care of the "check" to put the class only when needed.
* Maybe we should have a similar behavior with `removeRevealOpenClasses`: being callable on all Reveal destructions and take care of the "check" to only remove the class when needed. So with [the check](https://github.com/zurb/foundation-sites/pull/11041/files#diff-5b5e0861de0a3864db2c3c28919700b7L411) inside the function.

But it's a _maybe_, and I don't want to loose too much time on a so small thing 😄

ncoden (approved):
👍

Issue #11050, opened 2018-03-13
Breakpoint function bug


rediris:
#### How to reproduce this bug:

1. Step one
Enter the following SASS:
```
@include breakpoint(large) {
  /*
  expected: should be 64em
  actual: 64em
  */
}
@include breakpoint(large down) {
  /*
  expected: should be 63.9375em
  actual: 74.9375em
  */
}

@media screen and #{breakpoint(large)} {
  /*
  expected: should be 64em
  actual: 64em
  */
}
@media screen and #{breakpoint(large down)} {
  /*
  expected: should be 63.9375em
  actual: 74.9375em
  */
}
```

2. Run your SASS compiler
3. Review the results:
```
@media print, screen and (min-width: 64em) {
  /*
  expected: should be 64em
  actual: 64em
  */ }

@media screen and (max-width: 74.9375em) {
  /*
  expected: should be 63.9375em
  actual: 74.9375em
  */ }

@media screen and (min-width: 64em) {
  /*
  expected: should be 64em
  actual: 64em
  */ }

@media screen and (max-width: 74.9375em) {
  /*
  expected: should be 63.9375em
  actual: 74.9375em
  */ }
```

#### What should happen:
What ought to happen is a proper breakpoint calculation for `max-width` breakpoints.

#### What happened instead:
The calculation was performed on the `next` breakpoint

#### Browser(s) and Device(s) tested on:
Tested in local CLI environment

#### Foundation Version(s) you are using:
`6.4.3`

#### Test case link:
Please recreate in your local CLI environment. Codepen doesn't support Foundation SASS compilation.

ncoden:
`[breakpoint] down` means "for this breakpoint and below". `breakpoint(large down)` must generate a breakpoint for large and below, which mean that the maximum width must be _just before the next breakpoint_.

```js
├─ xlarge max
│
├─ xlarge min
├─ large max    ┓
│               │
├─ large min    │
├─ medium max   │
│               │   "xlarge down" Breakpoint
├─ medium min   │
├─ small max    │
│               │
...
```
 
So it's not a bug. It is consistent with `breakpoint(large)` meaning "for this breakpoint and above".

If you don't want the large breakpoint to be included, use `breakpoint(medium down);`.

PR #11051, opened 2018-03-13
Breakpoint function bugfix for #11050

rediris:
This is intended to repair the bug illustrated here https://github.com/zurb/foundation-sites/issues/11050


ncoden:
Not a bug. See https://github.com/zurb/foundation-sites/issues/11050#issuecomment-372720035.

Issue #11049, opened 2018-03-13
[Toggler] Incorrect use of aria-expanded attribute invalids accessibility compliance


fleps:
#### How to reproduce this bug:

1. Configure Toggler plugin using any of the steps described on the documentation
2. Open the page and check the generated code

#### What should happen:
The `aria-expanded` attribute is being set on the element that contains the expandable content / is being expanded.

#### What happened instead:
This attribute must be set on the element that expands and collapses content, not on the content that is expanded or collapsed. 

**Note**: The information above came directly from a specialized company that audits WCAG Accessibility and they mentioned that this is a common mistake / miss-use of the aria-expanded attribute.

#### Foundation Version(s) you are using:
6.4.1

#### Test case link:
https://codepen.io/anon/pen/rdxagp

fleps:
**Additions**

I investigated the plugin and fixed on my side, as we are on a hurry to have valid AA compliance. Maybe this helps for a future fix:

On the `init` function, changed from this:

```
var id = this.$element[0].id;
$('[data-open="' + id + '"], [data-close="' + id + '"], [data-toggle="' + id + '"]').attr('aria-controls', id);
// If the target is hidden, add aria-hidden
this.$element.attr('aria-expanded', this.$element.is(':hidden') ? false : true);
```

To this:
```
// Add ARIA attributes to triggers
var id = this.$element[0].id;
$('[data-open="' + id + '"], [data-close="' + id + '"], [data-toggle="' + id + '"]').attr({
    'aria-controls': id,
    'aria-expanded': this.$element.is(':hidden') ? false : true
});
// The line below is no longer needed and was actually saying "aria-hidden" while it adds "aria-expanded"
//this.$element.attr('aria-expanded', this.$element.is(':hidden') ? false : true);
```

And also a change on the `_updateARIA` function is needed to this:

```
var id = this.$element[0].id;
$('[data-open="' + id + '"], [data-close="' + id + '"], [data-toggle="' + id + '"]').attr('aria-expanded', isOn ? true : false);
```

fleps:
Hi @DanielRuf, I saw the open PR but I have the impression that it will fix only the initial attribute set.

Doesn't the _updateARIA function needs to be changed also to toggle true/false on the trigger, or i'm reading it wrong?

DanielRuf:
You are right. I've updated the PR and tests accordingly.

PR #11052, opened 2018-03-13
fix: add aria-expanded to the toggler trigger

DanielRuf:
The `aria-expanded` attribute should be on the trigger not the container of the toggled element(s).

References:
https://www.w3.org/WAI/GL/wiki/Using_aria-expanded_to_indicate_the_state_of_a_collapsible_element
http://3needs.org/en/testing/code/aria-expanded.html
http://www.3needs.org/en/testing/code/aria-expanded-code.html

Closes https://github.com/zurb/foundation-sites/issues/11049

### DIFF
--- a/test/javascript/components/toggler.js
+++ b/test/javascript/components/toggler.js
@@ -106,13 +106,22 @@ describe('Toggler', function() {
 
     it('updates aria-expanded after the class is toggled', function() {
       $html = $('<div id="toggler" data-toggler="class"></div>').appendTo('body');
+      var $triggers = $(`<div>
+          <a data-open="toggler">Open</a>
+          <a data-close="toggler">Close</a>
+          <a data-toggle="toggler">Toggle</a>
+        </div>`).appendTo('body');
       plugin = new Foundation.Toggler($html, {});
 
       plugin._toggleClass();
-      $('#toggler').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-close]').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
 
       plugin._toggleClass();
-      $('#toggler').should.have.attr('aria-expanded', 'false');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'false');
+      $triggers.find('[data-close]').should.have.attr('aria-expanded', 'false');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'false');
     });
   });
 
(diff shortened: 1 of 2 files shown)


test/javascript/components/toggler.js
```diff
@@ -58,17 +58,31 @@ describe('Toggler', function() {
 
     it('sets aria-expanded to true if the element is visible', function() {
       $html = $('<div id="toggler" data-toggler="class"></div>').appendTo('body');
+      var $triggers = $(`<div>
+          <a data-open="toggler">Open</a>
+          <a data-close="toggler">Close</a>
+          <a data-toggle="toggler">Toggle</a>
+        </div>`).appendTo('body');
```
ncoden: Can you use a generator here to avoid code duplication ?

test/javascript/components/toggler.js
```diff
@@ -58,17 +58,31 @@ describe('Toggler', function() {
...
       plugin = new Foundation.Toggler($html, {});
 
-      $('#toggler').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-close]').should.have.attr('aria-expanded', 'true');
+      $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
```
ncoden: `open`, `close`, `open`. Should't it be `toggle`?

Also, Cannot we use `('[data-open], [data-close], [data-toggle]')` here, or would this test for only the first element / one element and not every ?

test/javascript/components/toggler.js
```diff
@@ -107,11 +130,19 @@ describe('Toggler', function() {
...
+        </div>`).appendTo('body');
+
       plugin = new Foundation.Toggler($html, {});
 
       $html.on('on.zf.toggler', function() {
         $('#toggler').should.be.visible;
-        $('#toggler').should.have.attr('aria-expanded', 'true');
+        $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
+        $triggers.find('[data-close]').should.have.attr('aria-expanded', 'true');
+        $triggers.find('[data-open]').should.have.attr('aria-expanded', 'true');
```
ncoden: Unit tests. So let's tests only what the function should do alone and not all its side-effects. We tested that aria attribute are well set and updated above, no need to repeat it there.

test/javascript/components/toggler.js
```diff
@@ -121,11 +152,19 @@ describe('Toggler', function() {
...
+        </div>`).appendTo('body');
+
       plugin = new Foundation.Toggler($html, {});
 
       $html.on('off.zf.toggler', function() {
         $('#toggler').should.be.hidden;
-        $('#toggler').should.have.attr('aria-expanded', 'false');
+        $triggers.find('[data-open]').should.have.attr('aria-expanded', 'false');
+        $triggers.find('[data-close]').should.have.attr('aria-expanded', 'false');
+        $triggers.find('[data-open]').should.have.attr('aria-expanded', 'false');
```
ncoden: Same

ncoden (changes requested):

test/javascript/components/toggler.js
DanielRuf: Ok, will prepare it then.

test/javascript/components/toggler.js
DanielRuf: Not sure here about `toggler`. Will check it then.

test/javascript/components/toggler.js
DanielRuf: Ok.

DanielRuf:
Pushed an update with the proposed changes.

ncoden (approved):
👍 LGTM, thanks @DanielRuf  !

PR #11060, opened 2018-03-15
Replace single with double quote in code example

NeonArray:
To keep with the consistency of the rest of the code examples


ncoden (approved):
Hi @ocularrhythm, thank you for your contribution !

I did the following:
* Replace _all_ single quotes by double quotes in HTML attributes of the documentation (in example and in the docs code)
* Get the latest version of `master` so tests run well.

So LGTM 😄

Issue #11066, opened 2018-03-15
[Tooltip] deprecated positionClass


DanielRuf:
Can we remove the deprecated `positionClass` option?

Tooltip: positionClass

ncoden:
It was never supported since `v6.4.3` and is already removed from the docs. https://github.com/zurb/foundation-sites/pull/11038 remember ? 😄

DanielRuf:
> [Tooltip]

DanielRuf:
This is not the dropdown component ;-)

DanielRuf:
See https://github.com/zurb/foundation-sites/blob/master/js/foundation.tooltip.js#L382-L387

ncoden:
Ah f***

PR #11067, opened 2018-03-15
chore: remove the deprecated positionClass in tooltip

DanielRuf:
`positionClass` was marked as deprecated a while ago.

I propose to plan this for 6.5 as this will also introduce a change to the code and the deprecated code will be removed.

Closes https://github.com/zurb/foundation-sites/issues/11066

### DIFF
--- a/js/foundation.tooltip.js
+++ b/js/foundation.tooltip.js
@@ -102,7 +102,7 @@ class Tooltip extends Positionable {
    * @private
    */
   _buildTemplate(id) {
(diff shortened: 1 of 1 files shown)


ncoden (approved):
Like for #11038, `positionClass` was broke in v6.4.0 and never worked since. The position class is set on the template by `Tooltip` but only read on the element by `Positionable`. So LGTM 👍

ncoden:
And yes, good for v6.4.4 since we break nothing (already broken) but simply clean the code.

ncoden:
Note: `v6.4.4` and `v6.5.0` were moved to `v6.5.0` and `v6.6.0`

PR #11072, opened 2018-03-18
fix: add the search input cancel button in Chrome/Safari again

DanielRuf:
Currently all modern browsers support the search input cancel button.
So removing it in normalize-scss makes no more sense.

In fact normalize 8.0.0 added it back, see https://github.com/necolas/normalize.css/commit/f7c98c4c859c15363763f10a439b63d85b9afba0

But the used [normalize-scss](https://github.com/JohnAlbin/normalize-scss) still relies on an old version of normalize.

References:
https://github.com/JohnAlbin/normalize-scss/issues/131
https://github.com/necolas/normalize.css/issues/685

Closes https://github.com/zurb/foundation-sites/issues/11015

### DIFF
--- a/scss/_global.scss
+++ b/scss/_global.scss
@@ -43,6 +43,7 @@ $medium-gray: #cacaca !default;
 /// Color used for dark gray UI items.
 /// @type Color
 $dark-gray: #8a8a8a !default;
+$dark-gray: #8a8a8a !default;
 
 /// Color used for black ui items.
 /// @type Color
@@ -135,6 +136,11 @@ $global-color-pick-contrast-tolerance: 0 !default;
 
 @mixin foundation-global-styles {
   @include -zf-normalize;
+
+  // Add the search input cancel button in Chrome/Safari again
+  [type="search"]::-webkit-search-cancel-button {
+    -webkit-appearance: searchfield-cancel-button;
+  }
 
   // These styles are applied to a <meta> tag, which is read by the Foundation JavaScript
   .foundation-mq {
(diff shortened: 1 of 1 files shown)


scss/_global.scss
```diff
@@ -43,6 +43,7 @@ $medium-gray: #cacaca !default;
 /// Color used for dark gray UI items.
 /// @type Color
 $dark-gray: #8a8a8a !default;
+$dark-gray: #8a8a8a !default;
```
ncoden: Duplicate

scss/_global.scss
```diff
@@ -136,6 +137,11 @@ $global-color-pick-contrast-tolerance: 0 !default;
 @mixin foundation-global-styles {
   @include -zf-normalize;
 
+  // Add the search input cancel button in Chrome/Safari again
+  [type="search"]::-webkit-search-cancel-button {
+    -webkit-appearance: searchfield-cancel-button;
+  }
+
```
ncoden: > ...or overwrite this in our global styles

Sorry, I had in mind a cleaner version of https://github.com/zurb/foundation-sites/issues/11015#issuecomment-373959867 to resolve this issue, so this overwrite would not be required. I'll take care of it.

ncoden (changes requested):

scss/_global.scss
DanielRuf: Was not awara that I hit the shortcut to duplicate it.

ncoden:
> Sorry, I had in mind a cleaner version of #11015 (comment) to resolve this issue, so this overwrite would not be required. I'll take care of it.

DanielRuf:
Not yet sure what a cleaner approach would be. Let me know how it should be then =)

PR #11077, opened 2018-03-18
fix: fix "load" actions in dynamically added components

ncoden:
Cleaner versions of https://github.com/zurb/foundation-sites/pull/10947 - fix: support dynamically added components (@DanielRuf)

Prevent bugs in Accordion, Magellan, OffCanvas, Reveal, Sticky, Tabs and trigger utilities when the component is dynamically added because the window `load` event they rely on would never be triggered.

#### Changes:
* add `onLoad` utility to wait for window load even after it is already loaded
* fix various component listeners on window load when the window is already loaded

#### Notes:
* In addition to the bug fix, Accordion now smoothly scroll to the opened pane when the hash is changed
* When a Magellan component is dynamically created, the window DO NOT scroll to the target corresponding to the location hash as the magellan did not created the target and should appear without effects on it after the page already loaded.

### DIFF
--- a/js/foundation.util.core.js
+++ b/js/foundation.util.core.js
@@ -61,4 +61,18 @@ function transitionend($elem){
   }
 }
 
-export {rtl, GetYoDigits, RegExpEscape, transitionend};
+/**
+ * Call the given function once the window is loaded,
+ * or immediately if the window is already loaded.
+ * @function
+ *
+ * @param {Function} fn - function to call on window load.
+ */
+function onLoad(fn) {
+  if (document.readyState === 'complete')
+    setTimeout(() => fn(), 0);
+  else
+    $(window).one('load', () => fn());
+}
+
+export {rtl, GetYoDigits, RegExpEscape, transitionend, onLoad};
--- a/js/foundation.util.core.js
+++ b/js/foundation.util.core.js
@@ -62,17 +62,32 @@ function transitionend($elem){
 }
 
 /**
- * Call the given function once the window is loaded,
- * or immediately if the window is already loaded.
+ * Return an event type to listen for window load.
+ *
+ * If `$elem` is passed, an event will be triggered on `$elem`. If window is already loaded, the event will still be triggered.
+ * If `handler` is passed, attach it to the event on `$elem`.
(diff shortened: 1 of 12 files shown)


js/foundation.util.core.js
```diff
@@ -61,4 +61,18 @@ function transitionend($elem){
...
+/**
+ * Call the given function once the window is loaded,
+ * or immediately if the window is already loaded.
+ * @function
+ *
+ * @param {Function} fn - function to call on window load.
+ */
+function onLoad(fn) {
+  if (document.readyState === 'complete')
+    setTimeout(() => fn(), 0);
```
DanielRuf: This should be `fn();` as `setTimeout` creates a new context and does not bind `_this` and nother variables in this case.

DanielRuf: This should be `fn();` as `setTimeout` creates a new context and does not bind `_this` and nother variables in this case.

DanielRuf (changes requested):

js/foundation.util.core.js
ncoden: There is no context to preserve. `_this` and other variables are not passed through the context.

DanielRuf: Take a look at the failing unit test(s) due to this ;-)

ncoden: Right, but this has nothing to do with the context, as `this` is not used.

https://github.com/ncoden/foundation-sites/blob/f353fc2f1a6d556a511734fa7eeacd0221cd064d/js/foundation.offcanvas.js#L169-L173

ncoden: The error comes from the component being already destroyed when the function in `setTimeout` is executed. I don't exactly know why, but the best solution would be to handle this case as we do for others events listeners by unregistering it on `destroy`.

ncoden:
@DanielRuf I fixed the bug in tests caused by the `onLoad` event that was not cleared on component destroy (so the event is called once component is destroyed, with an empty this). Now I use an event in both cases, `load` or the custom `_didLoad` returned by `onLoad` so we can use:

```js
const listener = onLoad($elem, handler);
...
$elem.off(listener);
```

DanielRuf (approved):

ncoden:
Thanks for the review.

PR #11078, opened 2018-03-18
 fix: change the calculation of box dimensions, fixes IE behavior

DanielRuf:
IE11 does not properly calculate `getBoundingClientRect` and produces values like `199.9998779296875`

image

jQuery uses a different approach depending on the node type:
https://github.com/jquery/jquery/blob/master/src/dimensions.js#L35-L41

But we should probably use `clientHeight` and `clientWidth`:

https://github.com/nefe/You-Dont-Need-jQuery#2.2

### DIFF
--- a/js/foundation.util.box.js
+++ b/js/foundation.util.box.js
@@ -74,21 +74,27 @@ function GetDimensions(elem){
   }
 
   var rect = elem.getBoundingClientRect(),
-      parRect = elem.parentNode.getBoundingClientRect(),
-      winRect = document.body.getBoundingClientRect(),
+      par = elem.parentNode,
+      parRect = par.getBoundingClientRect(),
+      win = document.body,
+      winRect = win.getBoundingClientRect(),
+      elemWidth = Math.max(elem.scrollWidth, elem.offsetWidth, elem.clientWidth),
+      elemHeight = Math.max(elem.scrollHeight, elem.offsetHeight, elem.clientHeight),
+      parWidth = Math.max(par.scrollWidth, par.offsetWidth, par.clientWidth),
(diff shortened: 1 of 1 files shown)


js/foundation.util.box.js
```diff
@@ -74,21 +74,22 @@ function GetDimensions(elem){
...
+      parRect = par.getBoundingClientRect(),
       winRect = document.body.getBoundingClientRect(),
       winY = window.pageYOffset,
       winX = window.pageXOffset;
 
   return {
-    width: rect.width,
-    height: rect.height,
+    width: elem.clientWidth,
+    height: elem.clientHeight,
```
ncoden: > The Element.clientWidth property is zero for elements with no CSS or inline layout boxes, otherwise it's the inner width of an element in pixels. It includes padding but not the vertical scrollbar (if present, if rendered), border or margin.
>
>  https://developer.mozilla.org/en-US/docs/Web/API/Element/clientWidth

So border, margin, CSS transformations and zero on inline element. I guess this would introduce some bugs ;)

I would personally use jQuery for these things, as it take care for us of all the low-level and browser specific stuff.

You can also take a look at https://github.com/zurb/foundation-sites/search?q=getBoundingClientRect&type=Issues&utf8=%E2%9C%93 to ensure not to reopen old issues.

ncoden (changes requested):

js/foundation.util.box.js
DanielRuf: Not sure but I do not see any error. So .width / height directly? Well, but we would have to migrate this to ES5 then anyway.

DanielRuf: @JeremyEnglert @kball what solution should we choose here?

DanielRuf:
It seems this does not occur anymore so I'm closing this PR.

Issue #11059, opened 2018-03-15
removing gutters on cells (semantically) 


nitrokevin:
Hi I'm not sure if this is an issue or expected behaviour, but when I set the gutters to false on a cell the margin values are removed but the width rem-cal that takes into account the gutters remains
#### How to reproduce this bug:

1. Set the cell to have no gutters using false @include xy-cell(3,false); 
2. Check in browser, margin still appears due to width calculation
3. Change code to keep gutters but with 0 size  @include xy-cell(3,true,0);
4. Margins have now been removed completely

#### What should happen:
I would expect setting gutters to false would completely remove the gutters
#### Browser(s) and Device(s) tested on:
All
#### Foundation Version(s) you are using:
6.4.3

AtlantaJones:
@nitrokevin  I'm seeing the same thing. Did you find a solution?

DanielRuf:
Please test the linked PR.

AtlantaJones:
I'll try to test as well tonight. Is there an easy way to test something like that when you already have the latest pulled in via NPM?

DanielRuf:
@AtlantaJones you can manually clone my fork into the directory and checkout the branch.

DanielRuf:
Hi @AtlantaJones, do you need any help or are there some questions left about how you can test it on your side?

PR #11081, opened 2018-03-21
fix: respect $gutter-output in the -xy-cell-properties mixin

DanielRuf:
Currently the gutters are included in the width calculation even if `$gutter-output` is set to `false`.
This should not be the case.

Closes https://github.com/zurb/foundation-sites/issues/11059

### DIFF
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -95,7 +95,7 @@
 /// Creates a cell for your grid.
 ///
 /// @param {Keyword|Number} $size [full] - The size of your cell. Can be `full` (default) for 100% width, `auto` to use up available space and `shrink` to use up only required space.
-/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters
+/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters. Always `true` for margin gutters.
 /// @param {Number|Map} $gutters [$grid-margin-gutters] - Map or single value for gutters.
 /// @param {Keyword} $gutter-type [margin] - Map or single value for gutters.
 /// @param {List} $gutter-position [right left] - The position to apply gutters to. Accepts `top`, `bottom`, `left`, `right` in any combination.
@@ -143,7 +143,7 @@
 /// Creates a single breakpoint sized grid. Used to generate our grid classes.
 ///
 /// @param {Keyword|Number} $size [full] - The size of your cell. Can be `full` (default) for 100% width, `auto` to use up available space and `shrink` to use up only required space.
-/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters
+/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters. Always `true` for margin gutters.
 /// @param {Number|Map} $gutters [$grid-margin-gutters] - Map or single value for gutters.
 /// @param {Keyword} $gutter-type [margin] - Map or single value for gutters.
 /// @param {String} $breakpoint [null] - The name of the breakpoint size in your gutters map to get the size from. If using with the `breakpoint()` mixin this will be set automatically unless manually entered.
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -127,7 +127,7 @@
     // Base flex properties
     @include xy-cell-base($size);
 
-    @if($gutter-type == 'margin') {
+    @if($gutter-output == true && $gutter-type == 'margin') {
       @include -xy-cell-properties($size, $gutter, $vertical);
     }
     @else {
@@ -163,7 +163,7 @@
   $gutter: -zf-get-bp-val($gutters, $breakpoint);
   $gutter-position: if($vertical == true, top bottom, left right);
 
-  @if($gutter-type == 'margin') {
+  @if($gutter-output == true and $gutter-type == 'margin') {
     @include -xy-cell-properties($size, $gutter, $vertical);
   }
   @else {
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -127,7 +127,7 @@
     // Base flex properties
     @include xy-cell-base($size);
 
-    @if($gutter-output == true && $gutter-type == 'margin') {
+    @if($gutter-output == true and $gutter-type == 'margin') {
       @include -xy-cell-properties($size, $gutter, $vertical);
     }
     @else {
--- a/scss/xy-grid/_cell.scss
+++ b/scss/xy-grid/_cell.scss
@@ -127,14 +127,11 @@
     // Base flex properties
     @include xy-cell-base($size);
 
-    @if($gutter-output == true and $gutter-type == 'margin') {
-      @include -xy-cell-properties($size, $gutter, $vertical);
-    }
-    @else {
-      @include -xy-cell-properties($size, 0, $vertical);
-    }
+    $-gutter-output: if($gutter-type == 'margin', true, $gutter-output);
+    $-gutter-margin: if($gutter-type == 'margin', $gutter, 0);
 
-    @if $gutter-output {
+    @include -xy-cell-properties($size, $-gutter-margin, $vertical);
+    @if ($-gutter-output) {
       @include xy-gutters($gutter, $gutter-type, $gutter-position);
     }
   }
@@ -163,16 +160,11 @@
   $gutter: -zf-get-bp-val($gutters, $breakpoint);
   $gutter-position: if($vertical == true, top bottom, left right);
 
-  @if($gutter-output == true and $gutter-type == 'margin') {
-    @include -xy-cell-properties($size, $gutter, $vertical);
-  }
-  @else {
-    @include -xy-cell-properties($size, 0, $vertical);
-  }
+  $-gutter-output: if($gutter-type == 'margin', true, $gutter-output);
+  $-gutter-margin: if($gutter-type == 'margin', $gutter, 0);
 
-  // If we want to output the gutters
-  @if($gutter-output) {
-    // TODO: Figure out if we need to pass breakpoint in here too.
+  @include -xy-cell-properties($size, $-gutter-margin, $vertical);
+  @if ($-gutter-output) {
     @include xy-gutters($gutter, $gutter-type, $gutter-position);
   }
 }
--- a/scss/xy-grid/_layout.scss
+++ b/scss/xy-grid/_layout.scss
@@ -10,7 +10,7 @@
 ///
 /// @param {Number} $n - Number of elements to display per row.
 /// @param {String} $selector ['.cell'] - Selector(s) to use for child elements.
-/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters
+/// @param {Boolean} $gutter-output [true] - Whether or not to output gutters. Always `true` for margin gutters.
 /// @param {Number|Map} $gutters [$grid-margin-gutters] - Map or single value for gutters.
 /// @param {Keyword} $gutter-type [margin] - Type of gutter to output. Accepts `margin` or `padding`.
 /// @param {List} $gutter-position [right left] - The position to apply gutters to. Accepts `top`, `bottom`, `left`, `right` in any combination.


scss/xy-grid/_cell.scss
```diff
@@ -127,7 +127,7 @@
     // Base flex properties
     @include xy-cell-base($size);
 
-    @if($gutter-type == 'margin') {
+    @if($gutter-output == true && $gutter-type == 'margin') {
```
ncoden: This does not prevent the gutter to be output. This set the external margin to 0, this reset them. See the `-xy-cell-properties` mixin:

https://github.com/zurb/foundation-sites/blob/69e3a68abc2649f84c97a047473fb9ed1152dcef/scss/xy-grid/_cell.scss#L91

I guess this is part of the big problem they got with margin gutters while working on the XY grid. Poke @kball. As the width _is_ the gutter I don't see any way to set one or the other with margin gutters. Maybe all we can do is update the documentation ?

ncoden (changes requested):

scss/xy-grid/_cell.scss
DanielRuf: In general this line adjusts the width using calc - gutter.

ncoden: Yes. And width `calc(100% - 0)` = 0 gutter. This is different from "no gutter".

DanielRuf: Check the generated code, it does not include any `- 0` when the mixin is used (se3 linked issue). Or I do not understand what you mean.

ncoden: Yeah because Sass optimize calc, but it's the same. `100%` width in margin grid = 0-width gutter.

If you set margin gutters before on the cell, then call the mixin with `$gutter-output: false`, it will not "prevent to output gutters" but "output 0-width gutters", and your cell will not have gutters anymore.

DanielRuf: Then I'll also adjust the properties mixin.

DanielRuf: The mixin is correct.

Please check https://github.com/DanielRuf/foundation-sites/blob/565c77d5dcb80852271c7f769721979629e6d0df/scss/xy-grid/_cell.scss#L90 which clearly does not use calc and does not include the gutter in the calculation. Sass itself does not convert `calc( val1 - 0)` to `calc(val1)`

Also see 
bildschirmfoto 2018-03-22 um 10 25 04

DanielRuf: > My point is that margin gutters will always be included in width calculation because margin gutters are part of the width.

I don't see this case in the current code ;-)

ncoden: Oh I knew Sass made optimizations but maybe I'm wrong with mixed units.
Anyway, `100%` is generated and this still means "0 gutters" for margin grids.

DanielRuf: > Anyway, 100% is generated and this still means "0 gutters" for margin gutters.

Sure, and this is what should happen there according to the issue description and the rest of the code afaik.

DanielRuf: Or am I wrong? Correct me if there is another way and if I misinterpreted the issue.

ncoden: hmm it's complicated. The issue author expect `$gutter-output: false` to _remove gutters_ (that's what this PR does), but according to its definition `$gutter-output: false` should simply _prevent to generate any gutters_ (which is different from removing them)...

I think we should keep this separated and explicitely said that `$gutter-output` is not supported for margin grids, as the width/height is related to the margin and setting one without the other would create bugs (cells taking too much or not enough space). So maybe we should ignore it for margin gutters (and advice to set `$gutters: 0` instead).

ncoden: @DanielRuf What do you think ?

ncoden: Poke @DanielRuf.

DanielRuf: Not sure, what would @SassNinja say here?

SassNinja: @ncoden I think the expected behavior described in the issue (https://github.com/zurb/foundation-sites/issues/11059) is quite clear:
`@include xy-cell(3, false)` is expected to behave identical to `@include xy-cell(3, false, 0)` which currently is not happening.

Regarding the `-xy-cell-properties` mixin I took a look and noticed the `calc()` doesn't get used if `$margin-gutter == 0`.

So in my eyes the fix of @DanielRuf works (except that it has to be `and` instead of `&&`!) and generates the expected output (without any `calc()` in it).

Quite difficult to show before & after in the same codepen so I just made quickly two pens.

SassNinja: @DanielRuf as mentioned above it has to be `and` here instead of `&&`

SassNinja (changes requested):

scss/xy-grid/_cell.scss
ncoden: @SassNinja See https://github.com/zurb/foundation-sites/pull/11081#discussion_r176386304. I disagree with the assumption that `@include xy-cell(3, false)` should be equal to `@include xy-cell(3, false, 0)` for the following reasons.

1. Preventing to generate gutters (as `$gutter-output: false` do) and generating zero gutters (as `$gutters: 0` do) is two different things. So we can not expect `@include xy-cell(3, false)` and `@include xy-cell(3, true, 0)` to be equal
2. With margin gutters, an XY cell cannot technically separate its width from its gutter. We cannot generate XY width without generating its gutters too.
3. An XY cell cannot be functional without a width defined.
4. In conclusion from the previous points, we cannot define a XY cell using margin gutters without defining its gutters, the `$gutter-output` option cannot be set to `false`, so it is useless and should be ignored.

I propose the following API:
```scss
// Generate cell with zero gutters
@include xy-cell(12, $gutter-type: margin, $gutters: 0);
```

```scss
// Ignore $gutter-output and generate cell with default gutters
@include xy-cell(12, $gutter-type: margin, $gutter-output: false);
```

DanielRuf: Already done =)

ncoden: Ok, I'll try to be more clear about what is wrong in my opinon. :)

> Why not? Seems to be the expected behavior and also makes sense to me. My understanding of $gutter-output:false is that it simply overrides the $gutters param and uses 0 instead.

```css
.example { padding: 0 }
```
...and...
```css
.example { }
```

...are two different things, because in one case previously defined gutters get removed and not in the other. "Preventing to generate properties" is common and useful with responsive Sass to just set the properties you want in a bigger breakpoint. This is why I talked about a `xy-cell-size` mixin, it would allow to change the size/gutter of a cell across breakpoints without duplicating the flexbox stuff.

_Currently_ for margin gutters it is the same, yes. But according to the feature description "Whether or not to output gutters" this should not reset gutters but do nothing instead.

With the current behavior or the one you propose in your Codepen, the `$gutter-output` behavior is not the same following the gutter type. For margins this is a "reset" and for paddings this is a "do nothing".

> But I'm focusing more on the result here and that looks good to me (no visible gutter between the cells).

I understand and I agree with you that the result code looks good. This is not my point. My point is the consistency of the feature we provide and the relation between _what we ask_ and _what we get_. As I explained before, `$gutter-output` is inconsistent and does not (and cannot) follow the expected behavior for margin gutters.

> Looks somehow more complex and error-prone to me.

Every option control one thing. Some options does not make sense in some contexts (like `$gutter-output` for margin gutters) and should be ignored. IMHO this is more precise and less error-prone as the understanding of each option behavior allow to understand the global behavior without having to care about special cases. `$gutter-output` controls whether or not we output gutters, `$gutters` controls the gutters width we output.

Using `$gutter-output: false` as an alias for `$gutters: 0` is indeed simpler, but this is an illusion as the behavior behind that is actually inconsistent and with side effects. Changing `$gutter-type` in your code will change the behavior of `$gutter-output` and lead to changes you could not expect.

What do you think? Does this make sense to you?

SassNinja: > "Preventing to generate properties" is common and useful with responsive Sass to just set the properties you want in a bigger breakpoint. This is why I talked about a `xy-cell-size` mixin, it would allow to change the size/gutter of a cell across breakpoints without duplicating the flexbox stuff.

Alright I think I understand now what you mean. But what what is the `xy-cell-size` mixin you mentioned supposed to do?
There's already a `xy-cell-size` function. Would it only look like this?

```scss
@mixin xy-cell-size($size, $direction: width) {
  #{$direction}: xy-cell-size($size);
}
```

> My point is the consistency of the feature we provide and the relation between what we ask and what we get. As I explained before, `$gutter-output` is inconsistent and does not (and cannot) follow the expected behavior for margin gutters.

Ok, makes sense. If you don't ask for gutter it should be ignored instead of resetted. So `@include xy-cell(3,false)` should not generate the same output as `@include xy-cell(3,true,0)`

> Every option control one thing. Some options does not make sense in some contexts (like `$gutter-output` for margin gutters) and should be ignored. IMHO this is more precise and less error-prone as the understanding of each option behavior allow to understand the global behavior without having to care about special cases. `$gutter-output` controls whether or not we output gutters, `$gutters` controls the gutters width we output.

Sounds good but how exactly would you realize it?
Your proposed doesn't seem to change the mixin but just uses keyword arguments.
User could still do something like `@include xy-cell(12, $gutter-type: margin, $gutter-output: false, $gutter: 0)`.
We can only ignore the `$gutter` if `$gutter-output` is false (which already happening although it's internally just set to 0 but the `@include xy-gutters` gets skipped)

I've checked the current PR and even with the discussed behavior it seems to generate the expected output:

1. `@include xy-cell($size: 3, $gutter-output: true, $gutters: 0, $gutter-type: margin)` => output contains `margin-x: 0`
2. `@include xy-cell($size: 3, $gutter-output: false, $gutters: 0, $gutter-type: margin)` => output contains no margin (`$gutter` doesn't get used here or rather is ignored and internally 0 is used for calculation)

Padding behaves similar.
Thus I don't see a case where the current PR does not generate the right output.

Maybe you can provide an example which doesn't generate the right output?

> By the way, if this is not over engineering I don't know what it is

That's for sure 😄 But it's usually worth it if everything is clear afterwards and we can be sure to have the best solution

ncoden: > There's already a xy-cell-size function. Would it only look like this?
> (...)

No because for margin cells the gutter would be removed. `xy-cell-size` would also have to handle gutters. So it would basically be all the width/gutter stuff inside `xy-cell` without the call to `xy-cell-base`.

> Sounds good but how exactly would you realize it?

Like I said `$gutter-output` would be ignored for margin gutters. This would be a padding-only option as only padding gutters can be escaped. In the API I proposed, the only change is that `$gutter-output` is ignored.

So for example:
```scss
// Margin cell with default gutters
@include xy-cell(12, $gutter-type: margin, $gutter-output: true);
@include xy-cell(12, $gutter-type: margin, $gutter-output: false);

// Margin cell with zero-width gutters
@include xy-cell(12, $gutter-type: margin, $gutter-output: true, $gutter: 0);
@include xy-cell(12, $gutter-type: margin, $gutter-output: false, $gutter: 0);

// Padding cell with default gutters
@include xy-cell(12, $gutter-type: padding, $gutter-output: true);
// Padding cell without gutters set (they will be inherited)
@include xy-cell(12, $gutter-type: margin, $gutter-output: false);

// Padding cell with zero-width gutters
@include xy-cell(12, $gutter-type: padding, $gutter-output: true, $gutter: 0);
// Padding cell without gutters set (they will be inherited)
@include xy-cell(12, $gutter-type: margin, $gutter-output: false, $gutter: 0);
```

So:
* If `$gutter-type` is `$gutter-type`, `$gutter-output` is ignored (considered as true)
* If `$gutter-output` is `false`, `$gutter` and `$gutters` are ignored (not used)

Does it make more sense to you ?

ncoden: @DanielRuf @SassNinja

SassNinja: > So:
> - If `$gutter-type` is margin, `$gutter-output` is ignored (considered as true)
> - If `$gutter-output` is false, `$gutter and `$gutters` are ignored (not used)
>
> Does it make more sense to you ?

Yes @ncoden, makes sense to me :+1:

Thus the code has to be changed into this, right?

```scss
@if($gutter-type == 'margin') {
  $gutter-output: true;
}

@if($gutter-output == true) {
  @include -xy-cell-properties($size, $gutter, $vertical);
}
@else {
  @include -xy-cell-properties($size, 0, $vertical);
}

@if $gutter-output {
  @include xy-gutters($gutter, $gutter-type, $gutter-position);
}
```

So the bottom line is the include from the issue `@include xy-cell(3, false)`  will behave identical to `@include xy-cell(3, true)`.
And apart from that you can't get rid of the margin left/right anymore when using margin gutter because `@include xy-cell(3, true, 0)` will output margin left/right 0.

ncoden: The second argument of `-xy-cell-properties` is `$margin-gutter` so it would be more logical and simple to have:

```scss
$-gutter-output: if($gutter-type == 'margin', true, $gutter-output);
$-gutter-margin: if($gutter-type == 'margin', $gutter, 0);

@include -xy-cell-properties($size, $-gutter-margin, $vertical);
@if ($-gutter-output) {
  @include xy-gutters($gutter, $gutter-type, $gutter-position);
}
```

@DanielRuf What do you think ? BTW I that think using the git "co-authors" feature would be relevant here 😄

DanielRuf: I followed the discussion so far but have to reread the last 3 comments later to make a decision.
So far it makes sense to be more consistent here and introduce this breaking change that you describe.

DanielRuf:
Applied the proposed change by @ncoden

DanielRuf:
> Could you amend it ?

How should the commit message / title be / look like?

ncoden:
Well something like:

````
feat: ignore `gutter-output` for margin grid in `xy-cell` mixins

`gutter-output` cannot be set to `false` for margin grid as the gutter (margin)
is defined with the width. Gutters can only be set or reset to 0, which does
not fullful the role of `gutter-output`.

See discussion at https://github.com/zurb/foundation-sites/pull/11081

**BREAKING CHANGE**

Before:
```scss
@include xy-cell(12, $gutter-type: margin, $gutter-output: false);
@include xy-cell-static(full, $gutter-type: margin, $gutter-output: false);
```

After
```scss
@include xy-cell(12, $gutter-type: margin, $gutters: 0);
@include xy-cell-static(full, $gutter-type: margin, $gutters: 0);
```

Co-authored-by: Nicolas Coden <nicolas@ncoden.fr>
````

What do you think ? ;)

DanielRuf:
Done.

ncoden:
Teamwork 🎉 

screen shot 2018-04-30 at 10 07 37

Thank you @DanielRuf @SassNinja

ncoden:
https://github.com/DanielRuf/foundation-sites/blob/828481aa8eb67258a53c3a68ba41221388ce0045/scss/xy-grid/_cell.scss#L98

Documentation must be updated to mention this behavior. Like `ignored for margin gutters`.

SassNinja:
I agree regarding the documentation update:
Maybe something as `Whether or not to output gutters when using padding gutter type. Always true for margin.` ?

ncoden (approved):
I took care of the documentation changes. LGTM 👍.

Migration needed, so planned for `v6.6.0`

ncoden:
@SassNinja Could you review this ?

SassNinja (approved):

SassNinja:
LGTM @ncoden

I've tested it with padding & margin and it works as expected 👍

PR #11086, opened 2018-03-23
chore: drop `normalize-scss` and upgrade to normalize.css v8.0.0

ncoden:
`normalize-scss` is not maintained and actually pollute the Sass namespace. Move back to a as-hard-to-maintain approach with a Sass mixin hosting the normalize.css code.

Note: `/* */` comments where replaced by `//` comments to not being output.

#### Changes
* Remove `normalize-scss` dependencies
* Create `vendor/normalize` SCSS file hosting the normalize v8.0.0 CSS inside a mixin
* Make only the root `./vendor` folder being ignored by git

Closes https://github.com/zurb/foundation-sites/issues/11015
Replaces https://github.com/zurb/foundation-sites/pull/11016
Replaces https://github.com/zurb/foundation-sites/pull/11072


DanielRuf (approved):
Looks good so far but we may have to check all exaples for layout changes due to the new normalize version.

ncoden:
@DanielRuf See https://github.com/necolas/normalize.css/blob/master/CHANGELOG.md

DanielRuf:
Sidenote: https://github.com/zurb/foundation-sites/blob/ab4beb10bb91034c76dcf0457c105766d8ef8b47/scss/components/_reveal.scss#L42-L50 is added before the other styles. Should we move this to another place or is it still needed (there)?

See the generated foundation.css file at the very top.

ncoden:
@DanielRuf This is related to https://github.com/zurb/foundation-sites/pull/10856 and components sorting. It's at the top because placeholders code is generated (if used) where it is placed. The problem in this case is that it is used inside a modifier mixin (`reveal-modal-width`), not the main one (`foundation-reveal`) and I guess the modifier mixin should be usable without the main one...

So, it seems it's part of an "initialization" step (after import and before each component generation) that does not exist in our architecture (and most SCSS architectures). That's too deep for v6 but a good challenge for v7 :). It's does not cause any bugs, right ?

DanielRuf:
> It's does not cause any bugs, right ?

It adds the styles even if none of the mixins is used so it may cause issues.

ncoden:
hmm... No it shouldn't. That's a placeholder, it generate styles only if used. And empty breakpoints are removed by Sass.

DanielRuf:
> components sorting. It's at the top because placeholders code is generated (if used) where it is placed

This is not the case. The import statement alone adds this directly.

See https://github.com/zurb/foundation-sites/pull/11087

ncoden:
> This is not the case. The import statement alone adds this directly.

Did you test it ?

DanielRuf:
At least it makes not much sense to add it up there. The other code in my PR works too. Deduplication is done by libsass.

DanielRuf:
Anyway, we should check that the side effects of the normalize update on the components.

DanielRuf:
> This is not the case. The import statement alone adds this directly.
> 
> Did you test it ?

Probably it was added because the docs and default foundation.scss use it with the include statement.

PR #11088, opened 2018-03-24
chore: remove rem-base deprecation note

DanielRuf:
Currently the docs contain the following deprecation note:

> `$rem-base` was deprecated in version 6.1, in favor of using `$global-font-size` to define rem calculation.

So it is a very old notice and we can safely remove it now.


ncoden (approved):
👍 Thanks

PR #11089, opened 2018-03-24
docs: fix check for flexbox-mode is-active class in component list

DanielRuf:
Currently the sidenav with the component list is not open when we are on https://foundation.zurb.com/sites/docs/flexbox-mode.html

This PR solves it.


ncoden (approved):
👍

Issue #8012, opened 2016-01-26
[Reveal] Reveal problem when using deepLink and more nested modal windows


martinpesout:
**How can we reproduce this bug?**
1. Implement more nested modal windows http://foundation.zurb.com/sites/docs/reveal.html#nested-modal
2. Enable deepLink for all modals
3. Try nested modals and try browser "back button" to return to previous state.

**What did you expect to happen?**

When you go from the first modal window to another, I'd like to see URL with appropriate hash, like `#modal-1`. But because we have 2 nested modals, new modal will change current URL by https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L277 but previous modal will be closed right after opening. So URL will be changed again by https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L471

**Possible solution**

My temporary solution is replace this part of code https://github.com/zurb/foundation-sites/blob/develop/js/foundation.reveal.js#L469-L475 with something like this: 

```
    setTimeout(function() {
        if ($('.reveal').filter(':visible').length === 0) {
            if(_this.options.deepLink){
              if(window.history.replaceState){
                window.history.replaceState("", document.title, window.location.pathname);
              }else{
                window.location.hash = '';
              }
            }
        }
    }, 200);
```

I know that it isn't ideal but it's working for now.

DanielRuf:
Thanks for reporting this. I guess we have to refctor this a bit to get this case working like it should work.

DanielRuf:
What do you think @ncoden ? Revisit this for v7?

ncoden:
@DanielRuf  I don't even understand what the issue is :/

DanielRuf:
The issue is that nested reveals do not open when you go back in the history like they were opened but in reverse. It stops after the first last reveal.

DanielRuf:
Try seeing all previously opened reveals by hitting the history.back button.

https://codepen.io/DanielRuf/pen/OvPpzb

ncoden:
Ok thanks, it's _after_ modals got closed. I'll take a look at it tomorrow.

ncoden:
> tomorrow.

Ahah. On it.

ncoden:
See https://github.com/zurb/foundation-sites/pull/11093

PR #11093, opened 2018-03-26
fix: fix history browsing with nested Reveal and deep-linking

ncoden:
### Changes:
* update history when closing Reveal with `replaceState`
  When closing a Reveal with `replaceState: true`, push a new entry to reset the hash. So going back to the history reopen the modals opened before in reverse order.
* do not update history when opening Reveal with its hash already set (like when going back in history)
* add visual tests for nested Reveal with deep-linking

Closes: #8012
Visual tests are from: https://codepen.io/DanielRuf/pen/OvPpzb

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -216,8 +216,8 @@ class Reveal extends Plugin {
    */
   open() {
     // either update or replace browser history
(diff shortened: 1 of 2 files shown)


DanielRuf (approved):
Found no issues 👍

PR #11094, opened 2018-03-26
fix: reveal sass error with @extend within @media

cietho:
With foundation-sites@6.4.3/6.4.4-rc1 and gulp@4.0.0 with gulp-sass@3.2.1 I get the following error when I'm using reveal in my project: [@include foundation-reveal();]

> You may not @extend an outer selector from within @media.
> You may not @extend an outer selector from within @media. You may only @extend selectors within the same directive.
> From "@extend %reveal-centered" on line 105 of node_modules/foundation-sites/scss/components/_reveal.scss on line 45 of node_modules/foundation-sites/scss/components/_reveal.scss

To fix this issue I removed the unnecessary placeholder and applied the settings directly within the @media directive.

### DIFF
--- a/scss/components/_reveal.scss
+++ b/scss/components/_reveal.scss
@@ -39,17 +39,6 @@ $reveal-zindex: 1005 !default;
 $reveal-overlay-background: rgba($black, 0.45) !default;
 
 
-// Placeholder selector for medium-and-up modals
-// Prevents duplicate CSS when defining multiple Reveal sizes
-@include breakpoint(medium) {
-  %reveal-centered {
-    right: auto;
-    left: auto;
-    margin: 0 auto;
-  }
-}
-
-
 /// Adds styles for a modal overlay.
 /// @param {Color} $background [$reveal-overlay-background] - Background color of the overlay.
 @mixin reveal-overlay($background: $reveal-overlay-background) {
@@ -102,7 +91,9 @@ $reveal-overlay-background: rgba($black, 0.45) !default;
   $max-width: $reveal-max-width
 ) {
   @include breakpoint(medium) {
-    @extend %reveal-centered;
+    right: auto;
+    left: auto;
+    margin: 0 auto;
     width: $width;
     max-width: $max-width;
   }


ncoden (commented):
Hi @cietho.

You should not have this error as the placeholder and the elements extending it are all in the same media query: `breakpoint(medium)`. Sass should not have trouble to build a selector for that placeholder.

I tried to compile with the same version as you (`gulp-sass@3.2.1`) and it works fine.

Did you tried to extend `%reveal-centered` or to include `reveal-modal-width` or `foundation-reveal` from within a breakpoint (or any media query) ? This is the only use case imho that could give you this error.

DanielRuf:
It is just a warning on the terminal afaik.

cietho:
Hi @ncoden 

Thanks for the quick reply. I'm not extending `%reveal-centered` in any of my files. The error log always points to node_modules/foundation-sites/scss/components/_reveal.scss.

I will create an example repo where this error occurs and link it here later.

cietho:
Hi @ncoden,

I've created an example project where the error occurs: [foundation-sites_issue_11094.zip](https://github.com/zurb/foundation-sites/files/1856170/foundation-sites_issue_11094.zip)

Best regards,
Thomas

ncoden:
@cietho Please create a public repo the next time, it's way more simple to review.
I'll take a look at this.

ncoden:
Ok, this seems to be an issue with `gulp-sass 3.*`. Unlike `gulp-sass 2.*`, this version has some trouble to ignore "nested media-query" when the two media query are the same. A solution is to move @extend %reveal-centered` out of the breakpoint in `@mixin reveal-modal-width`

cietho:
Hmm ok, but I don't see why this has to be a placeholder at all. It contains three simple rules `right: auto; left: auto; margin: 0 auto;` and it's only used in one place. It is much easier to remove the placeholder and apply the rules directly, isn't it?

ncoden:
It is only used in one mixin, right. But this mixin can be called by the user wherever he want to create custom components (or modify their behaviors), like all Foundation mixins. So removing the placeholder and applying properties directly would result in duplicated properties for projects taking advantage of the Sass library.

Actually, we should even generalize this and use placeholders in all components for static properties.

ncoden:
Closing in favor of https://github.com/zurb/foundation-sites/pull/11106.

anvucodespot:
Sorry to boot it up. I have same problem with @cietho and I see your solution. But I want to notice to you how I get this error.
When I change the value of $breakpoints[medium] with different default value in _setting.scss file, I will get this error. If I set $breakpoints[medium] = 640px, it will works fine.

ncoden:
@anvucodespot This is a Sass error, for some reasons it doesn't detect that the placeholder and the place it is used are in the same breakpoint (I checked and the two generated media query are exactly the same). So there can be some weird cases where this work, but the best solution is to put the placeholder usage outside of any mediaquery.

Issue #11097, opened 2018-03-27
Foundation is adding role to the dropdown HTML - Google Lighthouse ADA Audit is failing


gkanukuntla:
We are using Foundation for our project and I’m working on an WCAG (accessibility) issue that is generated by the Google Lighthouse Chrome Extension (but this is the same as running an Accessibility audit via Chrome Inspect tool > Audit tab).
The HTML that is pasted below is the html developed by us for mobile navigation menu of our application. We didnot add any ADA roles or attributes added as a part of initial development.
But, looks like Foundation is adding the role=”menu” to the first <ul> element of the menu as per the compiled HTML loading in the DOM and failing by the lighthouse tool and the code snippet below. 

Here are some work arounds I’ve tried in an effort to resolve the ADA issue:
1.	I explicitly added role=”menubar” as per https://www.w3.org/TR/html52/dom.html#allowed-aria-roles-states-and-properties standards.
2.	I explicitly copied just the htmls for dropdown menu and accordion menu from foundation site and ran the lighthouse tool the results show the same error below. 

Google lighthouse error details in the attachment below. 
[google_lighthouse_error.txt](https://github.com/zurb/foundation-sites/files/1849943/google_lighthouse_error.txt)
and link to https://dequeuniversity.com/rules/axe/2.2/aria-allowed-attr?application=lighthouse

<ul class="vertical medium-horizontal menu accordion-menu" data-responsive-menu="accordion medium-dropdown" role="menu" aria-multiselectable="true" data-accordion-menu="7385d8-accordion-menu" data-mutate="2cnxqc-responsive-menu">

Any suggestions to fix this problem, or are there any known issues with Foundation that may be causing this?

**How to reproduce this bug:**
1. Not sure if this can be replicated it is very specific to HTML layout for accordion menu. But can try copying the menu html code snippet for menu we developed in the attachment. 
[dropdown_lighthouse_ADA_Issue_Menu_CodeSnippet.txt](https://github.com/zurb/foundation-sites/files/1849928/dropdown_lighthouse_ADA_Issue_Menu_CodeSnippet.txt)
 into a already setup foundation project. 
2. Add google lighthouse chrome extension to chrome browser
3. Open the html file, in the browser click on lighthouse extension and click on Generate report. 

#### What should happen:
You should get zero errors on lighthouse ADA audit. 
#### What happened instead:
See this error when ran google Audit. 
Google lighthouse error details in the attachment below.
[...]

PR #11099, opened 2018-03-27
fix: apply menubar aria role instead of menu

DanielRuf:
`menu` was used but this is not a valiad aria role. `menubar` is the right aria role.

Closes https://github.com/zurb/foundation-sites/issues/11097

### DIFF
--- a/js/foundation.util.nest.js
+++ b/js/foundation.util.nest.js
@@ -35,7 +35,7 @@ const Nest = {
           .addClass(`submenu ${subMenuClass}`)
           .attr({
             'data-submenu': '',
-            'role': 'menu'
+            'role': 'menubar'
           });
         if(type === 'drilldown') {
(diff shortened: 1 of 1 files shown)


ncoden:
@DanielRuf As said in https://github.com/zurb/foundation-sites/issues/11097, I am not able to reproduce the issue. Do you get `role="menu" applied somewhere ?

DanielRuf:
Well, this line still uses `role: menu` ;-)

ncoden:
I seen that. But for what ?

DanielRuf:
There is no `role: menu` in the WCAG spec. Just `menubar`. This is the only place where we still have `menu`.

DanielRuf:
I guess nested (sub)menues like the code says it.

ncoden (approved):
Ok, so LGTM as we have to do that change anyway but _this does not close #11097_.

PR #11101, opened 2018-03-27
feat: add class for active tab when no hash is set in Tabs

### LINKED ISSUE #11100: Tabs: Browser back doesn't take us back to default tab

1. Setup some tabs
2. Set these options: `data-deep-link="true"` and  `data-update-history="true"`. The 2nd option pushes state onto the browser history. 
3. Click a few random tabs
4. Click the browser back button and watch the URL bar
5. Repeat step 4 multiple times

The hash changes as you click the tabs. When clicking back the state reverses and the previous tab loads accordingly. However, when getting back to the original URL, the tab state doesn't update. 

* Landing page -> example.com/  -> active Tab 1
* Click Tab 2 -> example.com/#tab-2 -> active Tab 2
* Click browser back -> example.com/ -> active Tab 2 (BUG)

Desktop Chrome

6.4.3

#### Test case link:

https://codepen.io/anon/pen/LdeRmW

jamiechong:
Hey Guys,

This fixes #11100.

It adds a new optional option `linkDefaultClass` which defaults to `is-default`. When using deep linking (`data-deep-link`) in combination with history push state (`data-update-history`)we want to take the user back to the default tab if they back-button to a history state that doesn't have a `location.hash`. This PR achieves that. 

To utilize you must add the `is-default` class to your default tab. 

```
<ul class="tabs" data-deep-link="true" data-update-history="true" data-auto-focus="false" data-tabs id="deeplinked-tabs">
                                  ---------- 
  <li class="tabs-title is-active is-default"><a href="#panel1d" aria-selected="true">Tab 1</a></li>
                                  ---------- 
  <li class="tabs-title"><a href="#panel2d">Tab 2</a></li>
  <li class="tabs-title"><a href="#panel3d">Tab 3</a></li>
  <li class="tabs-title"><a href="#panel4d">Tab 4</a></li>
</ul>
```

See the issue, which includes a Codepen that replicates the bug, for more details.

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -96,8 +96,13 @@ class Tabs extends Plugin {
     }
 
      //current context-bound function to open tabs on page load or history hashchange
-    this._checkDeepLink = () => {
+    this._checkDeepLink = (useDefaultTabIfNoHash) => {
       var anchor = window.location.hash;
+
+      if (!anchor.length && useDefaultTabIfNoHash !== false) {
+        anchor = this.$element.find(`.${this.options.linkDefaultClass} a`).attr('href') || '';
+      }
+
       //need a hash and a relevant anchor in this tabset
       if(anchor.length) {
         var anchorNoHash = (anchor.indexOf('#') >= 0 ? anchor.slice(1) : anchor);
@@ -122,7 +127,7 @@ class Tabs extends Plugin {
 
     //use browser to open a tab, if it exists in this tabset
     if (this.options.deepLink) {
-      this._checkDeepLink();
+      this._checkDeepLink(false); // do not useDefaultTabIfNoHash on init
     }
 
     this._events();
@@ -306,7 +311,8 @@ class Tabs extends Plugin {
       .attr({
         'aria-selected': 'false',
         'tabindex': -1
-      });
+      })
+      .blur();  // blur to make sure tab loses focus state on deepLink hashchange
 
     $(`#${$target_anchor.attr('aria-controls')}`)
       .removeClass(`${this.options.panelActiveClass}`)
@@ -483,6 +489,15 @@ Tabs.defaults = {
    * @default 'is-active'
    */
   linkActiveClass: 'is-active',
+
+  /**
+   * Class applied to the default `li` in tab link list.
+   * Used for when deepLink is true
+   * @option
+   * @type {string}
+   * @default 'is-default'
+   */
+  linkDefaultClass: 'is-default',
 
   /**
    * Class applied to the content containers.
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -96,11 +96,12 @@ class Tabs extends Plugin {
     }
 
      //current context-bound function to open tabs on page load or history hashchange
-    this._checkDeepLink = (useDefaultTabIfNoHash) => {
+    this._checkDeepLink = () => {
       var anchor = window.location.hash;
 
-      if (!anchor.length && useDefaultTabIfNoHash !== false) {
-        anchor = this.$element.find(`.${this.options.linkDefaultClass} a`).attr('href') || '';
+      if (!anchor.length && this._defaultAnchor) {
+        // if we don't have an anchor, use the default.
+        anchor = this._defaultAnchor;
       }
 
       //need a hash and a relevant anchor in this tabset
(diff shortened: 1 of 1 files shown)


ncoden (changes requested):
Hi @jamiechong,

Thanks for the PR, that's a great first contribution ! 👍 

I am not sure a new `is-default` class is what we need to resolve the issue:
* when going back in history, I expect the tab the go back to its previous state, not a _default one_.
* If a hash is set, the corresponsing tab is opened. In the other case. In the other case, `.is-active` is left changed and indicate the tab to open by default. So according to its definition, `.is-default` is redundant with `.is-active`.

So I would rather go for a simple bug fix to make open the previous tab when going back in history even without hash.

I'm not sure what's the best way to revert a state with the History API, I'll take a look at it tomorrow.

jamiechong:
The `is-default` class is merely to indicate which tab should be active when there is no `location.hash` set.  When the `hashchange` event fires, `_checkDeepLink` is called, but it needs to know which tab to load up. If there is no hash present, how do we choose? I figure we let the developer explicitly indicate which tab should become active in this case .

ncoden:
I understand the issue. My point is that this is an option for a _very specific_ scenario and there is often in cases like this one a much more simple way to resolve the issue.

Let's evaluate the use cases:
* If you just arrive and no hash is set, the default tab to display is indicated by `is-active`. This is the tab to display when nothing is selected.
* If you go back to history, you expect to see the previous tab selected. If you go back to the original state, you expect to see the original tab selected.

So a solution could be to store the active tab on each `pushState` and retrieve it on each `popstate` when `updateHistory` is enabled.

jamiechong:
I simplified to use the active tab as our default anchor when there is none specified in `location.hash`. Let me know what you think!

jamiechong:
@ncoden I think that is a new bug because closing the modal should not reset the hash to `#`. It's almost as if Foundation needs a separate object to handle the UI state. It should be

* Land on 1st (is-active) tab -> `#`
* Open the 3rd tab -> `#panel3`
* Open the modal -> `#panel3|modal1`
* Close the modal -> `#panel3`
* Go to 2nd Tab -> `#panel2`
* Click back -> `#panel3`
* Click back -> `#panel3|modal1`
* Click back -> `#panel3`
* Click back -> `#`

ncoden:
hmm yeah, you're right 😄. Wrong example.

Let's say a simple link to reset the hash:
```html
<a href="#">Click me</a>
```

What should happen when clicking on it. Reset to tab 1 or stay on tab 3 ?

ncoden:
@jamiechong I like this approach. The state of a page is defined by its URL. Refreshing it should, in most case, lead to the same view.

@DanielRuf I don't understand 😄

DanielRuf:
Use `#tab1` as final state. `#` can be problematic as it is a general and empty anchor (which is also often used for other things like totop buttons).

DanielRuf:
Even if it would be `#` I would automatically replace the history state with `#tab1`.

ncoden:
I disagree with changing the URL on load without any user interaction. If the requested URI is `#`, we could consider it as equal to `#tab1` but we cannot change the url to `#hash1`. I think it's disturbing, almost intrusive, and also bug prone.

jamiechong:
If you want to try it out on your own dev builds (a clean 6.4.3, not this branch). Replace the `_checkDeepLink` function with this: 

```
    this._checkDeepLink = () => {
      var anchor = window.location.hash;
      if(anchor.length) {
        var anchorNoHash = (anchor.indexOf('#') >= 0 ? anchor.slice(1) : anchor);
        var $link = this.$element.find(`[href$="${anchor}"],[data-tabs-target="${anchorNoHash}"]`).first();
        if ($link.length) {
          this.selectTab($(anchor), true);

          //roll up a little to show the titles
          if (this.options.deepLinkSmudge) {
            var offset = this.$element.offset();
            $('html, body').animate({ scrollTop: offset.top }, this.options.deepLinkSmudgeDelay);
          }

          /**
            * Fires when the zplugin has deeplinked at pageload
            * @event Tabs#deeplink
            */
           this.$element.trigger('deeplink.zf.tabs', [$link, $(anchor)]);
        }
      } else if (this.options.updateHistory) {
        anchor = this.$tabTitles.filter(`.${this.options.linkActiveClass}`).find('a').attr('href');
        history.replaceState({}, '', anchor);
      }
    }
```

The `else if` block is responsible for the "intrusive" url change.

DanielRuf:
I would not change the hash but treat is as first tab.

ncoden:
Ok so we're just very bad at understanding each others 😄

jamiechong:
So my PR as it stands now does what you describe @DanielRuf.

ncoden:
Closing in favor of https://github.com/zurb/foundation-sites/pull/11477
@jamiechong I put you as co-author.

PR #11102, opened 2018-03-28
docs: remove box-shadow from variable table cells

DanielRuf:
This removes the wrong box-shadow on cells in the Sass variables table.

Currently it is like this:
bildschirmfoto 2018-03-28 um 12 13 58

See https://foundation.zurb.com/sites/docs/off-canvas.html

### DIFF
--- a/docs/assets/scss/docs.scss
+++ b/docs/assets/scss/docs.scss
@@ -485,3 +485,8 @@ a.docs-banner {
 .docs-nav-container {
   overflow: auto;
 }
+
+.docs-variable-table td.shadow {
+  box-shadow: none;
+}
+


ncoden (changes requested):
Why do we have `.shadow` applied in the first place ?

I think we should rather add a namespace to https://github.com/zurb/foundation-docs/blob/master/templates/partials/sass-reference.hbs#L18

DanielRuf:
Becuase of the `type` annotation in the code.

DanielRuf:
Not sure what you have in your mind but my proposed change would have solved it too ;-)

DanielRuf:
https://github.com/zurb/foundation-sites/blob/3eb3433ea81fde9d6c8c3cd49e3c85e3dca72bc6/scss/components/_off-canvas.scss#L26

ncoden:
Your proposed change hide one of the effect. This is different from removing the source of the unwanted effect.

ncoden:
What I had in mind was using
```html
<td class="zf-docs-type-{{toLower this.type}}">...</td>
```
instead of
```html
<td class="{{toLower this.type}}">...</td>
```

So we can still stylize variables according to their type in the doc getting side-effects from Foundation Prototype utilities. That's why namespaces are important in CSS.

DanielRuf:
Arey the even used there?

ncoden:
No they are not. See https://github.com/zurb/foundation-docs/blob/master/scss/_reference.scss.

PR #11103, opened 2018-03-28
more accurate -zf-zero-breakpoint error message

### LINKED ISSUE #9268: $-zf-zero-breakpoint error message
A zero breakpoint is required as the first key/value of the `$breakpoints` map. But [the error](https://github.com/zurb/foundation-sites/blob/12317b1854283b3a2708d586f1141d523684f8cd/scss/util/_breakpoint.scss#L24) is misleading. 

This:

``` scss
@error 'Your smallest breakpoint (defined in $breakpoints) must be set to "0".';
```

Should probably be:

``` scss
@error 'The first key in the $breakpoints map must have a value of "0".';
```

And a note should also be added to [the docs page](http://foundation.zurb.com/sites/docs/media-queries.html#changing-the-breakpoints).

andycochran:
This PR closes #9268. 

-----

**For version 7:**

- Consider allowing for custom notes within `settings.scss` ([as ncoden notes](https://github.com/zurb/foundation-sites/issues/9268#issuecomment-365112252)). 
- Consider more semantic, mobile-first naming for breakpoints. For example, `small` applies styles to all devices, so should the first key/value in the breakpoints map be `all: 0`?

### DIFF
--- a/scss/util/_breakpoint.scss
+++ b/scss/util/_breakpoint.scss
@@ -25,7 +25,7 @@ $-zf-zero-breakpoint: small !default;
 $-zf-breakpoints-keys: map-to-list($breakpoints, 'keys');
 
 @if nth(map-values($breakpoints), 1) != 0 {
-  @error 'Your smallest breakpoint (defined in $breakpoints) must be set to "0".';
+  @error 'The first key in the $breakpoints map must have a value of "0".';
 }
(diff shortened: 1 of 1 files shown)


ncoden (approved):
👍 Thanks @andycochran. Better late than never ;)

PR #11104, opened 2018-03-28
fix radius on input-group-label and input-group-button

### LINKED ISSUE #8833: Fix input-group-label's border-radius when input-group-label is the last child
Foundation properly applies border-radius to an `input-group-label` as the first child of `input-group-field`, but when `input-group-label` is the last child of `input-group`, no border-radius is applied.

Working properly as `:first-child`:

Working improperly as `:last-child`:

Intended behavior as `:last-child`:

HTML used as `:last-child`:

``` html
<div class="input-group">
    <input type="number" class="input-group-field">
    <span class="input-group-label">%</span>
</div>
```

This issue is caused by [line 41](https://github.com/zurb/foundation-sites/blob/df4d7d76e97bd934622beacbb65efacd600b2e26/scss/forms/_input-group.scss#L41) of _input-group.scss:

``` scss
> :last-child {
    > * {
        border-radius: if($global-text-direction == rtl, $global-radius 0 0 $global-radius, 0 $global-radius $global-radius 0);
    }
}
```

That line is targeting only children of `input-group`'s last child (rather than the last child itself). This solution seems to be implemented to support `input-group-button` as the last child, [as seen in the Forms documentation](http://foundation.zurb.com/sites/docs/forms.html#inline-labels-and-buttons).

A dirty hack that will solve both cases while not causing visual artifacts is as follows:

``` scss
> :last-child {
    border-radius: if($global-text-direction == rtl, $global-radius 0 0 $global-radius, 0 $global-radius $global-radius 0);
    > * {
        border-radius: if($global-text-direction == rtl, $global-radius 0 0 $global-radius, 0 $global-radius $global-radius 0);
    }
}
```

I'm sure there's a more elegant solution to allow `input-group-label` to be used as the last child.

andycochran:
This PR closes #8833 by explicitly setting styles for `.input-group-button` within `.input-group` so that it and `.input-group-label` have the expected border radius.


scss/forms/_input-group.scss
```diff
@@ -34,11 +34,21 @@ $input-prefix-padding: 1rem !default;
     }
 
     > :first-child {
-      border-radius: if($global-text-direction == rtl, 0 $input-radius $input-radius 0, $input-radius 0 0 $input-radius);
+      &:not(.input-group-button){
+        border-radius: if($global-text-direction == rtl, 0 $input-radius $input-radius 0, $input-radius 0 0 $input-radius);
+      }
+
+      &.input-group-button > * {
```
ncoden: Selectors can be merged here are the code in the same. So:
```scss
&:not(.input-group-button), &.input-group-button > * {
   ...
}
```
(don't do that, see the comment above).

Otherwise we're just duplicating code.

scss/forms/_input-group.scss
```diff
@@ -34,11 +34,21 @@ $input-prefix-padding: 1rem !default;
     }
 
     > :first-child {
-      border-radius: if($global-text-direction == rtl, 0 $input-radius $input-radius 0, $input-radius 0 0 $input-radius);
+      &:not(.input-group-button){
```
ncoden: Foundation provide a SASS library, not only CSS components.
* We can't expect Input group button to be implemented as `.input-group-button` (user should be able to create custom components with input group mixins)
* Why only targetting Input group button ?

So I would rather go for:
```scss
&, &  > * { ... }
```

To be honnest, the `> *` is ugly and should be removed. But we'll keep it for backward compatibility. Adding a comment here like "don't do this" could be useful so it does not became an example.

ncoden (changes requested):
Three things to keep in mind when adding selectors:
* Selectors specificity must be the lowest possbile
* Parents should not apply properties automatically on their children
* Duplicated properties

scss/forms/_input-group.scss
andycochran: I've DRY'd this. :)

scss/forms/_input-group.scss
andycochran: Since `.input-group-button` is hardcoded later in the `foundation-form-prepostfix` mixin, can't we expect it to be implemented as `.input-group-button`?

ncoden: Ah f**. I did not realized that Input Group was not _component-oriented_ at all !
Well, its still a problem: Why only targetting Input group button ?

This is generally a bad idea to make our CSS component having particular effects on each-others as the user should be able to use them with custom CSS alongside its own components. Input Group should have the same behavior no matter what is the class of the button we put it. This is the best way to make someone loose 1 hour: "Why when I put my own button there it is not rounded anymore."

andycochran:
@ncoden I say we live with this not being component-oriented, and make sure not to repeat this mistake in v7. Are there any remaining changes you require?

ncoden:
> Why only targetting Input group buttons ?

andycochran:
Because the docs explicitly state to place buttons inside the `input-group-button` wrapper.

andycochran:
We could probably get rid of the `> *` too: 
`&.input-group-button .button `

ncoden:
We should assume as little as possible to HTML structure inside our components. People should be able to freely use the classes of components/elements we provide without limited by self-references and dependencies (like `.button` in a selector that have nothing to do with our Button components).

> Because the docs explicitly state to place buttons inside the input-group-button wrapper.

Got it. So I think we can drop `&:not(.input-group-button)` and use:
```scss
&, &.input-group-button { ... }
```
As `.input-group-button` does not have any visual properties.

ncoden:
@andycochran What do you think ?

ncoden:
@andycochran Hey 👋. Will you have some time work on this ? I'd like your opinon on https://github.com/zurb/foundation-sites/pull/11104#issuecomment-377585825

ncoden:
I applied the requested changes

> We should assume as little as possible to HTML structure inside our components. People should be able to freely use the classes of components/elements we provide without limited by self-references and dependencies.
> 
> This commit simplify the selector used to apply border-radius on Input Group childreens to make all childreens affected, with a single exception for `.input-group-button` where sub-childreens are affected too.

Looks ready for stable v6.5.0

ncoden (approved):

DanielRuf:
Oh, node-sass failing on Node.js 10?

Issue #11082, opened 2018-03-22
[Orbit] Page jumping after each slide transition


logopogo:
#### How to reproduce this bug:

1. View the Orbit docs page (https://foundation.zurb.com/sites/docs/orbit.html) in Chrome on a MAC or on Android (not sure about Windows).
2. Scroll down so the top of the orbit slider demo is partially up out of view (above the site header) 
3. Wait for a slide transition to fire
4. The page will jump back down to reset the slider into full view

#### What should happen:
Page shouldn't jump at all

#### What happened instead:
The pages jumps for some reason

#### Browser(s) and Device(s) tested on:
I've only seen this happen in Chrome (MAC) Version 65.0.3325.162 and a few versions prior and on my Android device.
Have tested on latest Firefox and Safari on MAC and it doesn't seem to occur.

#### Foundation Version(s) you are using:
6.4.3

#### Test case link:
Foundations own docs:
https://foundation.zurb.com/sites/docs/orbit.html

DanielRuf:
It seems the change between the different `display` properties is the cause.

I am currently trying and this is one solution that works:

js/foundation.orbit.js

```
...
if (this.options.useMUI && !this.$element.is(':hidden')) {
        Motion.animateIn(
          //$newSlide.addClass('is-active').css({'position': 'absolute', 'top': 0}),
          $newSlide.addClass('is-active');//.css({'position': 'absolute', 'top': 0}),
          this.options[`animInFrom${dirIn}`],
          function(){
            $newSlide.attr('aria-live', 'polite');
            //$newSlide.attr('aria-live', 'polite');
        });

        Motion.animateOut(
          $curSlide.removeClass('is-active'),
          this.options[`animOutTo${dirOut}`],
          function(){
            $curSlide.removeAttr('aria-live');
            if(_this.options.autoPlay && !_this.timer.isPaused){
              _this.timer.restart();
            }
            //do stuff?
          });
      } else {
...
```

scss/components/orbit.scss
```
@mixin orbit-slide {
  width: 100%;
  position: absolute;

  &.no-motionui {
    &.is-active {
      top: 0;
      left: 0;
    }
  }
}
```

Not sure if this is the best approach. At least the orbit component has a few issues and the intent of the change between the display properties is not quite clear to me.

cc @ncoden @SassNinja @JeremyEnglert or do you see any other solutions and would this be the "correct" solution?

ncoden:
[...]

PR #11107, opened 2018-03-29
fix: always use absolute positioning to prevent jumping Orbit slider

DanielRuf:
Before and after a slide is animted its position attribute is changed between `absolute` and `relative` which can produce a jump of the element after a slide change.

Closes https://github.com/zurb/foundation-sites/issues/11082

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -344,11 +344,10 @@ class Orbit extends Plugin {
 
       if (this.options.useMUI && !this.$element.is(':hidden')) {
         Motion.animateIn(
-          $newSlide.addClass('is-active').css({'position': 'absolute', 'top': 0}),
+          $newSlide.addClass('is-active'),
           this.options[`animInFrom${dirIn}`],
           function(){
-            $newSlide.css({'position': 'relative', 'display': 'block'})
-            .attr('aria-live', 'polite');
+            $newSlide.attr('aria-live', 'polite');
         });
 
         Motion.animateOut(
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -149,8 +149,9 @@ class Orbit extends Plugin {
       temp = this.getBoundingClientRect().height;
       $(this).attr('data-slide', counter);
 
-      if (!/mui/g.test($(this)[0].className) && _this.$slides.filter('.is-active')[0] !== _this.$slides.eq(counter)[0]) {//if not the active slide, set css position and display property
-        $(this).css({'position': 'relative', 'display': 'none'});
+      // hide all slides but the active one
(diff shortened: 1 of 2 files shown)


js/foundation.orbit.js
```diff
@@ -344,11 +344,10 @@ class Orbit extends Plugin {
 
       if (this.options.useMUI && !this.$element.is(':hidden')) {
         Motion.animateIn(
-          $newSlide.addClass('is-active').css({'position': 'absolute', 'top': 0}),
+          $newSlide.addClass('is-active'),
           this.options[`animInFrom${dirIn}`],
           function(){
-            $newSlide.css({'position': 'relative', 'display': 'block'})
-            .attr('aria-live', 'polite');
+            $newSlide.attr('aria-live', 'polite');
```
ncoden: Slides are hidden when `motion-ui` is not used.

https://github.com/DanielRuf/foundation-sites/blob/dbdae96312bcccd0820d8ee43c088d5d6a1fab14/js/foundation.orbit.js#L153

We should keep `'display': 'block'`;

ncoden (changes requested):

js/foundation.orbit.js
DanielRuf: But this causes the jump then.

DanielRuf: Ah, just the block property?

ncoden: Yeah, and remove `'position': 'relative'` from line 153. But make some tests (with and without motionUI / `mui` classes), I'm not sure.

ncoden:
@DanielRuf Tell me when you done your tests or if I should take care of them 😉

DanielRuf:
> or if I should take care of them

Some help is really appreciated and would be great.

ncoden:
> Yeah, and remove 'position': 'relative' from line 153. But make some tests (with and without motionUI / mui classes), I'm not sure.

I found a bug before figuring out that `'position': 'relative'` was not removed. I took care of this.

> **7daaba5 fix: remove position reset of Orbit slides preventing the "full-absolute" new behavior**
>
> All Orbit slides should always have an `absolute` positioning to prevent jumping Orbit slider (see #11107). The `'position': 'relative'` reset is no longer necessary here and break the Orbit behavior as inline styles overrides `.orbit-slide` class styles.

ncoden (approved):
> Yeah, and remove 'position': 'relative' from line 153. But make some tests (with and without motionUI / mui classes), I'm not sure.

I found a bug before figuring out that `'position': 'relative'` was not removed. I took care of this.

> **7daaba5 fix: remove position reset of Orbit slides preventing the "full-absolute" new behavior**
>
> All Orbit slides should always have an `absolute` positioning to prevent jumping Orbit slider (see #11107). The `'position': 'relative'` reset is no longer necessary here and break the Orbit behavior as inline styles overrides `.orbit-slide` class styles.

I made some tests and everything seems to work 👍

PR #10986, opened 2018-02-25
docs: fix broken links #10985

ncoden:
Changes:
* fix broken link to "XY Grid" in "Float Grid" docs
* fix broken link to "Starter Projects" in "Installation" docs

Closes [#10985 - [The Grid] documentation has broken link](https://github.com/zurb/foundation-sites/issues/10985) (@nigelm)
Related to [#10977 - Fix broken link to XY grid](https://github.com/zurb/foundation-sites/pull/10977) (@ocularrhythm)

PR #11113, opened 2018-03-31
Fix link to XY Grid

schuhwerk:
The link to the XY Grid points here:
https://foundation.zurb.com/xy-grid.html (/xy-grid.html)
Point it to the right place:
https://foundation.zurb.com/sites/docs/xy-grid.html (./xy-grid.html)

### DIFF
--- a/docs/pages/grid.md
+++ b/docs/pages/grid.md
@@ -18,7 +18,7 @@ tags:
 ## Importing
 
 <div class="callout alert">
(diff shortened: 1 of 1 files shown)


ncoden:
Already fixed by https://github.com/zurb/foundation-sites/pull/10986/files. We just need to update the server to `master`. Poke @rafibomb

Issue #9952, opened 2017-04-13
[Grid] Uncentering columns breaks their float


andycochran:
How to replicate: 
- At the medium breakpoint, two columns that total > 12 and are both centered
- At the large breakpoint, these columns total ≤ 12 and are both uncentered
- The `.large-uncentered` class does not allow columns to float, as the `.medium-centered` class still applies `float:none;` and `clear:both;` 

https://codepen.io/andycochran/pen/mmyyzj

JeremyEnglert:
Here is a super rough way of fixing this, although, we may be able to make this a bit more generic so it applies to all of these types of instances. I'll give it a shot if we think this is a good path to go down.

```
@include breakpoint(large) {
	.medium-centered {
		&.large-uncentered {
			float: left;
			clear: none;
			&:last-child:not(:first-child) {
				float: right;
				clear: none;
			}
		}
	}
}
```

JeremyEnglert:
This is probably a bit better.

```SASS
.small-centered {
	@include breakpoint(medium) {
		&.medium-uncentered {
			float: left;
			clear: none;
			&:last-child:not(:first-child) {
				float: right;
				clear: none;
			}
		}
	}
	@include breakpoint(large) {
		&.large-uncentered {
			float: left;
			clear: none;
			&:last-child:not(:first-child) {
				float: right;
				clear: none;
			}
		}
	}	
}

.medium-centered {
	@include breakpoint(large) {
		&.large-uncentered {
			float: left;
			clear: none;
			&:last-child:not(:first-child) {
				float: right;
				clear: none;
			}
		}
	}
}
```

kball:
Sigh... feels like we're going down the specificity rabbit hole... @ncoden do you have any better ideas for how to address this?

andycochran:
> specificity rabbit hole

^ Yup. I'm wondering if we should just live with this and be better with specificity in v7.

andycochran:
@ncoden That plan works for me.

ncoden:
Hmmm... `$grid-column-align-edge` is `true` by default. Removing it can be a breaking change.
See https://foundation.zurb.com/sites/docs/grid.html#incomplete-rows

> In order to work around browsers' different rounding behaviors, Foundation will float the last column in a row to the right so the edge aligns. (...)

So this is not only about incomplete grid alignement, but also to "hide" in the gutter some browsers' precision bug.

[...]

PR #11115, opened 2018-03-31
fix: fix uncentered last column on Float Grid #9952

ncoden:
#### Changes:
* Reposition the last column on the left/right (according to `$grid-column-align-edge`) when uncentered.
* Refactor `@mixin grid-column-position` to handle all the float positioning and the last column hack in the same place.
* add visual test for Float Grid uncentered columns

Closes https://github.com/zurb/foundation-sites/issues/9952

### DIFF
--- a/scss/grid/_column.scss
+++ b/scss/grid/_column.scss
@@ -63,12 +63,8 @@
   // Gutters
   @include grid-column-gutter($gutters: $gutters);
 
-  // Last column alignment
-  @if $grid-column-align-edge {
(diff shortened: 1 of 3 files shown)


DanielRuf:
Will test later that day.

DanielRuf (approved):
LGTM 👍 
https://codepen.io/DanielRuf/pen/BrqRQo

ncoden:
Thanks for the review

PR #11116, opened 2018-04-01
fix: use more permissive URL pattern for Abide #9180

### LINKED ISSUE #9180: [Abide] URL validation fails when URL encoded commas are in the URL.
1. Create a URL validated text input
2. Use a URL that has a comma that is url Encoded.  An example would be "https://www.google.com/#q=Et%2Ctu%2Cbrute"
3. Shift focus off of the input.  

%2C is a valid URL encoded character.  I expect abide not to balk at the URL.

**What happened instead?**
Abide flags the input as invalid.

**Test case:**
http://codepen.io/jellisii/pen/amNZJN

This is also a problem with Foundation5 as well.

ncoden:
Use a more permissive URL regex in Abide to allow all escaped characters and various others valid URLs.

### Changes:
* Use [CommonRegexJS](https://github.com/talyssonoc/CommonRegexJS/blob/e2901b9f57222bc14069dc8f0598d5f412555411/lib/commonregex.js#L76) URL regex in Abide
* Add FTPS and SFTP to accepted protocols in Abide url
* Update Abide documentation

**Note:** the new RegExp is close to an "Standard URL validation" and allow a lot of valid URLs that may have to be forbidden on the end website (intrants, local domains, utf8 & unicode characters...).
For more restrictive RegExps, see https://mathiasbynens.be/demo/url-regex.

Closes https://github.com/zurb/foundation-sites/issues/9180

### DIFF
--- a/docs/pages/abide.md
+++ b/docs/pages/abide.md
@@ -321,8 +321,13 @@ cvv : /^([0-9]){3,4}$/,
 
 // http://www.whatwg.org/specs/web-apps/current-work/multipage/states-of-the-type-attribute.html#valid-e-mail-address
 email : /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$/,
+}[a-zA-Z0-9])?)+$/,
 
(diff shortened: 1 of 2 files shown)


docs/pages/abide.md
```diff
@@ -321,8 +321,13 @@ cvv : /^([0-9]){3,4}$/,
 
 // http://www.whatwg.org/specs/web-apps/current-work/multipage/states-of-the-type-attribute.html#valid-e-mail-address
 email : /^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)+$/,
+}[a-zA-Z0-9])?)+$/,
```
DanielRuf: Is this line added by accident?

docs/pages/abide.md
```diff
@@ -321,8 +321,13 @@ cvv : /^([0-9]){3,4}$/,
...
+
+// From CommonRegexJS (@talyssonoc)
+// https://github.com/talyssonoc/CommonRegexJS/blob/e2901b9f57222bc14069dc8f0598d5f412555411/lib/commonregex.js#L76
+// For more restrictive URL Regexs, see https://mathiasbynens.be/demo/url-regex.
+url: /^((?:(https?|ftp|file|ssh):\/\/|www\d{0,3}[.]|[a-z0-9.\-]+[.][a-z]{2,4}\/)(?:[^\s()<>]+|\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\))+(?:\((?:[^\s()<>]+|(?:\([^\s()<>]+\)))*\)|[^\s`!()\[\]{};:\'".,<>?\xab\xbb\u201c\u201d\u2018\u2019]))$/,
```
DanielRuf: What about sftp and ftps? Or should these not be covered here?

DanielRuf:
Rest looks good so far accoding to some regex tests.

docs/pages/abide.md
ncoden: These were not accepted before. but we can add them

DanielRuf (approved):

Issue #10468, opened 2017-07-26
Custom breakpoints + grid-margin-x = incorrect size ordering


wgstjf:
Hi guys,

Not sure if we have discovered a bug with grid-margin-x in 6.4.1 or whether we've got something wrong.  Here's the situation...we have custom breakpoints and custom breakpoint classes as below:

```scss
$breakpoints: (
  small: 0,
  smallx: 376px,
  medium: 668px,
  mediumx: 1024px,
  large: 1210px,
  xlarge: 1701px,
  xxlarge: 1921px,
);
$print-breakpoint: large;
$breakpoint-classes: (small smallx medium mediumx large xlarge);
```

When using grid-margin-x we are finding that the 'small' breakpoint is overriding our custom one 'mediumx'.  Looking at the compiled css it looks like the ordering is messed up, with the 'small' breakpoint stuff appearing (or at least being repeated) after the 'mediumx' styles thus leading to the overriding of what should be the correct sizes etc.  Below is a snippet of the compiled css showing the issue...

```css
  @media print, screen and (min-width: 64em) { /* mediumx */
    .grid-margin-x > .mediumx-auto {
      width: auto; }
    .grid-margin-x > .mediumx-shrink {
      width: auto; }
    .grid-margin-x > .mediumx-1 {
      width: calc(8.33333% - 2.25rem); }
    .grid-margin-x > .mediumx-2 {
      width: calc(16.66667% - 2.25rem); }
    .grid-margin-x > .mediumx-3 {
      width: calc(25% - 2.25rem); }
    .grid-margin-x > .mediumx-4 {
      width: calc(33.33333% - 2.25rem); }
    .grid-margin-x > .mediumx-5 {
      width: calc(41.66667% - 2.25rem); }
    .grid-margin-x > .mediumx-6 {
      width: calc(50% - 2.25rem); }
    .grid-margin-x > .mediumx-7 {
      width: calc(58.33333% - 2.25rem); }
    .grid-margin-x > .mediumx-8 {
      width: calc(66.66667% - 2.25rem); }
    .grid-margin-x > .mediumx-9 {
      width: calc(75% - 2.25rem); }
    .grid-margin-x > .mediumx-10 {
      width: calc(83.33333% - 2.25rem); }
    .grid-margin-x > .mediumx-11 {
      width: calc(91.66667% - 2.25rem); }
    .grid-margin-x > .mediumx-12 {
      width: calc(100% - 2.25rem); } }
[...]

PR #11117, opened 2018-04-01
fix: fix cell widths overridden by readjusted smaller breakpoints in XY Grid #10468

ncoden:
### Context

With margin gutters, a cell width is defined with its gutter. This is why we have to readjust all sizes and not only `.cell` when applying the new gutter in a bigger breakpoint. This must include all sizes for smaller breakpoints too. Before this commit there was an optimization that prevent gutters for smaller breakpoints without a new gutter defined to be regenerated. This can seems logical since no gutter was generated for this breakpoint so there is nothing to readjust.

### Bug 

Because the gutter is defined _with the width/height_, the width/height of readjusted gutters (with a custom gutter) overrides the width/height of non-readjusted  gutters (without a custom gutter) even if they are from a bigger breakpoint.

### Changes:
* Readjust gutters for all breakpoints even without new gutter defined
* Add explainaitions

There is no impact on the bundle `foundation.css` size because there is no breakpoint (outside of the last one) defined without a custom gutter. There can be a minimal impact on a user CSS size if ze uses a lot of custom breakpoints.

**Note**: there a reason why grid system always used paddings for grid system: it does not impact the width. All troubles (and a lot of code duplication and specificity issues) we have with the XY Grid comes from that. And apart from redicing by 1 the number of nested HTML tags in a complex grid layout, I don't see a single advantage of margin gutters. @brettsmason @kball Do you have an opinion on this ?

Closes https://github.com/zurb/foundation-sites/issues/10468

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -146,12 +146,12 @@
     @include -zf-each-breakpoint() {
 
       // This is purely for responsive gutters - the margin grid has to go back and adjust widths (or heights)
-      // for prior breakpoints based on the responsive gutter.
+      // for all prior breakpoints.
+      // As their gutter is defined with their width/height, even breakpoint without a new margin must be
+      // generated to not having their width/height overrided by re-adjusted smaller breakpoints.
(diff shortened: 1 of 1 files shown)


DanielRuf (approved):
LGTM 👍

Issue #10391, opened 2017-07-11
Accessibility of dropdown toggling in Foundation 5 vs 6


daneah:
Accessibility of toggling dropdowns using keyboard navigation seems to have changed from Foundation 5 to Foundation 6. It was suggested that we direct accessibility issues to the attention of @Owlbertz. Thanks for your time and help!

#### How to reproduce this bug:

Construct a dropdown per Foundation 6's [Dropdown doc](http://foundation.zurb.com/sites/docs/dropdown.html) (explicitly shown below).

#### What should happen:

Using keyboard navigation, TABing to the trigger element and hitting Enter opens the dropdown content. TABing back to the trigger element while the content is open and hitting Enter again should close the content.

#### What happened instead:

Attempting to close already-opened dropdown content by TABing to the trigger element and hitting Enter sends focus to the open content, which _remains_ open.

#### Browser(s) and Device(s) tested on:

macOS + latest Chrome

#### Foundation Version(s) you are using:

Foundation 6.4.1

#### Test case link:

CodePen seems to be having issues saving and forking at the moment. The simplest test case with Foundation JS + CSS presumed is:

```html
<a href="#" data-toggle="my-dropdown">Open Dropdown</a>

<div id="my-dropdown" class="dropdown-pane" data-dropdown>
    Hello
</div>
```

Owlbertz:
Hi @daneah, thank you for reporting this.
However, I am not sure if this is by design or a bug, as `ESC` can be used to close the Dropdown.

When checking the WAI-ARIA authoring practices, imo "Alert" seems to be the equivalent for Foundation's Dropdowns, which also do not have the behavior you expect (see [this example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/alert/index.html)).
Other resources work completely different. For example [this page](http://ianmcburnie.github.io/mindpatterns/disclosure/flyout/) auto-closes the panels automatically once they lost focus.

So while it seems to be not 100% standardized, I agree that Foundation's Dropdowns should close on a second `ENTER`-click as this would follow the mouse interactions.

daneah:
[...]

PR #11123, opened 2018-04-03
fix: toggle dropdown if there is no focusable element inside

DanielRuf:
Closes https://github.com/zurb/foundation-sites/issues/10391

### DIFF
--- a/js/foundation.dropdown.js
+++ b/js/foundation.dropdown.js
@@ -35,8 +35,8 @@ class Dropdown extends Positionable {
     this._init();
 
     Keyboard.register('Dropdown', {
-      'ENTER': 'open',
-      'SPACE': 'open',
+      'ENTER': 'toggle',
+      'SPACE': 'toggle',
       'ESCAPE': 'close'
     });
   }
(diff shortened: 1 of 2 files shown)


ncoden (approved):
LGTM 👍. I seen that when using the dropdown and it annoyed me as well.

It's good to go like this but it could be nice to have a `focus` option replacing `auto-focus` with:
* `true`: always focus the panel
* `"auto"`: (default) focus the panel if it contains any focusable element
* `"first"`: always focus the first focusable element
* `false`: never focus the panel

DanielRuf:
> It's good to go like this but it could be nice to have a focus option replacing auto-focus with:

But this would be a new feature or not?

ncoden:
> But this would be a new feature or not?

Yes this would be a new feature. So not in this PR. We still want to make "don't focus dropdown where ther's nothing in it" the default behavior for accessibility reasons.

DanielRuf:
> It's good to go like this but it could be nice to have a focus option replacing auto-focus with:
> 
> true: always focus the panel
> "auto": (default) focus the panel if it contains any focusable element
> "first": always focus the first focusable element
> false: never focus the panel

How can we track this (as issue) at best so we do not forget it then?

Issue #11124, opened 2018-04-04
in-canvas-for-x width:auto instead of width:inherit


RvWensen:
#### How to reproduce this bug:

<div class="off-canvas in-canvas-for-large position-left" data-off-canvas> inside of a cell with a width definition class for example "large-4"

so:

```
<div class="cell large-4">
   <div class="off-canvas in-canvas-for-large position-left" data-off-canvas>
```

#### What should happen:

The off-canvas should be full width if you ask me.
Instead it becomes also a large-4 so 33% of the width of the div it is in.
That because the widht of the class in-canvas-for-large is width:inherit.

So if i want to create what i think in most cases is desired behavior i need to do the following:

```
<div class="cell large-4">
    <div style="width:auto;">
       <div class="off-canvas in-canvas-for-large position-left" data-off-canvas>
```

#### idea:

Change width:inherit to width:auto for the class in-canvas-for-x

DanielRuf:
Using `grid-x` container on the off-canvas does not help here?

RvWensen:
@DanielRuf 

it won't work if you make like:

`
<div class="grid-x">
   <div class="cell large-4">
      <div class="off-canvas in-canvas-for-large position-left" data-off-canvas>
`
Because the in-canvas-for-large will still take over the size of the above div which is large-4 so 33%.

it will work if you make it like this:

`
<div class="grid-x">
   <div class="cell large-4">
      <div class="grid-x">
         <div class="cell">
            <div class="off-canvas in-canvas-for-large position-left" data-off-canvas>
`
But then you make even more extra div's then my above example with:

`<div style="width:auto;"> `

DanielRuf:
Why not add the needed class(es) from the extra divs to the offcanvas div?

RvWensen:
@DanielRuf 

Do you mean that i write a class and overwrite the default value of width:inherit to width:auto?
Sure i can do that, but i prefer to use default functionality of Foundation.
And there is no way to edit in the _settings.scss.

Is there a reason to use width:inherit? If so could you explain me why?

DanielRuf:
It is basically a reset and the default behavior and default useragent styles by browser vendors, see https://github.com/zurb/foundation-sites/blob/develop/scss/components/_off-canvas.scss#L414

In my opinion this is highly project related and not a general feature that is needed. What do the others think? @SassNinja @ncoden

RvWensen:
@DanielRuf 

[...]

PR #11141, opened 2018-04-06
in-canvas-for-x width:auto instead of width:inherit

RvWensen:
Issue fix.
Discussion can be found https://github.com/zurb/foundation-sites/issues/11124

### DIFF
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -416,10 +416,10 @@ $breakpoint: small
   visibility: visible;
   height: auto;
   position: static;
-  background: inherit;
+  background: none;
   width: auto;
-  overflow: inherit;
-  transition: inherit;
+  overflow: visible;
+  transition: none;
 
   // Increase CSS specificity
   &.position-left,
--- a/scss/components/_off-canvas.scss
+++ b/scss/components/_off-canvas.scss
@@ -417,7 +417,7 @@ $breakpoint: small
   height: auto;
   position: static;
   background: inherit;
-  width: inherit;
+  width: auto;
   overflow: inherit;
   transition: inherit;
 


ncoden:
Hi @RvWensen,

Here are some recommentations your future PRs:
* Use the same "format" as others PRs for your PR and commits titles: `feat:`, `fix:`, `docs:`, `refactor:`, `clean:`, `chore:`...
* Provide a list of the changes mades with their reasons (even if you end in repeating yourself with what was said in the related issue).
* If your PR closes an issue, said it explicitely (like "Closes #XXX" or "Fixes #XXX")

RvWensen:
Thanks for the feedback!

scss/components/_off-canvas.scss
```diff
@@ -417,7 +417,7 @@ $breakpoint: small
   height: auto;
   position: static;
   background: inherit;
-  width: inherit;
+  width: auto;
```
ncoden: This makes sense for `width: auto`, but @SassNinja was there any reasons for going for `inherit` in the first place for properties that does not inherit their value by default ?

ncoden: poke @SassNinja

ncoden: Done some research. The incanvas feature always used `inherit`, without explicit reasons given. Most font properties should inherit but not `background`, `width`, `overflow`, `transition` as this is not their default value.

SassNinja: @ncoden the in-canvas is supposed to just revert the off-canvas styles. Seems I was overusing the `inherit` a bit ;)
You're replacements of the properties in talk look good to me 👍

SassNinja (approved):

SassNinja:
lgtm 👍 

The `inherit` was simply overused. Looks better now!

Issue #11121, opened 2018-04-03
Block Grid not aligning correctly with grid-margin-x in 6.4.3


lrobson555:
#### How to reproduce this bug:

1. Create a block grid that only uses small-up-3, no medium or large classes, with grid-margin-x

#### What should happen:

Three columns should be displayed in a row with margins between.

#### What happened instead:

Only two columns are displayed in a row.
Looks like when calculating the width of each column using the percentage minus the margin, it's only taking into account a 20px margin instead of 30px.

#### Browser(s) and Device(s) tested on:

Chrome v 65.0.3325.181, macOS Sierra 10.12.6

#### Foundation Version(s) you are using:

Broken on 6.4.3, seems to work fine in 6.4.0.

#### Test case link:

Broken in first, works in second:
- [CodePen with Foundation 6.4.3 and XY Grid](https://codepen.io/lrobson555/pen/qoKeXJ)
- [CodePen with Foundation 6.4.0 and XY Grid](https://codepen.io/lrobson555/pen/dmKxZa)

DanielRuf:
```
$ git bisect bad
8a0ae1958a58a084e6fe0049482df483ce7a8b3c is the first bad commit
commit 8a0ae1958a58a084e6fe0049482df483ce7a8b3c
Author: Kevin Ball <kmball11@gmail.com>
Date:   Fri Jul 21 08:37:21 2017 -0700

    Merge pull request #10435 from zurb/xy-grid-layout-fix

    Add missing arguement for xy-grid-layout mixin

:040000 040000 7a37fc20973f7df6a5e3f83c089016c355353544 26fb8348cbcedb0ef5d783bb3bc5913fc2eef8d0 M      scss

Daniel@Daniel-MSI MINGW64 ~/GITHUB/foundation-sites ((8a0ae1958...)|BISECTING)
$ git show 8a0ae1958a58a084e6fe0049482df483ce7a8b3c
commit 8a0ae1958a58a084e6fe0049482df483ce7a8b3c (HEAD, refs/bisect/bad)
Author: Kevin Ball <kmball11@gmail.com>
Date:   Fri Jul 21 08:37:21 2017 -0700

    Merge pull request #10435 from zurb/xy-grid-layout-fix

    Add missing arguement for xy-grid-layout mixin

diff --git a/scss/xy-grid/_layout.scss b/scss/xy-grid/_layout.scss
index bd8c14284..3b7650e32 100644
--- a/scss/xy-grid/_layout.scss
[...]

PR #11145, opened 2018-04-08
fix: use the right breakpoints in block-grid for gutter calculation

DanielRuf:
https://github.com/zurb/foundation-sites/commit/8a0ae1958a58a084e6fe0049482df483ce7a8b3c produced a regression in the block-grids as this changed the mixin usage and order of arguments.

Also the margin grid part did not use the right breakpoint and so the calculation of the widths was wrong after the mentioned commit as now the breakpoint argument was passed through.

Closes https://github.com/zurb/foundation-sites/issues/11121

### DIFF
--- a/scss/xy-grid/_classes.scss
+++ b/scss/xy-grid/_classes.scss
@@ -188,7 +188,7 @@
     @include -zf-each-breakpoint {
       @for $i from 1 through $xy-block-grid-max {
         .#{$-zf-size}-up-#{$i} {
-          @include xy-grid-layout($i, '.cell', false, $gutter-type: padding);
+          @include xy-grid-layout($n: $i, $selector: '.cell', $gutter-output: false, $gutter-type: padding, $breakpoint: $-zf-size);
         }
       }
     }
@@ -203,7 +203,7 @@
           @each $bp in -zf-breakpoints-less-than($-zf-size) {
             @if(map-has-key($grid-margin-gutters, $bp)) {
               .grid-margin-x.#{$bp}-up-#{$i} {
-                @include xy-grid-layout($i, '.cell', false, $gutter-type: margin, $breakpoint: $bp);
+                @include xy-grid-layout($n: $i, $selector: '.cell', $gutter-output: false, $gutter-type: margin, $breakpoint: $-zf-size);
               }
             }
           }
@@ -211,7 +211,7 @@
       }
       @for $i from 1 through $xy-block-grid-max {
         .grid-margin-x.#{$-zf-size}-up-#{$i} {
-          @include xy-grid-layout($i, '.cell', false, $gutter-type: margin, $breakpoint: $-zf-size);
+          @include xy-grid-layout($n: $i, $selector: '.cell', $gutter-output: false, $gutter-type: margin, $breakpoint: $-zf-size);
         }
       }
     }
--- a/scss/xy-grid/_layout.scss
+++ b/scss/xy-grid/_layout.scss
@@ -14,6 +14,7 @@
 /// @param {Number|Map} $gutters [$grid-margin-gutters] - Map or single value for gutters.
 /// @param {Keyword} $gutter-type [margin] - Type of gutter to output. Accepts `margin` or `padding`.
 /// @param {List} $gutter-position [right left] - The position to apply gutters to. Accepts `top`, `bottom`, `left`, `right` in any combination.
+/// @param {String} $breakpoint [null] - The breakpoint to use for the cell generation.
 /// @param {Boolean} $vertical [false] - Set to true to output vertical (height) styles rather than widths.
 @mixin xy-grid-layout(
   $n,
@@ -22,7 +23,7 @@
   $gutters: $grid-margin-gutters,
   $gutter-type: margin,
   $gutter-position: right left,
-  $breakpoint: $-zf-zero-breakpoint,
+  $breakpoint: null,
   $vertical: false
 ) {
   $size: percentage(1/$n);
--- a/scss/xy-grid/_layout.scss
+++ b/scss/xy-grid/_layout.scss
@@ -28,6 +28,6 @@
   $size: percentage(1/$n);
 
   & > #{$selector} {
-    @include xy-cell($size, $gutter-output, $gutters, $gutter-type, $gutter-position, $breakpoint, $vertical);
+    @include xy-cell($size, $gutter-output: $gutter-output, $gutters: $gutters, $gutter-type: $gutter-type, $gutter-position: $gutter-position, $breakpoint: $breakpoint, $vertical: $vertical);
   }
 }
--- a/scss/xy-grid/_layout.scss
+++ b/scss/xy-grid/_layout.scss
@@ -28,6 +28,6 @@
   $size: percentage(1/$n);
 
   & > #{$selector} {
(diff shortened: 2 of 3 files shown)


scss/xy-grid/_layout.scss
```diff
@@ -28,6 +28,6 @@
   $size: percentage(1/$n);
 
   & > #{$selector} {
-    @include xy-cell($size, $gutter-output, $gutters, $gutter-type, $gutter-position, $breakpoint, $vertical);
+    @include xy-cell($size, $gutter-output: $gutter-output, $gutters: $gutters, $gutter-type: $gutter-type, $gutter-position: $gutter-position, $breakpoint: $breakpoint, $vertical: $vertical);
```
ncoden: I don't understand. The `xy-cell` mixin now is:
```scss
@mixin xy-cell(
  $size: full,
  $gutter-output: true,
  $gutters: $grid-margin-gutters,
  $gutter-type: margin,
  $gutter-position: right left,
  $breakpoint: null,
  $vertical: false
) {
```
Why using named parameters here ? Isn't the order right ?

scss/xy-grid/_classes.scss
```diff
@@ -188,7 +188,7 @@
     @include -zf-each-breakpoint {
       @for $i from 1 through $xy-block-grid-max {
         .#{$-zf-size}-up-#{$i} {
-          @include xy-grid-layout($i, '.cell', false, $gutter-type: padding);
+          @include xy-grid-layout($n: $i, $selector: '.cell', $gutter-output: false, $gutter-type: padding, $breakpoint: $-zf-size);
```
ncoden: Same. `xy-grid-layout` takes as arguments:
```scss
$n,
$selector: '.cell',
$gutter-output: true,
```

`$i, '.cell', false` seems to be the right order.

scss/xy-grid/_layout.scss
ncoden: Documentation for `xy-grid-layout` is out-of-date. `$breakpoint` description is missing.

scss/xy-grid/_classes.scss
ncoden: A better solution for `$breakpoint: $-zf-size` would be to make `$breakpoint` using `null` by default in `@mixin xy-grid-layout`, which mean "I will use `$-zf-size` if it exists, `$-zf-zero-breakpoint` otherwhise" as defined in `@mixin xy-cell`.

ncoden (changes requested):
I may be wrong as I don't undestand the bug, but I think there is changes in this PR with no impact and unrelated to the issuse it try to solve.

scss/xy-grid/_layout.scss
DanielRuf: The named params were set by kball, I will choose the short form here.

scss/xy-grid/_classes.scss
DanielRuf: No, xy-grid-layout was changed and affected by https://github.com/zurb/foundation-sites/commit/8a0ae1958a58a084e6fe0049482df483ce7a8b3c cc @kball

DanielRuf: Named params should be more clear here.

scss/xy-grid/_layout.scss
ncoden: No I mean: 
```diff
-  @include xy-cell($size, $gutter-output, $gutters, $gutter-type, $gutter-position, $breakpoint, $vertical); 
```
is the same as
```diff
+  @include xy-cell($size, $gutter-output: $gutter-output, $gutters: $gutters, $gutter-type: $gutter-type, $gutter-position: $gutter-position, $breakpoint: $breakpoint, $vertical: $vertical); 
```

Why changing ?

ncoden: > Named params should be more clear here.

Ah, same for here ?

DanielRuf: `$gutters` and `$breakpoint` were added so the order is not right anymory

ncoden: I don't get it.

Take a look at the call:
```diff
-  @include xy-cell($size, $gutter-output, $gutters, $gutter-type, $gutter-position, $breakpoint, $vertical); 
```
...and the mixin: 
```scss
@mixin xy-cell(
  $size: full,
  $gutter-output: true,
  $gutters: $grid-margin-gutters,
  $gutter-type: margin,
  $gutter-position: right left,
  $breakpoint: null,
  $vertical: false
) {
```

The order is the same.

DanielRuf: `xy-grid-layout($n: $i, $selector: '.cell', $gutter-output: false, $gutter-type: padding, $breakpoint: $-zf-size);`

It is about `xy-grid-layout` not the cell.

DanielRuf: See the current changes.

DanielRuf: And https://github.com/zurb/foundation-sites/pull/11145/files#diff-31c7029afafde5d1c8f61e2705198579R206 was the real bug.

DanielRuf: I can revert this single line but so far the line was not consistent. The bug was in another line (the one with the breakpoint).

ncoden: Yeah I seen that later.

DanielRuf: Also the order is not correct. Unnamed arguments could be problematic as there is `gutters` in between. https://github.com/zurb/foundation-sites/blob/develop/scss/xy-grid/_layout.scss#L18

ncoden: > as there is gutters in between

You are passing `$gutters`. You pass all parameters with the same name in the same order to `xy-cell`.

DanielRuf: You are still referring to the wrong code which is not in the changeset anymore afaik ;-) The change there was reverted.

AI am speaking about https://github.com/zurb/foundation-sites/pull/11145/files#diff-31c7029afafde5d1c8f61e2705198579R191

DanielRuf: Please check the current changeset here https://github.com/zurb/foundation-sites/pull/11145/files ;-)

I think there is some miscommunication.

ncoden: > Please check the current changeset here https://github.com/zurb/foundation-sites/pull/11145/files ;-)

Yes I seen that. But you seemed to still argue that these are not equivalent.

scss/xy-grid/_classes.scss
ncoden: So we keep `$breakpoint: $-zf-size` because "Named params should be more clear here" ?

Please at least change `$breakpoint: $-zf-zero-breakpoint` to `$breakpoint: null` in `xy-grid-layout` and add the missing argument documentation.

scss/xy-grid/_layout.scss
DanielRuf: Then it was a fault on my side, sorry.

scss/xy-grid/_classes.scss
DanielRuf: Done. Feedback is welcome.

Would be great to get some more feedback from the others.

ncoden (approved):
LGTM. Thanks for the changes.

PR #11168, opened 2018-04-12
Fix documentation for OffCanvas#closed event

NeonArray:
The documentation incorrectly stated that the event triggers when opened instead of when it closes.


DanielRuf:
Please use the master branch for docs.

ocularrhythm:
Right on

DanielRuf:
It seems you have to rebase locally and force-push to the remote.

ncoden:
@DanielRuf FYI we will get rid of the `master`/`develop` split by always pushing changes on `develop` and using `master` for stable releases only, like a standard GitFlow model.

DanielRuf:
> @DanielRuf FYI we will get rid of the master/develop split by always pushing changes on develop and using master for stable releases only, like a standard GitFlow model.

Sounds good 👍

PR #11171, opened 2018-04-12
Incorrect variable name in legacy Float Grid docs

englishtom:
In the docs it states that $global-flex needs to be set to false to enable the float grid, however the variable is $global-flexbox


DanielRuf (approved):

ncoden (approved):
Thanks @englishtom

PR #11174, opened 2018-04-13
build: use npm ci

DanielRuf:
`npm ci` can be faster and is meant for CI. This PR is meant for evaluating the new ci command and testig the performance and compare the logs of Travis CI.

https://docs.npmjs.com/cli/ci
http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,8 @@ matrix:
   - node_js: "stable"
     env: ZF_TRAVIS_COMMAND=test:javascript:browserstack
 
-install: npm install
+before_install:  npm install -g npm@latest
+install: npm ci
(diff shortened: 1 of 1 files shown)


ncoden:
Performances:
* `npm install`: 113s
  screen shot 2018-04-14 at 19 40 30
* `npm ci`: 58s (95% faster)
  screen shot 2018-04-14 at 19 40 10

ncoden (approved):
LGTM 👍 

`npm@lastest` is required there to enforce older node versions to use the NPM latest version. It should always be compatible as NPM supports all node LTS versions.

And for `npm ci` itself, well it's better, faster, stronger ;)

PR #11177, opened 2018-04-15
tests: update the dependencies for browsers

DanielRuf:
This updates the dependencies for the unit tests in browser environments.


ncoden:
@DanielRuf I think we should remove these dependencies from the repo on rely on NPM instead.

ncoden (changes requested):
@DanielRuf I think we should remove these dependencies from the repo and rely on NPM instead in an other PR (and resolve issues).

DanielRuf:
> I think we should remove these dependencies from the repo

We can't as these are the tests which are running in the "real" browsers (see package.json).
But the question is how we can test it better as I do not have access to this Browserstack instance and just have one at work.

DanielRuf:
The very long html error might be due to unfocus/focus issues in IE. The others are test related except a few.

DanielRuf:
I think I just misinterpreted it due to the wording. Instead of removing the dependencies we want to switch to the versions in node_modules and remove the obsolete selfhosted files (as long as we have them there).

Will work on this later.

DanielRuf:
```
Launching 18 worker(s) for 18 run(s).


[Windows 8.1, Internet Explorer 11.0] Error: "calls reflow on viewport size change once" failed, expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:145)

[Windows 8.1, Internet Explorer 11.0] Failed: 260 tests, 251 passed, 1 failed, 8 skipped; ran for 0ms


[Windows 8.1, Firefox 60.0 beta] Error: "calls reflow on viewport size change once" failed, Error: expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:142)

[Windows 8.1, Firefox 60.0 beta] Failed: 260 tests, 251 passed, 1 failed, 8 skipped; ran for 0ms


[ios 7.0, Mobile Safari null] Error: "calls reflow on viewport size change once" failed, AssertError: expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:145)

[ios 7.0, Mobile Safari null] Failed: 260 tests, 251 passed, 1 failed, 8 skipped; ran for 0ms


[Windows 7, Internet Explorer 10.0] Error: "calls reflow on viewport size change once" failed, expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:145)

[Windows 7, Internet Explorer 10.0] Failed: 260 tests, 251 passed, 1 failed, 8 skipped; ran for 0ms


[Windows 10, Edge 15.0] Error: "calls reflow on viewport size change once" failed, expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:145)
  Source: Error: expected _reflow to be called once but was called 0 times (http://localhost:8888/test/javascript/lib/sinon.js:145)

[Windows 10, Edge 15.0] Failed: 260 tests, 251 passed, 1 failed, 8 skipped; ran for 0ms
```

Some issue with `_reflow` is left.

ncoden:
> 6f59b81 tests: update the dependencies for browsers 

Wrong approach (use NPM instead). Please open a dedicated PR.

> 8e55fd3 tests: some browsers focus the body element when an element is unfocused

Not a nice commit name ;). Also conditional tests are often a bad idea as they hide what doesn't work the same way across browsers instead of forcing us to fix it. Please open a dedicated PR.

DanielRuf:
> Wrong approach (use NPM instead). Please open a dedicated PR.

I will do then, I'm just fixing the tests for now ;-) One after the other.

DanielRuf:
> Not a nice commit name ;). Also conditional tests are often a bad idea as they hide what doesn't work the same way across browsers instead of forcing us to fix it. Please open a dedicated PR.

What would you recommend then? I'm using this PR for testing the single commits and will then cherry-pick them to separate PRs after this.

DanielRuf:
Please keep it open for now so I can run the tests on the Browserstack instance.

I will take care of closing it afterwards.

DanielRuf:
PS: I'm at work so I can not yet test with the ZURB Browserstack instance for now. Also I saw no issues in our Browserstack instance so I guess there is just a race condition / performance issue.

DanielRuf:
What I am wondering is: why does the test keep running for so long even after the last Browserstack test was already done? @ncoden

DanielRuf:
@ncoden is the following just because of the reopened PR or what could be the cause?

```
Error from BrowserStack:  Error: Validation Failed - `device` invalid
    at IncomingMessage.<anonymous> (/home/travis/build/zurb/foundation-sites/node_modules/browserstack/lib/client.js:73:9)
    at IncomingMessage.emit (events.js:185:15)
    at endReadableNT (_stream_readable.js:1106:12)
    at process._tickCallback (internal/process/next_tick.js:178:19)
```

DanielRuf:
> Conditional tests are often a bad idea as they hide what doesn't work the same way across browsers instead of forcing us to fix it.

How would you "fix" this obvious browser bug / behavior which is not consistent across all browsers, well known and can not be fixed imho?

ncoden:
> How would you "fix" this obvious browser bug / behavior which is not consistent across all browsers, well known and can not be fixed imho?

Tests for the specific browser and versions affected by the bug, find a reference, and embed the test in a `try { ... } catch { ... }` with an error being throw for all non-affected browsers (according to the user agent).

DanielRuf:
> Tests for the specific browser and versions affected by the bug, find a reference, and embed the test in

There is no good way to detect IE10 and other browsers (and we should not do any useragent sniffing here imho).

DanielRuf:
Not yet sure how this can be done in the best possible way.

ncoden:
The goal is to still test the feature _and make it fail if our code is invalid_. Tests should fail. There is no point with having tests that bypass errors.

Then, for browsers with a specific bug
* Or we check for this specific bug, but this is should be distinguishable from a bug on our side. So the behavior we test should be an isolated scenario, split from our code.
* Or we check for the affected browsers. This is not the best solution (it goes against the "walk like a duck, qwack like a duck, this is a duck" principle), but it may be the only one. See http://browserhacks.com/

DanielRuf:
Hm, I still have no idea how this should look like in the code. Could need some support here.

ncoden:
> Hm, I still have no idea how this should look like in the code. Could need some support here.

Let's start by isolating the focus bug:

> some browsers focus the body element when an element is unfocused 

Which browsers. Any references ? How are we sure this is a browser bug ? What does "unfocused" mean (we only focus _others_ elements in the code) ?

DanielRuf:
> Which browsers. Any references ? How are we sure this is a browser bug ? What does "unfocused" mean (we only focus others elements in the code) ?

Take a look at the unit tests of the PR which updated the dependencies. It it the issue with the long message where IE, FF 60 and so on focus the body element, not the last focused element.

DanielRuf:
Regarding the focus issue: https://travis-ci.org/zurb/foundation-sites/jobs/367058960#L819

ncoden:
So it's not a browser bug but an actual bug ?

DanielRuf:
> So it's not a browser bug but an actual bug ?

I still think this is a difference in the browser implementations.

ncoden:
@DanielRuf Can we close this PR ?

PR #11179, opened 2018-04-16
tests: prevent race condition in _reflow test

DanielRuf:
This PR increases the timeouts in the test for `_reflow` and prevents some race conditions by stacking the functions.

Reference: https://github.com/zurb/foundation-sites/pull/11177

### DIFF
--- a/test/javascript/components/interchange.js
+++ b/test/javascript/components/interchange.js
@@ -170,12 +170,10 @@ describe('Interchange', function() {
     it('calls reflow on viewport size change once', function(done) {
       $html = $(generateTemplate('image')).appendTo('body');
       plugin = new Foundation.Interchange($html, {});
-      Foundation.IHearYou();
-      let spy = sinon.spy(plugin, '_reflow');
-
       setTimeout(function() {
-        $(window).trigger('resize');
+        Foundation.IHearYou();
       }, 1);
+      let spy = sinon.spy(plugin, '_reflow');
 
       setTimeout(function() {
         $(window).trigger('resize');
@@ -184,6 +182,10 @@ describe('Interchange', function() {
       setTimeout(function() {
         $(window).trigger('resize');
       }, 10);
+
+      setTimeout(function() {
+        $(window).trigger('resize');
+      }, 20);
 
       setTimeout(function() { // Wait for third trigger...
         sinon.assert.calledOnce(spy);


test/javascript/components/interchange.js
```diff
@@ -170,12 +170,10 @@ describe('Interchange', function() {
     it('calls reflow on viewport size change once', function(done) {
       $html = $(generateTemplate('image')).appendTo('body');
       plugin = new Foundation.Interchange($html, {});
-      Foundation.IHearYou();
-      let spy = sinon.spy(plugin, '_reflow');
-
       setTimeout(function() {
-        $(window).trigger('resize');
+        Foundation.IHearYou();
```
ncoden: Why should `Foundation.IHearYou` be called dynamically ?

test/javascript/components/interchange.js
```diff
@@ -185,10 +183,14 @@ describe('Interchange', function() {
...
 
+      setTimeout(function() {
+        $(window).trigger('resize');
+      }, 20);
+
       setTimeout(function() { // Wait for third trigger...
         sinon.assert.calledOnce(spy);
         done();
-      }, 50);
+      }, 150);
```
ncoden: Why `150` instead of `50` ? There should be no race condition there.

test/javascript/components/interchange.js
DanielRuf: Because it was already there and is also done like this in the other unit tests, see https://github.com/zurb/foundation-sites/pull/11179/files#diff-e0276fdab014830c6cb0753691ddc7f1L173

test/javascript/components/interchange.js
DanielRuf: Well, it seems the tests run slower on Travis and we might set this lower later again but so far this + the other changes fixed it in my other PR which includes all changes.

test/javascript/components/interchange.js
ncoden: I meant _dynamically ?_

DanielRuf: Because this seems to solve the race condition with the other changes (see the other PR which has all changes).

test/javascript/components/interchange.js
ncoden: I don't like having magic numbers like this one. There is no reasons for `trigger('resize')` to not trigger `_reflow` immediately.

test/javascript/components/interchange.js
DanielRuf: The tests successfully ran with these changes. ;-)

test/javascript/components/interchange.js
DanielRuf: Reset it to `50`.

test/javascript/components/interchange.js
ncoden: hmm... Ok. But we don't know _why_ ?
`IHearYou` works on `[data-resize}` which is set at `Interchange` Initialisation. This should be done synchronously and does not require any laps time.

ncoden: I got the "hey, it just works". But I don't think that's enough. We'll get into others trouble and weird bugs soon if we just to like "meh, ok it just works" without trying to understand where the real issue is.

test/javascript/components/interchange.js
ncoden: My point is that, according what I understand from the code, `1`, `2`, `3`, etc should be enough as delays, or even `setTimeout` without delays. If there is race conditions with that, well there's a bug in our code and we should not try to hide it in our tests.

DanielRuf:
Let me know if and when I can squash merge both commits into one and change the commit title.

test/javascript/components/interchange.js
ncoden: Ah yeah. `debounce` 10 by default.

ncoden (approved):
👍 Please squash commits.

ncoden:
Thanks

PR #11180, opened 2018-04-16
chore: use test dependencies from node_modules

DanielRuf:
This replaces https://github.com/zurb/foundation-sites/pull/11177 and updates the browser-based test file to use the dependencies from node_modules.


package.json
```diff
@@ -39,6 +39,7 @@
     "babel-preset-env": "^1.6.1",
     "browser-sync": "^2.18.12",
     "browserstack-runner": "^0.7.0",
+    "chai": "^4.1.2",
```
ncoden: We must update both `package-lock.json` and `yarn.lock`. Please do the following:
```bash
rm package-lock.json
yarn
npm i --package-lock-only
sed -i '' 's/http\:\/\//https\:\/\//g' package-lock.json
sed -i '' 's/http\:\/\//https\:\/\//g' yarn.lock
```

test/javascript/index.html
```diff
@@ -6,16 +6,16 @@
...
+    <link rel="stylesheet" href="../../node_modules/mocha/mocha.css" />
     <link rel="stylesheet" href="../../_build/assets/css/foundation-float.css">
     <link rel="stylesheet" href="../../node_modules/motion-ui/dist/motion-ui.css">
   </head>
   <body>
     <div id="mocha"></div>
-    <script src="lib/chai.js"></script>
-    <script src="lib/chai-jquery.js"></script>
-    <script src="lib/sinon.js"></script>
+    <script src="../../node_modules/chai/chai.js"></script>
```
ncoden: We use `chai` directly, so please install `chai` in devDependencies.

ncoden (changes requested):

test/javascript/index.html
DanielRuf: That's what I did (see the package.json)?

package.json
DanielRuf: Hm, thought we use snyp or something similar. Will do it when I come to the installation of yarn.

test/javascript/index.html
ncoden: hmmm I did not seen that, sorry.

package.json
ncoden: I took care of that.

ncoden (approved):
👍 LGTM, thank you @DanielRuf

ncoden:
I guess that tests fail here because of the focus bug you described at #11177 ?

DanielRuf:
> I guess that tests fail here because of the focus bug you described at #11177 ?

Exactly.

ncoden:
Ok. I'll take a look at this bug later. Like, why does this happens only now with updated test dependencies ?

DanielRuf:
> Ok. I'll take a look at this bug later. Like, why does this happens only now with updated test dependencies ?

It also happened before but the new dependencies clearly show the issue instead of hiding it. Before this we saw it only with the previous Browserstack setup and now the testsuite fails due to this (before this it got a timeout with an undefined exit code).

Issue #11195, opened 2018-04-20
Is there a way to set additional properties for header styles in settings.


asolopovas:
Hi,

I have been using foundation for a while now and would like to suggest a feature, I am not sure if it can work but if it would that would be awesome.

Currently, we have variable $header-styles, where we set 'font-size' at different breakpoints. It would be cool if we could place additional properties such as colour and padding. This would give great flexibility as you could change your heading paddings and colors right from settings.scss file. Like so

```scss
$header-styles: (
  sm: (
    'h1': ('font-size': 19, 'padding': 10),
  ),
  md: (
    'h1': ('font-size': 38, 'padding': 30),
  ),
);
```

ncoden:
Hi @asolopovas,

Thank you for your issue.
There is already discussions about this in https://github.com/zurb/foundation-sites/issues/9483. So I'm closing this for now. Please reopen if you think this is different.

Issue #10918, opened 2018-02-06
Support more or all properties in $header-styles


DanielRuf:
So far $header-styles only uses 'font-size', 'line-height', 'margin-top' and 'margin-bottom'.

It makes probably sense to support more or all properties to prevent duplicate declaration of header styles in more scss files which are not supported [here](
https://github.com/zurb/foundation-sites/blob/develop/scss/typography/_base.scss#L320-L360)

https://github.com/zurb/foundation-sites/blob/develop/scss/settings/_settings.scss#L140-L157

https://github.com/zurb/foundation-sites/blob/develop/scss/typography/_base.scss#L320-L360

ncoden:
See https://github.com/zurb/foundation-sites/issues/9483 and https://github.com/zurb/foundation-sites/pull/9500.

There was a lot of discussions about the subject, and I don't remember end up agreeing on something. The main problem is that `$header-styles` is a tool to help to quickly create header styles. With too many options or supported properties, it became easier to advise people to create their header styles manually.

JeremyEnglert:
Closing this as it is a duplicate of an existing ticket.

I think this is something we should definitely discuss for F7.

DanielRuf:
Right. I'll subscribe to the other two issues.

PR #11196, opened 2018-04-22
[WIP] feat: make header styling more flexible using placeholders

oxyc:
Based on @ncoden PR https://github.com/zurb/foundation-sites/pull/9500 and issue #9483.

There's been a few issues (#10918, #11195) about supporting more properties in the `$header-styles` map, but most of those use cases are not about applying different styles to different screen sizes, but just getting styles applied to all headers as well as the new header classes `.h1, .h2` etc. Therefore I don't think they belong in `$header-styles` but there should fit in somewhere for sure (eg. the use case of h1 having a different font-weight than h2-h6 comes up quite frequently on my projects).

Currently the only method is manually specifying each selector like so

```scss
h1, .h1,
h2, .h2,
h3, .h3,
h4, .h4,
h5, .h5,
h6, .h6 {
  text-decoration: underline
}
```

This is a bit tedious and if you ever wanted to eg. use the heading style for a different type of element or class, you would have to extend the class

```scss
select.filter {
  @extend .h1;
}
```

Extending classes is a terrible idea in my opinion and causes very difficult to find bugs. The main issue is how all [chained selectors gets applied](https://www.smashingmagazine.com/2015/05/extending-in-sass-without-mess/#what-happens-when-you-use-extend) as well. I put up a simple example illustrating the issue here: https://codepen.io/anon/pen/aGvpwG

Where I work we have a policy never to extend classes directly and always extend placeholders instead.

This PR would allow users to extend headers much more easily and I'd vote for the pattern being reused in many other parts of Foundation. Specifically buttons as I often end up wanting to reapply button styles to selectors I might not have control over (using WP for example).

Examples:

```scss
// Alternative to the first example posted in this message
%header {
  text-decoration: underline
}

// Apply an additional style to h1, .h1 as often requested in other issues.
%h1 {
  font-weight: 900;
}

// Applies the complete foundation + user styling of h1 headers.
select.filter {
  @extend %h1;
}
```

_Note: this is probably not mergable, the `_tmp.css` file is a stray file from ncodens PR and I'm mostly making this PR to illustrate my idea._


ncoden (commented):
Hi @oxyc. I did not know we could define placeholders multiple times and this is a really great news for Foundation. 👏 

A common issue for all our components is people wanting to add some styles and configuration in components with complex selectors. As discussed in #9500, we cannot add support for all CSS3 properties in our configuration file. Our current architecture have reached its limits.

I agree with everything you said. Extending selectors is not a good practices. User should be able to fully customize our components and the best way to do it is via its own CSS code, but we must give him tools to do it correctly. Defining placeholders to let the user extending it as it will the same way he/she would include mixins is definitely the best solution. It give him/her the choice of freely customizing the components we provide (without settings required) or creating new ones from them.

So thank you, this is a really good PR (and first contribution). 🎉 

Sadly, you can see by the "SAVE SUFF" commit that it is based on incomplete "WIP" work. It must be cleaned, rebased and documented. So I propose you two options:
* I try to finish #9500 and include your commit in it after that.
* I try to finish #9500, merge it, and you open a new PR based on it after that.

What do you prefer ?

PS: Thank you for the good explanations, in 2 years around Foundation this is the best pull request message I ever seen 👍!

ncoden:
> Note: this is probably not mergable, the _tmp.css file is a stray file from ncodens PR and I'm mostly making this PR to illustrate my idea.

Marking this as Work In Progress

oxyc:
> I try to finish #9500 and include your commit in it after that.

Feel free to include it, it's mostly based on your work! I can help out next week if you don't have enough time but otherwise I don't need the contribution :)

DanielRuf (approved):
LGTM, what can we do next here?

DanielRuf:
I guess we have to cleanup and rebase this one?

joeworkman (approved):

Issue #11193, opened 2018-04-20
Issue with TypeScript definitions for selectTab


glen-84:
#### How to reproduce this bug:

1. Use code like `$("#matchGames").foundation("selectTab", "#result");`

#### What should happen:

No TypeScript error – `elem` can be a string according to the [docs](https://foundation.zurb.com/sites/docs/tabs.html#selecttab).

#### What happened instead:

TypeScript error: `Argument of type '"#result"' is not assignable to parameter of type 'JQuery<HTMLElement>'.`

![image](https://user-images.githubusercontent.com/261509/39064419-c0291f72-44ce-11e8-8de5-6d9315b99e88.png)

The problem is here:

```ts
interface JQuery {
  foundation(method?: string|Array<any>, $element?: JQuery): JQuery;
}
```

... but changing that might not be ideal?

#### Browser(s) and Device(s) tested on:

n/a

#### Foundation Version(s) you are using:

6.4.3

#### Test case link:

n/a

glen-84:
@DaSchTour It looks like it does. Thanks!

PR #11199, opened 2018-04-24
Possible fix for selectTab defintion error #11193

DaSchTour:
Possible fix for #11193

### DIFF
--- a/dist/js/foundation.d.ts
+++ b/dist/js/foundation.d.ts
@@ -486,6 +486,7 @@ declare module FoundationSites {
 }
 
 interface JQuery {
+  foundation(method: "selectTab", $element: JQuery | string, historyHandled?: boolean): JQuery;
   foundation(method?: string|Array<any>, $element?: JQuery): JQuery;
 }
 


DaSchTour:
Verified by bug reporter

DanielRuf (approved):

dist/js/foundation.d.ts
```diff
@@ -486,6 +486,7 @@ declare module FoundationSites {
 }
 
 interface JQuery {
+  foundation(method: "selectTab", $element: JQuery | string, historyHandled?: boolean): JQuery;
```
ncoden: Why adding a special case of `selectTab` while all components methods can actually take their own parameters that may not be a jQuery element. We should make `foundation(method?: string|Array<any>, $element?: JQuery): JQuery;` more flexible instead.

ncoden (changes requested):

dist/js/foundation.d.ts
DanielRuf: Hm, isn't `historyHandled?: boolean` optional anyway and could be merged to the original line? I guess that is what you mean (not sure here if this would work due to the different interfaces).

ncoden: What I meant is:
```typescript
foundation(method?: string|Array<any>, ...args: any[]): JQuery;
```

DanielRuf: Isn't this too general and doesn't this not show the `$element` and `historyhandled` params?

ncoden: A lot of methods in Foundation does not have the `$element` param and have others params types. I see three solutions to support them:
* allow everything (like proposed above)
  ```typescript
  foundation(method?: string|Array<any>, ...args: any[]): JQuery;
  ```
* allow optional `$element` and everything
  ```typescript
  foundation(method?: string|Array<any>, $element?: JQuery, ...args: any[]): JQuery;
  ```
* support all Foundation components methods
  ```typescript
  foundation(method?: string|Array<any>, $element?: JQuery): JQuery;
  foundation(method: 'selectTab', $element: JQuery | string, historyHandled?: boolean): JQuery;
  foundation(method: 'removeRadioErrorClasses', groupName: string);
  foundation(method: 'validateText', $element: JQuery, pattern: string): boolean;
  foundation(method: 'validateRadio', $element: groupName: string): boolean;
  foundation(method: 'matchValidation', $element: validators: string, required: boolean): boolean;
  ```

I just want to be consistent.

DaSchTour: Well this should just be a fix for this reported issue. I didn't have the time to go through all the foundation methods and creating typings for them. We should definitely create as exact typings as possible. The first two solutions lose a lot of type safety and the second would still lead to the reported bug.

ncoden: The problem with the 3rd option is that typing is relative to `method` here, not the plugin. So if we have a method used in two different plugins (like `open`, `close`, `toggle`...), there can be incompatibilities.

We have a `.foundation` funciton that can take _anything_ as arguments. So it may be less "typing safe", but it is more logical and certain to accept _anything_ in the typing too (like the first option).

DaSchTour: Well then make it `foundation(...args: any[]): JQuery;`
I must admit, that we stopped using any of foundations JS because it's a hell using it in a typescript environment.

ncoden:
> @DaSchTour closed this 2 minutes ago

@DaSchTour Will you open an other PR ?

PR #11214, opened 2018-04-29
fix: check for activeElement before calling blur() on it

### LINKED ISSUE #11213: tests: failing unit tests in IE11 due to blur() being called on null / undefined
## Expected Behavior

The unit tests should successfully run in IE11

## Current Behavior

A few unit tests fail due to `blur()` being called on null / undefined.

> TypeError: Die Eigenschaft "blur" eines undefinierten oder Nullverweises kann nicht abgerufen werden.
   at Anonymous function (js-tests.js:2340:13)

## Possible Solution

Check if `document.activeElement` is defined at all.

How to reproduce:
1. run the unit tests in IE11 by opening the generated index.html in tests/javascript

## Your Environment

- Foundation version(s) used: develop
- Browser(s) name and version(s): IE11
- Operating System and version (desktop or mobile): Windows 10

## Checklist (all required):

- [x] This is a bug report or a feature request.

DanielRuf:
## Description

Currently the tests throw errors in IE11 on Windows 10 because there might be some race conditions or differences in the browser behaviors as document.activeElement is not always defined.

- Closes https://github.com/zurb/foundation-sites/issues/11213
## Motivation and Context

Fixes the unit tests in IE11 on Windows 10.

## Types of changes

- [ ] Documentation
      functionality to change)

## Checklist (all required):

- [x] The pull request targets the right branch (`develop` or `support/*`).

### DIFF
--- a/test/javascript/components/accordion.js
+++ b/test/javascript/components/accordion.js
@@ -27,7 +27,7 @@ describe('Accordion', function() {
 
   afterEach(function() {
     plugin.destroy();
-    if(document.activeElement)document.activeElement.blur();
+    if(document.activeElement) document.activeElement.blur();
     $html.remove();
   });
 
--- a/test/javascript/components/accordion.js
+++ b/test/javascript/components/accordion.js
@@ -27,7 +27,7 @@ describe('Accordion', function() {
 
   afterEach(function() {
     plugin.destroy();
-    document.activeElement.blur();
+    if(document.activeElement)document.activeElement.blur();
     $html.remove();
   });
 
--- a/test/javascript/components/interchange.js
+++ b/test/javascript/components/interchange.js
@@ -172,7 +172,7 @@ describe('Interchange', function() {
       plugin = new Foundation.Interchange($html, {});
       setTimeout(function () {
         Foundation.IHearYou();
-      }, 10);
+      }, 1);
       let spy = sinon.spy(plugin, '_reflow');
 
       setTimeout(function () {
--- a/test/javascript/components/interchange.js
+++ b/test/javascript/components/interchange.js
@@ -170,27 +170,19 @@ describe('Interchange', function() {
     it('calls reflow on viewport size change once', function(done) {
       $html = $(generateTemplate('image')).appendTo('body');
       plugin = new Foundation.Interchange($html, {});
-      setTimeout(function() {
+      setTimeout(function () {
         Foundation.IHearYou();
-      }, 1);
+      }, 10);
       let spy = sinon.spy(plugin, '_reflow');
 
-      setTimeout(function() {
+      setTimeout(function () {
         $(window).trigger('resize');
-      }, 5);
+      }, 50);
 
-      setTimeout(function() {
-        $(window).trigger('resize');
-      }, 10);
-
-      setTimeout(function() {
-        $(window).trigger('resize');
-      }, 20);
-
-      setTimeout(function() { // Wait for third trigger...
+      setTimeout(function () {
         sinon.assert.calledOnce(spy);
         done();
-      }, 50);
+      }, 150);
     });
   });
 
--- a/test/javascript/util/keyboard.js
+++ b/test/javascript/util/keyboard.js
@@ -20,7 +20,7 @@ describe('Keyboard util', function() {
   };
 
   afterEach(function() {
-    document.activeElement.blur();
+    if(document.activeElement)document.activeElement.blur();
   });
 
   it('exists on the Foundation API', function() {
--- a/test/javascript/util/keyboard.js
+++ b/test/javascript/util/keyboard.js
@@ -20,7 +20,7 @@ describe('Keyboard util', function() {
   };
 
   afterEach(function() {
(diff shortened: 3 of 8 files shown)


ncoden (commented):
Stupid mistake on my side, sorry about that.

But I don't understand why doesn't this appear in https://travis-ci.org/zurb/foundation-sites/jobs/372717914 ? I can reproduce it locally with `index.html`.

test/javascript/components/accordion.js
```diff
@@ -27,7 +27,7 @@ describe('Accordion', function() {
 
   afterEach(function() {
     plugin.destroy();
-    document.activeElement.blur();
+    if(document.activeElement)document.activeElement.blur();
```
ncoden:
```
if(document.activeElement)document.activeElement.blur();
//                        ^
//                     no space ?
```

DanielRuf:
I still guess that the tests on Browserstack run slower than on real devices.

ncoden:
> chore: add missing space between condition and statement 

Prefix depends on the type of change, not where it is. Here, it should be `style`

test/javascript/util/keyboard.js
```diff
@@ -20,7 +20,7 @@ describe('Keyboard util', function() {
   };
 
   afterEach(function() {
-    document.activeElement.blur();
+    if(document.activeElement)document.activeElement.blur();
```
ncoden:
```js
if(document.activeElement)document.activeElement.blur();
//                        ^
//                     no space ?
```

You can rebase this change with the previous commit I guess. The PR is clear enough. And use `style: ...`

test/javascript/components/interchange.js
```diff
@@ -170,27 +170,19 @@ describe('Interchange', function() {
...
         $(window).trigger('resize');
-      }, 5);
-
-      setTimeout(function() {
-        $(window).trigger('resize');
-      }, 10);
-
-      setTimeout(function() {
-        $(window).trigger('resize');
-      }, 20);
```
ncoden: This break the purpose of the test itself. `calls reflow on viewport size change once` means that with multiple `resize`, `_reflow` is called only once within a given delay (10ms by default for MutationObserver debounce). We cannot test this behavior without triggering `resize` once.

So we actually broke these test before with switching to delays over 10ms. We should triggers several `resize` in 10ms and expect reflow to be called once.

Also, this should not be in this PR.

ncoden (changes requested):
> 29d21c8 tests: always call blur() for asynchronous calls 

Why "asynchronous calls". I see nothing asynchronous in [test/javascript/components/dropdownMenu.js](https://github.com/DanielRuf/foundation-sites/blob/29d21c8b92c72f2fd1da94874a83aaabfcf58eba/test/javascript/components/dropdownMenu.js).

Also even if `.focus()` is called asynchronously, we wait for the test to finish before calling `.blur()` in `afterEach`.

So this line is not the issue.

test/javascript/components/interchange.js
DanielRuf: Right. But not fixing this in this PR will produce a failing Travis build.

What do you recommend because of this issue with reflow and the timing issues on the Travis containers?

DanielRuf:
> So this line is not the issue.

The testsuite tells differently but ok. How would you solve this here? Not sure if we can solve these issues related to timing functions when the browser / VM is too slow.

test/javascript/components/interchange.js
ncoden: > Right. But not fixing this in this PR will produce a failing Travis build.

Are you sure ? Tests runs after `tests: always call blur() for asynchronous calls`. Do we need both commits for tests to runs ?

I would recommend to actually test what we want to test and expect it to fail if the component does not work properly: trigger `resize` several times in 10ms and check after that if `_reflow` was called once.

This _shoud_ work. If it doesn't, Interchange has a bug. I'll take a look at this later.

DanielRuf: > I'll take a look at this later.

This would be really helpful, thanks.

ncoden:
> I'll take a look at this later.

Will do this tomorrow

ncoden:
So the bug with the previous Interchange tests came from the browser not able to respect the setTimeout delays, which lead to big differences between several timeout running in parrallel. If the difference is bigger than the `utils.trigger` debounce time (10ms), several reflow may be triggered.

DanielRuf:
Well, in general setTimeout is never exact when it comes to timing as it is often longer than what is set as timeout.

So what is the correct solution in your opinion? Keep in mind that the VM might be slower and timeouts may trigger later.

ncoden:
See https://github.com/zurb/foundation-sites/pull/11259

DanielRuf:
> If you're okay with these PRs

Which PRs? You linked just one? I thought these replace mine?

ncoden:
Sorry, see https://github.com/zurb/foundation-sites/pull/11259 and its required fix https://github.com/zurb/foundation-sites/pull/11258

ncoden:
Closed in favor of #11260

PR #11265, opened 2018-05-14
feat: filter elements with plugins from reflow method

thoughtassassin:
Filter elements with plugins from reflow method:
- Add filter method to find to line 152 of foundation.core.js
- Remove warning from subsequest each method

## Description
Foundation sites warns "Tried to initialize x on an element that already has a foundation plugin" in the reflow method in foundation.core.js. The elements that already have the foundation plugin can easily be filtered before applying the filter. If the elements are filtered properly the warning would be unnecessary.

A codepen to illustrate the issue is here:

https://codepen.io/thoughtassassin/pen/KRBqqa?editors=1011

Open the console to see the warnings when you click the "Toggle" button.                                                 

## Motivation and Context
In the React app that I am working on I call $(document).foundation() in the componentDidMount method and in componentDidUpdate method. componentDidMount initializes the elements when the app loads. componentDidUpdate initializes new elements in the DOM. However, when the method is called in componentDidMount, the elements that already have plugins initialized throw the warning.

The QA engineer that is testing the app doesn't like seeing these warnings. Even though it seems mostly harmless, it would be nice to remove the warnings when using $(document).foundation() to initialize elements that have not been initialized.

## Types of changes
Added a filter to the find and addBack methods on $elem in the reflow method in foundation.core.js to remove elements that already have a "zfPlugin." I also removed the warning that was thrown in the forEach method of $elem that follows.

- [ ] Documentation
      functionality to change)

## Checklist (all required):

- [x] The pull request targets the right branch (`develop` or `support/*`).

### DIFF
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -149,6 +149,8 @@ $.ajax('assets/partials/kitten-carousel.html', function(data) {
 });
 ```
 
+Plugins that are already initialized will be ignored. However for performances reasons, we recommend calling `.foundation()` only on the DOM nodes that changed.
+
 ---
 
 ## Adding Content to Plugins
--- a/docs/pages/javascript.md
+++ b/docs/pages/javascript.md
@@ -149,7 +149,7 @@ $.ajax('assets/partials/kitten-carousel.html', function(data) {
 });
 ```
 
-Plugins that are already initialized will be ignored. However for performances reasons, we recommend calling `.foundation()` only on the DOM nodes that changed.
+Plugins that are already initialized will be ignored. However for performance reasons, we recommend calling `.foundation()` only on the DOM nodes that changed.
 
 ---
 
--- a/js/foundation.core.js
+++ b/js/foundation.core.js
@@ -149,18 +149,15 @@ var Foundation = {
       var plugin = _this._plugins[name];
 
       // Localize the search to all elements inside elem, as well as elem itself, unless elem === document
-      var $elem = $(elem).find('[data-'+name+']').addBack('[data-'+name+']');
+      var $elem = $(elem).find('[data-'+name+']').addBack('[data-'+name+']').filter(function () {
+        return typeof $(this).data("zfPlugin") === "undefined";
+      });
 
       // For each plugin found, initialize it
       $elem.each(function() {
         var $el = $(this),
             opts = {};
-        // Don't double-dip on plugins
-        if ($el.data('zfPlugin')) {
-          console.warn("Tried to initialize "+name+" on an element that already has a Foundation plugin.");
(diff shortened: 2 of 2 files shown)


ncoden (commented):
I @thoughtassassin,

I understand the purpose of this pull request. It would be really useful for the case where we cannot simply keep the trace of what is already initialized or not, so we would just "initialize everything".

However this can quickly become a bad practice. It would be easy to re-browse the entire DOM three at each change anywhere on the page instead of only where and when it is needed.

This can have a huge performance impact so it is most often not recommanded. This is IMHO a good reason to throw a warning, but it should be removed in production.

thoughtassassin:
@ncoden,

I am adding another filter operation. That is what I am guessing may be the performance hit. However, The filter operation is not traversing the document tree. It simply removes a few elements that may be collected during the first traversal ($(elem).find().addBack)()). Therefore, concerns about additional document tree traversal seem misplaced here.

From what I could see in the app that I am working on, once the document traversal occurred during the _find_ and _addBack_ operation, there were at most two elements in the list...hardly a performance hit. However, I cannot say if my app is representative, or what the cumulative effect might be in a large and complex application. 

The simplest solution seems to be just to return from the forEach operation, as in the current implementation, and remove the warning. The check for $(this).data('zfPlugin') is already happening. All I did was add a filter operation and remove the warning. I did so because it seems illogical to iterate over an element that you know is not going to be initialized.

I agree that it would be nice to follow best practices by keeping track of which elements are initialized. The warning does little to enforce that practice. For relatively simple applications, and for casual users, this warning smacks of premature optimization (therefore unnecessary).

What I am proposing is providing the flexibility to use the framework without the heavy-handed pedantry. This change will not prevent users from optimizing their code as they see fit or as necessary. However, for those where this is not a concern, there is no choice but to see a warning that makes little sense in many scenarios.

SassNinja:
> This is IMHO a good reason to throw a warning, but it should be removed in production.

There's currently no dev/prod mode, is there?
(thinking abt sth as `Foundation.mode = 'development'` or `Foundation.debug = true` 😉)

Anyway I think throwing a warning does probably not keep most developers from doing multiple reinitialization bcz it doesn't break anything.
So for me this PR is fine.

ncoden:
Ok, you convinced me. We can remove the warning, but we have to add some explanations in the documentation about the possible performance issue (like "don't call `.foundation()` on document again and again"). I'll take care of that.

thoughtassassin:
Thank you, @ncoden!

ncoden (approved):
@thoughtassassin I added a warning in the documentation. I found nothing mentioning the current `.foundation()` limitation.

PR #11065, opened 2018-03-15
fix: do not add the scrollbar by default

DanielRuf:
Using reveal adds a double scrollbar as overflow: scroll is used. We should use overflow: auto instead.

https://codepen.io/DanielRuf/pen/RMRVjO

https://codepen.io/DanielRuf/debug/RMRVjO

Closes https://github.com/zurb/foundation-sites/issues/11064 and https://github.com/zurb/foundation-sites/issues/10824

Codepen with the changes: https://codepen.io/DanielRuf/pen/RMRgNN

PR #10583, opened 2017-08-26
fixed issue #7831

isapir:
based on https://stackoverflow.com/a/13891717/968244

PR #11341, opened 2018-06-17
fix: keep a scrollbar on document when Reveal opens #7831

### LINKED ISSUE #7831: [Reveal] Background content width problem.
On reveal modal docs page
1. open any modal link

To open modal and nothing more

**What happened instead?**
When I open modal the width of content is changing

**Test case**
Open any modal link

Give us a link to a [CodePen](http://codepen.io) or [JSFiddle](http://jsfiddle.net) that recreates the issue.
http://foundation.zurb.com/sites/docs/reveal.html

ncoden:
## Description

#11065 was intented to avoid double scrollbars from a Reveal modal, but removing all scrollbars when the modal opens changes the page content width and the content shift. This is a regression of #7831.

The current solution is a mix between the two previous fixes:
* Modal may have a scrollbar or not, following its content height
* Body has no scrollbar and prevent any scroll
* Document may have a scrollbar, following its content height (controlled in JS)

So the body content never shift, as the document has a scrollbar when a Reveal modal is open only if it already had one before.

## Changes:
* add the zf-has-scroll global html modifer and toggle it on Reveal opening/closing and on window resizing
* always prevent scroll on body instead of html tag like in #10583

### DIFF
--- a/js/foundation.reveal.js
+++ b/js/foundation.reveal.js
@@ -324,6 +324,17 @@ class Reveal extends Plugin {
     this.$element.trigger('open.zf.reveal');
   }
 
+  /**
+   * Adds classes and listeners on document required by open modals.
+   *
+   * The following classes are added and updated:
+   * - `.is-reveal-open` - Prevents the scroll on document
+   * - `.zf-has-scroll`  - Displays a disabled scrollbar on document if required like if the
+   *                       scroll was not disabled. This prevent a "shift" of the page content due
+   *                       the scrollbar disappearing when the modal opens.
+   *
+   * @private
+   */
   _addGlobalClasses() {
     const updateScrollbarClass = () => {
       $('html').toggleClass('zf-has-scroll', !!($(document).height() > $(window).height()));
@@ -334,6 +345,10 @@ class Reveal extends Plugin {
     $('html').addClass('is-reveal-open');
   }
 
+  /**
+   * Removes classes and listeners on document that were required by open modals.
+   * @private
+   */
   _removeGlobalClasses() {
     this.$element.off('resizeme.zf.trigger.revealScrollbarListener');
     $('html').removeClass('is-reveal-open');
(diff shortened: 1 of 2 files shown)


ncoden:
Poke @isapir for review

DanielRuf:
This does probably not respect viewport changes.

DanielRuf:
> Poke @isapir for review

Was the review request meant for @isapir or me?

ncoden:
@DanielRuf Both

ncoden:
> This does probably not respect viewport changes.

You're right, we should watch for viewport change. I don't like doing things like this in JS but this seems to be the only solution.

ncoden:
Some comments to describe that "disabled scrollbar" behavior would be useful tho.

isapir (approved):
@ncoden @DanielRuf  It looks fine to me though I didn't test it by running.

IMO in the edge case that the viewport changes while the Reveal is open then it's OK to show a slight shift and/or disabled scrollbar.

ncoden:
@isapir @DanielRuf I made the required changes and some refactor.

> 96141b743 docs: add doc for Reveal methods _addGlobalClasses/_removeGlobalClasses
> deda9dc3a refactor: rename Reveal internal methods for a better clarity
> 82a04933a fix: show/hide Reveal scrollbar on window resizing

DanielRuf:
> fix: show/hide Reveal scrollbar on window resizing

I guess `orientationchange` is missing.

ncoden:
> I guess orientationchange is missing.

Doesn't it count as a resize ?

ncoden:
I got my answer https://stackoverflow.com/a/23996839/4317384

> Some devices/browsers do, some not.

Web development often sounds like biology classes. There's rule for everything and exception for every rule.

ncoden:
@DanielRuf Actually I am using `resizeme` from `utils.triggers`, and the entire framework rely on this. So if we decide to use it (according to browser support), we'll not do it in this PR.

DanielRuf:
Ok.

DanielRuf (approved):

PR #11388, opened 2018-07-07
docs: add pnpm

DanielRuf:
## Description

Adds pnpm install command. pnpm is "performant npm".

## Types of changes

- [x] Documentation
      functionality to change)

## Checklist (all required):

- [ ] The template is fully and correctly filled.
- [ ] My code follows the code style of this project.
- [ ] All new and existing tests passed.


ncoden (changes requested):
https://pnpm.js.org/docs/en/limitations.html

> npm-shrinkwrap.json and package-lock.json are ignored. Unlike pnpm, npm can install the same name@version multiple times and with different sets of dependencies. npm's shrinkwrap file is designed to reflect the node_modules layout created by npm. pnpm cannot create a similar layout, so it cannot respect npm's lockfile format.

This is a serious issue. Anyone collaborating on a project with `pnpm` would ignore lockfiles, install different dependencies, may discover bugs specific to its even environment and introduce bugs to the project because of that.

As every npm package is already installable by `pnpm` (like yarn), adding the `pnpm` installation command would only be promotion and I don't think we should promote it because of its limitations.

What do you think ?

DanielRuf:
In general pnpm takes a totally different approach than npm and yarn (symlinked dependencies) and does many things better.

@zkochan how do you see this?

ncoden:
Hi @zkochan 

We do not have the same issue with Yarn as it is now pretty popular and have a `yarn.lock` file that is used in most popular project. It is simple to keep it updated alongside `package-lock.json` and some GitHub dependency managers like [dependabot](dependabot.com) support it.

I took at better look at your documentation and seen that you have your own lockfile `shrinkwrap.yaml`. I think we can add `pnpm` installation command to the documentation, but not before adding a `shrinkwrap.yaml` lockfile compatible with others lockfiles, like we do for Yarn.

However, I don't find any way to do that. Unlike `yarn import` and `npm i --package-lock-only` that generate a lockfile from others lockfiles or installed `node_modules`, `pnpm i --shrinkwrap-only` will link `node_modules` _then_ generate `shrinkwrap.yaml` in such a way that modules are always updated to the latest versions.

Doing the installation with pnpm first then generating the npm/yarn lockfiles from it is not a reliable solution neither as our contributors may not want to use `pnpm` and rebuild their `node_modules` folder, and because the linked `node_modules` make npm and yarn considering all installed package dependencies as pinned (I would not be able to say why, I guess it comes from the way dependencies are organized in the `pnpm` store).

Would you know how we could continue to support and update both npm, yarn and pnpm lockfiles ?
Something like `pnpm import` similar to [yarn import](https://yarnpkg.com/en/docs/cli/import) would be useful, like making `pnpm i --shrinkwrap-only` working with non-linked `node_modules`.

What do you think ?

zkochan:
I created an issue in the pnpm repo to implement `pnpm import`.

ref https://github.com/pnpm/pnpm/issues/1266

zkochan:
pnpm now has a `pnpm import` command that can convert a `package-lock.json`/`npm-shrinkwrap.json` to `shrinkwrap.yaml`

ncoden (approved):
LGTM 👍 
Thank you @DanielRuf and @zkochan!

PR #11393, opened 2018-07-09
fix: set matchMedia on window to fix issue on IE11, fixes #11358

### LINKED ISSUE #11358: Polyfill for mediaQuery breaks in IE11
## What should happen?
Foundation should load all of its JavaScript (in this case, trying to initialize an accordion) in IE11 without throwing any exceptions.

## What happens instead?
In IE11, get error **`SCRIPT65535: Invalid Calling object`** in the console window and the accordion does not work.

## Possible Solution
The cause seems to be the polyfill for `window.matchMedia`.  There is a local variable `matchMedia`, to which `window.matchMedia` is being assigned, and calling this as a function in _[foundation.util.mediaQuery.js](https://github.com/zurb/foundation-sites/blob/v6.4.3/js/foundation.util.mediaQuery.js#L163)_ is what seems to trigger the error.

The preferable solution would be to remove the polyfill from foundation altogether (just use the native `window.matchMedia`), and leave a note in the setup guide specifying that for older browsers a third-party polyfill will be required ([https://github.com/paulirish/matchMedia.js](https://github.com/paulirish/matchMedia.js))... Foundation should not be baking in its own polyfills.

## Steps to Reproduce
How to reproduce:
1. Define markup for an accordion.
2. Call `$(document).foundation()`
3. Navigate to site in IE11
4. Observe that clicking on the accordion does not cause it to expand or collapse, and there is an error in the console.

## Context

I'm just trying to use the Accordion plugin.  It works in Chrome, but not IE11 b/c of the aforementioned polyfill.

## Your Environment

- Foundation version(s) used: 6.4.3
- Browser(s) name and version(s): Internet Explorer 11
- Operating System and version (desktop or mobile): Windows 10 Enterprise (desktop)

## Checklist (all required):

DanielRuf:
## Description

This fixes an issue in IE11 with a conflict between window.matchMedia and matchMedia.

- Closes #11358

## Types of changes

- [ ] Documentation
      functionality to change)

## Checklist (all required):

- [] The pull request targets the right branch (`develop` or `develop-v...`).

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -17,9 +17,10 @@ const defaultQueries = {
 
 
 // matchMedia() polyfill - Test a CSS media type/query in JS.
-// Authors & copyright (c) 2012: Scott Jehl, Paul Irish, Nicholas Zakas, David Knight. Dual MIT/BSD license
-window.matchMedia = window.matchMedia || (function() {
-  'use strict';
+// Authors & copyright(c) 2012: Scott Jehl, Paul Irish, Nicholas Zakas, David Knight. MIT license
+/* eslint-disable */
+window.matchMedia || (window.matchMedia = (function () {
+  "use strict";
 
   // For browsers that support matchMedium api such as IE 9 and webkit
   var styleMedia = (window.styleMedia || window.media);
@@ -33,14 +34,18 @@ window.matchMedia = window.matchMedia || (function() {
     style.type  = 'text/css';
     style.id    = 'matchmediajs-test';
 
-    script && script.parentNode && script.parentNode.insertBefore(style, script);
+    if (!script) {
+      document.head.appendChild(style);
+    } else {
+      script.parentNode.insertBefore(style, script);
+    }
 
     // 'style.currentStyle' is used by IE <= 8 and 'window.getComputedStyle' for all other browsers
     info = ('getComputedStyle' in window) && window.getComputedStyle(style, null) || style.currentStyle;
 
     styleMedia = {
-      matchMedium(media) {
-        var text = `@media ${media}{ #matchmediajs-test { width: 1px; } }`;
+      matchMedium: function (media) {
+        var text = '@media ' + media + '{ #matchmediajs-test { width: 1px; } }';
 
         // 'style.styleSheet' is used by IE <= 8 and 'style.textContent' for all other browsers
         if (style.styleSheet) {
@@ -52,7 +57,7 @@ window.matchMedia = window.matchMedia || (function() {
         // Test if media query is true or false
         return info.width === '1px';
       }
-    }
+    };
   }
 
   return function(media) {
@@ -60,8 +65,9 @@ window.matchMedia = window.matchMedia || (function() {
       matches: styleMedia.matchMedium(media || 'all'),
       media: media || 'all'
     };
-  }
-})();
+  };
+})());
+/* eslint-enable */
 
 var MediaQuery = {
   queries: [],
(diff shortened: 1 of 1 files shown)


DanielRuf:
See https://github.com/paulirish/matchMedia.js/blob/master/matchMedia.js#L3

ncoden (commented):
https://github.com/paulirish/matchMedia.js/blob/cd9df4c529a3c5c95c6d66b77bb7cc3d2419c460/matchMedia.js#L3 is a polyfill that will often be explicitely imported in the project. This is not the case with the embed version inside Foundation. Some could consider this as an undisired and dangerous side-effect, and that this `matchMedia` should only be used within the Foundation scope without affecting `window`.

If I understand the issue correctly, renaming `matchMedia` to `_matchMedia` would work too without introducing any side-effect, right ?

ncoden:
We could use the opportunity to update this polyfill to the latest version :)

DanielRuf:
> If I understand the issue correctly, renaming matchMedia to _matchMedia would work too without introducing any side-effect, right ?

Why would you do that? A polyfill is meant to directly polyfill the window object, it is now correct in the upstream version but not ours ;-)

This polyfills the native matchMedia.

DanielRuf:
It just polyfills / sets it if it is not yet available on the window object.

ncoden:
Does this polyfill perfectly reproduce the native `matchMedia` specs ? If yes, ok that's fine 👍. However, this could introduce bugs in an other package relying on its own polyfill but which would use our (buggy) one thinking it is the native one.

DanielRuf:
> Does this polyfill perfectly reproduce the native matchMedia specs ? If yes, ok that's fine

It should. It is only used if the browser does not support it.
https://github.com/paulirish/matchMedia.js/issues/65#issuecomment-133665167

https://caniuse.com/#feat=matchmedia

It should still work after these years. I trust @paulirish

There are not many polyfills, see https://github.com/search?o=desc&q=matchmedia&s=stars&type=Repositories and https://github.com/Financial-Times/polyfill-service/tree/master/packages/polyfill-library/polyfills/matchMedia

ncoden (approved):

kb3eua:
I still don't understand why Foundation is forcing this polyfill on users.  Some sites may only be targeting evergreen browsers that already support this natively, in which case this polyfill adds unnecessary bloat to the JavaScript file.  And I agree with @ncoden about the possibility of a conflict with other polyfills (not to mention that in this case you'd shipping 2 polyfills to do the same job... more unnecessary code bloat users will have to download).  The Foundation installation guide should mention that a polyfill is required for older versions of IE should the user want to support them, but it should put the onus on the developer to install the polyfill themselves separate from Foundation.

DanielRuf:
> that already support this natively, in which case this polyfill adds unnecessary bloat to the JavaScript file.

Polyfills are only applied if the polyfilled feature is not supported so this is not the case.

DanielRuf:
And it is not much code ;-)

kb3eua:
Also, the browsers that require this polyfill are so ancient and have such a small percentage of market share that it makes no sense to ship this polyfill.  IE11 supports this natively and has ~1.9% market share... older IEs that might need this polyfill have ~0.1% market share... at what point do you drop support for older IEs?

ncoden:
@kb3eua For now, Foundation support IE >= 9 (unless for flex features). It is already planned to deprecate oldersIE 9/10 in future versions, but keep in mind that Foundation is known for its very good browser support and often used for it.

DanielRuf:
> Also, the browsers that require this polyfill are so ancient and have such a small percentage of market share that it makes no sense to ship this polyfill. IE11 supports this natively and has ~1.9% market share... older IEs that might need this polyfill have ~0.1% market share... at what point do you drop support for older IEs?

This is totally different in different countries.

DanielRuf:
> Polyfills may be ignored during execution but they are still bloating the Javascript file

I know but this is not much compared to the rest. We can also apply a patch using patch-package to remove the code / polyfill ;-)

ncoden:
>  We can also apply a patch using patch-package to remove the code / polyfill ;-)

It works, but we should not rely on it as package maintainers, this is not an "healthy" and maintainable solution in the long term for our users.

DanielRuf:
Sure, this is why I proposed the polyfill-service and then drop internal polyfills ;-)

PR #11411, opened 2018-07-17
fix: resolve dropdown menu disappearing for #11410

arloduff:
## Description
Currently, if a dropdown menu contains an input field which displays suggestions to select from a dropdown (such as an email input on Chrome), the mouseleave event is triggered if the user hovers over one of the choices, causing the entire dropdown to disappear. This adds a check for the case of `e.relatedTarget === null` which will be the case when one of the input options is hovered over, but not when the mouse leaves through another event.

- Closes https://github.com/zurb/foundation-sites/issues/11410

## Motivation and Context
Allows users to choose an input option when their dropdown menu contains an input without the entire dropdown disappearing, which can be a bad user experience.

## Types of changes

Dropdowns containing input fields that offer suggestions will now stay open long enough for the user to choose one. I have found one bug with this change, which is that in Firefox, if a user hovers over one of the input suggestions and then uses the keyboard to navigate to a different window, the dropdown will remain open. The dropdown will auto close again once focus is returned to the browser window, but will remain open until then. However, since current functionality is for the dropdown to disappear when one of those suggestions is hovered over, I wouldn't consider this change to break existing functionality.

- [ ] Documentation
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing
      functionality to change)

## Checklist (all required):

- [x] I have read and follow the [CONTRIBUTING](CONTRIBUTING.md) document.
- [x] There are no other pull request similar to this one.
- [x] The pull request title is descriptive.
- [x] The template is fully and correctly filled.
- [x] The pull request targets the right branch (`develop` or `develop-v...`).
- [x] My commits are correctly titled and contain all relevant information.
- [x] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly to my changes (if relevant).
- [ ] I have added tests to cover my changes (if relevant).
- [ ] All new and existing tests passed.

PR #11418, opened 2018-07-23
fix: factorize mouseleave on browser UI fix and expand it to Dropdown & Tooltip

ncoden:
## Description

Factorize the bug fix introduced in https://github.com/zurb/foundation-sites/pull/11411 in DropdownMenu into an easily reusable helper and use it to patch all the containers that may encounter the same bug: Dropdowns and Tooltips.

In addition to the #11411 bug fix, the `ignoreMousedisappear` also wait for the mouse to reappear and check if it is still in the element or actually left it. This fix the case when the browser UI is in the border of the watched element and the mouse move trough it to leave it. With the previous fix in DropdownMenu, the menu would stay open indefinitely.

#### `ignoreMousedisappear(handler, options)`

Retuns an handler for the `mouseleave` that ignore disappeared mouses.

If the mouse "disappeared" from the document (like when going on a browser UI element, See https://git.io/zf-11410),
the event is ignored.
- If the `ignoreLeaveWindow` is `true`, the event is ignored when the user actually left the window
  (like by switching to an other window with [Alt]+[Tab]).
- If the `ignoreReappear` is `true`, the event will be ignored when the mouse will reappear later on the document
  outside of the element it left.

```js
$('.my-container').on('mouseleave', ignoreMousedisappear((e) => {
  // Will not be called when the mouse goes on a browser UI element
}));
```

### All changes:
- Add the `ignoreMousedisappear` core utility.
- Use `ignoreMousedisappear` for Dropdown, DropdownMenu and Tooltip.

Note: Orbit also rely on the "mouseleave" event but do not contains anything that may display a browser UI element.

## Types of changes

- [ ] Documentation
      functionality to change)

## Checklist (all required):

- [ ] ~~I have updated the documentation accordingly to my changes (if relevant).~~ (internal behavior)


ncoden:
Todo: add a visual test. Unit tests would be very hard - if not impossible - to make correctly. We have no way to control the mosue behavior on browser UI elements.

ncoden:
I added a visual test for the Dropdown. This made me find a bug I expected: the `mouseleave` event debouncing make the handler sometimes called after the non-filtered `mouseenter` event handler. We always assume in our codebase that `mouseleave` will always be called before the related `mouseenter`.

I am looking for a solution to detect windows switching in Firefox without event debouncing.

ncoden:
The `mouseenter` bug is resolved, Dropdown should work fine. However, there is still a race condition in the window leaving detection in Firefox. I'll search for a more suitable way to detect it, but for now I only found the lost of focus on window, that is triggered _after_ `mouseleave` and `mouseout`.

DanielRuf:
> @DanielRuf Could you review this ?

Planned for this week.

ncoden:
Hi @DanielRuf! Will you have time this week to review this pull request ?

ncoden:
@DanielRuf Let me know if you do not have the time to review this.

ncoden:
@DanielRuf @SassNinja 👋 Will you have some time to review this ? Thanks :)

PR #11440, opened 2018-08-10
fix: remove invalid "aria-multiselectable" on AccordionMenu #11437 

### LINKED ISSUE #11437: [Accessibility] Aria-Selected Attribute missing from Accordion
## What should happen?
Using accordion should add aria-selected attribute in order to pass AA and AAA accessibility conformance.

## What happens instead?
Accordion is created without aria-selected attribute which causes it to get flagged in SiteImprove Accessibility Checker. 

## Possible Solution
When using accordion class names, add aria-selected="false" attribute to nested <li> elements. 

## Test Case and/or Steps to Reproduce
Test Case:
Here is a screen shot depicting the accessibility issue for the missing aria-selected attribute. 

https://imgur.com/a/THYiONr

How to reproduce:
1. Create menu using accordion class names.
2. Use SiteImprove to test AA and AAA accessibility.
3. Go through SiteImprove to see the alerts for missing aria states or properties.

## Context
Effected myself and my team when trying to create a fully accessible web-based application. 

## Your Environment
- Foundation version(s) used: V6.4.1
- Browser(s) name and version(s): Chrome
- Operating System and version (desktop or mobile): Microsoft Windows 10 Pro

## Checklist (all required):

Thank you all for the great framework! Hope this helps improve accessibility for us all.

ncoden:
### Remove invalid `aria-multiselectable` in AccordionMenu 

As a menu, links inside the AccordionMenu plugin are only hidden for practical visual reasons. Opening a link "expand it" (make nested links visible), but no information (other than strictly visual) if added about the expanded link itself or the value it may represent.

For this reason, AccordionMenu's links cannot be considered as "selected" when opened but only "expanded", and the "aria-multiselectable" attribute should be removed.

Closes #11437 

### Remove useless `aria-multiselectable` in Drilldown 

The default value for `aria-multiselectable` is `false`, and this attribute is not used elsewhere in Foundation. I do not see any reason to keep it.

### Add `aria-selected` attribute to all `treeitem` elements

`aria-selected` is required for `treeitem` elements but Foundation plugins using this role does not have a "selected" state.

This pull request split the search for component elements from the setup or a11y attributes in Drilldown initialization, and add `[aria-selected=false]` to all `treeitem` elements in AccordionMenu and Drilldown.

## Types of changes

- [ ] Documentation
      functionality to change)

## Checklist (all required):

- ~~I have updated the documentation accordingly to my changes (if relevant).~~

### DIFF
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -53,10 +53,7 @@ class AccordionMenu extends Plugin {
     var _this = this;
 
     this.$element.find('[data-submenu]').not('.is-active').slideUp(0);//.find('a').css('padding-left', '1rem');
-    this.$element.attr({
-      'role': 'tree',
-      'aria-multiselectable': this.options.multiOpen
-    });
+    this.$element.attr({ 'role': 'tree' });
 
     this.$menuLinks = this.$element.find('.is-accordion-submenu-parent');
     this.$menuLinks.each(function(){
--- a/js/foundation.accordionMenu.js
+++ b/js/foundation.accordionMenu.js
@@ -85,9 +85,14 @@ class AccordionMenu extends Plugin {
         'id': subId
       });
     });
+
+    // Set a11y attributes for all items
     this.$element.find('li').attr({
-      'role': 'treeitem'
+      'role': 'treeitem',
+      // [aria-selected] is required for `treeitem`. See https://git.io/zf-11440.
+      'aria-selected': false,
     });
+
     var initPanes = this.$element.find('.is-active');
     if(initPanes.length){
       var _this = this;
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -54,10 +54,7 @@ class Drilldown extends Plugin {
       this.$element.addClass('drilldown');
     }
 
-    this.$element.attr({
-      'role': 'tree',
-      'aria-multiselectable': false
-    });
+    this.$element.attr({ 'role': 'tree' });
     this.$submenuAnchors = this.$element.find('li.is-drilldown-submenu-parent').children('a');
     this.$submenus = this.$submenuAnchors.parent('li').children('[data-submenu]').attr('role', 'group');
     this.$menuItems = this.$element.find('li').not('.js-drilldown-back').attr('role', 'treeitem').find('a');
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -56,14 +56,23 @@ class Drilldown extends Plugin {
 
     this.$element.attr({ 'role': 'tree' });
     this.$submenuAnchors = this.$element.find('li.is-drilldown-submenu-parent').children('a');
-    this.$submenus = this.$submenuAnchors.parent('li').children('[data-submenu]').attr('role', 'group');
-    this.$menuItems = this.$element.find('li').not('.js-drilldown-back').attr('role', 'treeitem').find('a');
+    this.$submenus = this.$submenuAnchors.parent('li').children('[data-submenu]');
+    this.$menuLis = this.$element.find('li').not('.js-drilldown-back');
+    this.$menuItems = this.$menuLis.find('a');
 
     // Set the main menu as current by default (unless a submenu is selected)
     // Used to set the wrapper height when the drilldown is closed/reopened from any (sub)menu
     this.$currentMenu = this.$element;
 
     this.$element.attr('data-mutate', (this.$element.attr('data-drilldown') || GetYoDigits(6, 'drilldown')));
+
+    // Set a11y attributes
+    this.$submenus.attr('role', 'group');
+    this.$menuLis.attr({
+      'role': 'treeitem',
+      // [aria-selected] is required for `treeitem`. See https://git.io/zf-11440.
+      'aria-selected': false,
+    });
 
     this._prepareMenu();
     this._registerEvents();


Owlbertz:
Looks good to me. But I tested that using the SiteImprove accessibility tool and it still claims the `aria-selected` is missing:
image

This is most likely the case because the `aria-selected` is stated as **required** in the [W3C documentation for treeitem](https://www.w3.org/TR/wai-aria-1.1/#treeitem). However, the [W3C documentation for aria-selected](https://www.w3.org/TR/wai-aria-1.1/#aria-selected) states that the attribute is not required for single selection containers if selection equals the focus.

ncoden:
I did not found any meaning in the W3C documentation about the "selected" state and when we should consider an item as "selected", but when reading between the lines of the [W3C documentation for treeitem](https://www.w3.org/TR/wai-aria-1.1/#treeitem) you shared, I found the following:

- A tree is a recursive list of items that can be collapsed and expanded.
- Multiple items can be "collapsed" or "expanded".
- If `aria-selected` represents the collapsing state of the items like `aria-expanded`, then `aria-multiselectable` should always be true thus should not be an option for `tree`.
- If `aria-selected` represents a state different from the "expanded" state, then it provide no information required by the definition of a `treeitem` and should not be required.

According to @mcking65 (https://github.com/w3c/aria-practices/issues/667) this is a bug, but when he checked in June `aria-selected` was not required. I can see it as required today in both the [ARIA 1.1 spec](https://www.w3.org/TR/wai-aria-1.1/#treeitem) and the [ARIA 1.2 Editor Release](https://w3c.github.io/aria/#treeitem).

ncoden:
@Owlbertz I did not found any meaning in the W3C documentation about the "selected" state and when we should consider an item as "selected", but when reading between the lines of the [W3C documentation for treeitem](https://www.w3.org/TR/wai-aria-1.1/#treeitem) you shared, I found the following:

- A tree is a recursive list of items that can be collapsed and expanded.
- Multiple items can be "collapsed" or "expanded".
- If `aria-selected` represents the collapsing state of the items like `aria-expanded`, then `aria-multiselectable` should always be true thus should not be an option for `tree`.
- If `aria-selected` represents a state different from the "expanded" state, then it provide no information required by the definition of a `treeitem` and should not be required.

According to @mcking65 (https://github.com/w3c/aria-practices/issues/667) this is a bug, but when he checked in June `aria-selected` was not required. I can see it as required today in both the [ARIA 1.1 spec](https://www.w3.org/TR/wai-aria-1.1/#treeitem) and the [ARIA 1.2 Editor Release](https://w3c.github.io/aria/#treeitem).

If it is not a bug they want us to always specify whether an item is selectable or not, and I don't see the reason behind this decision as "selectable" is not part of the concept of a tree item.

A solution could be to set `aria-selected=undefined` or `aria-selected=false` to all AccordionMenu items. What do you think ?

Owlbertz (dismissed):
I agree that in our case `aria-multiselectable` is wrong as only one item - the one currently having focus - can be selected at once.

ncoden:
I found more informations there: https://www.levelaccess.com/how-not-to-misuse-aria-states-properties-and-roles/

> You might be tempted to put aria-selected on links to simulate toggle buttons or to indicate that a checkbox is checked. These are incorrect uses of aria-selected –  a checked and selected state are not the same. Use aria-pressed for toggle buttons – that is the only place where aria-pressed can be used.
>
> Selection indication is most often used **when a control can have multiple items selected such as options in a listbox or rows in a grid**, or to indicate that **something is active but not focused** such as a page tab.

I think I got it. So it indicated a "selected" state despite of focus, and make sense for tabs because the tab button is "active" even when not focused. For our menu, things are expandable but not "active" and an expanded menu does not represent any value (like a list of selected options would).

Also according to the [ARIA 1.1 Spec](https://www.w3.org/TR/wai-aria-1.1/#requiredState):

> States and properties specifically required for the role and subclass roles. Content authors MUST provide a non-empty value for required states and properties. Content authors MUST NOT use the value undefined for required states and properties, unless undefined is an explicitly-supported value of that state or property.

So I'll set `[aria-selected=false]` to all 'treeitem'.

ncoden:
@Owlbertz I made some change. Please tell me if everything is okay before I merge it 😉

Owlbertz:
As stated above I do not see any necessity to set this as we only have the focused item as selected.

ncoden:
Okay, I actually got things wrong:

> (...) unless undefined is an explicitly-supported value of that state or property.

`undefined` is an explicitly-supported value for `aria-selected` and means that the item is not selectable. Which is our case of our treeitems. Se we should use `[aria-selected=undefined]` instead 😞

ncoden:
Actually.... we may have used the wrong role there:

> #### [tree]()
> A widget that allows the user to **select** one or more items from a hierarchically organized collection.

So a "tree" according to the ARIA definition is not only a recursive list of items, but a recursive list of _selectable_ items. Which is why `aria-selected` is required by the spec itself (and not only recommended for `aria-multiselectable` trees)

Instead, we should use:

> #### [menu](https://www.w3.org/TR/wai-aria-1.2/#menu)
> A type of widget that offers a list of choices to the user.
>
> (...)
>
> #### [menuitem](https://www.w3.org/TR/wai-aria-1.2/#menuitem)
> An option in a set of choices contained by a menu or menubar.

I found the following to support this idea:
- The [W3C example](https://www.w3.org/TR/wai-aria-practices-1.1/examples/menubar/menubar-1/menubar-1.html) for a menu bar use nested `ul[role=menu]` and `li[role=menuitem]` with the same DOM structure than our AccordionMenu component.
- AccordionMenu fit perfectly into the very restrictive definitions given by the two following articles explaining where `[role=menu]` should (and should not) be used.

- http://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html
- https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/

@Owlbertz What do you think ?

Owlbertz:
I remember that we switched the role from `menu` to `tree` a few months ago via a PR because it seemed that (at least for the Drilldown) the tree-role is more fitting because it reflects the hierarchical structure better - which I agree with. Based on the definition of `tree` that you provided (especially the part with selection) I tend to think that we might have to roll back to `menu` as this would be more appropriate.

ncoden:
@Owlbertz As described in the two articles given above, most developers tends to misunderstand and misuse roles and others aria attributes.

I found the following, talking about a menu structure like
```html
<ul role="menu">
  <li>
    <a role="menuitem">
       (...)
```

> We want the user to know that they are using a link and can expect link behavior, so this is no good. Like I said, true menus are for (JavaScript driven) application behavior.
>
> What we’re left with is a kind of hybrid component, which isn’t quite a true menu but at least tells users whether the list of links is open, thanks to the aria-expanded state. This is a perfectly satisfactory pattern for navigation menus.

But the W3C seems fine with this model, and if the implicit semantics is not perfect (`li` are implicitely perceved as list items), the explict attributes (`role`) seems to be the more suitable for this kind of component.

mcking65:
When choosing the pattern to use, one of the most important considerations is the keyboard interaction model. If you are not implementing a menu-like or tree-like keyboard interaction model, then do not use those roles and their associated patterns as described in the ARIA Authoring Practices. If you are implementing one of those keyboard interaction models, then please use the appropriate pattern so that screen reader users will be aware of the interactions that are supported. Also, please provide visual design that helps sighted users be aware of the keyboard interactions that are available.

The primary advantage of composite widgets like menu, tree, grid, treegrid, and toolbar are that they collapse what might ordinarily be bazillions of tab stops into one tab stop and tell the user, "Hey, use the arrow keys to get around here". Which one to use depends on the type of keyboard interaction model you want and to some degree the appearance you want.

Menus and trees are good for accessible hierarchical navigation structures that are repeated consistently on each page included in the navigation system. As demonstrated in the [navigation menu example](http://w3c.github.io/aria-practices/examples/menubar/menubar-1/menubar-1.html), menuitems can behave like links. They are read by screen readers as menuitems, not links, but menus are commonly used for navigation in all desktop apps. While opening another web page is an action that can be launched from a menu, it is not so good for an item in a menu to both work as a parent menuitem and as a link. It can be done, but is problematic.

If you do not plan to condense the navigation system into a single tab stop, you can condense it into a few with a series of elements that show/hide navigation links. In that case, consider using the [disclosure pattern](http://w3c.github.io/aria-practices/examples/grid/LayoutGrids.html). A series of disclosures is not so nice for a multi-level hierarchical navigation system, but works perfectly well for a flat structure.

Similar to a series of disclosure buttons is the [accordion pattern](http://w3c.github.io/aria-practices/#accordion), which can also be used for navigation where each panel in the accordion is just a list of links that the user can tab through. The accordion pattern in the Authoring Practices is currently a set of forms, but it could be very easily repurposed to work for a navigation structure.

Another good option for condensing the tab sequence is a grid. For example, a disclosure or accordion header could show a grid. Grid is better than a standard list for a set of links because it is a single tab stop so the user does not have to tab through every link in the list. Grid also has the advantage that a gridcell does not suppress the link semantics. A menuitem can work like a link, but it is a menuitem. A treeitem can work like a link, but it is announced by screen readers as a treeitem. A gridcell, on the other hand, is a container. So, you place focus on the link inside the gridcell and the link is announced as a link. And, the grid provides the arrow key navigation. You may want to checkout the first example on the [layout grids](http://w3c.github.io/aria-practices/examples/grid/LayoutGrids.html) example page.

Your decision primarily boils down to whether your navigation structure will be a single composite widget like a menu or tree that occupies only one tab stop in the page, making it very easy for keyboard users to skip by the entire thing. Or, is it a series of things that the user tabs through like a set of disclosures or an accordion.

ncoden:
@mcking65 

Thank you for these informations. I am not sure I understood everything, but I found them very useful.

The component we are working on is the [Accordion Menu](https://foundation.zurb.com/sites/docs/accordion-menu.html). It is a list of links that can be expanded/collapsed, nested, and supports keyboard navigation to navigate within. It is intended to be used as a vertical side menu, but I don't know what we should expect regarding to the tab stop. I am not an user of assistive technologies myself and I don't know what would be the the more comfortable for our users, between "single" and a "multiple" tab stop behaviors for this type of component. Your opinion would be welcome.

Beside that, your description of each role is quite helpful. The `tree` pattern causes us some issues for the reasons given above. I planned to switch to `menu` but from your presentation `grid` seems at least as suitable as the `menu`, and without the downsides of the link not presented as a link (like explained in the section "The menu and menuitem roles" of the [SmashingMagazine article](https://www.smashingmagazine.com/2017/11/building-accessible-menu-systems/)).

Thank you again for your help. We try to do our best at Foundation to make our component following the best accessibility recommendations. However, we are not Accessibility experts and do not use screen readers on a regular basis so, despite our long researches in the W3C documentation, we do not always know how to do things correctly. Would you mind if we poke you time to time in future (interesting I hope) accessibility issues ?

jimmynotjim:
Is the accordion menu essentially a collapsed section of content? If you follow this example, it would work pretty much right out of the gate, there's no need for extra roles when `aria-expanded` shows the state of the content and `tab` would navigate through the links as always. https://inclusive-components.design/collapsible-sections/.

Here's a quick demo in CodePen https://codepen.io/jimmynotjim/pen/YjmEQE?editors=1000

jimmynotjim:
Not sure if it's helpful, but I needed to improve the accordion nav at work to fix some issues we are seeing when navigating via VoiceOver. I took the previous quick demo and improved it a bit. 

This demo uses native JS but could easily be updated for jQuery querying and animations. It has the ability to toggle the accordions or let them operate independently, and includes a fallback for when JS isn't available or fails for some reason.

https://codepen.io/jimmynotjim/full/aaBgMm/

@ncoden I'm still new to how Foundation is managed, is it helpful to PR something like this? I know it's sorta in this semi-private/semi-public sphere and don't want to step on any toes but I could probably have it ported to work with Foundation if you'd like.

kball:
Since this is marked still as WIP from @ncoden and AFAIK he is no longer involved, going to close this out.

mlangone:
However this is still causing siteimprove failures. accordion-menu is still showing as tree/tree-item. Has all development stopped on foundation? I would assume with the onslaught of lawsuits surrounding accessibility there has to be a fix for this. Please anyone have any direction on how to either revert to menu menuitem or fix the aria requirements for tree treeitem

DanielRuf:
> However this is still causing siteimprove failures. accordion-menu is still showing as tree/tree-item. Has all development stopped on foundation? I would assume with the onslaught of lawsuits surrounding accessibility there has to be a fix for this. Please anyone have any direction on how to either revert to menu menuitem or fix the aria requirements for tree treeitem

Hi, we are currently a rather small team of people who work on Foundation in their free time.

If you want you can provide a new PR and I will review it.

Issue #11401, opened 2018-07-12
Importing Foundation component as ES6 module includes the whole code and not only the code which is necessary for the component


marvinhuebner:
Currently I'm updating our boilerplate project to 6.5.0-rc2. As i've read in #10864 it should now be possible to include a JavaScript component with this way:

```
import $ from 'jquery';
import { Reveal } from 'foundation-sites';

$(document).foundation();
```

## What should happen?

Only the code which is necessary for the Reveal should be included in the bundle.

## What happens instead?

The whole Foundation JavaScript gets included in the bundle. 

## Possible Solution

Use the "old" solution, as described here: https://github.com/zurb/foundation-zurb-template/blob/master/src/assets/js/lib/foundation-explicit-pieces.js

## Test Case and/or Steps to Reproduce

Test Case:

_See code example above, or the example:_

[![Edit #11401 Example](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/8y19xzqr89?module=%2Fsrc%2Findex.js)

## Your Environment

- Foundation version(s) used: 6.5.0.rc.2
- Browser(s) name and version(s): Chrome
- Operating System and version (desktop or mobile): MacOS

## Checklist (all required):

- [X] I have read and follow the CONTRIBUTING.md document.
- [X] There are no other issues similar to this one.
- [X] The issue title is descriptive.
- [ ] The template is correctly filled.

marvinhuebner:
I've added an example here:

[![Edit #11401 Example](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/8y19xzqr89?module=%2Fsrc%2Findex.js)

mattbag:
https://codesandbox.io/s/5w114vn64k
Looking at the zurb template I usually use, this is what they do.
It's in js/lib/ docs need to be more explicit

ncoden:
Hello @marvinhuebner.

I seen your issue some days ago and planned to work on it this weekend. I watched at [zurb/foundation-zurb-template:src/assets/js/lib/foundation-explicit-pieces.js](https://github.com/zurb/foundation-zurb-template/blob/master/src/assets/js/lib/foundation-explicit-pieces.js) and don't see how this would resolve the issue.

I'll make some tests to see how treeshaking is handled by most used builders.

ncoden:
[...]

PR #11445, opened 2018-08-12
chore: improve "tree shaking" support and documentation

ncoden:
## Description

This pull request change some Foundation entries so they better follow the recommended practices and should be better analyzed by bundlers in order to enable tree-shaking. Also, it adds a section in the "Javascript" documentation with advices to reduce the final bundler size.

### Changes
* Add a simpler export file `js/foundation.js`
* Use `js/foundation.js` instead of `js/entries/foundation.js`
* Add default export as it may be sometimes required for tree-shaking
* Add "Tree Shaking" section in javascript documentation

### Others changes
* Add missing export of `Move` in "foundation-plugins" entry

- Related to https://github.com/zurb/foundation-sites/issues/11401

## Types of changes

- [x] Documentation
      functionality to change)

## Checklist (all required):


DanielRuf (commented):
Not sure if we really need two exports as only the default export is needed (afaik).

DaSchTour:
Well I think using default export is bad practice, because it allows doing very creepy things and also breaks checks for instanceof.
Here are just some other ideas why default export never were a good idea
https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad
https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

Another thing to consider is, that named exports also help with not breaking the API. If you have a default export every change to this export is possibly a breaking change.

So just never ever use default exports.

DanielRuf:
@DaSchTour I'm aware of this but it seems even Twitter Bootstrap (and many others) use `export default`, see https://github.com/twbs/bootstrap/blob/v4-dev/js/src/carousel.js#L520

ncoden:
It is not recommended to use default **imports**. But I remember that in some environnements they may be the only possible solution, and seems to work better with tree shaking. I'll try to find back the articles explaining this.

SassNinja:
@DaSchTour default export is not bad imo – if used correctly.

If someone misuses it to bundle several export in one default export it's of course bad practice and interferes with tree shaking. So you shouldn't do

**`utils.js`**
```javascript
function triple(val) {
  return val * 3;
}
function double(val) {
  return val * 2;
}
export default {
  triple: triple,
  double: double
};
```
**`app.js`**
```javascript
import utils from './utils';
console.log(utils.triple(10));
```

but instead do

**`utils.js`**
```javascript
export function triple(val) {
  return val * 3;
}
export function double(val) {
  return val * 2;
}
```
**`app.js`**
```javascript
import { triple } from './utils';
console.log(triple(10));
```

In contrast if you've a module that has just one export because it doesn't make sense to import partials of it then it's absolutely ok in my eyes to use export default. This happens if there's just one main function or class.

**`super-carousel.js`**
```javascript
class SuperCarousel {
  constructor() {
    console.log(`I'm a super carousel!`);
  }
}
export default SuperButton;
```
**`app.js`**
```javascript
import SuperCarousel from './super-carousel';
const carousel = new SuperCarousel();
```

docs/pages/javascript.md
```diff
@@ -63,6 +63,19 @@ var Foundation = require('foundation-sites');
...
 ```
 
+#### Tree Shaking
+
+Many bundlers like Webpack, Rollup or Parcel support tree shaking. It consists of the removing the unused code parts from your codebase or your dependencies. Take a look at these articles to know how it works and how to enable it: [How To Clean Up Your JavaScript Build With Tree Shaking](https://www.engineyard.com/blog/tree-shaking), [Why Webpack 2's Tree Shaking is not as effective as you think](https://advancedweb.hu/2017/02/07/treeshaking/) and [Reduce JavaScript Payloads with Tree Shaking](https://developers.google.com/web/fundamentals/performance/optimizing-javascript/tree-shaking/).
+
+Sadly, the "static analysis" promised by these bundlers to detect unused code in dependencies does not usually work, so we recommend you to manually import the Foundation plugins instead, like the following:
+
+```js 
+import Dropdown from 'foundation-sites/js/foundation.dropdown';
```
SassNinja: @ncoden I don't like this way of using foundation...

Is there a way to enable something more user friendly (while still being tree shakable)? Such as
```javascript
import { Reveal } from 'foundation-sites';
```

SassNinja (changes requested):

docs/pages/javascript.md
marvinhuebner: @SassNinja i've also tested this pr (see [here](https://github.com/zurb/foundation-sites/issues/11401#issuecomment-412483242)). Would also prefer if there is a way when it's enough to do something like 
```js
import { Reveal } from 'foundation-sites';

$(document).foundation();
```
But it seems that this is not working, not with the current implementation.

Currently i have to do this to get it working: 
```js
import Foundation from 'foundation-sites/js/foundation';
import Reveal from 'foundation-sites/js/foundation.reveal';

var test = new Foundation.Reveal($('.reveal'));
```

DanielRuf: Foundation needs more than just the plugin to work properly, if you mean that.

ncoden: @SassNinja Sadly, as explained in the paragraph just above, the following _should work in theory_...
```js
import { Reveal } from 'foundation-sites';
```

...but **in practice** does not, as "pure modules" are not well detected by bundlers, and a direct import is recommended. For example the `lodash` library faced the same problem and they created [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) to convert things like `import map from 'lodash'` to `import _map from 'lodash/map'`.

So at best it depends on the user stack to choose whether he can use the named import and try to keep tree shaking working.

ncoden: Poke @jdalton, It would be nice to get your opinon on this 😉.

ncoden: @marvinhuebner I prefer the `import { Reveal } from 'foundation-sites';` approach too. But I just want to make something that works 😉

ncoden:
@SassNinja I removed "grouped" named exports:

> Default exports should only be used for "primary" objects or classes, not for libraries for which individual named exports should be prefered.

Also, I kept named exports at the end of the files. See https://github.com/airbnb/javascript/issues/1275#issuecomment-274367523

ncoden:
@DaSchTour IMHO default imports are not a good import practice, but it is not a bad export one. There is many situation where default imports is the only solution, like when trying to simply transition from ES5 to ES6 (so from CommonJS to ESM modules) or when module imports are handled and/or transpiled by the framework's CLI (like for Ember).

docs/pages/javascript.md
jdalton: For side-effect free ability: `lodash-es` is detected as pure by webpack v4 and parcel.js with the `lodash-es`'s use of the package.json field `"sideEffects": false,`.

ncoden: I tested with `"sideEffects": false,` and followed all the ressources I found on tree shaking. I am unable to make this work with imports from `foundation-sites`.

Take a look at https://github.com/ncoden/foundation-sites-tree-shaking. Please tell me if I got something wrong. The project rely on this branch `chore/tree-shaking`, but locally it's linked to my local `foundation-sites` clone.

jdalton: `sideEffects: false` is a per package setting not an application setting. This is why it's set in `lodash-es` because I, as the package author, am stating that its modules are side-effect free.

DaSchTour:
@ncoden I know this is a constructed scenario. But one thing that can happen with default exports, for example because of somebody making a mistake while resolving merge conflicts, you end up with code like this
```js
import Reveal from 'foundation-sites/js/foundation.dropdown';
import Dropdown from 'foundation-sites/js/foundation.reveal';
```
This is not an error and with JS you might search for hours to find the reason why everything is messed up.

docs/pages/javascript.md
ncoden: > `sideEffects: false` is a per package setting

Yes I know. I set `sideEffects: false` locally in `foundation-sites`

> but locally it's linked to my local foundation-sites clone with sideEffects set.

ncoden:
@DaSchTour Yes. Default imports should be avoided when possible for this reason. But we should still propose default import for the reasons given above, even if we do not _encourage_ it.

docs/pages/javascript.md
DaSchTour: I'm curious. As Foundation uses jQuery and extends it and also relies on extending global Foundation object I'm not really sure if Foundation can be considered sideEffects free.

DaSchTour:
@ncoden I don't see any reason why default imports should be possible. In general I wonder why put so much effort into making foundation 6 tree shakable or why put any effort into foundation 6 at all while everybody is waiting for 7.

docs/pages/javascript.md
ncoden: Foundation plugins can be considered as without side effects, as they must be imported and initialized on a component to work. It's the same for the Foundation Core that is manually added to jQuery `addTojQuery()`. I guess the "no side effects" only means: do not rely on others components without importing them. But Foundation core and plugins import all their dependencies.

ncoden:
> why put so much effort into making foundation 6 tree shakable

Because fixing and cleaning the ESM+AMD+CJS+static imports (which is a big "feature" of the v6.5.0 release) goes with providing efficient ways to use them in each of these environnement. And IMHO this means allowing to import single plugins without importing the whole framework (which is unexpected when you have the ability to cherry-pick feature).

>  why put any effort into foundation 6 at all while everybody is waiting for 7

Good question. Personally, I'd like to get Foundation 6 _stable_ before we take 6-12  months to prepare and work on Foundation 7. But I recognize that this step took way much time than expected. I am sorry about that.

SassNinja:
> Good question. Personally, I'd like to get Foundation 6 stable before we take 6-12 months to prepare and work on Foundation 7. But I recognize that this step took way much time than expected. I am sorry about that. – @ncoden 

I don't think it's wrong to enable using Foundation v6 with module bundler the way users expect (import only certain plugins).

It may seem a bit waste of time regarding v7 but I think it's more important to ensure a good usability of the existing version. If users get pissed of using Foundation they may not wait until v7...

docs/pages/javascript.md
ncoden: @jdalton @DaSchTour Would you know what I am missing here (https://github.com/ncoden/foundation-sites-tree-shaking) ? I don't see what I could configure make tree shaking working with imports from `foundation-sites`.

DaSchTour:
@SassNinja well I guess some users already won't be waiting for v7. It's called Osborne effect.

ncoden:
@DaSchTour @SassNinja Could we not enter into meta-discussions about the global Foundation roadmap and focus on the current issue ? Time matters, everyone agree on that, but we're not saving any with this kind of discussion. For now I'd like to get this PR approved/rejected to release RC3.

docs/pages/javascript.md
ncoden: @SassNinja @DaSchTour  So what's your opinon on this ?

DaSchTour: @ncoden I have no idea. I never had any problems with tree shaking but I also haven't used or dependend on any globals for a long time.

ncoden:
@SassNinja 

I adapted the message so direct imports are only recommended if tree-shaking doesn't work, "depending on your build environment". Are you good with this ?

@jdalton 

I added the `sideEffects` flag, I think this is the best I can do.

SassNinja:
> I adapted the message so direct imports are only recommended if tree-shaking doesn't work, "depending on your build environment". Are you good with this ? – @ncoden 

Not really :/

But I've now set up a simple webpack config using your feature branch and tree shaking did work without problems. The only downside is all plugins still appear in the stats but the emitted JS code looks good.

```javascript
// webpack.example.js

const path = require('path');

module.exports = {
    mode: 'production',
    devtool: 'none',
    entry: {
        'example': './example.js'
    },
    output: {
        path: path.resolve(__dirname, 'dist-example'),
        filename: '[name].js'
    },
    externals: {
        jquery: 'jQuery'
    },
    optimization: {
        minimize: false,
        runtimeChunk: 'single'
    }
};
```

```javascript
// example.js

import { Abide } from './js/foundation.js';

console.log(typeof Abide);
```

How did your setup look where tree shaking isn't working?

ncoden:
@SassNinja See https://github.com/ncoden/foundation-sites-tree-shaking

ncoden:
@SassNinja I cannot make tree shaking work with your configuration. Could you share your dummy project in a repo ?

SassNinja:
> I cannot make tree shaking working with your webpack configuration. Could you share your dummy project in a repo ?

Here's an example repo:
https://github.com/SassNinja/example-foundation-sites-tree-shaking

I think I know what went wrong on your side because I also faced an issue when creating that example repo. To work the bundler needs to take the `module` field in the `package.json` what webpack does not do by default.

So you need to override it like this
```javascript
resolve: {
    mainFields: ['module', 'browser', 'main']
}
```

SassNinja:
> Thank you. Too bad that the webpack documentation about tree shaking doesn't mention it.

Maybe we should adjust our package.json to fix this without the need for additional bundler config. The `browser` field looks promising to me.

See https://github.com/webpack/webpack/issues/4674#issuecomment-355853969

What do you think?

SassNinja:
I've simply replaced the field name `module` with `browser`.
Afterwards it's working for me without further bundler config (tested webpack).

SassNinja:
A radical idea: what about removing the whole `entries` folder?

I think if you use a module bundler (and therefore probably want tree shake) you have to include the desired plugins via
```javascript
import { Abide, OffCanvas } from 'foundation-sites';
```
We can provide a list with all plugins imports for copy & paste.

And people who wanna import the whole framework without a module bundler (e.g. when using gulp without any bundler in the task) may use the dist file directly.

We could also cover the dist case with an export default but I don't see any useful reason to offer something like

```javascript
import FoundationEverything from '../dist/js/foundation.js';
export default FoundationEverything;
```

ncoden:
> I've simply replaced the field name module with browser.
> Afterwards it's working for me without further bundler config (tested webpack).

Thank you, I'll test this tonigh and update the pull request.

> A radical idea: what about removing the whole `entries` folder?

Hmmm, this would break backward compatibility so it's definitely not for this pull request. For v6.6.0, maybe if it enables anything useful. But for now I don't see any advantages for us or our users.

ncoden:
I removed the `browser` field it and it works well. According to the [specs](https://github.com/defunctzombie/package-browser-field-spec) there is no reason for us to use a `browser` field as the package itself it only for browsers (or any environement with a DOM). With this configuration, ESM-compliant transpilers/bundlers will pick `module`, and the others will pick `main`.

There is still one problem too: the "picked" code is not only a ESM bundle, but also ES6 code. Webpack will use `module` because it's an ESM bundle but will not transpile the ES6 stuff...

So I think we should also provide our distribution files in the `.esm.js` format: a transpiled file BUT with ESM exports untouched, and use it for the `module` field.

See for example the [Vue dist folder](https://github.com/vuejs/vue/tree/dev/dist): `vue.js` (UMD), `vue.common.js` (commonJs) and `vue.esm.js` (ESM)

What do you think ?

SassNinja:
> According to the specs there is no reason for us to use a browser field as the package itself it only for browsers (or any environement with a DOM).

Alright then let's ditch `browser` and only use `main` and `module`

> Webpack will use module because it's an ESM bundle but will not transpile the ES6 stuff...

I don't understand this – what do you mean?

The code gets picked up by the bundler which understands ES6 and resolves all dependencies. If necessary it can be transpiled to ES5 using e.g. the babel-loader.

> See for example the Vue dist folder: vue.js (UMD), vue.common.js (commonJs) and vue.esm.js (ESM)

I think a transpiled ESM file is nice to speed the build process up as there's no need then to transpile node_modules yourself (while preserving tree shake support).
However we should also offer an untouched file imo (see `vue.esm.browser.js`) for people who don't need any pre transpilation.

ncoden:
> However we should also offer an untouched file imo (see vue.esm.browser.js) for people who don't need any pre transpilation.

You are right. I'm going for the following: `foundation.js` (UMD), `founcation.cjs.js` (commonJs), `foundation.esm.js` (transpiled ESM module), `foundation.es6.js` (untouched ESM module).

DaSchTour:
What is the difference between transpiled ESM module and untouched ESM module? Transpiling ESM module to let's say ES5 would make it "remove" modules because ES5 doesn't have modules. And I think any kind of processing might possibly hurt treeshaking and also minification.

SassNinja:
> What is the difference between transpiled ESM module and untouched ESM module? Transpiling ESM module to let's say ES5 would make it "remove" modules because ES5 doesn't have modules. – @DaSchTour 

That's why I wrote `foundation.esm.js` would still contain ES6 code, precisely the import/export stuff.
But compared to `foundation.es6.js` (@ncoden maybe we should rename it to avoid confusion?) it's compiled to ES5 except of all import/export.

> And I think any kind of processing might possibly hurt treeshaking – @DaSchTour 

afaik webpack requires ES6 import/export for tree shaking. If we preserve that on transpilation it shouldn't be a problem and improves the build time because users can exclude foundation from their own ES5 transpilation then.
Anyway this has to be tested of course (haven't done this yet).

ncoden:
Some bundlers (webpack, rollup...) supports the ES6 module format and require to use it for tree shaking but will leave the other ES6 stuff intact (they are bundlers, not transpilers). `foundation.esm.js` is transpiled ES5 code embeded in ESM modules for those bundlers. `foundation.es6.js` is untouched ES6 code and modules that can be manually transpiled by our users depending on their targets.

ncoden:
@SassNinja So I made (a lot of) tests:
- ESM6+ bundles (es6+ in es modules) are correctly shaked by both Webpack and Rollup.
- ESM5 bundles (transpilled es5 in es modules) are not shaked, despites being marked as `__PURE__` by Rollup. This comes from static properties defined outside of the class like `<plugin>.defaults = { ... }`, and plugin initializations in entry files.

There is no other way to define static properties for now. There is a [TC39 proposal](https://github.com/tc39/proposal-class-fields) for this but the [Babel plugin](https://babeljs.io/docs/en/next/babel-plugin-proposal-class-properties.html) generate code outside of the class closure, making it [incompatible with tree sharking](https://github.com/babel/babel/issues/5902).

For now I think we can go with only an ES6+ version directly from the sources (so Webpack/Rollup users will have to transpile the code), and I'll do some refactor later to allow us to provide shakable transpiled bundles.

I removed `pkg.browser`, and improved the "tree shaking" section in the installation documentation to promote named imports over "manual" imports. What do you think ?

SassNinja:
> There is no other way to define static properties for now. There is a TC39 proposal for this but the Babel plugin generate code outside of the class closure, making it incompatible with tree sharking.
>
> For now I think we can stay with the ES6 sources in pkg.modules (so Webpack/Rollup users will have to transpile the code), and I'll do some refactor later to allow us to provide shakable transpiled bundles.

Ok, then let's stay with untranspiled ES6 code in the module field.
Would be of course much better if users don't need to transpile node modules themselves but does not make sense if tree shaking doesn't work correctly anymore.

> I removed pkg.browser, and improved the "tree shaking" section in the installation documentation to promote named imports over "manual" imports. What do you think ?

Docs lgtm :+1: 

The only thing I still doesn't like is the usage of named **and** default export in the plugins that export the same.
I've only tested webpack and there it also works without default export.
But you've also tested other bundlers and noticed it's necessary to have default exports, haven't you?

SassNinja (approved):

ncoden:
@SassNinja Thank you for the review :)

> But you've also tested other bundlers and noticed it's necessary to have default exports, haven't you?

I remember that yes, but I'll check again just to be sure.

ncoden:
I tested, and issues actually come fron side-effects in the "fully-initialized" browser entry `js/entries/foundation.js`. But we can't move it from there as this is how Foundation is used: `import { Foundation } from "foundation-sites"; // already initialized`

PR #11451, opened 2018-08-14
fix: prevent sasslint warning when using qualifying element with class

DanielRuf:
## Description

Fixes a warning due to a qualifying element with a classname in the dropdown-menu component.

- Closes 

## Motivation and Context

Fixes a warning due to a qualifying element with a classname in the dropdown-menu component.

## Types of changes

- [ ] Documentation
      functionality to change)

## Checklist (all required):

### DIFF
--- a/.sass-lint.yml
+++ b/.sass-lint.yml
@@ -114,7 +114,7 @@ rules:
   no-qualifying-elements:
     - 1
     - allow-element-with-attribute: false
-      allow-element-with-class: false
+      allow-element-with-class: true
       allow-element-with-id: false
 
   no-trailing-zero: 1
--- a/scss/components/_dropdown-menu.scss
+++ b/scss/components/_dropdown-menu.scss
@@ -152,8 +152,8 @@ $dropdown-menu-item-background-active: transparent !default;
 
     // Top-level item
     > li > a {
-      background: $dropdownmenu-background;
       padding: $dropdownmenu-padding;
+      background: $dropdownmenu-background;
     }
 
     // Top-level item active state


.sass-lint.yml
```diff
@@ -114,7 +114,7 @@ rules:
   no-qualifying-elements:
     - 1
     - allow-element-with-attribute: false
-      allow-element-with-class: false
+      allow-element-with-class: true
```
ncoden: This is a bad practice and should stay at `false`. In cases like this one (when we know this is bad but, meh, there is no other solution), we should use `sass-lint:disable` instead.

ncoden (changes requested):
Take a look at `gulp lint` 😜

.sass-lint.yml
DanielRuf: > we should use sass-lint:disable

You were against setting this per line in files and meant that we should only set the config options in the yml file. I read some issues regarding this rule and many recommend and use this setting.

DanielRuf:
> Take a look at gulp lint 😜

What's the issue with it?

DanielRuf:
> Take a look at gulp lint 😜

It seems these are never shown during `commit` and `push`.

DanielRuf:
It seems we have many new issues which are suppressed during `commit` and `push` and also in `npm run test`. So I guess I'll take another approach and check why it is now different than before. I thought we fixed all these warnings.

.sass-lint.yml
ncoden: I was against using `sass-lint:disable` for `force-element-nesting` and `force-pseudo-nesting` because there can be many case when we want to avoid nesting (like: because we just have a single property to set in a deeply-nested element and don't want to take 10 lines for that). So instead of adding "exceptions" everything, let's change the rule because the rule is bad.

For `allow-element-with-class` I think it is very useful to force us to take care of the specificity. It does not depends on the way we want to organize our properties to get the best readibility but on the selector itself, the CSS that is exposed to the user and the whole architecture we (badly) try to follow: like avoiding heavy selectors, magical properties applied from the parent and everything leading to the specificity hell.

Does it makes sense to you ?

PR #11477, opened 2018-08-26
fix: return to initial state when reseting hash in Accordion & Tabs #11100

### LINKED ISSUE #11100: Tabs: Browser back doesn't take us back to default tab

ncoden:
## Description

When deep-linking is enabled and the user goes back in the history up to when no hash is set, the active panel or tab does not change and is still the one that was selected the first.

This pull request make Accordion and Tabs displaying the panel or tab that was active at the component initialization.

### Changes:
- Save up the active panel/tab at Accordion/Tabs initialization (`_initialAnchor`)
- Display the initial panel/tab when the hash change to `""`.

- Closes https://github.com/zurb/foundation-sites/issues/11100
- Replaces [#11101 - feat: add class for active tab when no hash is set in Tabs ](https://github.com/zurb/foundation-sites/issues/11101) (@jamiechong)

## Types of changes

- [ ] Documentation

## Checklist

- ~~I have updated the documentation accordingly to my changes (if relevant).~~

### DIFF
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -118,6 +118,10 @@ class Tabs extends Plugin {
       if ($anchor && $anchor.length && $link && $link.length) {
         this.selectTab($anchor, true);
       }
+      // Otherwise, collapse everything
+      else {
+        this._collapse();
+      }
 
       // Roll up a little to show the titles
       if (this.options.deepLinkSmudge) {
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -73,6 +73,11 @@ class Tabs extends Plugin {
         'aria-labelledby': linkId
       });
 
+      // Save up the initial hash to return to it later when going back in history
+      if (isActive) {
+        _this._initialAnchor = `#${hash}`;
+      }
+
       if(!isActive) {
         $tabContent.attr('aria-hidden', 'true');
       }
@@ -85,6 +90,7 @@ class Tabs extends Plugin {
         });
       }
     });
+
     if(this.options.matchHeight) {
       var $images = this.$tabContent.find('img');
 
@@ -98,8 +104,14 @@ class Tabs extends Plugin {
      //current context-bound function to open tabs on page load or history hashchange
     this._checkDeepLink = () => {
       var anchor = window.location.hash;
+
+      // if there is no anchor, return to the initial tab
+      if (!anchor.length && this._initialAnchor) {
+        anchor = this._initialAnchor;
+      }
+
       //need a hash and a relevant anchor in this tabset
-      if(anchor.length) {
+      if (anchor.length) {
         var anchorNoHash = (anchor.indexOf('#') >= 0 ? anchor.slice(1) : anchor);
         var $link = this.$element.find(`[href$="${anchor}"],[data-tabs-target="${anchorNoHash}"]`).first();
         if ($link.length) {
--- a/js/foundation.tabs.js
+++ b/js/foundation.tabs.js
@@ -234,18 +234,10 @@ class Tabs extends Plugin {
    */
   _handleTabChange($target, historyHandled) {
 
-    /**
-     * Check for active class on target. Collapse if exists.
-     */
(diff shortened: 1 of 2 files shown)


SassNinja (changes requested):
@ncoden I've tested the accordion and it works – but only if an item is initially active!

If no item has initially the class `is-active` is fails.
I think we should cover that case, too.

js/foundation.tabs.js
```diff
@@ -98,8 +104,14 @@ class Tabs extends Plugin {
...
+      if (!anchor.length && this._initialAnchor) {
+        anchor = this._initialAnchor;
+      }
+
       //need a hash and a relevant anchor in this tabset
-      if(anchor.length) {
+      if (anchor.length) {
         var anchorNoHash = (anchor.indexOf('#') >= 0 ? anchor.slice(1) : anchor);
         var $link = this.$element.find(`[href$="${anchor}"],[data-tabs-target="${anchorNoHash}"]`).first();
         if ($link.length) {
```
SassNinja: In the course of testing the tabs I noticed the focus style remains when using history back.

tabs_deep_linking_480px_low

What do you think about auto focusing the link to fix this?

```javascript
if ($link.length) {
  $link.focus();
  this.selectTab($(anchor), true);
```

SassNinja (changes requested):

js/foundation.tabs.js
ncoden: This is an other issue, not related to this pull request. Why not, but I wonder what's the impact on accessibility to move the focus like this.

ncoden:
@SassNinja I refactored the component to clean up a bit the logic and split the options processing/component logic from simple DOM manipulations. It should resolve the first bug you pointed out.

SassNinja:
> Could you review this again please ? :)

It's working better now: history back adjust the adressbar as expected.

But the active state is not working as expected:
when I start without an initially opened tab, click on a tab and go back in history, I don't get the original active state (none tab opened). Instead the previously clicked tab is active (although the page url has changed).

Shall I provide an example repo or is it clear what I mean?

ncoden:
> Shall I provide an example repo or is it clear what I mean?

I understood the issue, and through I resolved this issue. If you look at the code, you can see that `_closeAll()` is called if we go back to a state where no link is active.

ncoden:
Actually, I fixed this issue for the Accordion only. :p

SassNinja:
Just to be clear: the accordion component works as expected.
The issues described above only refers to the tabs components.

Here's a quick gist with the content of my `\test\visual\tabs\deep-linking.html`
https://gist.github.com/SassNinja/c194da0830de81a510061c99ffcb8938

ncoden:
> Just to be clear: the accordion component works as expected.
> The issues described above only refers to the tabs components.

Yeah I got it. I need to do the same refactor in Tabs as I did in Accordion to resolve it. I'll take care of this sunday.

ncoden:
@SassNinja Done.

ncoden:
@SassNinja 👋 Could you test it again?

SassNinja:
@ncoden lgtm 👍 

Just two things don't look nice
- focus state doesn't get reseted on history back (as mentioned earlier)
- the hash tag in the browser url doesn't get removed when I toggle a tab, so that no panel is opened (when using `data-active-collapse="true"`)

But nothing related to this PR – so good to merge imo.

SassNinja (approved):

PR #11478, opened 2018-08-28
Set colors if values are not already set

timgremore:
## Description

SASS supports conditional variable assignment with the `!default` flag. Adding `!default` enables developers to assign project-specific values without editing foundation `settings.scss`.

[SASS variable defaults](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#variable_defaults_default)

## Types of changes

## Checklist

### DIFF
--- a/scss/settings/_settings.scss
+++ b/scss/settings/_settings.scss
@@ -74,12 +74,12 @@ $foundation-palette: (
   success: #3adb76,
   warning: #ffae00,
   alert: #cc4b37,
-);
-$light-gray: #e6e6e6;
-$medium-gray: #cacaca;
-$dark-gray: #8a8a8a;
-$black: #0a0a0a;
-$white: #fefefe;
+) !default;
+$light-gray: #e6e6e6 !default;
+$medium-gray: #cacaca !default;
+$dark-gray: #8a8a8a !default;
+$black: #0a0a0a !default;
+$white: #fefefe !default;
 $body-background: $white;
 $body-font-color: $black;
 $body-font-family: 'Helvetica Neue', Helvetica, Roboto, Arial, sans-serif;


SassNinja:
Thank you @timgremore for the PR!

But to be honest I don't see why this is necessary.

The settings file is not meant to be overridden but overrides the sass components (which use `!default`) itself.
It's an optional file for copy & paste to easily adjust the frameworks variables.

timgremore:
@SassNinja I'm trying to avoid editing the settings file directly so future upgrades to Foundation can be made without the need to reconcile my settings with the new version. Generally speaking, maintaining my project settings apart from library settings is a helpful separation of concerns. And since there is no cost to using `!default`, I think it's worth considering.

SassNinja:
> I'm trying to avoid editing the settings file directly

Then just don't use it!
It's just an optional file to make it as easy as possible to customize foundation.

Since your PR adds `!default` to some colors I assume you're overriding those colors in your own sass code?
If so just create your own settings file where you only specify / override the desired colors to be as future safe as possible.

It doesn't make sense to use `!default` in a file that is not meant to be overridden by other sass code.

SassNinja:
> The colors defined in settings are referenced in other assignments in the same file 

It doesn't matter when you don't import that settings file yourself.
Foundation does **not** import that file!

If a component uses `$light-gray` it does **not** rely on the settings file but on a global variable (which already uses `!default`)
https://github.com/zurb/foundation-sites/blob/develop/scss/_global.scss#L37

For my projects I always use a settings file because it's easier to adjust some variables if you don't need to look them all up first. Upgrading is not much complicater bcz of it.
But if you prefer to not depend on such a settings file it's also fine.

Again, the settings file you've adjusted in your PR does not affect foundation unless you import it yourself.

timgremore:
@SassNinja Thanks for pointing out `_global.scss` - I had missed that obviously. Thanks again!

PR #11543, opened 2018-10-17
fix: avoid overhead when calling MediaQuery._init() multiple times

### LINKED ISSUE #11540: MediaQuery.queries array seems "overcrowded"
## What is missing?
By resizing window the Foundation.MediaQuery.queries array will be filled with a lot of queries. They vary by changing breakpoints by resizing the window (sometimes more, sometimes less). I was expecting that there are only my defined "css" breakpoints available.

snip_20181017095355

## What is wrong now?
I dont think that there should be an array of 217 queries. The amount should not exceed my amount of css breakpoints.

## Your Environment
- Foundation 6.4.3
- Chrome 69 
- Windows 10

## Checklist
- [X] here are no other issues similar to this one.
- [X] The issue title and template are correctly filled.

SassNinja:
## Description

This makes sure Foundation.MediaQuery.queries doesn't contain several duplicates as it gets called by many plugins.

- Closes https://github.com/zurb/foundation-sites/issues/11540

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -80,6 +80,14 @@ var MediaQuery = {
    * @private
    */
   _init() {
+
+    // make sure the initialization is only done once to avoid overhead when calling _init() several times
+    if (this.isInitialized === true) {
+      return;
+    } else {
+      this.isInitialized = true;
+    }
+
     var self = this;
     var $meta = $('meta.foundation-mq');
     if(!$meta.length){
--- a/js/foundation.util.mediaQuery.js
+++ b/js/foundation.util.mediaQuery.js
@@ -81,7 +81,7 @@ var MediaQuery = {
    */
   _init() {
 
-    // make sure the initialization is only done once to avoid overhead when calling _init() several times
+    // make sure the initialization is only done once when calling _init() several times
     if (this.isInitialized === true) {
       return;
     } else {
@@ -99,6 +99,8 @@ var MediaQuery = {
 
     namedQueries = parseStyleToObject(extractedStyles);
 
+    self.queries = []; // reset
+
     for (var key in namedQueries) {
       if(namedQueries.hasOwnProperty(key)) {
         self.queries.push({
@@ -114,11 +116,12 @@ var MediaQuery = {
   },
 
   /**
-   * Reinitializes the media query helper what might be necessary if your CSS is not (completely) loaded while _init() is executed.
+   * Reinitializes the media query helper.
+   * Useful if your CSS breakpoint configuration has just been loaded or has changed since the initialization.
    * @function
    * @private
    */
-  _reinit() {
+  _reInit() {
     this.isInitialized = false;
     this._init();
   },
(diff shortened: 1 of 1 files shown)


ncoden (commented):
Thank you @SassNinja for the pull request 👍,

What if the CSS configuration is not availaible at the first initialization and appear later? It would not work anymore with a single initialization.

We could fix #11540 by reloading `queries` instead.

What do you think ?

ncoden:
> Usually the CSS should be ready before the JS but I guess it may be a problem if loading CSS via dynamic imports.

_Usually_ 😉 

> So you would prefer to reset the queries before populating it (again)?

Yes, this or adding a method to manually reinitialize MediaQuery. Do you have any objection ?

SassNinja:
> Yes, this or adding a method to manually reinitialize MediaQuery. Do you have any objection?

A) reset

- pro: impossible to break any existing project
- con: not performance optimized (meta elem gets added and `parseStyleToObject`  executed multiple times)

B) _reinit method

- pro: better performance (no unnecessary reinitialization)
- con: might break existing projects if the developer wasn't aware his CSS is loaded too late (he has to fix it himself with the new _reinit method)

I think I'd go with option B because it's more important to be as fast as possible instead of slowing everything down due to rare cases (CSS loaded too late).
We can include this in the release notes so devs know what's wrong when they are facing MediaQuery issues after updating.

Do you agree?

ncoden:
Possible breaking change. Planned for v6.6.0

js/foundation.util.mediaQuery.js
```diff
@@ -105,6 +113,16 @@ var MediaQuery = {
...
   },
 
+  /**
+   * Reinitializes the media query helper what might be necessary if your CSS is not (completely) loaded while _init() is executed.
+   * @function
+   * @private
+   */
+  _reinit() {
+    this.isInitialized = false;
+    this._init();
```
ncoden: This would populate the `queries` array again.
You could remove `queries: []` and set up `queries` in the `_init()` method.

js/foundation.util.mediaQuery.js
```diff
@@ -105,6 +113,16 @@ var MediaQuery = {
     this._watcher();
   },
 
+  /**
+   * Reinitializes the media query helper what might be necessary if your CSS is not (completely) loaded while _init() is executed.
+   * @function
+   * @private
+   */
+  _reinit() {
```
ncoden: `_reInit()` (with an uppercase)

js/foundation.util.mediaQuery.js
```diff
@@ -105,6 +113,16 @@ var MediaQuery = {
     this._watcher();
   },
 
+  /**
+   * Reinitializes the media query helper what might be necessary if your CSS is not (completely) loaded while _init() is executed.
```
ncoden:
```
Reinitializes the media query helper.
Useful if your CSS breakpoint configuration has just been loaded or has changed since the initialization.
```

ncoden (changes requested):

js/foundation.util.mediaQuery.js
SassNinja: done

ncoden (approved):
LGTM! Thank you @SassNinja 😄

aj1991:
Nice one 👍

PR #11592, opened 2018-11-19
fix: use concat return value for Trigger names

DanielRuf:
## Description

It seems the return value of the concat is not used in the Triggers which creates the trigger name. We should check if this is the correct approach as the unit tests do not fail with and without this change.

- Closes 

## Types of changes

- [ ] Documentation

## Checklist


ncoden (approved):
:+1: Thx @DanielRuf

Issue #10684, opened 2017-09-27
Off-Canvas Overlay Issues


KnightYoshi:
So I encountered an issue while using off-canvas, where the overlay when off-canvas is open does not appear.

If you scale from large or medium down to small and click the navigation icon the off-canvas will open, but the overlay used to close it when clicked/tapped is missing. I've tested and was able to reproduce this with consistency in Firefox, FF Dev, and Chrome.
https://codepen.io/KnightYoshi/pen/YrZWoQ/

Here's an animated GIF of what I'm talking about.
![off-canvas](https://user-images.githubusercontent.com/6134576/30889549-17dd127c-a2f5-11e7-82f2-9c40580deaf9.gif)

After I resize the window to be medium/large and reload it to reset the page and then scale it down the the off-canvas overlay is missing.

rafibomb:
Hmm, it seems to be your specific setup. It works a regular off-canvas as expected. https://codepen.io/rafibomb/pen/rGyXaB

The overlay is based on the `<div class="off-canvas-content" data-off-canvas-content>` which appears to be missing

Knight-Yoshi:
No, it's not my set up.
I did forget the `.off-canvas-content` in my example
I added it:
https://codepen.io/KnightYoshi/pen/YrZWoQ 

Still behaves the same.

Knight-Yoshi:
![off-canvas2](https://user-images.githubusercontent.com/6134576/30928443-aff84088-a389-11e7-87d0-03bb8ba0d1e9.gif)

Knight-Yoshi:
And it always happens when sizing down the window to small.

Knight-Yoshi:
This is still an issue that has not been acknowledged.

bonovski:
Encountered the same issue. 
I'm using an off-canvas which is going in-canvas on large.

The overlay isn't triggered when going from large to small, it works OK in the opposite direction.

So if you open the page on desktop size, you won't get the overlay when resizing to small.

DanielRuf:
> Encountered the same issue.
> I'm using an off-canvas which is going in-canvas on large.
> 
> The overlay isn't triggered when going from large to small, it works OK in the opposite direction.
> 
> So if you open the page on desktop size, you won't get the overlay when resizing to small.

Hi, thanks for the report.

Can you provide a working codepen which reproduces this?

KnightYoshi:
[...]

PR #11603, opened 2018-11-21
refactor: check and fails for missing DOM elements

ncoden:
## Description

Throw an error when the DOM elements that are required by components cannot be found, instead of letting the component crashing or going into a weird behavior.

### Changes
- Fail for OffCanvas `[data-off-canvas-content]`
- Fail for Slider `[data-slider-handle]
- Simplify and clean up OffCanvas methods.

### ⚠️ TODO
- [ ] Also check attributes that target an element.

Related to https://github.com/zurb/foundation-sites/issues/10684

## Types of changes

- [ ] Documentation

## Checklist

- [ ] I have read and follow the CONTRIBUTING.md document.
- [ ] The pull request title and template are correctly filled.
- [ ] The pull request targets the right branch (`develop` or `develop-v...`).
- [ ] My commits are correctly titled and contain all relevant information.

### DIFF
--- a/js/foundation.offcanvas.js
+++ b/js/foundation.offcanvas.js
@@ -65,27 +65,26 @@ class OffCanvas extends Plugin {
    * @private
    */
   _init() {
-    var id = this.$element.attr('id');
+    const id = this.$element.attr('id');
 
-    this.$element.attr('aria-hidden', 'true');
-
-    // Find off-canvas content, either by ID (if specified), by siblings or
-    // by closest selector (fallback)
+    // Find off-canvas content, either by:
     if (this.options.contentId) {
-      this.$content = $('#' + this.options.contentId);
+      // 1. The specified ID (if any)
+      this.$content = $(`#${this.options.contentId}`);
+      // (throw an error if not found)
       if (!this.$content.length) throw new Error(`Could not find the required OffCanvas content element with the given contentId "${this.options.contentId}".`);
     }
     else {
+      // 2. The siblings
       this.$content = this.$element.siblings('[data-off-canvas-content]').first();
+      this.nested = !this.$content.length;
+      // 3. The closest parent
       if (!this.$content.length) this.$content = this.$element.closest('[data-off-canvas-content]').first();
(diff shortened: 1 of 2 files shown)


DanielRuf:
`Error` correctly throws a real error and `console.error` just formats a console output.

https://developer.mozilla.org/en-US/docs/Web/API/console/error
https://console.spec.whatwg.org/#error
https://console.spec.whatwg.org/#logger
https://console.spec.whatwg.org/#loglevel-severity

First result for `throw Error vs console.error javascript`: https://stackoverflow.com/questions/25377115/what-is-the-difference-between-throw-error-and-console-error

ncoden:
@DanielRuf Ahah, I perfectly know what these methods _does_. What I am looking for is documentation about their recommended usage. I know that raising an error is a good thing when something is wrong ("fails early"), but is this still true for DOM-based frameworks? Shouldn't we in this case let the component run the best it can and display errors in the console?

SassNinja:
> Shouldn't we in this case let the component run the best it can and display errors in the console?

Imo that's exactly what we should do!

A `console.error` points out something is definitely wrong. 
But I wouldn't be that strict and stop execution. If the warning lights in a car are on it's up to the driver to decide if he fixes it or continues to drive.

ncoden:
Note: If we throw an error in the component to stop its own programmatic ("Something is wrong, stop everything"), it will not break others components. Errors are cached and logged by the Foundation core.

DanielRuf:
> Note: If we throw an error in the component to stop its own programmatic ("Something is wrong, stop everything"), it will not break others components. Errors are catch and logged by the Foundation core.

This should be fine then. In this case we can prevent wrongly used components if something is completely wrong / off.

SassNinja:
This PR seems to include more changes than just implementing `throw new Error()`...

I still have the opinion the framework should do the best it can to keep running (and provide error message if appropriate) instead of throwing real errors. Though this might be topic of discussion.

Anyway I'm closing this PR because it has conflicts and should be discussed as issue first.

PR #11613, opened 2018-11-27
feat: add support of [formnovalidate] in Abide

### LINKED ISSUE #11426: abide: provide ability bypass validation on certain submit actions
## What is missing?
The ability to bypass Abide validation for certain submit actions, e.g. an Abide equivalent of the `formnovalidate` attribute in HTML5.

## Why is it important?
It is ofen used to implement *Save* and related actions where invalid input is acceptable.

## Possible Solution
s/`$('[data-abide] [type=submit]')`/`$('[data-abide] [type=submit]:not([data-no-validate])`/

## Context
I was attempting to use Abide with a form where HTML 5 validation was already bypassed with `formnovalidate` on certain submit actions,

## Your Environment
- Foundation version(s) used: 6.4.3, 6.5.0-rc.2

## Checklist (all required):

SassNinja:
## Description

This PR adds `formnovalidate` support to Abide. Precisely it checks for `[type="submit"][formnovalidate]` and disables the whole validation if clicked (emulates the native behavior https://www.w3schools.com/tags/att_input_formnovalidate.asp)
In the course of this I've added the method `enableValidation` that provides an easy way to enable/disable the validation.

- Closes https://github.com/zurb/foundation-sites/issues/11426

## Types of changes

- [ ] Documentation

## Checklist


js/foundation.abide.js
```diff
@@ -92,6 +102,14 @@ class Abide extends Plugin {
     this._init();
   }
 
+  /**
+   * Enables (true) or disables (false) the whole validation
+   * @param {Boolean} enable
+   */
+  enableValidation(enable) {
```
ncoden: Choose one:

```js
myform.enableValidation();
myform.disableValidation();
```

```js
myform.setValidation(true);
myform.setValidation(false);
```

```js
myform.isValidationEnabled = true;
myform.isValidationEnabled = false;
```

js/foundation.abide.js
```diff
@@ -59,6 +61,14 @@ class Abide extends Plugin {
         return this.validateForm();
       });
 
+    this.$submitDisabler
+      .off('click.zf.abide')
+      .on('click.zf.abide', (e) => {
+        e.preventDefault();
+        this.enableValidation(false);
+        this.$element.submit();
+      });
```
ncoden: The `[formnovalidate]` attribute on an input element make the validation being bypassed _when the form is submitted via this element_. This does not enable or disable the form validation entirely and does not affect others validation methods.

See https://codepen.io/ncoden/pen/LXJybV?editors=1111

So if we want to support this attribute for Abide, we should search for a `[formnovalidate]` attribute on the `currentTarget` during the submit event, and bypass the validation if found. 

Also, any input can have `[formnovalidate]`, since using <kb>Enter</kb> on an `input[type="text"]` will trigger the form submission.

This does not prevent to have a private `isValidationEnabled` state next to that, or to dynamically check at each submission for `data-enable-validation` on the form element. But it should not be affected by `[formnovalidate]`, which is specific to the trigger of the form submission.

ncoden (dismissed):
Hi @SassNinja 👋,

Thank you for the pull request!
I have some remarks though regarding of the implementation.

js/foundation.abide.js
SassNinja: I go with the first (already pushed it)

Initially I thought about naming it `toggleValidation` (same as `setValidation`) but didn't like it because it's no clear enough imo.
Two individual methods sound good to me.

js/foundation.abide.js
SassNinja: > This does not enable or disable the form validation entirely

It does! When the submit is triggered by an input with `formnovalidate` the whole validation is skipped (`required`, `type="email"`etc.)
Or have I misunderstood you?

> we should search for a `[formnovalidate]` attribute on the `currentTarget` during the submit event

The `currentTarget` would be the form which doesn't have any `formnovalidate` attribute. I don't see another way than listening to the submit input/button.

> Also, any input can have `[formnovalidate]`, since using Enter on an `input[type="text"]` will trigger the form submission.

It's useless on any other input than the submit input. You can test is in your codepen:
- remove `formnovalidate` from the submit input
- add `formnovalidate` to the required email input

The validation won't get bypassed.
Pressing the Enter key will trigger the next submit input in the form's markup.
Imagine there are two submit inputs in the form, after the required email input.
- if the first submit input has `formnovalidate` the validation is bypassed
- if only the second submit input has `formnovalidate`(and the first not) the validation is not bypassed

ncoden: >  Or have I misunderstood you?

I afraid so. I meant that if you have a form with 2 submit inputs, one with `[formnovalidate]` and one without, submitting the form once with the first input will only disable the validation for this submission, and will not disable it for a second submission later via the second input. It's why I talked about a "form-level state" vs a "submission-level state". See the provided Codepen.

> The `currentTarget` would be the form

You are right, I forgot about that. I wanted to retrieve the element that triggered the form submission to handle `[formnovalidate]` like the browser does. In Firefox this would be `originalEvent.explicitOriginalTarget` (but it's an experimental feature).

> It's useless on any other input than the submit input

Again you are right. Only `input[type="submit"]` elements are taken into account for `[formnovalidate]`. When entering <kb>Enter</kb> on `input[type="text"]`, the event will be intercepted by the first `input[type="submit"]` element found in the form and the `submit` event will be submitted from it, then retargeted by the form element (so we loose the trace of the original submit input).

The hyperform package use the [following helper](https://github.com/hyperform/hyperform/blob/master/src/tools/catch_submit.js) to retrieve the submit button that were used to submit the form. It's a bit heavy though.

---

My point is only that whether the validation will be bypassed on depends on the submit button used directly or indirectly for the current submission only, regardless of "whether the form validation is enabled" (at the form level).

ncoden:
@SassNinja Could you add unit tests after we agreed on the expected behavior of the form at https://github.com/zurb/foundation-sites/pull/11613#pullrequestreview-178902704 ? 😄

js/foundation.abide.js
SassNinja: > submitting the form once with the first input will only disable the validation for this submission, and will not disable it for a second submission later via the second input.

if the form is submitted by the first input (bypassed validation) the page will probably reload so it doesn't matter if the validation is enabled again afterwards or not.
But I got your point – the disabled validation should be bound to the `formnovalidate` input and not affect the other inputs.

> I wanted to retrieve the element that triggered the form submission

Since all those elements can be found by `[type="submit"][formnovalidate]` I think the safest way is listening to those elements.

> The hyperform package use the following helper to retrieve the submit button that were used to submit the form. It's a bit heavy though.

Too heavy – I prefer event listener on submit inputs as described above.

> My point is only that whether the validation will be bypassed on depends on the submit button used directly or indirectly for the current submission only, regardless of "whether the form validation is enabled" (at the form level).

Ok, got it! I'll try to emulate this behavior.

ncoden: > Since all those elements can be found by [type="submit"][formnovalidate] I think the safest way is listening to those elements.

To behave like the native `[formnovalidate]` have to support:
- Click on the submit input
- <kb>Space</kb>/<kb>Enter</kb>
- <kb>Enter</kb> on any field of the form
- Any other form submission process

This is heavy. I would recommend you to drop the `[formnovalidate]` support and simply use a dynamic option (checked at the form submission) like `data-validate`. What do you think?

SassNinja: I've reworked this PR: Abide behaves now 100% identical to native formnovalidate

js/foundation.abide.js
```diff
@@ -385,6 +427,11 @@ class Abide extends Plugin {
         validator = $el.attr('data-validator'),
         equalTo = true;
 
+    // skip validation if disabled
```
kball: Why do we need this here as well? Don't we skip validating inputs below in validateForm before we would ever get to this?

js/foundation.abide.js
```diff
@@ -92,6 +105,35 @@ class Abide extends Plugin {
...
+   * Checks whether the submitted form should be validated or not, consodering formnovalidate and isEnabled
+   * @returns {Boolean}
+   * @private
+   */
+  _validationIsDisabled() {
+    if (this.isEnabled === false) { // whole validation disabled
+      return true;
+    } else if (typeof this.formnovalidate === 'boolean') { // triggered by $submit
+      return this.formnovalidate;
+    } else { // triggerd by Enter in non-submit input
```
kball: minor typo in comment. Should read 'triggered'.

kball (approved):
Broadly looking good - a couple quick notes/questions but I don't think they block merging.

js/foundation.abide.js
SassNinja: thanks, fixed now

js/foundation.abide.js
SassNinja: both are public methods, meaning you may call `validateInput` without triggering `validateForm`

furthermore Abide itself calls only `validateInput` on change of an input
(`validateForm` gets only called on submit)

that's why I've to skip validation in both methods

SassNinja: within `validateForm` I'm resetting this.formnovalidate which I use to recognize if the submit was triggered by a formnovalidate submit (otherwise it would affect future events)

joeworkman:

PR #11622, opened 2018-12-07
fix: fix SmoothScroll not applying custom options

### LINKED ISSUE #11620: SmoothScroll options can't be set
## What should happen?
I should be able to set smoothScroll options via data attributes.
`data-offset="100" data-animation-duration="1"`

## What happens instead?
These attributes are ignored

## Possible Solution
I believe problem is in `_handleLinkClick` method https://github.com/zurb/foundation-sites/blob/develop/js/foundation.smoothScroll.js#L60
`this` is always undefined. So `this.options` is also undefined. So only default options will ever be used.

Codepen in Foundation 6.5.1: https://codepen.io/anon/pen/QJPpbQ
Note that there is no offset and the scroll does take 500ms - instead of 1ms.
Also the scroll stops before actually reaching the element.

## Your Environment

- Foundation version(s) used: 6.5.1

ben-z:
## Description

SmoothScroll no longer takes into account the options passed in to its constructor because the event handler doesn't have the correct `this`. It worked prior to v6.5 because [before the refactor, the event handler uses `_this` from the closure instead of `this` directly](https://github.com/zurb/foundation-sites/blob/140abba37384784f2c6fc38d0ca8b1812c0eba16/js/foundation.smoothScroll.js#L47).

Please refer to https://github.com/zurb/foundation-sites/issues/11620 for a detailed description of this bug.

This PR binds `this` to the event handlers to solve the problem.

- Closes #11620

## Types of changes

- [ ] Documentation

## Checklist

It's my first time contributing. As far as I know there aren't any unit tests for this, so I didn't add any tests for my changes. Please let me know if I need to do anything else for this PR!

### DIFF
--- a/js/foundation.smoothScroll.js
+++ b/js/foundation.smoothScroll.js
@@ -39,8 +39,9 @@ class SmoothScroll extends Plugin {
      * @private
      */
     _events() {
-        this.$element.on('click.zf.smoothScroll', this._handleLinkClick.bind(this));
-        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', this._handleLinkClick.bind(this));
+        this._linkClickListenner = this._handleLinkClick.bind(this);
+        this.$element.on('click.zf.smoothScroll', this._linkClickListenner);
+        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', this._linkClickListenner);
     }
 
     /**
@@ -97,8 +98,8 @@ class SmoothScroll extends Plugin {
      * @function
      */
     _destroy() {
-        this.$element.off('click.zf.smoothScroll', this._handleLinkClick)
-        this.$element.off('click.zf.smoothScroll', 'a[href^="#"]', this._handleLinkClick);
+        this.$element.off('click.zf.smoothScroll', this._linkClickListenner)
+        this.$element.off('click.zf.smoothScroll', 'a[href^="#"]', this._linkClickListenner);
     }
 }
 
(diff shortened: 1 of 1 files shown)


DanielRuf (approved):

ncoden (changes requested):
With `.bind()`, a new function will be created and attached to the event. So when we want to clear the handlers in `_destroy` with `this._handleLinkClick`, the function will not be detached from the event.

You could save `this._handleLinkClick.bind(this)` as `this._linkClickListenner` then reuse it in `_destroy` to clear it.

ben-z:
@ncoden good catch! Fixed as suggested. Since `_events` seems to only be called from the constructor, I don't think we need to worry about `this._linkClickListenner` being overwritten when there are multiple calls to `_events`.

js/foundation.smoothScroll.js
```diff
@@ -39,8 +39,9 @@ class SmoothScroll extends Plugin {
      * @private
      */
     _events() {
-        this.$element.on('click.zf.smoothScroll', this._handleLinkClick.bind(this));
-        this.$element.on('click.zf.smoothScroll', 'a[href^="#"]', this._handleLinkClick.bind(this));
+        this._linkClickListenner = this._handleLinkClick.bind(this);
```
DanielRuf: Listener instead of Listenner

DanielRuf (changes requested):

js/foundation.smoothScroll.js
ben-z: awks 😅. Fixed

ncoden (approved):
LGTM. Thank you @ben-z 👍

ncoden:
Waiting for @DanielRuf to approve the changes as well.

DanielRuf (approved):

joeworkman:

PR #11669, opened 2019-02-03
fix: orbit component's transition end conflict issue

yanwsh:
orbit component's transition end should not be affected by its children's transition end.

## Description

For obit component, adding transition animation to each slide's child element. When user click child element and then click next slide, current slide will disappear and only show next slide. You can reproduce this issue [here](https://codepen.io/yanwsh/pen/gqWqeW). Click "slider 1" and then click "next arrow" button, you will see "slider 1" disappear and only have "slider 2" animation.

This is because there is a conflict between child's transition end and parent's transition end. Orbit slider only listen to `transitionend` event and can't figure out which event coming from. Add `stopPropagation` to prevent children's `transitionend` event propagate to parent.

You can verify the fix [here](https://codepen.io/yanwsh/pen/rPmbLw)

Let me know if you have any questions.  :blush:

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -5,7 +5,7 @@ import { Keyboard } from './foundation.util.keyboard';
 import { Motion } from './foundation.util.motion';
 import { Timer } from './foundation.util.timer';
 import { onImagesLoaded } from './foundation.util.imageLoader';
-import { GetYoDigits } from './foundation.core.utils';
+import { GetYoDigits, transitionend } from './foundation.core.utils';
 import { Plugin } from './foundation.core.plugin';
 import { Touch } from './foundation.util.touch'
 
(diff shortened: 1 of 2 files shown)


DanielRuf:
Hi, thanks for reporting this.

(Also I guess transitioning all properties is not what should happen.)

Can you provide a PR to show the exact changes?

DanielRuf:
I see no slider in the codepen with the fix, can you check this?

yanwsh:
Here is the issue.

When child contains transition (transition all or whatever) and you click on the slider, it will focus on the child element. Then, when you click "next slide", child will lose its focus and use transition to go back to original state. Then it will conflict with parent transition end event. 

Can you reproduce the issue with my [first link](https://codepen.io/yanwsh/pen/gqWqeW)? 

I think the fix link need take a while base on the network speed since it's not use CDN.

yanwsh:
Let me know if you can't reproduce this issue. I will upload a video to clarify this issue.

DanielRuf:
> adding transition animation to each slide's child element

This sounds like a custom requirement and use case.

A video would be probably help. Also please rebase the branch on our `develop` branch.

PR #11672, opened 2019-02-12
some more changes

### LINKED ISSUE #11671: Form validation with custom patterns fails if javascript is disabled
This happens with sites that use custom values for the pattern attribute, such as "alpha_numeric" which are not part of the html5 spec (which requires a regexp).

## What should happen?
Form should still be usable even if javascript is disabled or foundation.js is blocked.

## What happens instead?
Fields not matching the regular expression prevent the form from being submitted. In the case of the previous form, first field, it would pass it if you write "alpha_numeric" as the field content.

## Possible Solution

Please do not advise people to use the pattern field in ways that don't comply with the html5 standard! You seem to already have some other way to do it with custom attributes, so please don't use the standard one for this, as it breaks functionality for people opting out of your library.

Test Case: https://amazonwatch.org/take-action/urge-ecuadors-new-oil-friendly-environmental-minister-to-reject-yasuni-oil-drilling

That is not my site, I'm a web browser developer.

How to reproduce:
1. Disable javascript or install a script blocker such as NoScript.
2. Visit https://amazonwatch.org/take-action/urge-ecuadors-new-oil-friendly-environmental-minister-to-reject-yasuni-oil-drilling
3. Fill up the form with proper values and try to submit.

## Context

I maintain web browsers and extensions that focus on privacy and software freedom, and I was reported that sites using your library may fail to validate forms if javascript is disabled.

## Your Environment
- Foundation version(s) used:  https://amazonwatch.org/assets/v3/foundation.min.js
- Browser(s) name and version(s): Firefox 64
- Device, Operating System and version: not relevant
- Link to your project: That is not my site, I'm a web browser developer.

## Checklist
- [ ] There are no other issues similar to this one.
There is one, but you closed it! #6501
Please don't close this ticket until it is corrected, you are encouraging web developers to break standards.
- [X] The issue title and template are correctly filled.

DanielRuf:
invalid

### DIFF
--- a/js/foundation.accordion.js
+++ b/js/foundation.accordion.js
@@ -62,7 +62,7 @@ class Accordion extends Plugin {
       $content.attr({'role': 'tabpanel', 'aria-labelledby': linkId, 'aria-hidden': true, 'id': id});
     });
 
-    var $initActive = this.$element.find('.is-active').children('[data-tab-content]');
+    var $initActive = this.$element.find('.is-active [data-tab-content]');
     if ($initActive.length) {
       // Save up the initial hash to return to it later when going back in history
       this._initialAnchor = $initActive.prev('a').attr('href');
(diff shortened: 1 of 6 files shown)


DanielRuf:
Still have to write some tests when I find the time.

PR #11679, opened 2019-02-14
Mixins css triangle

esanzgar:
## Description

Some improvements to the css-triangle.

Rules are more specific and less verbose

- Closes 

## Types of changes

- [ ] Documentation

## Checklist

- [ ] I have read and follow the CONTRIBUTING.md document.
- [ ] The pull request title and template are correctly filled.
- [ ] The pull request targets the right branch (`develop` or `develop-v...`).
- [ ] My commits are correctly titled and contain all relevant information.

### DIFF
--- a/scss/util/_mixins.scss
+++ b/scss/util/_mixins.scss
@@ -44,29 +44,25 @@
   width: 0;
   height: 0;
 
-  border-style: inset;
+  border-style: solid;
   border-width: $triangle-size;
 
   content: '';
 
   @if ($triangle-direction == down) {
     border-bottom-width: 0;
-    border-top-style: solid;
     border-color: $triangle-color transparent transparent;
   }
   @if ($triangle-direction == up) {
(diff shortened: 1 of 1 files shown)


DanielRuf (commented):
We need this verbose code, for example for rtl support. And we only want to set the needed properties. Also `inset` is missing now.

esanzgar:
My understanding is that all the borders become transparent except the one that is visible (which is solid). Does it matter the style of the rest of the borders if they become transparent?

esanzgar:
Shouldn't be better to supply the four values for the border-color, just to avoid any default value?

DanielRuf:
Not sure what you mean, all the properties are already different. See the order of their values.

rafibomb (dismissed):
I'll check this out but seems to work. Need to test with RTL

joeworkman (approved):

PR #11761, opened 2019-05-28
Event trigger problem in minified version

pprossi:
https://github.com/zurb/foundation-sites/issues/11743#issuecomment-496658744

### DIFF
--- a/js/foundation.core.plugin.js
+++ b/js/foundation.core.plugin.js
@@ -44,11 +44,7 @@ function hyphenate(str) {
 }
 
 function getPluginName(obj) {
-  if(typeof(obj.constructor.name) !== 'undefined') {
-    return hyphenate(obj.constructor.name);
-  } else {
     return hyphenate(obj.className);
-  }
 }
 
 export {Plugin};
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -370,6 +370,12 @@ class Orbit extends Plugin {
           this.timer.restart();
         }
       }
+     /**
+      * Triggers when the slide has finished animating in.
+      * does'nt realy make any sense, keep event for compatibility reasons.
+      * @event Orbit#slidechange
+      */
+      this.$element.trigger('slidechange.zf.orbit', [$newSlide]);      
     }
   }
 
--- a/js/foundation.orbit.js
+++ b/js/foundation.orbit.js
@@ -349,6 +349,7 @@ class Orbit extends Plugin {
           this.options[`animInFrom${dirIn}`],
           function(){
             $newSlide.css({'display': 'block'}).attr('aria-live', 'polite');
+            _this.$element.trigger('endcurrentslidechange.zf.orbit', [$curSlide]);
         });
 
         Motion.animateOut(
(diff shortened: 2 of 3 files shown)


DanielRuf (commented):
Not sure about the slidechange changes, this would be a breaking change.

js/foundation.core.plugin.js
```diff
@@ -44,11 +44,7 @@ function hyphenate(str) {
 }
 
 function getPluginName(obj) {
-  if(typeof(obj.constructor.name) !== 'undefined') {
```
DanielRuf: Not sure but I think there was some reason for this in the past (IE10 and 11?). We'll see in the tests what will happen.

DanielRuf:
Please fill out the PR template. At least describe wjat this PR does, solves and why it does this way.

DanielRuf:
The slidechange tests fail.

> fires slidechange.zf.orbit event

Also it seems like we have to update some dependencies to get the tests running on Node 12.

pprossi:
```
   changeSlide()
      ✓ changes slides to the next one for ltr is true
      ✓ changes slides to the last one for ltr is false
      ✓ changes slides to the chosen one
      ✓ toggles ARIA attributes
      ✓ fires beforeslidechange.zf.orbit event
      1) fires slidechange.zf.orbit event
```
the slidechange.zf.orbit event does'nt make any sense. 
But for compatibility reasons you simple can ceep the removed part in the dunction.

pprossi:
the remaining errors seem to depend on node modules node-sass and node-fsevents.

js/foundation.orbit.js
```diff
@@ -368,11 +370,12 @@ class Orbit extends Plugin {
...
         }
       }
-    /**
-    * Triggers when the slide has finished animating in.
-    * @event Orbit#slidechange
-    */
-      this.$element.trigger('slidechange.zf.orbit', [$newSlide]);
+     /**
+      * Triggers when the slide has finished animating in.
+      * does'nt realy make any sense, keep event for compatibility reasons.
```
DanielRuf: Not sure what you nean that it does not make any sense. We can get the index and wait for onanimation/transitionend to work with this.

js/foundation.orbit.js
```diff
@@ -349,13 +349,15 @@ class Orbit extends Plugin {
           this.options[`animInFrom${dirIn}`],
           function(){
             $newSlide.css({'display': 'block'}).attr('aria-live', 'polite');
+            _this.$element.trigger('endcurrentslidechange.zf.orbit', [$curSlide]);
```
DanielRuf: These should be documented, see the current docs, `@fires` and the comments that we use.

DanielRuf (changes requested):
We should not add any new events in this PR. Let's discuss the event naming and events in another issue and PR.

js/foundation.orbit.js
pprossi: > slidechange.zf.orbit | Triggers when the slide has finished animating in.

Thats wrong. is triggert after animation is started and right after beforeslidechange.

Anyway, i had some proposals. Think about it and use it according to your wish. ;)

DanielRuf: Then we should change it in the docs. Please do not mix multiple issues in this PR so we can fix one issue after another and this makes reviews much easier.

PR #11870, opened 2019-10-19
fix: use offset property for button-dropdown mixin - closes #9030

### LINKED ISSUE #9030: button-dropdown not respecting $offset
Hi there,

When using the button-dropdown mixin, the last parameter (offset/distance from arrow icon to button text) isn't respected - it always gets the default value (the right padding of the button).

Seems like an easy fix: the component mixin actually gets `get-side($button-padding, right)` instead of `$offset` on line 163 of button.scss

Thanks!

DanielRuf:
## Description

This PR fixes the wrong `$offset` value in the `button-dropdown` mixin.

- Closes #9030

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/scss/components/_button.scss
+++ b/scss/components/_button.scss
@@ -294,7 +294,7 @@ $button-responsive-expanded: false !default;
 @mixin button-dropdown(
   $size: 0.4em,
   $color: $white,
-  $offset: get-side($button-padding, right)
+  $offset: $offset
 ) {
   &::after {
     @include css-triangle($size, $color, down);


scss/components/_button.scss
```diff
@@ -294,7 +294,7 @@ $button-responsive-expanded: false !default;
 @mixin button-dropdown(
   $size: 0.4em,
   $color: $white,
-  $offset: get-side($button-padding, right)
+  $offset: $offset
```
joeworkman: I think that we need to create a new variable for the default offset so that it can be referenced. 

```
$dropdown-offset : get-side($button-padding, right);

@mixin button-dropdown(
  $size: 0.4em,
  $color:white,
  $offset: $dropdown-offset
) {
```

joeworkman (changes requested):

scss/components/_button.scss
DanielRuf: I think I misunderstood the initial issue. We can close this as this parameter is already an optional one and the parameter is correctly passed.

PR #11878, opened 2019-10-21
The real master

joeworkman:
In this PR, I took the v6.5.3 release tag and merged it with the master branch. This should make master the true master branch now. The master branch should always contain the current production version of the Foundation framework.


DanielRuf (approved):
too much to review - should be ok

.travis.yml
```diff
@@ -7,10 +7,9 @@ os:
   - osx
 
 node_js:
-  - 6
   - 8
   - 10
-  - node
```
DanielRuf: Generelly we should test 8, 10, 12 and node. Because we want to fail early =)

DanielRuf: Generally we should test 8, 10, 12 and node. Because we want to fail early =)

DanielRuf: Because 12 is also LTS now. And we have many reports of users who say that it does not work on NodeJS 12.

DanielRuf: The only issue that I see is node-sass. Please do not add further changes to this branch, otherwise we have to rebase and merge other PRs. We should resolve the failing build in develop and not do this "hack" here =) Keep the branch as it is atm =) (before your changes)

PR #11885, opened 2019-10-24
Fix/responsive accordion tabs initial methods

### LINKED ISSUE #11884: responsive accordion tabs doesn't provide all methods initially
In our team we've spotted a bug in the responsive accordion tabs plugin

## What should happen?

After having initialized an instance, it should be possible to call the methods of the Accordion/Tabs plugin.
For example `$(selectorA).foundation('toggle', $(selectorB))`

## What happens instead?

The methods of the Accordion/Tabs plugin a not available initially but only after having transformed the instance from Accordion/Tabs into Tabs/Accordion (on breakpoint change).

## Possible Solution

The problem seems be located in foundation's core, precisely it's reflow method:
it overrides the zfPlugin data attribute what interferes with how the responsive according tabs plugin works.
I'm working on a solution...

Test Case: https://foundation.zurb.com/sites/docs/responsive-accordion-tabs.html

How to reproduce:
try to call an accordion method
```js
const $elem = $('[data-responsive-accordion-tabs="accordion medium-tabs large-accordion"]');

$elem.foundation('down', $elem.find('[data-tab-content]').eq(1)); // error
```
it doesn't work until you've changed the viewport (so it transforms into tabs) and then change again to get the accordion back.

## Your Environment

- Foundation version(s) used: 6.5.3
- Browser(s) name and version(s): any
- Device, Operating System and version: any

SassNinja:
## Description

The Responsive Accordion Tabs plugin is special because (in contrast to all other foundation plugins) it changes its data zfPlugin attribute into another plugin (either Accordion or Tabs).

Foundation's core doesn't consider this and executes the `reflow` method for each new plugin instance:
https://github.com/foundation/foundation-sites/blob/develop/js/foundation.core.js#L168

So the responsive accordion tabs plugin initially sets the correct zfData but gets overriden by reflow. As a result the transformed plugin's method (e.g. `down` for accordion) are initially not available when calling it programmatically.

My PR fixes this by returning the instance stored in zfData (instead of the ResponsiveAccordionTabs instance).

I've bound this new behavior to a `reflow` option so there's no breaking change and people who use `new Foundation.ResponsiveAccordionTabs()` still get the same result.

- Closes #11884

## Types of changes

- [ ] Documentation

## Checklist

### DIFF
--- a/js/foundation.core.js
+++ b/js/foundation.core.js
@@ -156,8 +156,8 @@ var Foundation = {
       // For each plugin found, initialize it
       $elem.each(function() {
         var $el = $(this),
-            opts = {};
-            
+            opts = { reflow: true };
+
         if($el.attr('data-options')){
           var thing = $el.attr('data-options').split(';').forEach(function(e, i){
             var opt = e.split(':').map(function(el){ return el.trim(); });
--- a/js/foundation.responsiveAccordionTabs.js
+++ b/js/foundation.responsiveAccordionTabs.js
@@ -36,6 +36,11 @@ var MenuPlugins = {
  */
 
 class ResponsiveAccordionTabs extends Plugin{
+  constructor(element, options) {
+    super(element, options);
+    return this.storezfData || this;
+  }
+
   /**
    * Creates a new instance of a responsive accordion tabs.
    * @class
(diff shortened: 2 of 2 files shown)


SassNinja:
just noticed this also fixes https://github.com/foundation/foundation-sites/issues/10181

DanielRuf (approved):

joeworkman:

PR #11886, opened 2019-10-24
Merge pull request #11886 from SassNinja:fix/get-yo-digits

### LINKED ISSUE #11883: GetYoDigits() pads with zeros when asked for more than 10 characters.
## What should happen?

In certain circumstances, when asking for more than 10 characters from `GetYoDigits()`, the string should continue to be random characters.

## What happens instead?

The string has 10 random characters at the front of the string, and the rest of the characters after the first 10 are 0.

Test Case: https://codepen.io/jellisii/pen/OJJpwvO?editors=1011

## Context

This only appears to happen in Chrome/Chromium.  I cannot test Safari or IE easily.  Firefox behaves as expected.  My current workaround is to generate and concatenate strings of 10 characters or less.  

## Your Environment

- Foundation version(s) used: 6.5.3
- Browser(s) name and version(s): Firefox 69.0.2 on Linux, Chromium 77.0.3865.120
- Device, Operating System and version:  Ubuntu Linux 18.04

## Checklist

- [x] The issue title and template are correctly filled.

SassNinja:
## Description

Prior to this there was unnecessary complicated math used in the `GetYoDigits` util and it didn't work if length > 10

Furthermore this change is a huge performance improvement!

Thanks to @DanielRuf for the nice test setup 👍 
https://jsperf.com/test4444

- Closes https://github.com/foundation/foundation-sites/issues/11883

## Types of changes

- [ ] Documentation

### DIFF
--- a/js/foundation.core.utils.js
+++ b/js/foundation.core.utils.js
@@ -19,9 +19,13 @@ function rtl() {
  * @default {String} '' - if no plugin name is provided, nothing is appended to the uid.
  * @returns {String} - unique id
  */
-function GetYoDigits(length, namespace){
-  length = length || 6;
-  return Math.round((Math.pow(36, length + 1) - Math.random() * Math.pow(36, length))).toString(36).slice(1) + (namespace ? `-${namespace}` : '');
+function GetYoDigits(length = 6, namespace){
+  let str = '';
+  const chars = '0123456789abcdefghijklmnopqrstuvwxyz';
+  for (let i = 0; i < length; i++) {
+    str += chars[Math.floor(Math.random() * chars.length)];
+  }
+  return namespace ? `${str}-${namespace}` : str;
 }
 
 /**
--- a/js/foundation.core.utils.js
+++ b/js/foundation.core.utils.js
@@ -22,8 +22,9 @@ function rtl() {
 function GetYoDigits(length = 6, namespace){
   let str = '';
   const chars = '0123456789abcdefghijklmnopqrstuvwxyz';
+  const charsLength = chars.length;
   for (let i = 0; i < length; i++) {
-    str += chars[Math.floor(Math.random() * chars.length)];
+    str += chars[Math.floor(Math.random() * charsLength)];
   }
   return namespace ? `${str}-${namespace}` : str;
 }


js/foundation.core.utils.js
```diff
@@ -19,9 +19,13 @@ function rtl() {
...
  * @returns {String} - unique id
  */
-function GetYoDigits(length, namespace){
-  length = length || 6;
-  return Math.round((Math.pow(36, length + 1) - Math.random() * Math.pow(36, length))).toString(36).slice(1) + (namespace ? `-${namespace}` : '');
+function GetYoDigits(length = 6, namespace){
+  let str = '';
+  const chars = '0123456789abcdefghijklmnopqrstuvwxyz';
+  for (let i = 0; i < length; i++) {
+    str += chars[Math.floor(Math.random() * chars.length)];
```
DanielRuf:
```suggestion
    str += chars[Math.floor(Math.random() * 36)];
```

DanielRuf: Either we cache it (do it once, the value never changes) or use directly `36`.

DanielRuf (changes requested):

js/foundation.core.utils.js
SassNinja: I wouldn't use directly `36` because it's not that obvious it's the chars length (might introduce issue if someone extends the chars).

What do you mean with "cache it"?
Are you talking about something as `const charsLength = chars.length` before the for loop?

SassNinja: Doesn't seem to make much difference if you use `String().length` more than once:
https://jsperf.com/whats-the-price-of-string-length

DanielRuf: Yes, because some VMs cache it but we should do that before. Also older browsers behave different. Newer ones like the current v8 engine are optimized.

DanielRuf: > Are you talking about something as `const charsLength = chars.length` before the for loop?

Yes, exactly. Please do it like this.

SassNinja: alright, I've changed & pushed it

DanielRuf (approved):
LGTM 👍

DanielRuf:
> Merge pull request #11886 from SassNinja:fix/get-yo-digits

Normally this is done when we create a merge commit using `--no-ff` or the GitHub UI =)

PR #11979, opened 2020-02-03
Implement SonarCloud

joeworkman:
integrated with Travis


DanielRuf:
No, please use the GH integration from the marjetplace as we should also move to GitHub Actions =)

joeworkman:
Roger that. I will kill it...

PR #11986, opened 2020-02-05
chore: better sass dependency management

joeworkman:
## Description
This is another attempt at https://github.com/foundation/foundation-sites/pull/9542

I could not get sass:foundationSCSS to work with gulp-shopify-sass and eyeglass. The code is still in there but I removed the taks from being ran. Maybe we can figure it out one day. 

The sassy-lists libraries are imported from node_modules now.

Normalize.css is still manually added into vendor/normalize. The reason that this needs to be there is that would need to be imported from within a mixin. Sass does not allow this yet. There for it needs to be made into its own mixin. A better way would be to simply have a separate import command for this. However, this would break existing implementations and we do not want to do that. The normalize.css does not change frequently at all. So this is not too big of a deal right now. Its definitely something that could be fixed in F7 though. 

- Closes 

## Types of changes
- [ ] Documentation


DanielRuf (commented):
This is a big breaking change regarding the changes. Also this will break the Sass version as users have to run `npm i` now. There is a good reason why the vendor folder is in the published packages.

I would close this PR and use the correct way and improve it in small steps without such changes.

gulp/config.js
```diff
@@ -24,21 +24,11 @@ module.exports = {
...
-    'node_modules/@(sassy-lists)/stylesheets/functions/_replace.scss',
-    'node_modules/@(sassy-lists)/stylesheets/functions/_to-list.scss'
-  ],
-
   SASS_DOC_PATHS: [
     'scss',
     'node_modules/motion-ui/src',
-    'node_modules/foundation-docs/scss'
+    'node_modules/foundation-docs/scss',
+    'node_modules/sassy-lists/stylesheets'
```
DanielRuf: This is for docs.

gulp/config.js
```diff
@@ -66,6 +56,7 @@ module.exports = {
   ],
 
   DIST_FILES: [
+    './_build/assets/scss/foundation.scss',
```
DanielRuf: Makes not much sense, this is not a dist file.

gulp/tasks/customizer.js
```diff
@@ -63,32 +64,18 @@ gulp.task('customizer:loadConfig', function(done) {
...
-      'node_modules/@(sassy-lists)/stylesheets/functions/remove',
-      'node_modules/@(sassy-lists)/stylesheets/functions/replace',
-      'node_modules/@(sassy-lists)/stylesheets/functions/to-list'
-    ])
-    .pipe(gulp.dest('_vendor'));
-});
-
 // Creates a Sass file from the module/variable list and creates foundation.css and foundation.min.css
-gulp.task('customizer:sass', gulp.series('customizer:loadConfig', 'customizer:prepareSassDeps', function() {
+gulp.task('customizer:sass', gulp.series('customizer:loadConfig', function() {
```
DanielRuf: Removing things to fix them is not the correct approach.

gulp/tasks/customizer.js
```diff
@@ -63,32 +64,18 @@ gulp.task('customizer:loadConfig', function(done) {
...
 
   return stream
-    .pipe(sass({
+    .pipe(sass(eyeglass({
       includePaths: [
         'scss',
         'node_modules/motion-ui/src'
       ]
-    }))
+    })))
```
DanielRuf: Badly formatted and hard to read.

gulp/tasks/deploy.js
```diff
@@ -22,7 +22,7 @@ var NEXT_VERSION;
 gulp.task('deploy', gulp.series('deploy:prompt', 'deploy:version', 'deploy:dist', 'deploy:plugins', 'deploy:settings', 'deploy:commit', 'deploy:templates'));
 
 gulp.task('deploy:prep', gulp.series('deploy:prompt', 'deploy:version', 'deploy:dist', 'deploy:plugins', 'deploy:settings'));
-gulp.task('deploy:dist', gulp.series('sass:foundation', 'javascript:foundation', 'deploy:dist:files', 'deploy:dist:bundles'));
+gulp.task('deploy:dist', gulp.series('sass:foundationCSS', 'javascript:foundation', 'deploy:dist:files', 'deploy:dist:bundles'));
```
DanielRuf: Renaming tasks will break implementations and docs.

gulp/tasks/deploy.js
```diff
@@ -52,11 +52,12 @@ gulp.task('deploy:version', function() {
 
 // Generates compiled CSS and JS files and sourcemaps and puts them in the dist/ folder
 gulp.task('deploy:dist:files', function() {
-  var cssFilter = filter(['**/*.css'], { restore: true });
-  var jsFilter  = filter(['**/*.js'], { restore: true });
+  var scssFilter         = filter(['**/*.scss'], { restore: true });
```
DanielRuf: Please don't do this type of indention.

gulp/tasks/deploy.js
```diff
@@ -77,6 +78,10 @@ gulp.task('deploy:dist:files', function() {
     // * Create minified-sourcemaps based on standard sourcemaps.
     //   Sourcemaps are initialized before the ".min" renaming to be able retrieve
     //   original sourcemaps from source names.
+    .pipe(scssFilter)
+      .pipe(gulp.dest('./dist/scss'))
```
DanielRuf: This should never land in dist.

gulp/tasks/sass.js
```diff
@@ -1,45 +1,46 @@
-'use strict';
```
DanielRuf: Disabling strict mode allows more broken JS.

gulp/tasks/sass.js
```diff
@@ -1,45 +1,46 @@
...
-// Prepare dependencies
-gulp.task('sass:deps', function() {
-  return gulp.src(CONFIG.SASS_DEPS_FILES)
-    .pipe(gulp.dest('_vendor'));
+// Concat Foundation Sass and its dependencies into a single SCSS file
+// TODO: This is not ran with the default deploy and sass tasks. There are still issues with eyeglass and gulp-shopify-sass
+gulp.task('sass:foundationSCSS', function() {
+  return gulp.src(['assets/foundation.scss'])
+    .pipe(gss(eyeglass()))
+    .pipe(rename('foundation.scss'))
```
DanielRuf: Makes it all unnecessary complicated.

package.json
```diff
@@ -159,5 +161,8 @@
     "commitizen": {
       "path": "./node_modules/cz-conventional-changelog"
     }
+  },
+  "dependencies": {
+    "normalize.css": "^8.0.1"
```
DanielRuf: This should not be there.

scss/foundation.scss
```diff
@@ -7,13 +7,7 @@
...
 // --- Dependencies ---
 @import 'vendor/normalize';
-@import '../_vendor/sassy-lists/stylesheets/helpers/missing-dependencies';
-@import '../_vendor/sassy-lists/stylesheets/helpers/true';
-@import '../_vendor/sassy-lists/stylesheets/functions/contain';
-@import '../_vendor/sassy-lists/stylesheets/functions/purge';
-@import '../_vendor/sassy-lists/stylesheets/functions/remove';
-@import '../_vendor/sassy-lists/stylesheets/functions/replace';
-@import '../_vendor/sassy-lists/stylesheets/functions/to-list';
+@import '../node_modules/sassy-lists/stylesheets/SassyLists';
```
DanielRuf: Will not work in the pure scss version.

scss/vendor/normalize.scss
```diff
@@ -1,281 +1,184 @@
...
 
-  // Remove the inner padding in Chrome and Safari on macOS.
-
-  [type="search"]::-webkit-search-decoration {
-    -webkit-appearance: none;
+  [type=search]::-webkit-search-decoration {
+    -webkit-appearance: none
   }
 
-  // 1. Correct the inability to style clickable types in iOS and Safari.
```
DanielRuf: Normally there should be still the comments.

Please use the original file, not the production file.

https://github.com/necolas/normalize.css/blob/8.0.1/normalize.css#L304

joeworkman:
OK. There went quite a few hours of work... oh well. I will close the other one as well.

PR #12025, opened 2020-03-26
Fix Keyboard trap in drilldown subnav

### LINKED ISSUE #11641: Drilldown Sub-Menus won't let you Tab/Shift Tab out of Sub-Menu
## What should happen?
While in the submenu of a drilldown menu you should be able to shift-tab or tab to items below or above the menu items.

You should also be able to tab/shift-tab to an already open submenu.

## What happens instead?
Right now you get stuck inside the submenu when you navigate to it via the keyboard keys.

If you try to tab to an already open submenu it won't let you get to it as well.

## Possible Solution

In the drilldown.js file you are setting what the keyboard keys should do:

_setup(element, options) {
    this.$element = element;
    this.options = $.extend({}, Drilldown.defaults, this.$element.data(), options);
    this.className = 'Drilldown'; // ie9 back compat

    this._init();

    Keyboard.register('Drilldown', {
      'ENTER': 'open',
      'SPACE': 'open',
      'ARROW_RIGHT': 'next',
      'ARROW_UP': 'up',
      'ARROW_DOWN': 'down',
      'ARROW_LEFT': 'previous',
      'ESCAPE': 'close',
      'TAB': 'down',
      'SHIFT_TAB': 'up'
    });
  }

I would strongly recommend removing the tab/shift tab lines. This will let the default action for those keys still do what you want it today without manually setting them.

For being able to tab into the submenu I would recommend setting the whole ul to visibility:hidden and having a visibility: visible on the is-active drilldown allowing for the correct tabing/shifting keyboard commands to work correctly.

Test Case: 
https://codepen.io/ndonley/pen/YdEOzb

How to reproduce:
1. Open the sub menu for the drilldown with the keyboard
2. Try to tab above or below the drilldown submenu
3. Try using the keyboard to tab into the open drilldown submenu

## Context

So it looks like this was brought up before in:
https://github.com/zurb/foundation-sites/issues/9597
Looks like you tried to fix it but seems to still be an issue. The main reason this is an issue is for accessibility. The user needs to be able to navigate all menus with only their keyboard. As of right now they get stuck with the keyboard alone. I would recommend not changing the native command keys since this seems to cause an issue.

## Checklist

- [X] There are no other issues similar to this one.
- [X ] The issue title and template are correctly filled.

trishasalas:
## Description

1. Remove TAB/SHIFT_TAB from script to allow browser to manage tabbing

While the tab key moves focus appropriately in the dom order within this component. The focus is not visually discernible when the subnav is opened. Since the method to move back up within the dom tree is present via arrow keys and the link at the top of the subnav, I agree with the solution that was proposed in https://github.com/foundation/foundation-sites/issues/11641
Broad information about keyboard best practice can be found here: https://www.w3.org/TR/wai-aria-practices/#kbd_generalnav
2. Set root ul visibility:hidden and set subnav to visibility:visible when subnav is open

This will remove it from the dom allowing the user to tab into the next component/area when the subnav is open/visible

3. Add `.visible` class for use in drilldown subnav when opened.

These changes are recommented by https://github.com/ntdonley in https://github.com/foundation/foundation-sites/issues/11641

- Closes https://github.com/foundation/foundation-sites/issues/11641

## Types of changes

- [ ] Documentation
- [X] Breaking change (anything that would change an existing functionality)

## Checklist

- [X] The pull request title and template are correctly filled.
- [X] The pull request targets the right branch (`develop` or `develop-v...`).
- [X] My commits are correctly titled and contain all relevant information.

### DIFF
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -38,8 +38,6 @@ class Drilldown extends Plugin {
       'ARROW_DOWN': 'down',
       'ARROW_LEFT': 'previous',
       'ESCAPE': 'close',
-      'TAB': 'down',
-      'SHIFT_TAB': 'up'
     });
   }
 
@@ -243,16 +241,6 @@ class Drilldown extends Plugin {
             }, 1);
           });
           return true;
-        },
-        up: function() {
-          $prevElement.focus();
-          // Don't tap focus on first element in root ul
-          return !$element.is(_this.$element.find('> li:first-child > a'));
-        },
-        down: function() {
-          $nextElement.focus();
-          // Don't tap focus on last element in root ul
-          return !$element.is(_this.$element.find('> li:last-child > a'));
         },
         close: function() {
           // Don't close on element in root ul
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -210,18 +210,7 @@ class Drilldown extends Plugin {
     var _this = this;
 
     this.$menuItems.add(this.$element.find('.js-drilldown-back > a, .is-submenu-parent-item > a')).on('keydown.zf.drilldown', function(e){
-      var $element = $(this),
-          $elements = $element.parent('li').parent('ul').children('li').children('a'),
-          $prevElement,
-          $nextElement;
-
-      $elements.each(function(i) {
-        if ($(this).is($element)) {
-          $prevElement = $elements.eq(Math.max(0, i-1));
-          $nextElement = $elements.eq(Math.min(i+1, $elements.length-1));
-          return;
-        }
-      });
+      var $element = $(this);
 
       Keyboard.handleKey(e, 'Drilldown', {
         next: function() {
--- a/js/foundation.drilldown.js
+++ b/js/foundation.drilldown.js
@@ -149,26 +149,26 @@ class Drilldown extends Plugin {
     var _this = this;
 
     $elem.off('click.zf.drilldown')
-    .on('click.zf.drilldown', function(e){
-      if($(e.target).parentsUntil('ul', 'li').hasClass('is-drilldown-submenu-parent')){
-        e.preventDefault();
-      }
+         .on('click.zf.drilldown', function(e){
+           if($(e.target).parentsUntil('ul', 'li').hasClass('is-drilldown-submenu-parent')){
+             e.preventDefault();
+           }
 
-      // if(e.target !== e.currentTarget.firstElementChild){
-      //   return false;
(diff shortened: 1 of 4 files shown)


scss/components/_visibility.scss
```diff
@@ -70,6 +70,10 @@
     visibility: hidden;
   }
 
+  .visible {
+    visibility: visible !important;
```
DanielRuf: Why `!important`?

js/foundation.drilldown.js
```diff
@@ -38,8 +38,6 @@ class Drilldown extends Plugin {
       'ARROW_DOWN': 'down',
       'ARROW_LEFT': 'previous',
       'ESCAPE': 'close',
-      'TAB': 'down',
-      'SHIFT_TAB': 'up'
```
DanielRuf: So how can `up` be triggered then?

DanielRuf:
Hi @trishasalas,

thanks for your contribution.
Please take a look at the code smells by the changes and resolve them.

scss/components/_visibility.scss
trishasalas: Good call, tested and no need for `!important` here.

js/foundation.drilldown.js
trishasalas: `SHIFT+TAB` is default keyboard behavior as well and will work without the need to define it in the component. You can see this PR in action at utulsa.edu

trishasalas:
Thank you, @DanielRuf. Changes have been made as requested.

DanielRuf:
Hi @trishasalas,

please check if the following is enabled on the right. Otherwise I can not update your branch with our current `develop` branch.

Bildschirmfoto 2020-03-28 um 12 34 20

DanielRuf:
Please also updatethe failing tests.

trishasalas:
The 'allow edits from maintainer' checkbox is not visible for me. I have done some searching and can't figure out why it might not be there, any thoughts? 

Testing is new-ish for me, I am working on it and will have a fix asap. It's a good opportunity for me, thanks for your patience and working with me on this!

DanielRuf:
You can ignore the browserstack error because this happens because forks have no access to the credentials / secrets due to the security policies.

trishasalas:
Thank you for the clarification regarding Browserstack. I’ve asked GitHub directly about the missing checkbox.

trishasalas:
I am closing this pull request. Please see https://github.com/foundation/foundation-sites/pull/12038

Issue #11754, opened 2019-05-20
Magellan with data-offset does not trigger correctly on the first item


mingyeungs:
## What should happen?
Magellan with `data-offset` should should be triggered at offset positions

## What happens instead?
All menu items, but not the first one, is triggered at the correct position.

## Possible Solution

update `/js/foundation.magellan.js` line 170:

from 
`if(newScrollPos < this.points[0]){ /* do nothing */ }`

to: 
`if(newScrollPos < this.points[0] - this.options.offset - (isScrollingUp ? this.options.threshold : 0)){ /* do nothing */ }`

## Test Case and/or Steps to Reproduce (for bugs)

Test Case: https://codepen.io/mingyeungs/full/oRerpB

## Your Environment

- Foundation version(s) used: 6.5.1
- Browser(s) name and version(s): Chrome 74.0.3729
- Device, Operating System and version: macOS High Sierra (10.13.4)

## Checklist

- [x] I have read and follow the CONTRIBUTING.md document.
- [x] There are no other issues similar to this one.
- [x] The issue title and template are correctly filled.

JPustkuchen:
Yup, we can confirm this problem! Any chance to fix this?

thomasfrobieter:
I can confirm that @mingyeungs fix works fine!

DanielRuf:
Thanks for confirming that this resolves the issue. PRs which fix this are very welcome.

PR #12084, opened 2020-05-28
Magellan with data-offset does not trigger correctly on the first ite…

JPustkuchen:
See #11754


DanielRuf:
Hi @JPustkuchen,

thanks for your PR.

Please use our PR template and provide more information so we can review the changes.

Is it also possible for you to provide a test which fails without this change and succeeds with it?

JPustkuchen:
Thank you very very very much Daniel! Really small but very important fix. Sorry for my late reply, I was very busy. Sorry!

PR #12139, opened 2020-08-27
fix (doc) title-bar fragment in link

pine3ree:
`#title-bar` is not a valid fragment of the _off-canvas_ page


DanielRuf (approved):
Nice catch.

LGTM

PR #12256, opened 2021-08-31
use passive event-listeners for touch/wheel events if possible

wfehr:
Credits to @aivaras82 for this approach:
https://github.com/foundation/foundation-sites/issues/11517#issuecomment-617965760

Problem/solution etc. are the same as described in linked issue (and comment).

This still exists with new version 6.7.0.

### DIFF
--- a/js/foundation.util.touch.js
+++ b/js/foundation.util.touch.js
@@ -13,7 +13,19 @@ var startPosX,
     elapsedTime,
     startEvent,
     isMoving = false,
-    didMoved = false;
+    didMoved = false,
+    supportsPassive = false;
+
+// Test via a getter in the options object to see if the passive property is accessed
+try {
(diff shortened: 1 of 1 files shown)


DaSchTour:
I would suggest to drop IE support and simply and try avoiding this hacky feature check.

joeworkman:
I have not tested this. Does IE11 not simply ignore the option if passed? Do we need to even have the feature check?

joeworkman:
I tested the following code in IE11 and it did not produce any errors inside of the console. 

```js
document.addEventListener("touchstart", function(e) {
    console.log(e.defaultPrevented);  // will be false
    e.preventDefault();   // does nothing since the listener is passive
    console.log(e.defaultPrevented);  // still false
  }, {passive: true});
```

PR #12326, opened 2021-11-10
Fix: Do not return unitless zeroes from -zf-to-rem, as these do not work as values within css calc()

### LINKED ISSUE #12325: Zero width gutters lead to invalid calc() values for XY-grid cells
## What should happen?
Zero-width (`0px`) gutters should be valid for XY grid.

## What happens instead?
`-zf-to-rem` may produce values without an unit suffix as an apparent output optimization, but these are invalid as inputs to the css `calc()` function. See steps to reproduce for a simple example.

## Possible Solution
My suggestion would be to remove the suffix-lacking zero return "optimization" from `-zf-to-rem`. I will create a pull request with the proposed change myself, but I realize that I know relatively little of the foundation codebase and that this might not be the best way to fix these problems. To my sensibilities however it seems that a function called "X to Y" should not return something that is not a Y.

How to reproduce:
1. Set `$grid-margin-gutters: (small: 0px, medium: 0px);`
2. Create an XY-grid with one or more cells
```scss
.test-grid {
  @include xy-grid;
  .test-bigcell {
    @include xy-cell(12);
    background: red;
  }
}
```
3. The `rem-calc` in `xy-cell-size-css` produces an invalid `calc()` instruction for the cells' width, which browsers will then ignore.

## Your Environment
- Foundation version(s) used: 6.7.4
- Browser(s) name and version(s): Chrome 97.0.4681.0 and probably all others
- Device, Operating System and version: PC

## Checklist
- [x] The issue title and template are correctly filled.

noorus:
## Description
Zero gutters used to work fine with the XY grid.
After I recently upgraded several sites to Foundation 6.7.4, I found that at some point there's been a breaking change. I tracked the problem down to invalid `calc()`s being generated for cell widths, due to `-zf-to-rem()` returning zero values without a unit suffix. This patch fixed my problems; See the issue for further discussion though.

- Closes #12325 

## Types of changes
- [ ] Documentation

## Checklist

### DIFF
--- a/scss/util/_unit.scss
+++ b/scss/util/_unit.scss
@@ -95,11 +95,6 @@ $global-font-size: 100% !default;
     $value: divide(strip-unit($value), strip-unit($base)) * 1rem;
   }
 
-  // Turn 0rem into 0
-  @if $value == 0rem {
-    $value: 0;
-  }
-
   @return $value;
 }
 


joeworkman:
I can see this as a bug. However, your fix feel a bit heavy handed. the rem-calc function is used throughout the SCSS. I feel the better solution would be to add exception to this rule before each calc(). I will get that done. Thank you.

PR #12433, opened 2022-05-30
Added online version of autoprefixer.

rahulkumarsingh73690:
Added online version of Autoprefixer. So users can instantly use it without installing it.


joeworkman:
The online tool linked to looks nice. However, I am not sure it's good to link to some random online tool that I do not know who is behind it.

